PR #3, opened 2020-03-17
Better conditional tooltip usage

regoarrarr:
Incorporating changes Ross made to better handle the conditional tooltip usage

### DIFF
--- a/gallery/src/components/NxTextInput/NxTextInputPage.tsx
+++ b/gallery/src/components/NxTextInput/NxTextInputPage.tsx
@@ -59,9 +59,7 @@ const NxTextInputPage = () =>
             <td>string[]</td>
             <td>No</td>
             <td>
-              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid.
-              If the form is invalid, the entire <code className="nx-code">NxTextInput</code> element will be wrapped in
-              an <code className="nx-code">NxTooltip</code> element, with error styling.
+              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid
             </td>
           </tr>
           <tr>
--- a/gallery/src/components/NxTextInput/NxTextInputPage.tsx
+++ b/gallery/src/components/NxTextInput/NxTextInputPage.tsx
@@ -59,7 +59,8 @@ const NxTextInputPage = () =>
             <td>string[]</td>
             <td>No</td>
             <td>
-              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid
+              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid.
+              If the field is invalid, it will receive error styling and a tooltip displaying the first error message.
             </td>
           </tr>
           <tr>
--- a/gallery/src/components/NxTooltip/NxTooltipPage.tsx
+++ b/gallery/src/components/NxTooltip/NxTooltipPage.tsx
@@ -41,7 +41,7 @@ export default function NxTooltipPage() {
               <td className="nx-cell">title</td>
               <td className="nx-cell">React Node (e.g. VDOM or string)</td>
               <td className="nx-cell">Yes</td>
-              <td className="nx-cell">The tooltip content</td>
+              <td className="nx-cell">The tooltip content. Defaults to empty string</td>
             </tr>
             <tr className="nx-table-row">
               <td className="nx-cell">placement</td>
--- a/gallery/src/components/NxTooltip/NxTooltipPage.tsx
+++ b/gallery/src/components/NxTooltip/NxTooltipPage.tsx
@@ -41,7 +41,7 @@ export default function NxTooltipPage() {
               <td className="nx-cell">title</td>
               <td className="nx-cell">React Node (e.g. VDOM or string)</td>
               <td className="nx-cell">Yes</td>
-              <td className="nx-cell">The tooltip content. Defaults to empty string</td>
+              <td className="nx-cell">The tooltip content. If empty, the tooltip is not shown.</td>
             </tr>
             <tr className="nx-table-row">
               <td className="nx-cell">placement</td>
--- a/lib/src/components/NxTextInput/NxTextInput.tsx
+++ b/lib/src/components/NxTextInput/NxTextInput.tsx
@@ -33,8 +33,10 @@ const NxTextInput = forwardRef<HTMLInputElement | HTMLTextAreaElement, Props>(
        * `trimmedValue` is a hidden property in `props`
        * We need to remove it so react doesn't complain when we pass the object
        * with the props to the `createElement` method below.
+       * `title` is removed to prevent errors when `NxTextInput`, which itself uses` NxTooltip`, is wrapped
+       * in another `NxTooltip`
        */
-      const newProps = omit(['trimmedValue'], attrs);
+      const newProps = omit(['trimmedValue', 'title'], attrs);
       const isTextArea = type === 'textarea',
           element = isTextArea ? 'textarea' : 'input',
           typeAttr = isTextArea ? undefined : (type || 'text'),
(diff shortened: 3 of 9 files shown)


gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -59,9 +59,7 @@ const NxTextInputPage = () =>
             <td>string[]</td>
             <td>No</td>
             <td>
-              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid.
-              If the form is invalid, the entire <code className="nx-code">NxTextInput</code> element will be wrapped in
-              an <code className="nx-code">NxTooltip</code> element, with error styling.
+              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid
```
rpokorny:
```suggestion
              Zero or more validation error messages.  If empty or not defined, the field is considered to be valid.
              If the field is invalid, it will receive error styling and a tooltip displaying the first error message.
```

gallery/src/components/NxTooltip/NxTooltipPage.tsx
```diff
@@ -41,7 +41,7 @@ export default function NxTooltipPage() {
               <td className="nx-cell">title</td>
               <td className="nx-cell">React Node (e.g. VDOM or string)</td>
               <td className="nx-cell">Yes</td>
-              <td className="nx-cell">The tooltip content</td>
+              <td className="nx-cell">The tooltip content. Defaults to empty string</td>
```
rpokorny:
```suggestion
              <td className="nx-cell">The tooltip content. If empty, the tooltip is not shown.</td>
```

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -16,28 +16,22 @@ describe('NxTextInput', function() {
...
 
-  it('renders a text input by default', function() {
-    expect(getShallowComponent()).toMatchSelector('input');
-    expect(getShallowComponent()).toHaveProp('type', 'text');
-    expect(getShallowComponent().find(NxTooltip)).not.toExist();
-  });
-
-  it('renders a text input wrapped in a NxTooltip if there are validation errors', function() {
-    expect(getShallowComponent({ validationErrors: 'foo' })).toMatchSelector(NxTooltip);
+  it('renders a text input wrapped in a NxTooltip by default', function() {
```
rpokorny: It always does, right?

```suggestion
  it('renders a text input wrapped in a NxTooltip', function() {
```

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -67,16 +69,14 @@ const NxTextInput = forwardRef<HTMLInputElement | HTMLTextAreaElement, Props>(
         onKeyPress: inputOnKeyPress
       });
 
-      const invalidTextElementWithTooltip = (
+      return (
         <NxTooltip open={isInvalid}
                    title={firstValidationError}
                    className="nx-tooltip--validation-error"
                    placement="top-end">
           { textElement }
```
rpokorny: There's no longer really any reason to break out the textElement as a separate variable

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
rpokorny: > It always does, right?

Oh I see, "text input" as opposed to "password input".

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -61,7 +55,13 @@ describe('NxTextInput', function() {
   });
 
   it('sets the nx-tooltip--validation-error class on the tooltip', function() {
-    expect(getShallowComponent({ validationErrors: 'foo' })).toHaveClassName('nx-tooltip--validation-error');
+    expect(getShallowComponent()).toHaveClassName('nx-tooltip--validation-error');
```
rpokorny: This seems like it reverses the meaning of the test.  Also, either way the test seems inadequate because it doesn't test the case when the `nx-tooltip--validation-error` class _shouldn't_ be present

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -61,7 +55,13 @@ describe('NxTextInput', function() {
...
+  });
+
+  it('sets the tooltips `open` prop to false if there are no validation errors', function() {
+    expect(getShallowComponent()).toHaveProp('open', false);
+    expect(getShallowComponent({ validationErrors: null })).toHaveProp('open', false);
+    expect(getShallowComponent({ validationErrors: [] })).toHaveProp('open', false);
```
rpokorny: To mirror the test below, this should probably test the `title` as well as `open`

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -30,12 +30,20 @@ describe('NxTooltip', function() {
     expect(component).toHaveProp('onClose', onClose);
   });
 
-  it('replaces null props with undefined', function() {
-    const component = getNxTooltip({ className: null, open: null, onOpen: null, onClose: null, placement: null });
+  it('replaces most null props with undefined, and title with empty string', function() {
```
rpokorny:
```suggestion
  it('replaces null title with empty string, and all other null props with undefined', function() {
```

lib/src/components/NxTextInput/NxTextInput.tsx
regoarrarr: > There's no longer really any reason to break out the textElement as a separate variable

Put it back into one statement in 9e7cb880b09b7d8ba0e700a4e8f247329abe5618

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
regoarrarr: > This seems like it reverses the meaning of the test. Also, either way the test seems inadequate because it doesn't test the case when the `nx-tooltip--validation-error` class _shouldn't_ be present

Interestingly the code as currently established (i.e. before I made any changes) shows the `nx-tooltip--validation-error` class unconditionally. 

I don't remember changing this test (though obviously I did according to the log) but currently there is no scenario in which that class is not present. I guess that goes back to the whole discussion of whether the tooltip should be displayed conditionally or not

raveneyex: > currently there is no scenario in which that class is not present.

That sounds like a bug.

regoarrarr: > > currently there is no scenario in which that class is not present.
> 
> That sounds like a bug.

Maybe? If a tooltip is not open, does it matter what CSS class is on it?

rpokorny: > Maybe? If a tooltip is not open, does it matter what CSS class is on it?

No it doesn't, and I hadn't looked closely enough before to understand that.  This tooltip _is_ always a validation error tooltip.  It just isn't visible if there is in fact no validation error.  I think the test as-is in this PR is fine then.

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
rpokorny: > c45c435

> and title to null

That's not right though, and I'd hope the test isn't passing like that.  It sets it to empty string.  MUI `Tooltip` doesn't tolerate null or undefined

regoarrarr: > > [c45c435](https://github.com/sonatype/sonatype-react-shared-components/commit/c45c4354f7c8f54016e9ea928e33a4a7d250bef2)
> 
> > and title to null
> 
> That's not right though, and I'd hope the test isn't passing like that. It sets it to empty string. MUI `Tooltip` doesn't tolerate null or undefined

This is if there are no validation errors so the title of the NxTextInput is null. The NxTooltip takes that and converts it to empty string (to pass into MUI Tooltip) but this test is just on the NxTextInput

PR #5, opened 2020-04-08
NxFilter input displays incorrectly when set to box-sizing: border-box

Blacktiger:
In NXRM, the React components are children of an ExtJS element that sets everything to `box-sizing: border-box`. My understanding is that we'd like all components to be sized that way, but rather than make a bunch of styling changes here, I'm just fixing the styling to the box-sizing value that works.


rpokorny:
Styles that are needed to override other frameworks that you are using in your app should be implemented in said app.  For example, IQ uses bootstrap, and bootstrap sometimes messes up the styling of RSC components.  We fix that in a stylesheet in IQ, not here, in order to keep the RSC library styles clean.  RSC of course cannot predict every framework it might be used with and the complications that they might cause.

Blacktiger:
> Styles that are needed to override other frameworks that you are using in your app should be implemented in said app. For example, IQ uses bootstrap, and bootstrap sometimes messes up the styling of RSC components. We fix that in a stylesheet in IQ, not here, in order to keep the RSC library styles clean. RSC of course cannot predict every framework it might be used with and the complications that they might cause.
I currently have an override in nxrm for this. However, according to @glenhunter the IQ styles should be designed to work with `box-sizing: border-box`. I think it's worth having the component styles do the right thing regardless of the environment they find themselves in.

rpokorny:
Alright, after talking more we've decided that this is the best way to go after all.  The goal going forward is that any RSC style that only works right with a particular `box-sizing` setting should explicitly specify that setting.

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -9,6 +9,10 @@
 
 $nx-filter-input-height: 20px;
 
+.nx-filter-input, .nx-filter-input * {
```
glenhunter, line 12: This could also be added to the style below via:
`&, & * {`
this has the benefit of keeping the styles grouped.

glenhunter:
```
.nx-filter-input {

  display: inline-block;
  vertical-align: middle;
  white-space: nowrap;

  &, & * {
    box-sizing: content-box;
  }
}
```

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -9,6 +9,10 @@
...
+  box-sizing: content-box;
+}
+
 .nx-filter-input {
   @include container-horizontal;
```
rpokorny: Selecting more precisely:

```suggestion
.nx-filter-input {
  @include container-horizontal;

  > * {
    box-sizing: content-box;
  }
```

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -15,6 +15,10 @@ $nx-filter-input-height: 20px;
   display: inline-block;
   vertical-align: middle;
   white-space: nowrap;
+  
+  &, & > * {
```
rpokorny, line 19: I don't believe the first `&` is necessary; only the children have heights

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -15,6 +15,10 @@ $nx-filter-input-height: 20px;
   display: inline-block;
   vertical-align: middle;
   white-space: nowrap;
+  
+  & > * {
+    box-sizing: content-box !important;
```
Blacktiger, line 20: Needed the !important here to ensure that the correct box-sizing overrides a different box-sizing regardless of selector specificity.

gallery/src/components/NxFilterInput/NxFilterInputPage.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ .nx-filter-input-examples * {
+  box-sizing: border-box;
```
Blacktiger, line 8: Override the box-sizing to demonstrate the problem NxFilterInput has when the box-sizing is set incorrectly. This lets me verify the fix while also helping us identify the problem in the future if the box-sizing settings are messed with.

lib/src/components/NxFilterInput/NxFilterInput.scss
rpokorny: > Needed the !important here to ensure that the correct box-sizing overrides a different box-sizing regardless of selector specificity.

I'm not sure we want to go this far.  We really, really don't like `!important`

glenhunter: > Without important, the example that has a globally set `box-sizing: border-box` doesn't work correctly. This should be safe to do here since NxFilterInput cannot have children other than the ones it renders.

Yeah, I'm with Ross on this, we should never use `!important`. How exactly is the global style in NXRM implemented?

glenhunter: To elaborate, IMO adding the box-sizing attribute to this element makes sense because it's necessary for the element to display correctly (and there's a small amount of debate about even that). Adding `!important` though isn't necessary for the component to display correctly everywhere, it's necessary to get the component to display correctly *in a specific product*. In such a case the override belongs in that specific product, not in RSC.

rpokorny: > Without important, the example that has a globally set box-sizing: border-box doesn't work correctly. This should be safe to do here since NxFilterInput cannot have children other than the ones it renders.

I'd be in favor of specifying the `box-sizing` explicitly on the elements that need it, and using the parent class as well to get a higher specificity.  For example `.nx-filter-input .nx-filter-input__addon` beats `.x-body *`

Blacktiger: That makes more sense and should alleviate the need for `!important`. The style in nxrm is basically the exact same `.someclass * { box-sizing: border-box }`.

Blacktiger: Ok, I made another change. I moved the child classes (`.nx-filter-input__add-on`, etc) inside the parent class `.nx-filter-input` and put the two necessary box-sizing declarations on the appropriate child classes. This makes them more-specific than the global override so they apply without the need for `!important`.

glenhunter: > That makes more sense and should alleviate the need for `!important`. The style in nxrm is basically the exact same `.someclass * { box-sizing: border-box }`.

This is a better solution than !important

gallery/src/components/NxFilterInput/NxFilterInputPage.scss
rpokorny: > Override the box-sizing to demonstrate the problem NxFilterInput has when the box-sizing is set incorrectly.

Maybe include something about this as a code comment

gallery/src/components/NxFilterInput/NxFilterInputPage.scss
```diff
@@ -0,0 +1,10 @@
...
+ }
+
```
rpokorny, line 10: Missing final newline

gallery/src/components/NxFilterInput/NxFilterInputPage.scss
glenhunter: I'm not sure that I understand why we have this? We're demonstrating how to break the component?

glenhunter: > > Override the box-sizing to demonstrate the problem NxFilterInput has when the box-sizing is set incorrectly.
> 
> Maybe include something about this as a code comment

I don't think this is necessary

glenhunter: Also the indentation is screwed up

gallery/src/components/NxFilterInput/NxFilterInputPage.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-filter-input-examples * {
+ box-sizing: border-box;
```
glenhunter:
```suggestion
  box-sizing: border-box;
```

gallery/src/components/NxFilterInput/NxFilterInputPage.scss
Blacktiger: Well, this demonstrates the component in a broken state when the styling is incorrect. Since we mentioned that we should expect components to work with `box-sizing: border-box` I needed a way to validate that the component worked correctly in that environment. I can either remove this style, or I could add a global style that affects all components in the gallery.

PR #6, opened 2020-04-08
NEXUS-22079 NxTable Components

Blacktiger:
Screenshots of the current examples:
Screen Shot 2020-04-08 at 3 47 57 PM

Screen Shot 2020-04-08 at 3 48 03 PM

Screen Shot 2020-04-08 at 3 48 22 PM

Screen Shot 2020-04-08 at 3 48 31 PM

Screen Shot 2020-04-08 at 3 48 39 PM


lib/src/components/NxTable/NxTable.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {Props} from './types';
+export {Props} from './types';
+
+import './NxTable.scss';
+
+const NxTable = function NxTableElement(props: Props) {
+  const {className, children, ...attrs} = props;
+  return (
+    <table className={classnames('nx-table', className)} {...attrs}>
+      {children}
```
rpokorny, line 19: There's actually no need to pull children out separately.  If you left them out of the destructure above, so they were still in `attrs`, then just `<table className={classnames('nx-table', className)} {...attrs} />` would work

Blacktiger: Good point, I can simplify the components a bit that way.

lib/src/components/NxTableBody/NxTableBody.tsx
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react';
+
+import {Props} from './types';
+export {Props} from './types';
+
+const NxTableBody = function NxTableBody(props: Props) {
+  const {className, ...attrs} = props;
+
+  return (
+    <tbody className={className} {...attrs} />
```
rpokorny, line 16: If you aren't setting any additional classnames there's no need to pull out the `className` prop.  In fact, I'd argue there's no reason to have this component

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const NxTableCell = function NxTableCell(props: Props) {
+  const {
+    isHeader = false,
+    isEmpty = false,
+    isError = false,
+    isNumeric = false,
+    isSortable = false,
+    hasIcon = false,
+    sortDir, className, children, ...attrs} = props;
```
rpokorny, line 24: This would be our typical formatting:
```suggestion
  const {
      isHeader = false,
      isEmpty = false,
      isError = false,
      isNumeric = false,
      isSortable = false,
      hasIcon = false,
      sortDir, 
      className, 
      children, 
      ...attrs
  } = props;
```

lib/src/components/NxTableBody/NxTableBody.tsx
Blacktiger: I guess the only argument I have for keeping this component is consistency. Without this component you end up with stuff like this.

```
<NxTable>
  <NxTableHead>
    ...
  </NxTableHead>
  <tbody>
    ...
  </tbody>
</NxTable>
```

I'll go ahead and remove the `NxTableBody` component.

gallery/src/components/NxTable/NxTableClickableExample.tsx
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
```
joedragons, line 2: I assume this is auto generated but should we be 2020 now?=)  Just noticed randomly, definitely not a blocker.

Blacktiger: That's why it's `-present`. I just copy-pasted the license header and I assume our license check would fail if I dared to change it.

joedragons: I guess I assumed that present covered us for all time but I am skeptical we'd be able to copyright new code in the past without someone legal saying that.
You're right tho, needs to pass check, so beyond scope...just stuck out in my mind :)

lib/src/components/NxTableBody/NxTableBody.tsx
rpokorny: > I'll go ahead and remove the NxTableBody component.

I mean, consistency is potentially a valid argument

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import classnames from 'classnames';
+import { faSort, faSortDown, faSortUp } from '@fortawesome/free-solid-svg-icons';
+
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+
+import { Props } from './types';
+export {Props} from './types';
```
raveneyex, line 14: Minor: unbalanced parenthesis spacing. 

```suggestion
export { Props } from './types';
```

dawidsawa:
Would having an alternating row colours be considered responsibility of this component or that's a product level decision? Same for handling an empty state - no data to show, but no error either.

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (sortDir === 'asc') {
+      sortIcon = <NxFontAwesomeIcon icon={faSortUp} fixedWidth/>;
+    }
+    else if (sortDir === 'desc') {
+      sortIcon = <NxFontAwesomeIcon icon={faSortDown} fixedWidth/>;
+    }
+    else {
+      sortIcon = <NxFontAwesomeIcon icon={faSort} fixedWidth/>;
+    }
+  }
```
raveneyex, line 49: I think we could reduce code duplication in these lines:

```
let sortIcon = faSort;
if (sortDir === 'asc') {
  sortIcon = faSortUp;
}
else if (sortDir === 'desc') {
  sortIcon = faSortDown;
}

const sortComponent = isSortable ? <NxFontAwesomeIcon icon={faSort} fixedWidth/> : null;
```

lib/src/components/NxTable/NxTable.tsx
```diff
@@ -0,0 +1,22 @@
...
+import './NxTable.scss';
+
+const NxTable = function NxTableElement(props: Props) {
+  const {className, ...attrs} = props;
+  return (
+    <table className={classnames('nx-table', className)} {...attrs} />
+  );
+};
+
+export default NxTable;
```
raveneyex, line 22: You're missing the `propTypes` definition  — we usually do both `Props` and `propTypes` in case the consuming project isn't using TS.

lib/src/components/NxTableHead/NxTableHead.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { Props } from './types';
+export {Props} from './types';
+
+const NxTableHead = function NxTableHead(props: Props) {
+  const {className, children, ...attrs} = props;
+
+  return (
+    <thead className={className} {...attrs}>
+      {addPropsToChildren(children, {isHeader: true})}
```
raveneyex, line 19: What's the use-case for this `addPropsToChildren` fn?

Blacktiger: Both `NxTableHead` and `NxTableRow` use it to do automatic prop-drilling for the `isHeader` prop. This means you just use `NxTableCell` without having to pass the isHeader prop to it to get the correct css classes applied. I added that function to reduce the duplicated code and it felt potentially useful in other situations where we want a component to apply arbitrary props to children.

lib/src/components/NxTable/NxTable.tsx
Blacktiger: Good catch, I'll added `propTypes`.

lib/src/components/NxTableCell/NxTableCell.tsx
Blacktiger: Good point, I've simplified the component.

gallery/src/pageConfig.ts
```diff
@@ -86,6 +87,7 @@ const pageConfig: PageConfig = {
     NxFontAwesomeIcon: NxFontAwesomeIconPage,
     NxVulnerabilityDetails: NxVulnerabilityDetailsPage,
     NxSubmitMask: NxSubmitMaskPage,
+    NxTablePage: NxTableComponentPage,
```
rpokorny, line 90:
```suggestion
    NxTable: NxTableComponentPage,
```

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -0,0 +1,54 @@
...
+const tableSortableExample = require('!!raw-loader!./NxTableSortableExample').default;
+const tableLoadingExample = require('!!raw-loader!./NxTableLoadingExample').default;
+const tableErrorExample = require('!!raw-loader!./NxTableErrorExample').default;
+
+export default function NxTablePage() {
+  return (
+    <>
+      <GalleryDescriptionTile>
+        <p>
+          NxTable is the top-level component to use when displaying tables of data.
```
rpokorny, line 29: The description should be fleshed out a lot more.  Some sort of listing of each of the components and its props would be in line with what we've done for other components

lib/src/components/NxTable/NxTable.scss
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../base-styles/nx-globals';
+@import '../../base-styles/nx-tables';
```
rpokorny, line 8: These imports will result in the `nx-table` styles being present twice in the output css.  Given the way the RSC styles are intended to work, I think you can simply get rid of this file

lib/src/components/NxTableBody/NxTableBody.tsx
rpokorny: > > I'll go ahead and remove the NxTableBody component.
>
> I mean, consistency is potentially a valid argument

I really did see the value of the consistency, maybe even liked it better that way.  What do others think? @sonatype/iq-laurel

gallery/src/components/NxTable/NxTableErrorExample.tsx
```diff
@@ -0,0 +1,42 @@
...
+          <NxTableCell>Header 1</NxTableCell>
+          <NxTableCell>Header 2</NxTableCell>
+          <NxTableCell>Header 3</NxTableCell>
+          <NxTableCell isNumeric>Header 4</NxTableCell>
+          <NxTableCell>Header 5</NxTableCell>
+        </NxTableRow>
+      </NxTableHead>
+      <tbody>
+        <NxTableRow>
+          <NxTableCell isError colSpan={5}>
```
rpokorny, line 32: I wonder if `NxTable` could have an interface similar to `LoadWrapper`, where it is given information about when it is loading or in error.  Then it could automatically (and consistently) display the error row and loading row, computing their `colSpan` based on the number of header cells

gallery/src/components/NxTable/NxTableSortableExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+const NxTableSortableExample = () => {
+  const [rows, setRows] = useState(initialState);
+  const [sortDir, setSortDir] = useState<'desc' | 'asc' | null>(null);
+
+  function sort() {
+    if (sortDir === null) {
+      setSortDir('asc');
+      setRows(rows.slice().sort((a, b) => {
+        return a > b ? 1 : -1;
+      }));
```
rpokorny, line 27:
```suggestion
      setRows(rows.slice().sort((a, b) => a - b));
```

gallery/src/components/NxTable/NxTableSortableExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+} from '@sonatype/react-shared-components';
+
+const initialState = ['A', '1', 'B', '2', 'C', '3', 'D', '4'];
+
+const NxTableSortableExample = () => {
+  const [rows, setRows] = useState(initialState);
+  const [sortDir, setSortDir] = useState<'desc' | 'asc' | null>(null);
+
+  function sort() {
+    if (sortDir === null) {
```
rpokorny, line 23: I think the logic of the null -> asc -> desc -> null -> ... ordering of the sort button cycling should be encapuslated into the RSC lib in some way.  Seems like a candidate for a "stateHelper" as patterned in `NxTextInput`

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+import React from 'react';
+import { shallow } from 'enzyme';
+import NxTable from '../NxTable';
+
+describe('NxTable', function() {
+  it('renders a table with the expected class names', function() {
+    const component = shallow(<NxTable className="test" />);
+
+    expect(component).toMatchSelector('table.nx-table.test');
+  });
```
rpokorny, line 16: Test pass-through of some other props

lib/src/components/NxTable/types.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import {TableHTMLAttributes} from 'react';
+
+// Final Props are the HTMLProps & our re-definitions
```
rpokorny, line 9: Comment copied from another file?  There aren't any redefinitions here

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,63 @@
...
+    if (sortDir === 'asc') {
+      sortIcon = faSortUp;
+    }
+    else if (sortDir === 'desc') {
+      sortIcon = faSortDown;
+    }
+    else {
+      sortIcon = faSort;
+    }
+  }
```
rpokorny, line 49: If you wanted to stick to immutable variables, this could be replaced with a lookup into an object that maps from sortDir to icon, along with a default fallback.

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const Tag = isHeader ? 'th' : 'td';
+
+  return (
+    <Tag className={classes} {...attrs}>
+      {children}
+      {sortIcon ? <NxFontAwesomeIcon icon={sortIcon} fixedWidth/> : null}
```
rpokorny, line 56:
```suggestion
      {sortIcon && <NxFontAwesomeIcon icon={sortIcon} fixedWidth/>}
```

lib/src/components/NxTableCell/__tests__/NxTableCell.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+ */
+import React from 'react';
+import { shallow } from 'enzyme';
+import NxTableCell from '../NxTableCell';
+import { faSort, faSortDown, faSortUp } from '@fortawesome/free-solid-svg-icons';
+import NxFontAwesomeIcon from '../../NxFontAwesomeIcon/NxFontAwesomeIcon';
+
+describe('NxTableCell', function () {
+  it('renders a table with the expected class names', function () {
+    expect(shallow(<NxTableCell className="test"/>)).toMatchSelector('td.nx-cell.test');
```
rpokorny, line 15: See the way we use `enzymeUtils#getShallowComponent` in most of other other tests to cut down on the repetetiveness of writing these `shallow` invocations

lib/src/components/NxTableHead/__tests__/NxTableHead.tsx
```diff
@@ -0,0 +1,22 @@
...
+describe('NxTableHead', function () {
+  it('sets the correct class names', function() {
+    expect(shallow(<NxTableHead className="test" />)).toMatchSelector('thead.test');
+  });
+
+  it('sets the isHeader property on children', function () {
+    const component = shallow(<NxTableHead><NxTableRow /></NxTableHead>);
+
+    expect(component.contains(<NxTableRow isHeader />)).toBe(true);
+  });
```
rpokorny, line 21: Test other props pass-through

rpokorny:
I notice that the sort icons here look a bit different in IQ: the non-active sort direction here disappears entirely rather than greying out.  I see that that's a product of you just using the `faSort`, `faSortUp`, and `faSortDown` icons directly.  Is that what we want @glenhunter @dturner32 ?  

Interestingly, there are fontawesome icons that would look essentially the way we have it in IQ, but they are in the premium duotone icon pack which we can't use here in RSC.

gallery/src/pageConfig.ts
Blacktiger: I did a bit of renaming to match the naming style we used for NxTextInput now, see if you like it better.

lib/src/components/NxTable/NxTable.scss
Blacktiger: If I do that, then I'll have to build this scss file in nxrm when I use these components which is fine. Outside the scope of this PR, but we're inconsistent in when I need to do that and when not. I'll remove this.

gallery/src/components/NxTable/NxTableErrorExample.tsx
Blacktiger: Yeah, I was thinking about this last night and I think it could be done to simplify things.

gallery/src/components/NxTable/NxTableSortableExample.tsx
Blacktiger: It may depend on the UI that uses the table. In one case you might want one column to always show the sort indicator and have the other columns indicated they are unsorted but sortable. In another case you might want the user to be able to clear all column sorting. Maybe we want to think about codifying a particular kind of UX though? We might want to think about adopting a library like react-table to provide that stateful behavior while keeping things pretty customizable. I though it better to avoid dealing with sorting logic for now. We can always come back and introduce something later.

lib/src/components/NxTable/types.ts
Blacktiger: It sure is. :P

lib/src/components/NxTableBody/NxTableBody.tsx
Blacktiger: It's easy enough to bring back if we think it's better.

Blacktiger:
> Would having an alternating row colours be considered responsibility of this component or that's a product level decision? Same for handling an empty state - no data to show, but no error either.

I don't think we currently have styles for striped tables. Settting the `TableRow` to isClickable does help to identify a row when you hover over it with the mouse due to the row getting a "highlighted" appearance.

lib/src/components/NxTableHead/NxTableHead.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const {children, ...attrs} = props;
+
+  return (
+    <thead {...attrs}>
+      {addPropsToChildren(children, {isHeader: true})}
+    </thead>
+  );
+};
+
+export default NxTableHead;
```
raveneyex, line 24: Missing propTypes declaration

Blacktiger: What would the propTypes even be? There aren't any special props for this component.

lib/src/components/NxTableCell/__tests__/NxTableCell.test.tsx
Blacktiger: done

lib/src/components/NxTableHead/__tests__/NxTableHead.tsx
Blacktiger: done

lib/src/components/NxTableCell/NxTableCell.tsx
Blacktiger: done

lib/src/components/NxTableCell/NxTableCell.tsx
Blacktiger: done

lib/src/components/NxTable/types.ts
Blacktiger: removed

lib/src/components/NxTableHead/NxTableHead.tsx
raveneyex: > What would the propTypes even be? There aren't any special props for this component.

We not only list `propTypes` for special props; any prop that the component takes should be there.
See `lib/src/components/NxTextInput/types.ts` as an example.

rpokorny:
For me, something very weird is going on with eslint on this branch.  It says there are multiple line-length violations in `NxTablePage.tsx`, even though there aren't, and some of the line numbers it gives are beyond the end of the actual file.

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+export default function NxTablePage() {
+  return (
+    <>
+      <GalleryDescriptionTile>
+        <h5>NxTable</h5>
+
+        <p>
+          The top-level component to use when displaying tables of data.
+          It can have <code>&lt;NxTableHead&gt;</code> and <code>&lt;NxTableBody&gt;</code> components as children.
```
rpokorny, line 40: Add `className=".nx-code"` to all `<code>` elements for consistent styling with other pages

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -0,0 +1,190 @@
...
+        </p>
+
+        <hr />
+
+        <h5>NxTableBody</h5>
```
rpokorny, line 45: Missing `NxTableHead`

gallery/src/components/NxTable/NxTableSortableExample.tsx
rpokorny: > In one case you might want one column to always show the sort indicator and have the other columns indicated they are unsorted but sortable.

I don't follow what you're describing here.

> We might want to think about adopting a library like react-table to provide that stateful behavior while keeping things pretty customizable. I though it better to avoid dealing with sorting logic for now. We can always come back and introduce something later.

To clarify, I wasn't suggesting that we encapsulate the actual sorting of the data, only the standard sequence of sort button states as you click on it

gallery/src/pageConfig.ts
```diff
@@ -24,7 +24,7 @@ import NxThreatBarPage from './components/NxThreatBar/NxThreatBarPage';
 import NxRadioPage from './components/NxRadio/NxRadioPage';
 import NxVulnerabilityDetailsPage from './components/NxVulnerabilityDetails/NxVulnerabilityDetailsPage';
 import NxListPage from './styles/NxList/NxListPage';
-import NxTablePage from './styles/NxTable/NxTablePage';
+import NxTableStylePage from './styles/NxTable/NxTableStylePage';
```
rpokorny, line 27: Don't forget to rename the `NxTablePage` within the `NxTableStylePage` file

lib/src/components/NxTableBody/NxTableBody.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <NxFontAwesomeIcon icon={faExclamationTriangle} fixedWidth/>
+        {error}
+      </NxTableCell>
+    </NxTableRow>
+  );
+
+  return (
+    <tbody {...attrs}>
+      {isLoading && loadingSpinnerRow}
+      {error && errorRow}
```
rpokorny, line 48:
```suggestion
      {error && !isLoading && errorRow}
```

lib/src/components/NxTableRow/__tests__/NxTableRow.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  it('sets the clickable class', function () {
+    expect(shallow(<NxTableRow isClickable />)).toMatchSelector('tr.nx-clickable');
+  });
+
+  it('passes through arbitrary props', function() {
+    expect(shallow(<NxTableRow id="test" />)).toMatchSelector('#test');
+  });
+
+  it('sets the isHeader property on children', function () {
+    expect(shallow(<NxTableRow isHeader><NxTableCell /></NxTableRow>)).toContainReact(<NxTableCell isHeader />);
```
rpokorny, line 30: Test when `isHeader` isn't set as well

lib/src/util/childUtil.ts
```diff
@@ -0,0 +1,27 @@
...
+export function addPropsToChildren(children: ReactNode, props: object) {
+  return React.Children.map(children, (child: ReactNode) => {
+    if (React.isValidElement(child)) {
+      return React.cloneElement(child, props);
+    }
+    return child;
+  });
+}
+
+export function only(children: ReactNode, type: React.FC): React.ReactElement | null {
```
rpokorny, line 19: A comment, or perhaps more descriptive name, would be helpful here

rpokorny, line 19: Is `React.FC` the right type here?  I would've thought it would be `React.ComponentType`

gallery/src/components/NxTable/NxTablePage.tsx
Blacktiger: fixed

gallery/src/components/NxTable/NxTableSortableExample.tsx
Blacktiger: I see now, though I'm not convinced it's going to be all that useful. If you're ok with shipping this without I'd prefer to.

gallery/src/pageConfig.ts
Blacktiger: done

lib/src/components/NxTable/__tests__/NxTable.test.tsx
Blacktiger: done

lib/src/components/NxTableRow/__tests__/NxTableRow.test.tsx
Blacktiger: done

lib/src/util/childUtil.ts
Blacktiger: I'll switch to React.ComponentType

Blacktiger: Added jsdoc, any suggestions on a better name to use?

lib/src/components/NxTableBody/NxTableBody.tsx
raveneyex: > I really did see the value of the consistency, maybe even liked it better that way. What do others think? @sonatype/iq-laurel

I vote for consistency. The example above without this component looks really weird.

raveneyex:
@Blacktiger when answering a comment could you please quote it and link the commit(s) that answer it?
It makes for a much more easy review.

lib/src/components/NxTableRow/__tests__/NxTableRow.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it('does not set the isHeader property on children', function () {
```
raveneyex, line 33: Better spec name.

```suggestion
  it('does not set the isHeader property on children if prop is not supplied', function () {
```

lib/src/components/NxTableBody/NxTableBody.tsx
Blacktiger: I added it back for the loading/error functionality anyway.

lib/src/components/NxTable/NxTable.scss
rpokorny: > Outside the scope of this PR, but we're inconsistent in when I need to do that and when not.

Where are we inconsistent about it?  We do have consistent rules about it as described in the README.

The two use cases that we aim to best support are:
1. Those that are having their styles built automatically based on the imports they are using.  It is possible that this currently doesn't as efficiently as intended, due to the way side-effect imports work, importing `index.ts` might be bringing in all component styles even those for components that aren't being used.
2. Those that just want to import all RSC styles as a precompiled css file, which we generate during the build.

If you are doing something in between either of those two options, then you'll have to work things out manually to some extent, but generally the idea would be that you (the consuming application) need to import all of the base styles and then the component-specific styles for all components that you are using.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
rpokorny: > done

I'm not seeing it

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  it('does not pass columns to NxTableBody', function () {
+    const component = shallow(
+      <NxTable>
+        <NxTableBody />
+      </NxTable>
+    );
+
+    expect(component.findWhere(child => child.type() === NxTableBody)).not.toHaveProp('columns');
```
rpokorny, line 44:
```suggestion
    expect(component.find(NxTableBody)).not.toHaveProp('columns');
```

lib/src/util/childUtil.ts
rpokorny: > Added jsdoc, any suggestions on a better name to use?

With the docs I think the name is fine

gallery/src/pageConfig.ts
rpokorny: > done

Doesn't seem to be

lib/src/components/NxTable/NxTable.scss
Blacktiger: Right, but I'm doing neither right now. I'm importing/exporting individual components which sometimes include their own styles and sometimes not. As you say though, we're not targeting that as a best practice so it's probably fine. It's been a little while since I've looked at it, I may soon be able to just import all the RSC styles now that I've started using an override file.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
Blacktiger: Passing children is now covered by the 'computes the columns in NxTableHead and passes them to NxTableBody' test. Though I suppose I didn't add one to check that the id is passed through.

Blacktiger: 88c798c840845469dec8d575fff01f160d84db4a

gallery/src/pageConfig.ts
Blacktiger: bfc30d6

lib/src/components/NxTable/NxTable.scss
rpokorny: > sometimes include their own styles and sometimes not.

Please elaborate on this.  Outside of the `base-styles`, I don't know of any styles required by components which the components themselves fail to pull in.  You should be pulling in all of the `base-styles` if you are using any part of RSC; that's  one of the assumptions we make.

gallery/src/components/NxTable/NxTableSortableExample.tsx
rpokorny: > I see now, though I'm not convinced it's going to be all that useful. If you're ok with shipping this without I'd prefer to.

What do other think? @sonatype/shared-components-group

glenhunter:
> I notice that the sort icons here look a bit different in IQ: the non-active sort direction here disappears entirely rather than greying out. I see that that's a product of you just using the `faSort`, `faSortUp`, and `faSortDown` icons directly. Is that what we want @glenhunter @dturner32 ?
> 
> Interestingly, there are fontawesome icons that would look essentially the way we have it in IQ, but they are in the premium duotone icon pack which we can't use here in RSC.

As long as @dturner32 is OK with the selected icons I am. I do however think that we should always show but up and down arrows with some indication of which is selected.

glenhunter:
> > I notice that the sort icons here look a bit different in IQ: the non-active sort direction here disappears entirely rather than greying out. I see that that's a product of you just using the `faSort`, `faSortUp`, and `faSortDown` icons directly. Is that what we want @glenhunter @dturner32 ?
> > Interestingly, there are fontawesome icons that would look essentially the way we have it in IQ, but they are in the premium duotone icon pack which we can't use here in RSC.
> 
> As long as @dturner32 is OK with the selected icons I am. I do however think that we should always show but up and down arrows with some indication of which is selected.

To clarify: we should always show both and up and down sort icons. The selected sort state should be differentiated (we usually do this by colour).

Blacktiger:
> > > I notice that the sort icons here look a bit different in IQ: the non-active sort direction here disappears entirely rather than greying out. I see that that's a product of you just using the `faSort`, `faSortUp`, and `faSortDown` icons directly. Is that what we want @glenhunter @dturner32 ?
> > > Interestingly, there are fontawesome icons that would look essentially the way we have it in IQ, but they are in the premium duotone icon pack which we can't use here in RSC.
> > 
> > 
> > As long as @dturner32 is OK with the selected icons I am. I do however think that we should always show but up and down arrows with some indication of which is selected.
> 
> To clarify: we should always show both and up and down sort icons. The selected sort state should be differentiated (we usually do this by colour).

The font awesome icons don't really have a way to do that. If we want to do that we probably don't want to use fontawesome? Instead we should probably be using other styles.

lib/src/components/NxTable/NxTable.scss
Blacktiger: The NxButton component doesn't pull in any styles by just importing it. I need to look at just importing all the styles by default now to see if it causes any problems as it's been a while since I tried it.

rpokorny: > The NxButton component doesn't pull in any styles by just importing it. I need to look at just importing all the styles by default now to see if it causes any problems as it's been a while since I tried it.

Note that the styles that `NxButton` relies on, `nx-btn`, have their own page in the gallery.  They are considered base styles.  This is the same situation here with `NxTable` and `nx-table`.

gallery/src/components/NxTable/NxTableClickableExample.tsx
mpiggott: I brought this up many years ago but its a limitation in the maven license plugin, so we typically use the inception of the project in the header instead of the creation of the code/file.

Blacktiger:
> > > > I notice that the sort icons here look a bit different in IQ: the non-active sort direction here disappears entirely rather than greying out. I see that that's a product of you just using the `faSort`, `faSortUp`, and `faSortDown` icons directly. Is that what we want @glenhunter @dturner32 ?
> > > > Interestingly, there are fontawesome icons that would look essentially the way we have it in IQ, but they are in the premium duotone icon pack which we can't use here in RSC.
> > > 
> > > 
> > > As long as @dturner32 is OK with the selected icons I am. I do however think that we should always show but up and down arrows with some indication of which is selected.
> > 
> > 
> > To clarify: we should always show both and up and down sort icons. The selected sort state should be differentiated (we usually do this by colour).
> 
> The font awesome icons don't really have a way to do that. If we want to do that we probably don't want to use fontawesome? Instead we should probably be using other styles.
Actually, it turns out I can use the "masking" stuff from fontawesome to create a sort icon that highlights the sort direction.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <NxTableHead>
+          <NxTableRow>
+            <NxTableCell />
+          </NxTableRow>
+        </NxTableHead>
+        <NxTableBody />
+      </NxTable>
+    );
+
+    expect(component.findWhere(child => child.type() === NxTableBody)).toHaveProp('columns', 1);
```
rpokorny, line 34:
```suggestion
    expect(component.find(NxTableBody)).toHaveProp('columns', 1);
```

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it('does not pass columns to NxTableBody', function () {
```
rpokorny, line 37: Is this more accurate?

```suggestion
  it('does not pass columns to NxTableBody when there is no NxTableHead', function () {
```

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+  else {
+    maskedSort = <NxFontAwesomeIcon icon={faSort} />;
+  }
+
+  const Tag = isHeader ? 'th' : 'td';
+
+  return (
+    <Tag className={classes} {...attrs}>
+      {children}
```
rpokorny, line 64: I recommend wrapping the `children` in a call to reactUtils#ensureElement

rpokorny:
> Actually, it turns out I can use the "masking" stuff from fontawesome to create a sort icon that highlights the sort direction.

Nice solution!

lib/src/components/NxTableCell/NxTableCell.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <span className="nx-cell__sort-icons fa-layers fa-fw">
```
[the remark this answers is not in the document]
Blacktiger: Probably, it does reserve a small 15px area otherwise. 17701191f19786490029f899abe295e498e395a8

lib/src/components/NxTableCell/NxTableCell.tsx
Blacktiger: 3d5cbe4

gallery/src/components/NxTable/NxTableSortableExample.tsx
raveneyex: > If you're ok with shipping this without I'd prefer to.

I'm fine with not shipping this in this PR, as long as we don't lose track of it. Maybe create a ticket to fllow up on this?

PR #7, opened 2020-04-14
Fix open-sans issues RSC-87 

glenhunter:
https://issues.sonatype.org/browse/RSC-87

Fix a few minor font issues that are also appearing in IQ, mostly descenders that are cut off due to changes in line-height, but also vertical alignment in nx-counter

### DIFF
--- a/lib/src/base-styles/_nx-counter.scss
+++ b/lib/src/base-styles/_nx-counter.scss
@@ -15,9 +15,9 @@
   color: $nx-text-color-light;
   display: inline-block;
   font-size: 11px;
-  line-height: 1.3; // vertically centres text in pill
+  line-height: 1.4; // vertically centres text in pill
   margin-left: 8px;
-  padding: 1px 4px;
+  padding: 0 4px 1px 4px;
   text-align: center;
   min-width: 8px;
 
--- a/lib/src/base-styles/_nx-counter.scss
+++ b/lib/src/base-styles/_nx-counter.scss
@@ -15,7 +15,7 @@
   color: $nx-text-color-light;
   display: inline-block;
   font-size: 11px;
-  line-height: 1.4; // vertically centres text in pill
+  line-height: 1.3; // vertically centres text in pill
   margin-left: 8px;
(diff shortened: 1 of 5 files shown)


lib/src/base-styles/_nx-counter.scss
```diff
@@ -15,7 +15,7 @@
   color: $nx-text-color-light;
   display: inline-block;
   font-size: 11px;
-  line-height: 1.4; // vertically centres text in pill
+  line-height: 1.3; // vertically centres text in pill
```
rpokorny, line 18: Could you clarify the need for this change.  I see what it does (makes the pill shorter) but I'm not seeing why it's noticeably better or worse this way

glenhunter: > Could you clarify the need for this change. I see what it does (makes the pill shorter) but I'm not seeing why it's noticeably better or worse this way

That was a crappy solution, this is a better one: https://github.com/sonatype/sonatype-react-shared-components/pull/7/commits/4f7960f1e18e9a82d4025e3e0c95e430741773c4

rpokorny: > That was a crappy solution, this is a better one: 4f7960f

Is it though?  Either way you're changing the actual height of the pill, not just "centering" the text.  I've been digging around in fontforge and as far as I can tell, opensans simply tends to have its characters a bit low in their box.  I haven't come up with a great solution though.

glenhunter: > > That was a crappy solution, this is a better one: [4f7960f](https://github.com/sonatype/sonatype-react-shared-components/commit/4f7960f1e18e9a82d4025e3e0c95e430741773c4)
> 
> Is it though? Either way you're changing the actual height of the pill, not just "centering" the text. I've been digging around in fontforge and as far as I can tell, opensans simply tends to have its characters a bit low in their box. I haven't come up with a great solution though.

Yeah you're right, though removing the top padding does *look* better than changing the `line-height`.

PR #10, opened 2020-04-20
CLM-14538 Add Esc button support to NxModal

jlopez-sonatype:


raveneyex:
Your branch is out of sync — you need merge from master.

jlopez-sonatype:
> Your branch is out of sync — you need merge from master.

Updated from master on 1a3a7fc

lib/src/components/NxModal/NxModal.tsx
```diff
@@ -4,22 +4,41 @@
...
+      onClose();
+    }
+  };
+
+  useEffect(function() {
+    document.addEventListener('keydown', modalCloseListener);
+
+    currentModalCloseHandlers.push(onClose);
+
+    return function() {
```
raveneyex, line 31: Give a name to this function  — it'll be easier to identify in case an error ever shows up.

lib/src/components/NxModal/types.ts
```diff
@@ -4,11 +4,16 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import {WeakValidationMap} from 'react';
+import React, {WeakValidationMap} from 'react';
```
raveneyex, line 7:
```suggestion
import { HTMLAttributes, WeakValidationMap } from 'react';
```

lib/src/components/NxModal/types.ts
```diff
@@ -4,11 +4,16 @@
...
 import * as PropTypes from 'prop-types';
 
-export type Props = React.HTMLAttributes<HTMLDivElement>;
+export type CloseHandler = () => void;
+
+export type Props = React.HTMLAttributes<HTMLDivElement> & {
```
raveneyex, line 12:
```suggestion
export type Props = HTMLAttributes<HTMLDivElement> & {
```

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +38,50 @@ describe('NxModal', function() {
...
+      );
+
+      act(() => {
+        mount(nxModal, {attachTo: container});
+      });
+
+      document.dispatchEvent(new KeyboardEvent('keydown', {key: 'Escape'}));
+      expect(mockCallBack).toHaveBeenCalled();
+    });
+
```
raveneyex, line 73: No test for the stacked-modals scenario?

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +38,50 @@ describe('NxModal', function() {
...
+          <NxModal id="modal-id" onClose={mockCallBack}/>
+        </div>
+      );
+
+      act(() => {
+        mount(nxModal, {attachTo: container});
+      });
+
+      document.dispatchEvent(new KeyboardEvent('keydown', {key: 'Escape'}));
+      expect(mockCallBack).toHaveBeenCalled();
```
raveneyex, line 71: Prefer the matcher `toHaveBeenCalledWith()` for more control.

lib/src/components/NxModal/types.ts
```diff
@@ -4,11 +4,16 @@
...
+  onClose: CloseHandler;
+};
 
 export const propTypes: WeakValidationMap<Props> = {
-  className: PropTypes.string
+  className: PropTypes.string,
+  onClose: PropTypes.func.isRequired
```
raveneyex, line 18: I'm not sure if this prop should be mandatory. What do others think?

lib/src/components/NxModal/NxModal.tsx
jlopez-sonatype: > Give a name to this function — it'll be easier to identify in case an error ever shows up.

Named on bf2a3cf

raveneyex: > > Give a name to this function — it'll be easier to identify in case an error ever shows up.
> 
> Named on [bf2a3cf](https://github.com/sonatype/sonatype-react-shared-components/commit/bf2a3cf46face787fbfc956b2057abe89243d65f)

You could've named it inline:

```
useEffect(() => {
    ...
    return function cleanup() { ... }
});
```

lib/src/components/NxModal/types.ts
raveneyex: > I'm not sure if this prop should be mandatory. What do others think?

Per wall-walk discussion closing this topic. We do want it to be mandatory.

lib/src/components/NxModal/types.ts
jlopez-sonatype: > Minor: this suggestion along with the one below.

Better import strategy suggested: c08c0a4

lib/src/components/NxModal/types.ts
jlopez-sonatype: > 

Better import strategy suggested: c08c0a4

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > Prefer the matcher `toHaveBeenCalledWith()` for more control.

Since the function doesn't accept parameters (because it's the modalCloseHandler, not the actual listener that it's wired on the modal) I've used instead the `toHaveBeenNthCalledWith` to ensure it was only called once and with no parameters

raveneyex: > > Prefer the matcher `toHaveBeenCalledWith()` for more control.
> 
> Since the function doesn't accept parameters (because it's the modalCloseHandler, not the actual listener that it's wired on the modal) I've used instead the `toHaveBeenNthCalledWith` to ensure it was only called once and with no parameters: [0965fe5](https://github.com/sonatype/sonatype-react-shared-components/commit/0965fe5f2debfca69a24523bb35fa7d5fffaa13f)

Prefer `toHaveBeenCalledTimes(1)` instead. 
`toHaveBeenNthCalledWith` is used to check the parameters supplied to the function on an nth call. It only works in your test because you're not supplying the fn with the rest of parameters.
See https://jestjs.io/docs/en/expect#tohavebeennthcalledwithnthcall-arg1-arg2-

lib/src/components/NxModal/NxModal.tsx
```diff
@@ -4,22 +4,42 @@
...
 
 import './NxModal.scss';
 
-const NxModal: FunctionComponent<Props> = ({className, ...attrs}) => {
+const currentModalCloseHandlers: CloseHandler[] = [];
 
+const NxModal: FunctionComponent<Props> = ({className, onClose, ...attrs}) => {
   const modalClasses = classnames('nx-modal', className);
+
+  const modalCloseListener = ({ key }: { key: string}) => {
```
rpokorny, line 19: The parameter type would be `KeyboardEvent` (the one from react, not the native one)

lib/src/components/NxModal/NxModal.tsx
```diff
@@ -4,22 +4,42 @@
...
+  };
+
+  const removeCloseHandlerListener = () => {
+    const idx = currentModalCloseHandlers.indexOf(onClose);
+    currentModalCloseHandlers.splice(idx, 1);
+    document.removeEventListener('keydown', modalCloseListener);
+  };
+
+  useEffect(function() {
+    document.addEventListener('keydown', modalCloseListener);
```
rpokorny, line 33: Why the deviation from my suggestion where the listener was registered outside of the `NxModal` invocation?  This needlessly registers it one time per modal instead of one time globally

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -5,12 +5,17 @@
...
-import NxModal, { Props } from '../NxModal';
-import {shallow} from 'enzyme';
+import {mount, shallow} from 'enzyme';
+import {act} from 'react-dom/test-utils';
+
+import NxModal, {Props} from '../NxModal';
+import {CloseHandler} from '../types';
 
 describe('NxModal', function() {
+  const dummyCloseHandler: CloseHandler = () => {};
```
rpokorny, line 15:
```suggestion
  const dummyCloseHandler = jest.fn();
```

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +38,50 @@ describe('NxModal', function() {
...
     expect(nxModal.find('.nx-modal').prop('data-foo')).toEqual('bar');
   });
+
+  describe('NxModal event listener support', () => {
+    let container: HTMLDivElement | null;
+
+    it('executes onClose method when pressing ESC key', function () {
+      const mockCallBack = jest.fn();
+      const nxModal = (
+        <div>
```
rpokorny, line 61: Is this div necessary

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +38,50 @@ describe('NxModal', function() {
...
+          <NxModal id="modal-id" onClose={mockCallBack}/>
+        </div>
+      );
+
+      act(() => {
+        mount(nxModal, {attachTo: container});
+      });
+
+      document.dispatchEvent(new KeyboardEvent('keydown', {key: 'Escape'}));
+      expect(mockCallBack).toHaveBeenNthCalledWith(1);
```
rpokorny, line 71: `.toHaveBeenCalledOnce()`?

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +38,50 @@ describe('NxModal', function() {
...
+    });
+
+    beforeEach(function () {
+      // Rendering container for the component in test.
+      container = document.createElement('div');
+      document.body.appendChild(container);
+    });
+
+    afterEach(function () {
```
rpokorny, line 80: `beforeEach` and `afterEach` should preferably go above the test itself

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: Changed validation of number of times called: d945d0a

jlopez-sonatype: > > > Prefer the matcher `toHaveBeenCalledWith()` for more control.
> > 
> > 
> > Since the function doesn't accept parameters (because it's the modalCloseHandler, not the actual listener that it's wired on the modal) I've used instead the `toHaveBeenNthCalledWith` to ensure it was only called once and with no parameters: [0965fe5](https://github.com/sonatype/sonatype-react-shared-components/commit/0965fe5f2debfca69a24523bb35fa7d5fffaa13f)
> 
> Prefer `toHaveBeenCalledTimes(1)` instead.
> `toHaveBeenNthCalledWith` is used to check the parameters supplied to the function on an nth call. It only works in your test because you're not supplying the fn with the rest of parameters.
> See https://jestjs.io/docs/en/expect#tohavebeennthcalledwithnthcall-arg1-arg2-

Changed validation of number of times called: d945d0a

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: Test created and refactored on:

jlopez-sonatype: > No test for the stacked-modals scenario?

Test created and refactored on:
d945d0a
8eff1da
Not sure if doing too many validations on a single test. Do you think it should be broken down into more simple ones?

lib/src/components/NxModal/NxModal.tsx
jlopez-sonatype: > Why the deviation from my suggestion where the listener was registered outside of the `NxModal` invocation? This needlessly registers it one time per modal instead of one time globally

Discussing the approach with @raveneyex he mentioned that might introduce a memory leak since it's always registered. With the current approach once there are no modals there are no more calls to the validation logic either which seems cleaner

lib/src/components/NxModal/NxModal.tsx
jlopez-sonatype: > The parameter type would be `KeyboardEvent` (the one from react, not the native one)

This change would prevent the function being added as a listener to the document. I can use the native one instead (the typescript interface defined in global) if that's allright.

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > `beforeEach` and `afterEach` should preferably go above the test itself

Moved utility functions to the top of the describe: 8eff1da

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > `.toHaveBeenCalledOnce()`?

I cannot find the definition of this method, only the `toHaveBeenCalled `which checks that it has been called at least once but not necessarily only once. I expected it to be available in the same object as `toHaveBeenNthCalledWith`, is it used in a different way?

lib/src/components/NxModal/NxModal.tsx
raveneyex: > my suggestion where the listener was registered outside of the `NxModal` invocation

Who would clean up such a listener once all the modals are closed?

lib/src/components/NxModal/__tests__/NxModal.test.tsx
raveneyex: > > `.toHaveBeenCalledOnce()`?
> 
> I cannot find the definition of this method, only the `toHaveBeenCalled `which checks that it has been called at least once but not necessarily only once. I expected it to be available in the same object as `toHaveBeenNthCalledWith`, is it used in a different way?

Try with `.toHaveBeenCalledTimes(1)` instead.

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > 

Changed dummy close handler into jest mock bb0d1b8

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > > > `.toHaveBeenCalledOnce()`?
> > 
> > 
> > I cannot find the definition of this method, only the `toHaveBeenCalled `which checks that it has been called at least once but not necessarily only once. I expected it to be available in the same object as `toHaveBeenNthCalledWith`, is it used in a different way?
> 
> Try with `.toHaveBeenCalledTimes(1)` instead.

Adressed on d945d0a from a previous comment

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > Is this div necessary

Div was not necessary. Refactored and removed creator function: 3f1b2a4

lib/src/components/NxModal/NxModal.tsx
rpokorny: > I can use the native one instead (the typescript interface defined in global) if that's allright.

Oh right, I was mistaken in telling you to use the react one.  The native one is the correct one here.

lib/src/components/NxModal/NxModal.tsx
rpokorny: > Who would clean up such a listener once all the modals are closed?

No one.  It is registered when the page first loads and it stays there until the browser tears everything down when the page is closed.  I am fine with this.

However, I do recognize that the ideal scenario would be to have zero listeners registered when there are no modals open, and exactly one listener registered when there are one or more modals open - a combination which neither of our solutions achieves.  I think it could be achieved fairly easily by putting some list-size-based conditions around the registering and unregistering of the listener in your impl.

lib/src/components/NxModal/__tests__/NxModal.test.tsx
rpokorny: > Try with .toHaveBeenCalledTimes(1) instead.

Ah `toHaveBeenCalledOnce` must only be a jasmine thing.

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +37,92 @@ describe('NxModal', function() {
...
+
+      if (containerSecondaryModal) {
+        document.body.removeChild(containerSecondaryModal);
+        containerSecondaryModal = null;
+      }
+    });
+
+    it('executes onClose method when pressing ESC key', function () {
+      const mockCallBack = jest.fn();
+      const nxModal = <NxModal id='first-modal-id' onClose={mockCallBack}/>;
```
rpokorny, line 81: Minor: we use double quotes in HTML and JSX

```suggestion
      const nxModal = <NxModal id="first-modal-id" onClose={mockCallBack}/>;
```

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +37,92 @@ describe('NxModal', function() {
...
+
+      document.dispatchEvent(new KeyboardEvent('keydown', {key: 'Escape'}));
+      expect(mockCallBack).toHaveBeenCalledTimes(1);
+    });
+
+    it('executes onClose method for each modal that has been opened, preserving order call', function () {
+      const firstMockCallBack = jest.fn();
+      const secondMockCallBack = jest.fn();
+      const firstNxModal = <NxModal id='first-modal-id' onClose={firstMockCallBack}/>;
+      const secondNxModal = <NxModal id='second-modal-id' onClose={secondMockCallBack}/>;
```
rpokorny, line 95:
```suggestion
      const firstNxModal = <NxModal id="first-modal-id" onClose={firstMockCallBack}/>;
      const secondNxModal = <NxModal id="second-modal-id" onClose={secondMockCallBack}/>;
```

jlopez-sonatype: > 

Use double quotes on JSX: a88dab1

lib/src/components/NxModal/__tests__/NxModal.test.tsx
jlopez-sonatype: > Minor: we use double quotes in HTML and JSX

Use double quotes on JSX: a88dab1

lib/src/components/NxModal/NxModal.tsx
raveneyex: > I think it could be achieved fairly easily by putting some list-size-based conditions around the registering and unregistering of the listener in your impl.

Can you elaborate on this? I don't quite follow.

jlopez-sonatype: > > I think it could be achieved fairly easily by putting some list-size-based conditions around the registering and unregistering of the listener in your impl.
> 
> Can you elaborate on this? I don't quite follow.

I think it refers to only registering the listener in the use effect if there is something on the closeHandler array and only deregistering it when there is nothing else on the array. I had tried something similar but couldn't get it to work at that time. I'm looking into the approach again. @rpokorny let me know if you had a different idea for this

jlopez-sonatype: > > > I think it could be achieved fairly easily by putting some list-size-based conditions around the registering and unregistering of the listener in your impl.
> > 
> > 
> > Can you elaborate on this? I don't quite follow.
> 
> I think it refers to only registering the listener in the use effect if there is something on the closeHandler array and only deregistering it when there is nothing else on the array. I had tried something similar but couldn't get it to work at that time. I'm looking into the approach again. @rpokorny let me know if you had a different idea for this

Improved the listener registration so it's only registered once per modal stacking: 463660d

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +37,92 @@ describe('NxModal', function() {
...
+
+      // Unmount second modal to remove it's listener from available listeners to execute
+      act(() => {
+        secondModalWrapper!.unmount();
+      });
+      // "Close" the first modal
+      document.dispatchEvent(new KeyboardEvent('keydown', {key: 'Escape'}));
+      expect(secondMockCallBack).toHaveBeenCalledTimes(2);
+      expect(firstMockCallBack).toHaveBeenCalledTimes(1);
+    });
```
rpokorny, line 126: I'd add one more esc press at the end here to assert that it doesn't call either callback an additional time when all the modals are already closed.

jlopez-sonatype: > I'd add one more esc press at the end here to assert that it doesn't call either callback an additional time when all the modals are already closed.

Additional assert for handler execution: 110be86

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -33,17 +37,102 @@ describe('NxModal', function() {
...
+      // Unmount the first modal to remove it's listener from available listeners to execute
+      act(() => {
+        firstModalWrapper!.unmount();
+      });
+      // Ensure there are no more calls to the registered close handlers with new keypresses
+      document.dispatchEvent(new KeyboardEvent('keydown', {key: 'Escape'}));
+      expect(secondMockCallBack).toHaveBeenCalledTimes(2);
+      expect(firstMockCallBack).toHaveBeenCalledTimes(1);
+    });
+  });
```
rpokorny, line 137: Oh, also test that keypresses other than ESC don't close the modal

jlopez-sonatype: > Oh, also test that keypresses other than ESC don't close the modal

Test handler is only called with Escape key and not pressing any other common keys: dad801c

PR #12, opened 2020-04-22
Bugfix empty error table body

Blacktiger:
If the error string is falsey, it would cause the following warning:

> Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <tbody>. Make sure you don't have any extra whitespace between tags on each line of your source code.

The NxLoadWrapper doesn't show the NxLoadError component if the error string is falsey so I want to keep that behavior here but hide the warning.

### DIFF
--- a/lib/src/components/NxTableBody/NxTableBody.tsx
+++ b/lib/src/components/NxTableBody/NxTableBody.tsx
@@ -45,7 +45,7 @@ const NxTableBody = function NxTableBody(props: Props) {
   return (
     <tbody {...attrs}>
       {isLoading && loadingSpinnerRow}
-      {error && !isLoading && errorRow}
+      {!!error && !isLoading && errorRow}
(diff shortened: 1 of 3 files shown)


lib/src/components/NxTableBody/NxTableBody.tsx
```diff
@@ -45,7 +45,7 @@ const NxTableBody = function NxTableBody(props: Props) {
   return (
     <tbody {...attrs}>
       {isLoading && loadingSpinnerRow}
-      {error && !isLoading && errorRow}
+      {!!error && !isLoading && errorRow}
```
Blacktiger, line 48: Convert the falsey error string to an actual boolean value here. Otherwise when it's an empty string this ends up rendering whitespace in the `<tbody>` and react gives us a warning.

PR #13, opened 2020-04-24
Refactor main header to match new design - RSC-82

glenhunter:
https://issues.sonatype.org/browse/RSC-82

Change Page Header design to match the latest dimensions from 8 Shapes as employed in IQ.

Any colour/logo changes for RSC are pending from the design group.

See https://issues.sonatype.org/browse/RSC-88 for creation of an RSC Page Header example page - there are some design considerations pending from the Design Group.


lib/src/base-styles/_nx-page-header.scss
```diff
@@ -8,65 +8,74 @@
 @import '../scss-shared/nx-text-helpers';
 @import '../scss-shared/nx-container-helpers';
 
+$main-header-background: $nx-sonatype-blue;
+$main-header-color: $nx-off-white;
+$main-header-highlight: $nx-sonatype-orange;
```
rpokorny, line 13: unused

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -8,65 +8,74 @@
...
+$main-header-height: 72px;
+$main-header-branding-height: 52px;
+$main-header-product-grid-rows: $main-header-branding-height / 2;
```
rpokorny, line 16: Is this right, considering the 2px gap?  Also would it be better called `$main-header-product-grid-rows`?

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -8,65 +8,74 @@
...
 }
 
-.nx-title__brand {
+.nx-product__name {
   @include semi-bold();
 
   -ms-grid-row-align: end;
   align-self: end;
-  font-size: 18px;
+  font-size: 23px;
```
rpokorny, line 77: Are you aware that the font size is larger than the line height here, and that the descenders stick out of the `<div>` and even down into the next row of the grid layout?

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -76,14 +85,17 @@
...
-  font-size: 12px;
+  font-size: 11px;
   -ms-grid-row: 3;
   -ms-grid-column: 3;
   grid-area: version;
   -ms-grid-column-align: left;
   justify-self: left;
-  line-height: 12px;
+  line-height: 11px;
+  padding-top: $nx-spacing-xxs;
```
rpokorny, line 100: This seems like a hack to mitigate the height issues of the box above

glenhunter: > This seems like a hack to mitigate the height issues of the box above

https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/ddbab71786239da9d3fc391fde51098a068df712

lib/src/base-styles/_nx-page-header.scss
glenhunter: > Are you aware that the font size is larger than the line height here, and that the descenders stick out of the `<div>` and even down into the next row of the grid layout?

https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/ddbab71786239da9d3fc391fde51098a068df712

lib/src/base-styles/_nx-page-header.scss
glenhunter: > Is this right, considering the 2px gap? Also would it be better called `$main-header-product-grid-rows`?

https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/ddbab71786239da9d3fc391fde51098a068df712

lib/src/base-styles/_nx-page-header.scss
glenhunter: > unused

https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/ddbab71786239da9d3fc391fde51098a068df712

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -10,10 +10,10 @@
 
 $main-header-background: $nx-sonatype-blue;
 $main-header-color: $nx-off-white;
-$main-header-highlight: $nx-sonatype-orange;
 $main-header-height: 72px;
-$main-header-branding-height: 52px;
-$main-header-product-grid-rows: $main-header-branding-height / 2;
+$main-header-product-height: 52px;
+$main-header-product-title-row-height: $main-header-product-height * .66;
+$main-header-product-version-row-height: $main-header-product-height * .33;
```
rpokorny, line 16:
```suggestion
$main-header-product-title-row-height: 2fr;
$main-header-product-version-row-height: 1fr;
```

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -41,61 +41,60 @@ $main-header-product-grid-rows: $main-header-branding-height / 2;
...
+  grid-template-rows: $main-header-product-title-row-height $main-header-product-version-row-height;
+  -ms-grid-columns: $main-header-product-height $nx-spacing-xxs auto;
+  grid-template-columns: $main-header-product-height auto;
   grid-template-areas:
     "logo title"
     "logo version";
-  height: $main-header-branding-height;
+  height: $main-header-product-height;
   margin-top: 10px;
   padding-right: 12px;
```
rpokorny, line 54: Is this right? Should it be margin instead?

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -41,61 +41,60 @@ $main-header-product-grid-rows: $main-header-branding-height / 2;
...
+  -ms-grid-rows: $main-header-product-title-row-height $main-header-product-version-row-height;
+  grid-template-rows: $main-header-product-title-row-height $main-header-product-version-row-height;
+  -ms-grid-columns: $main-header-product-height $nx-spacing-xxs auto;
+  grid-template-columns: $main-header-product-height auto;
   grid-template-areas:
     "logo title"
     "logo version";
-  height: $main-header-branding-height;
+  height: $main-header-product-height;
   margin-top: 10px;
```
rpokorny, line 53: Use a variable?

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -41,61 +41,60 @@ $main-header-product-grid-rows: $main-header-branding-height / 2;
...
   align-self: center;
   box-sizing: border-box;
-  -ms-grid-row: 1;
-  -ms-grid-row-span: 3;
-  -ms-grid-column: 1;
   grid-area: logo;
-  height: $main-header-branding-height;
-  width: $main-header-branding-height;
+  height: $main-header-product-height;
+  width: $main-header-product-height;
```
rpokorny, line 68:
```suggestion
  height: 100%;
  width: 100%;
```
(Only works correctly in conjunction with my `fr` suggestion above, otherwise half a pixel of height is lost to rounding)

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -41,61 +41,60 @@ $main-header-product-grid-rows: $main-header-branding-height / 2;
...
+  align-self: center;
   font-size: 23px;
-  text-transform: lowercase;
-  -ms-grid-row: 1;
-  -ms-grid-column: 3;
   grid-area: title;
   -ms-grid-column-align: left;
   justify-self: left;
-  line-height: 18px;
+  line-height: 100%;
```
rpokorny, line 83: Not that it matters for anything except the boxes as seen in the dev tools, but this seems to work best:

```suggestion
  line-height: normal;
```

lib/src/base-styles/_nx-page-header.scss
glenhunter: > Use a standard variable?

Neither 12px nor 8px looked right.

rpokorny: > Neither 12px nor 8px looked right.

Hmm, maybe margin isn't the right solution here anyway then.  How about `align-self: center`?

lib/src/base-styles/_nx-page-header.scss
glenhunter: > Is this right? Should it be margin instead?

Actually it's unnecessary: https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/c0c87699a0ea1b85f99c8cef03317a178d653ef1

lib/src/base-styles/_nx-page-header.scss
glenhunter: > > Neither 12px nor 8px looked right.
> 
> Hmm, maybe margin isn't the right solution here anyway then. How about `align-self: center`?

Ah, yes, that is a much better solution: https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/dbffb953b81bcfb70bf8e102cdeda9ecc7a59067

lib/src/base-styles/_nx-page-header.scss
rpokorny: > Actually it's unnecessary: c0c8769

Are you sure?  In an application with buttons to the right of this (like IQ), what maintains the space between the product name and the buttons?

glenhunter: > > Actually it's unnecessary: [c0c8769](https://github.com/sonatype/sonatype-react-shared-components/commit/c0c87699a0ea1b85f99c8cef03317a178d653ef1)
> 
> Are you sure? In an application with buttons to the right of this (like IQ), what maintains the space between the product name and the buttons?

Well in IQ nothing does, there is no right margin or padding present on the product block (or its children).

I don't have any issue with having any elements to the right of the product block using left margin for spacing. That seems more intentional than 12px of right margin here "just in case" even if it is zeroed out by `container-horizontal`.

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -8,82 +8,92 @@
...
+  @include bold();
+
+  -ms-grid-row-align: start;
+  align-self: start;
+  font-size: 11px;
   grid-area: version;
   -ms-grid-column-align: left;
   justify-self: left;
-  line-height: 12px;
+  line-height: 100%;
```
rpokorny, line 98: FYI descenders stick out of this box too

glenhunter: > FYI descenders stick out of this box too

I switched to "normal" here: https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/1c003acbd02b1d00ec23906843d06370bf80ab99

glenhunter: > > FYI descenders stick out of this box too
> 
> I switched to "normal" here: [1c003ac](https://github.com/sonatype/sonatype-react-shared-components/commit/1c003acbd02b1d00ec23906843d06370bf80ab99)

After trying "normal", "100%", and various numerical values, I opted for "1.1" as the best compromise between position and descender coverage. https://github.com/sonatype/sonatype-react-shared-components/pull/13/commits/6f321eb7ffa7069bc96911f210ed649c94b50e49

PR #14, opened 2020-04-24
Rename nx-root-container - RSC-89

glenhunter:
https://issues.sonatype.org/browse/RSC-89

Simple change to make the intention of this wrapping class clearer. 

`nx-root-container` isn't very clear as a class name, the purpose of this `<div>` is to act as a container for `.nx-page-sidebar` and `.nx-page-main`. I think `.nx-page-content-container` is more clear in its intent.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "0.43.1",
+  "version": "0.43.2",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -45,7 +45,7 @@ function Application() {
   return (
     <Router>
       <PageHeader />
-      <div className="nx-page-content-container">
+      <div className="nx-page-content">
         <aside className="nx-page-sidebar" id="gallery-sidebar">
           <GalleryNav />
         </aside>
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -45,7 +45,7 @@ function Application() {
   return (
     <Router>
       <PageHeader />
-      <div className="nx-root-container">
+      <div className="nx-page-content-container">
         <aside className="nx-page-sidebar" id="gallery-sidebar">
           <GalleryNav />
         </aside>
--- a/gallery/src/PageHeader.tsx
+++ b/gallery/src/PageHeader.tsx
@@ -12,8 +12,8 @@ const sonatypeLogo = require('@sonatype/react-shared-components/assets/img/SON_h
 
 function PageHeader() {
   return (
-    <header className="nx-header">
-      <div className="nx-header__inner">
+    <header className="nx-page-header">
+      <div className="nx-page-header__inner">
         <a href="#" className="nx-product">
           <img src={sonatypeLogo} className="nx-product__logo"/>
           <div className="nx-product__name">
--- a/gallery/src/index.html
+++ b/gallery/src/index.html
@@ -14,7 +14,7 @@
     <link rel="stylesheet" href="gallery.css"/>
   </head>
   <body class="nx-body">
-    <div id="ui" class="nx-body-container"></div>
+    <div id="ui" class="nx-page"></div>
     <script src="bundle.js"></script>
   </body>
 </html>
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "0.43.1",
+  "version": "0.43.2",
   "description": "Sonatype shared UI components and utilities written in React",
   "main": "index.js",
   "repository": {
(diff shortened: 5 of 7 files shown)


Blacktiger:
Personally I'd expect the main content to go into `nx-page-content-container` based on the name. `nx-page-main` feels like what the whole page would go into. :shrug:

glenhunter:
> Personally I'd expect the main content to go into `nx-page-content-container` based on the name. `nx-page-main` feels like what the whole page would go into. 🤷

I drew up the attached image to clarify the class names. Now is the time to make these changes, before the styles become too embedded in projects to easily change.

Screen Shot 2020-04-24 at 4 00 19 PM

Blacktiger:
> So I think what you're suggesting is that `nx-page-main` should become something like `nx-page-content`?

I'm just saying that's what I was expecting I guess. That might be because in nxrm we don't have a content sidebar really (sometimes an info panel on the right) because our left sidebar is used for navigation. We only have a main content area, so I guess the names are confusing to me. I think we should make sure the page layout classes/components include that diagram somewhere so people can figure out how things are structured since some things may not be clear just based on the name.

Blacktiger:
> > So I think what you're suggesting is that `nx-page-main` should become something like `nx-page-content`?

So maybe the class names should be `.nx-body`, `.nx-body-container`, `.nx-page` (new div needed for this), `.nx-page-header`, `.nx-page-container`, `.nx-page-container-sidebar` (or maybe `.nx-page-sidebar`), and `.nx-page-container-content` (or maybe `.nx-page-content`)? It would be nice if the class names related to the dom structure a bit more clearly to aid in understanding where/how the class should be used.

rpokorny:
> So maybe the class names should be .nx-body, .nx-body-container, .nx-page (new div needed for this), .nx-page-header, .nx-page-container, .nx-page-container-sidebar (or maybe .nx-page-sidebar), and .nx-page-container-content (or maybe .nx-page-content)? It would be nice if the class names related to the dom structure a bit more clearly to aid in understanding where/how the class should be used.

Something along thesee lines seems sensible to me, except that I don't think an additional div is necessary; we can just change `.nx-body-container` to `.nx-page`

glenhunter:
> > So maybe the class names should be .nx-body, .nx-body-container, .nx-page (new div needed for this), .nx-page-header, .nx-page-container, .nx-page-container-sidebar (or maybe .nx-page-sidebar), and .nx-page-container-content (or maybe .nx-page-content)? It would be nice if the class names related to the dom structure a bit more clearly to aid in understanding where/how the class should be used.
> 
> Something along these lines seems sensible to me, except that I don't think an additional div is necessary; we can just change `.nx-body-container` to `.nx-page`

I've tried to take the various comments into account and make the terms as clear as possible. I changed "container" to "wrapper" because I think that "wrapper" is a bit clearer that the `<div>` in question is meant to "wrap around" other content, but not have any other content itself.

Screen Shot 2020-04-27 at 10 38 13 AM

rpokorny:
> I've tried to take the various comments into account and make the terms as clear as possible. I changed "container" to "wrapper" because I think that "wrapper" is a bit clearer that the <div> in question is meant to "wrap around" other content, but not have any other content itself.

What was wrong with just `nx-page` for the top one (what you no have as `nx-page-wrapper`)?

rpokorny:
I notice we have a familiar pattern here: header, content, and then footer.  Just like in `nx-tile` and `nx-modal`.  I think it would be valuable to align the names even more closely to the way they are in those cases, by renaming `nx-page-content-wrapper` to simply `nx-page-content`.

glenhunter:
> I notice we have a familiar pattern here: header, content, and then footer. Just like in `nx-tile` and `nx-modal`. I think it would be valuable to align the names even more closely to the way they are in those cases, by renaming `nx-page-content-wrapper` to simply `nx-page-content`.

Yes, you're right, I lost sight of that a bit, but I think the naming of this is clearer:
Screen Shot 2020-04-27 at 10 56 05 AM

My goal with the "wrapper" suffix was to try and make it clear that wrappers don't hold content, they wrap other `<div>`s which contain the content. Keeping a consistent:
Parent
-- Header
-- Content
-- Footer
naming structure probably does a better job of that.

mpiggott:
At least how we use it in NXRM the sidebar isn't really part of the page content anymore than the header or the footer

Blacktiger:
> At least how we use it in NXRM the sidebar isn't really part of the page content anymore than the header or the footer

We probably can't use these layout classes in nxrm (or at least not all of them) until we rethink the navigation/layout anyway. I suspect we'll need to have our own custom layout components/classes at least initially.

glenhunter:
> > At least how we use it in NXRM the sidebar isn't really part of the page content anymore than the header or the footer
> 
> We probably can't use these layout classes in nxrm (or at least not all of them) until we rethink the navigation/layout anyway. I suspect we'll need to have our own custom layout components/classes at least initially.

My thinking is that these styles are mostly for new projects and IQ. Having spent some time in the RM codebase it seems unlikely that you'd be able to implement such extreme changes to your page structure in the near or even medium term.

glenhunter:
It's also worth pointing out that these page layout styles are not meant to be proscriptive/mandatory. They are meant to be an aid in getting a new project up and running quickly. If existing projects can use them for refactoring or new development that's a bonus.

rpokorny:
the header is currently `nx-header` not `nx-page-header`.

glenhunter:
> the header is currently `nx-header` not `nx-page-header`.

Changed here: https://github.com/sonatype/sonatype-react-shared-components/pull/14/commits/995ce5dd3c3d99afbaee683cc6532bf2c1b4a004

PR #18, opened 2020-05-05
Better fix for nested nx lists - RSC-81 

glenhunter:
https://issues.sonatype.org/browse/RSC-81

As part of the work to use `nx-table` and `nx-list` in our example documentation I created an overly broad class to deal with the issue where `nx-list` is applied to a `<ul>` or `<ol>` instead of a wrapping `<div>`, in particular where lists are nested within lists. This resulted in a display bug in IQ. This PR fixes that class, improves the styling of nested, bulleted lists, and provides examples for nested tables.

### DIFF
--- a/gallery/src/styles/NxList/NxListBulletedExample.tsx
+++ b/gallery/src/styles/NxList/NxListBulletedExample.tsx
@@ -17,9 +17,34 @@ const NxListWithBullets = () =>
       </li>
       <li className="nx-list__item">
         Item 2
+        <ul>
+          <li className="nx-list__item">
+            item 1
+          </li>
+          <li className="nx-list__item">
+            Item 2
+          </li>
+          <li className="nx-list__item">
+            item 3
+            <ul>
+              <li className="nx-list__item">
+                item 1
+              </li>
+              <li className="nx-list__item">
+                Item 2
+              </li>
+              <li className="nx-list__item">
+                item 3
+              </li>
+            </ul>
+          </li>
+          <li className="nx-list__item">
+            Item 4
+          </li>
+        </ul>
       </li>
       <li className="nx-list__item">
-        item 3
+        Item 3
       </li>
     </ul>
   </div>;
--- a/gallery/src/styles/NxList/NxListBulletedExample.tsx
+++ b/gallery/src/styles/NxList/NxListBulletedExample.tsx
@@ -11,13 +11,13 @@ const NxListWithBullets = () =>
     <h4 className="nx-list__title">
       List title
     </h4>
-    <ul>
+    <ul className="nx-list nx-list--bulleted">
       <li className="nx-list__item">
         item 1
       </li>
       <li className="nx-list__item">
         Item 2
-        <ul>
+        <ul className="nx-list nx-list--bulleted">
           <li className="nx-list__item">
             item 1
           </li>
@@ -26,7 +26,7 @@ const NxListWithBullets = () =>
           </li>
           <li className="nx-list__item">
             item 3
-            <ul>
+            <ul className="nx-list nx-list--bulleted">
               <li className="nx-list__item">
                 item 1
               </li>
--- a/lib/src/base-styles/_nx-lists.scss
+++ b/lib/src/base-styles/_nx-lists.scss
@@ -10,8 +10,9 @@
 .nx-list {
   @include container-vertical;
   margin: $nx-spacing-sm 0 $nx-spacing-md 0;
+  padding: 0;
 
-  &, > ul, > ol {
+  ul, ol {
     list-style: none;
     margin: 0;
     padding: 0;
@@ -64,6 +65,22 @@
     overflow: visible;
     padding: 6px 10px 6px 0;
     white-space: normal;
+  }
+
+  .nx-list__item {
+    list-style-type: disc;
+
+    ul {
+      margin-top: $nx-spacing-sm;
+    }
+
+    .nx-list__item {
+      list-style-type: circle;
+
+      .nx-list__item {
+        list-style-type: square;
+      }
+    }
   }
 }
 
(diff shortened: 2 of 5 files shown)


lib/src/base-styles/_nx-lists.scss
```diff
@@ -10,8 +10,9 @@
 .nx-list {
   @include container-vertical;
   margin: $nx-spacing-sm 0 $nx-spacing-md 0;
+  padding: 0;
 
-  &, > ul, > ol {
+  ul, ol {
```
rpokorny, line 15: Why this change?

lib/src/base-styles/_nx-lists.scss
```diff
@@ -65,6 +66,22 @@
     padding: 6px 10px 6px 0;
     white-space: normal;
   }
+
+  .nx-list__item {
+    list-style-type: disc;
+
+    .nx-list {
+      margin-top: $nx-spacing-sm;
```
rpokorny, line 74: This is suspicious.  Shouldn't this be handled by our normal margin and padding rules.  E.g, shouldn't `.nx-list` have a standard amount of top margin, which, when it's the first child in an `.nx-list__item`, gets removed, and then separately `.nx-list__item` has a certain amount of top padding?

lib/src/base-styles/_nx-lists.scss
glenhunter: > Why this change?

This is what caused the problem reported by Andrey in IQ. In this case `<ul className="nx-list">` loses the margin that are applied to `.nx-list`. This is incorrect.

lib/src/base-styles/_nx-lists.scss
glenhunter: > This is suspicious. Shouldn't this be handled by our normal margin and padding rules. E.g, shouldn't `.nx-list` have a standard amount of top margin, which, when it's the first child in an `.nx-list__item`, gets removed, and then separately `.nx-list__item` has a certain amount of top padding?

If `.nx-list__item` has top padding it will appear above the text of the list item, not between that text and the embedded `.nx-list`. I think it's clearer it you look at the example HTML.

gallery/src/styles/NxList/NxListBulletedExample.tsx
```diff
@@ -11,15 +11,40 @@ const NxListWithBullets = () =>
     <h4 className="nx-list__title">
       List title
     </h4>
-    <ul>
+    <ul className="nx-list nx-list--bulleted">
       <li className="nx-list__item">
         item 1
       </li>
       <li className="nx-list__item">
         Item 2
```
rpokorny, line 19: Mixing text nodes and elements as siblings

lib/src/base-styles/_nx-lists.scss
rpokorny: > If .nx-list__item has top padding it will appear above the text of the list item, not between that text and the embedded .nx-list. I think it's clearer it you look at the example HTML.

Right, after looking at the example I'm back to my same question, but focusing on a standard margin being the correct solution.  Why doesn't `.nx-list` have a top margin by default?

gallery/src/styles/NxList/NxListBulletedExample.tsx
```diff
@@ -11,15 +11,40 @@ const NxListWithBullets = () =>
     <h4 className="nx-list__title">
       List title
     </h4>
-    <ul>
+    <ul className="nx-list nx-list--bulleted">
```
rpokorny, line 14: So in this example the `.nx-list` styles are on _both_ the `div` and the top `ul`?  That doesn't seem right

lib/src/base-styles/_nx-lists.scss
rpokorny: > In this case <ul className="nx-list"> loses the margin that are applied to .nx-list. This is incorrect.

In what valid scenario would a `<ul classname="nx-list">` be an immediate child of a `<div className="nx-list">`?

rpokorny:
I'm not sure the top and bottom spacing [here](https://github.com/sonatype/sonatype-react-shared-components/pull/18/files#diff-e6ab612d764d04d059eee79c79caea47R39) should be done as padding.  Playing with the gallery, I notice that it bunches up at the tail of nested lists.  As seen here: 
image

lib/src/base-styles/_nx-lists.scss
glenhunter: > > In this case  loses the margin that are applied to .nx-list. This is incorrect.
> 
> In what valid scenario would a `<ul classname="nx-list">` be an immediate child of a `<div className="nx-list">`?

Adjusted here: https://github.com/sonatype/sonatype-react-shared-components/pull/18/commits/08dfe7f04fdd3ffabf59b3d45d6adab914c0dc84

gallery/src/styles/NxList/NxListBulletedExample.tsx
glenhunter: > Mixing text nodes and elements as siblings

Bare nodes now wrapped with `<span>`'s: https://github.com/sonatype/sonatype-react-shared-components/pull/18/commits/08dfe7f04fdd3ffabf59b3d45d6adab914c0dc84

gallery/src/styles/NxList/NxListBulletedExample.tsx
glenhunter: > So in this example the `.nx-list` styles are on _both_ the `div` and the top `ul`? That doesn't seem right

https://github.com/sonatype/sonatype-react-shared-components/pull/18/commits/08dfe7f04fdd3ffabf59b3d45d6adab914c0dc84

rpokorny:
LGTM but at this point I can't approve it because I substantially contributed to it

PR #19, opened 2020-05-06
Alphabetize component gallery

Blacktiger:
I find it hard to locate the component example pages, so I alphabetized the list.


gallery/src/pageConfig.ts
```diff
@@ -73,34 +73,34 @@ const pageConfig: PageConfig = {
...
     NxCheckbox: NxCheckboxPage,
-    NxRadio: NxRadioPage,
-    NxTextInput: NxTextInputPage,
-    NxBackButton: NxBackButtonPage,
-    NxLoadingSpinner: NxLoadingSpinnerPage,
+    NxDropdown: NxDropdownPage,
+    NxFilterInput: NxFilterInputPage,
+    NxFontAwesomeIcon: NxFontAwesomeIconPage,
     NxLoadError: NxLoadErrorPage,
+    NxLoadingSpinner: NxLoadingSpinnerPage,
```
joedragons, line 84: I think this should be one line down?

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "0.44.4",
+  "version": "0.43.5",
```
marushkevych, line 3: did you mean 0.44.5?

gallery/src/pageConfig.ts
glenhunter: > I think this should be one line down?

I fixed as part of: https://github.com/sonatype/sonatype-react-shared-components/pull/19/commits/49adb42cab53cc92a0a462713c993db778241c47

glenhunter:
I alphabetized the styles, added a "layouts" section for the form layouts (and soon to be page layouts), and moved the stateful components to be adjacent to their stateless counterparts.

joedragons:
I find the stateful merged in a bit harder to read than it was, but if @Blacktiger doesn't care, I'm fine.  +1 holds.

gallery/package.json
Blacktiger: 28aed3e

gallery/src/pageConfig.ts
```diff
@@ -53,57 +53,57 @@ import NxStatefulCheckboxPage from './components/NxStatefulCheckbox/NxStatefulCh
...
-  },
-  'Stateful React Components': {
-    NxStatefulCheckbox: NxStatefulCheckboxPage,
-    NxStatefulTextInput: NxStatefulTextInputPage,
-    NxStatefulSubmitMask: NxStatefulSubmitMaskPage,
-    NxStatefulDropdown: NxStatefulDropdownPage,
     NxStatefulTreeViewMultiSelect: NxStatefulTreeViewMultiSelect,
-    NxStatefulTreeViewRadioSelect: NxStatefulTreeViewRadioSelectPage
-  }
+    NxTreeViewRadioSelect: NxTreeViewRadioSelectPage,
```
Blacktiger, line 103: The only downside to this is I was able to simply sort the components alphabetically in my text editor, but with the stateful components next to their non-stateful versions we just have to make sure we stick to the correct ordering in the future. It's probably fine now that we've done the ordering once.

glenhunter: > The only downside to this is I was able to simply sort the components alphabetically in my text editor, but with the stateful components next to their non-stateful versions we just have to make sure we stick to the correct ordering in the future. It's probably fine now that we've done the ordering once.

And we have to remember when we add new stateful components to place them next to their non-stateful equivalents, but I think the grouping is worth it.

marushkevych: > The only downside to this is I was able to simply sort the components alphabetically in my text editor, but with the stateful components next to their non-stateful versions we just have to make sure we stick to the correct ordering in the future. It's probably fine now that we've done the ordering once.

When we discussed this change earlier didn't we decide to rename Stateful components to put "Stateful" at the end?

joedragons: I think that'd be less confusing conceptually, I may change my mind if I saw it ;)

PR #21, opened 2020-05-07
RSC-94 used yarn upgrade commonmark@0.29.1

wwannemacher:
am a react / javascript amateur, go easy :)

https://issues.sonatype.org/browse/RSC-94

Although ultimately we don't believe that NXRM was vulnerable, we're trying to start upgrading components even when we're not vulnerable (to decrease the chances that we introduce a change later that would make us vulnerable).

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -50,6 +50,7 @@
     "@fortawesome/fontawesome-svg-core": "1.2.17",
     "@fortawesome/free-solid-svg-icons": "=5.8.1",
     "@fortawesome/react-fontawesome": "=0.1.4",
+    "@material-ui/core": "^4.9.13",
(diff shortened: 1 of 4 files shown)


wwannemacher:
argh, I give up - I'll leave it to smarter folks to figure this out.

rpokorny:
We don't depend directly on `commonmark`, it is a transitive dependency brought in by `react-commonmark`.  We are using the latest version of `react-commonmark`, and its dependency specification for `commonmark` does not include v0.29.1, so there is no upgrade path here.

The only usage of `react-commonmark` in RSC is in the `NxVulnerabilityDetails` component.  If you aren't using that, you aren't vulnerable.  Further, typical usage of that component would be with markdown exclusively brought in from our own backend servers, not markdown coming from users, avoiding the risk of malicious input.

PR #24, opened 2020-05-25
Adjust default line height - RSC-93

glenhunter:
https://issues.sonatype.org/browse/RSC-93

Per directions from Derrek the default line-height on `nx-body` was changed to `1.5`. Then I changed the `h#` to unitless values and changed `nx-p` and bulleted list items to use the default values.

There is a small difference in the line-heights of bulleted list items and paragraphs, and very little difference elsewhere.

### DIFF
--- a/lib/src/base-styles/_nx-lists.scss
+++ b/lib/src/base-styles/_nx-lists.scss
@@ -63,7 +63,7 @@
     @include regular();
 
     border: none;
-    line-height: 1.5;
+    line-height: $nx-line-height;
     list-style: disc outside;
     margin: $nx-spacing-sm 10px $nx-spacing-sm 36px;
     overflow: visible;
--- a/lib/src/base-styles/_nx-lists.scss
+++ b/lib/src/base-styles/_nx-lists.scss
@@ -63,6 +63,7 @@
     @include regular();
 
     border: none;
+    line-height: 1.5;
     list-style: disc outside;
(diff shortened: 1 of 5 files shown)


lib/src/base-styles/_nx-lists.scss
```diff
@@ -63,6 +63,7 @@
     @include regular();
 
     border: none;
+    line-height: 1.5;
```
rpokorny, line 66: If this is specifically meant to be in sync with the main line-height should it use a variable?

rpokorny:
I assume you looked into all of the other custom line-heights throughout the RSC styles and determined that they were still necessary/desirable?

glenhunter:
> I assume you looked into all of the other custom line-heights throughout the RSC styles and determined that they were still necessary/desirable?

Yes, all of the other values are essentially magic numbers. I feel like we could do better with our headings - there's no typographic scale there that I can discern. But that's up to @dturner32 and can be handled (if at all) in a later task.

lib/src/base-styles/_nx-lists.scss
glenhunter: > If this is specifically meant to be in sync with the main line-height should it use a variable?

Sure: https://github.com/sonatype/sonatype-react-shared-components/pull/24/commits/b2e9b5427ab6596367636ec80aee3a05d85f9c0e

PR #28, opened 2020-05-28
Page layout guidelines RSC-83

glenhunter:
https://issues.sonatype.org/browse/RSC-83


gallery/src/main.scss
```diff
@@ -61,3 +61,12 @@
 
   position: relative;
 }
+
+.gallery-example-image {
+  height: auto;
```
rpokorny, line 66: Necessary?

glenhunter: > Necessary?

Not really, just habit: https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/8a9c70da52cbaa5a410297b32f848f1439cdcf02

gallery/src/main.scss
```diff
@@ -61,3 +61,11 @@
 
   position: relative;
 }
+
+.gallery-example-image {
+  max-width: 100%;
+}
+
+.nx-code.nx-code--pre-block {
+  white-space: pre;
```
rpokorny, line 70: Is this still used?

gallery/src/pageConfig.ts
```diff
@@ -69,7 +70,8 @@ const pageConfig: PageConfig = {
     'nx-container-helpers': NxContainerHelpersPage
   },
   'Layout Examples': {
-    'Form Layout Styles': NxFormLayoutPage
+    'Form Layout Styles': NxFormLayoutPage,
+    'Page Layout Guidelines': PageLayoutPage
```
rpokorny, line 74: Should this be named "... Guidelines" if it's not in the Guildelines section?

gallery/src/styles/PageLayout/PageLayoutPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const PageLayoutGuidelinesPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p>
+        HTML and SCSS are provided for both single and two column page layouts. We use{' '}
+        <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox">
+          flexbox
+        </a>
+        to handle our layouts.
```
rpokorny, line 24: This'll lose a space.
```suggestion
        </a>
        {' '}to handle our layouts.
```

gallery/src/styles/PageLayout/PageLayoutPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </p>
+      <p>
```
rpokorny, line 26: Don't all these need `.nx-p`?

gallery/src/styles/PageLayout/PageLayoutPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+        One challenge of flexbox layouts is that the order and nesting of the tags that make up the layouts
+        must be nested consistently or the layout will break. This often makes flexbox layouts challenging with
```
rpokorny, line 28:
```suggestion
        One challenge of flexbox layouts is that direct parent-child relationships between elements are crucial, and adding intermediate ancestors can easily break the layout. This often makes flexbox layouts challenging with
```

gallery/src/styles/PageLayout/PageLayoutPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+        sidebar element. When the sidebar element is not present the
+        {' '}<code className="nx-code">nx-page-content-main</code> block will expand to fill the available space.
+      </p>
+      <h2 className="nx-h2">Page layout illustrations</h2>
+      <h3 className="nx-h3">Standard two column layout</h3>
+      <p>
+        Our standard two column layout consists of a page header area, a page sidebar, main content area, and a page
+        footer. Many of our apps do not use a page footer and this can be safely omitted.
+      </p>
+      <p>
```
rpokorny, line 44: I'm not sure this `<p>` makes semantic sense.  It would probably be fine to omit, maybe with `display: block` added to the img

gallery/src/styles/PageLayout/PageLayoutPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
+import RawHtmlExample from '../../CodeExample';
+
+const SingleColumnLayoutExampleCode = require('!!raw-loader!./SingleColumnLayoutExample.html').default,
+    TwoColumnLayoutExampleCode = require('!!raw-loader!./TwoColumnLayoutExample.html').default,
+    ReactRootLayoutExampleCode = require('!!raw-loader!./ReactRootLayoutExample.html').default;
```
rpokorny, line 14: These variables should all start with lowercase letters

gallery/src/styles/PageLayout/PageLayoutPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {' '}<code className="nx-code">nx-page-content-main</code> block will expand to fill the available space.
+      </p>
+      <h2 className="nx-h2">Page layout illustrations</h2>
+      <h3 className="nx-h3">Standard two column layout</h3>
+      <p>
+        Our standard two column layout consists of a page header area, a page sidebar, main content area, and a page
+        footer. Many of our apps do not use a page footer and this can be safely omitted.
+      </p>
+      <p>
+        <img src="../resources/page-layout-mocks/page-layout-2col.png" className="gallery-example-image"/>
```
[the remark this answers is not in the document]
rpokorny: > I'm pretty sure this isn't going to work as-is in the built copy of the gallery. We need to copy the images to the dist dir

See the way I did the images for the line-height doc

gallery/src/main.scss
glenhunter: > Is this still used?

Nope: https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/f592da532d902417ba296aad137ca99885bbf644

gallery/src/pageConfig.ts
glenhunter: > Should this be named "... Guidelines" if it's not in the Guildelines section?

All the guidelines are gone here: https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/e982547a0fd7c62941a7a6975f5b823f25ee551b and here: https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/e982547a0fd7c62941a7a6975f5b823f25ee551b

gallery/src/styles/PageLayout/PageLayoutPage.tsx
glenhunter: > Don't all these need `.nx-p`?

Yes. The `<p>` tags are the one place where I want to go back on my "never style HTML tags rule". Not sure why `<p>` tags bug me in particular. https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/e982547a0fd7c62941a7a6975f5b823f25ee551b

gallery/src/styles/PageLayout/PageLayoutPage.tsx
glenhunter: > I'm not sure this `<p>` makes semantic sense. It would probably be fine to omit, maybe with `display: block` added to the img

I removed the `<p>` tags to no ill effect

gallery/src/styles/PageLayout/SingleColumnLayoutExample.html
```diff
@@ -0,0 +1,18 @@
...
+
+-->
+<body class="nx-body">
+  <div class="nx-page">
+    <div class="nx-page-header">
+    </div>
+    <div class="nx-page-content">
+      <div class="nx-page-content-main">
+      </div>
+    </div>
```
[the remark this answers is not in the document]
glenhunter: > In keeping with our recommendation that `.nx-page` be the react root, I'm wondering if we should demonstrate its children as react components instead of HTML.

If we did that then I'd suggest we have two sections, the first is just the HTML skeleton illustrated, the second is how you'd break up that skeleton into react components.

gallery/src/styles/PageLayout/PageLayoutPage.tsx
glenhunter: > These variables should all start with lowercase letters

https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/97efbdff0dea563d0e39985933b6dfd795b29104

gallery/src/styles/PageLayout/PageLayoutPage.tsx
rpokorny: > Yes. The <p> tags are the one place where I want to go back on my "never style HTML tags" rule.

You missed the one this comment is on.

> Not sure why <p> tags bug me in particular.

Probably because there's a lot of them.

> e982547

Wrong commit btw

rpokorny: > Probably because there's a lot of them.

One approach that consuming apps could take, including the gallery even, would be to use scss like `p { @extend .nx-p; }` if they are sure they want nx styles globally.  We just can't have it that way out of the box.

Or as another idea, we chould create some "root" class (`nx` maybe) that consuming apps should add to the top of each part of their app that uses RSC styles, and then we could add styles for `.nx p` and so on.

gallery/src/styles/PageLayout/SingleColumnLayoutExample.html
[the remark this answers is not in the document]
rpokorny: > If we did that then I'd suggest we have two sections, the first is just the HTML skeleton illustrated, the second is how you'd break up that skeleton into react components.

Eh nevermind.  Recommending "this is where the react part should start" is one thing, but I don't really want to be recommending how apps should be breaking down their react code into different components.

gallery/src/styles/PageLayout/PageLayoutPage.tsx
glenhunter: > > Probably because there's a lot of them.
> 
> One approach that consuming apps could take, including the gallery even, would be to use scss like `p { @extend .nx-p; }` if they are sure they want nx styles globally. We just can't have it that way out of the box.
> 
> Or as another idea, we chould create some "root" class (`nx` maybe) that consuming apps should add to the top of each part of their app that uses RSC styles, and then we could add styles for `.nx p` and so on.

https://github.com/sonatype/sonatype-react-shared-components/pull/28/commits/4186f9636ce50591b5adf13d305112df1a2c240e

PR #29, opened 2020-05-28
Raw html examples - RSC-101

rpokorny:
https://issues.sonatype.org/browse/RSC-101

Switch the examples of plain CSS styles (not react components) to use actual HTML rather than react components.

This PR now also includes the removal of the license headers from each of the source code displays


glenhunter:
Love that you got the license remove in there too.

PR #30, opened 2020-05-29
Page header redesign RSC-103

glenhunter:
https://issues.sonatype.org/browse/RSC-103

Match RSC to new design requirements from SDS


gallery/src/PageHeader.tsx
```diff
@@ -16,12 +16,10 @@ function PageHeader() {
...
-          <div className="nx-product__name">
-            Sonatype Component Library
-          </div>
-          <div className="nx-product__version">
-            Version: {packageJson.version}
-          </div>
+          <div className="nx-product__sonatype-wordmark">sonatype</div>
+          <div className="nx-product__vertical-divider">|</div>
+          <div className="nx-product__name">react shared component library</div>
+          <div className="nx-product__version">Version: {packageJson.version}</div>
```
rpokorny, line 22: Should these be spans at this point?

gallery/src/PageHeader.tsx
```diff
@@ -16,12 +16,10 @@ function PageHeader() {
...
         <a href="#" className="nx-product">
           <img src={sonatypeLogo} className="nx-product__logo"/>
-          <div className="nx-product__name">
-            Sonatype Component Library
-          </div>
-          <div className="nx-product__version">
-            Version: {packageJson.version}
-          </div>
+          <div className="nx-product__sonatype-wordmark">sonatype</div>
+          <div className="nx-product__vertical-divider">|</div>
```
rpokorny, line 20: Is there a reason this wasn't done as a border?

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -8,8 +8,8 @@
 @import '../scss-shared/nx-text-helpers';
 @import '../scss-shared/nx-container-helpers';
 
-$main-header-background: $nx-sonatype-blue;
-$main-header-color: $nx-off-white;
+$main-header-background: $nx-off-white;
+$main-header-color: $nx-off-black;
```
rpokorny, line 12: The text in zeplin seems to use several different colors none of which are this one.

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -39,61 +39,44 @@ $main-header-product-version-row-height: 1fr;
...
 
 .nx-product__version {
-  -ms-grid-row: 2;
-  -ms-grid-column: 3;
-  @include bold();
-
-  -ms-grid-row-align: start;
-  align-self: start;
+  align-self: center;
   font-size: 11px;
```
rpokorny, line 79: 12 in zeplin

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -39,61 +39,44 @@ $main-header-product-version-row-height: 1fr;
...
-  -ms-grid-row: 1;
-  -ms-grid-row-span: 2;
-  -ms-grid-column: 1;
-  -ms-grid-row-align: center;
   align-self: center;
   box-sizing: border-box;
-  grid-area: logo;
-  height: 100%;
-  width: 100%;
+  height: 20px;
```
rpokorny, line 52: 21px tall in zeplin.  Also this icon isn't as wide as it is tall so the `width` should probably be left unspecified

rpokorny:
It occurs to me that the layout of the proposed headers for the actual products is substantially different from the one for the gallery.  These styles are therefore really just for the gallery.  This leads to the following questions:

1. Should these styles should really be in the RSC lib, or should they only be in the gallery
2. Where and when will the styles for the for the application headers be implemented and proven out?  
  2a. If the application header styles are to be demonstrated in the gallery, then where would that happen if this other header style is used for the gallery's actual header?
  2b. If the application header styles aren't to be demonstrated in the gallery, then where are they to be demonstrated?
3. Why are we putting effort into a separate header style for the RSC gallery?

rpokorny:
In addition to my above comment, I know the mockups of the application headers include elements that the gallery header has simply never had: the buttons/links in the middle and to the right.  Historically, we've left those parts up to IQ.  If we want to include them in the standardization provided by RSC however, they too need to be added to the RSC header styles, and demonstrated somewhere.

gallery/src/PageHeader.tsx
glenhunter: > Should these be spans at this point?

They could be I guess. Since we're using flex box I don't see much reason to switch them

gallery/src/PageHeader.tsx
glenhunter: > Is there a reason this wasn't done as a border?

This worked first try and seemed simpler than applying a border to the box, shrinking the box to match the required height, and then potentially shifting the box around to match the required positioning

lib/src/base-styles/_nx-page-header.scss
glenhunter: > The text in zeplin seems to use several different colors none of which are this one.

Yeah, I talked to Derrek about the colours. We don't have a great solution for that yet so for the text colours I have been using our defaults until he can give us a full palette

rpokorny: > Yeah, I talked to Derrek about the colours. We don't have a great solution for that yet so for the text colours I have been using our defaults until he can give us a full palette

I'd rather set it to the right colors through magic numbers now, then when we get variables in place later it's a straightforward search and replace to clean them up.  Rather than setting it to the wrong color now and then having to re-research what colors everything should be when the variables are available.

glenhunter:
> In addition to my above comment, I know the mockups of the application headers include elements that the gallery header has simply never had: the buttons/links in the middle and to the right. Historically, we've left those parts up to IQ. If we want to include them in the standardization provided by RSC however, they too need to be added to the RSC header styles, and demonstrated somewhere.

@dturner32 is aware. We're going to need a separate page for the page header, and potentially the footer when it arrives.

lib/src/base-styles/_nx-page-header.scss
glenhunter: > > Yeah, I talked to Derrek about the colours. We don't have a great solution for that yet so for the text colours I have been using our defaults until he can give us a full palette
> 
> I'd rather set it to the right colors through magic numbers now, then when we get variables in place later it's a straightforward search and replace to clean them up. Rather than setting it to the wrong color now and then having to re-research what colors everything should be when the variables are available.

Colours properly identified here: https://github.com/sonatype/sonatype-react-shared-components/pull/30/commits/f7c7f5512a24c62f07672cbc503805e9e55f2a97

lib/src/base-styles/_nx-page-header.scss
glenhunter: > 21px tall in zeplin. Also this icon isn't as wide as it is tall so the `width` should probably be left unspecified

Moot bc we replaced the whole thing with an image as discussed in Slack.

lib/src/base-styles/_nx-page-header.scss
glenhunter: > 12 in zeplin

Fixed: https://github.com/sonatype/sonatype-react-shared-components/pull/30/commits/3549fe76834f168a1be07fb99bd3835ef3ae6c57

glenhunter:
> It occurs to me that the layout of the proposed headers for the actual products is substantially different from the one for the gallery. These styles are therefore really just for the gallery. This leads to the following questions:
> 
> 1. Should these styles should really be in the RSC lib, or should they only be in the gallery?
> 2. Where and when will the styles for the for the application headers be implemented and proven out?
>    2a. If the application header styles are to be demonstrated in the gallery, then where would that happen if this other header style is used for the gallery's actual header?
>    2b. If the application header styles aren't to be demonstrated in the gallery, then where are they to be demonstrated?
> 3. Why are we putting effort into a separate header style for the RSC gallery?

As discussed in wall walk on June 1, this is now the "default" page header, the IQ and RM headers are variations. The "Nexus" header will either get its own page in the Gallery, or will be a variation of the standard page header

rpokorny:
Note the 14px and 22px here:
image

I think you've got the header elements vertically centered which isn't correct.  This probably ties in to why you needed the top margin on the image

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -39,61 +40,41 @@ $main-header-product-version-row-height: 1fr;
...
-  height: 100%;
-  width: 100%;
 }
 
-.nx-product__name {
-  -ms-grid-row: 1;
-  -ms-grid-column: 3;
-  @include semi-bold();
+.nx-product__wordmark {
+  height: 21px;
```
rpokorny, line 55: Do we have a copy of the image that either fits this height exactly, or is scalable?  The current image is natively 42px tall, which at least is exactly double so the scaling shouldn't be ugly, but I'd think we'd want one that's pre-scaled

gallery/src/PageHeader.tsx
rpokorny: > This worked first try and seemed simpler than applying a border to the box, shrinking the box to match the required height, and then potentially shifting the box around to match the required positioning

I guess that's alright styling-wise but semantically I'm not a huge fan.  Could you move the pipe character into a `::before` so it isn't in the DOM?

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -39,61 +40,41 @@ $main-header-product-version-row-height: 1fr;
...
+  margin-top: 13px; // fix vertical centering
+}
 
-  -ms-grid-row-align: center;
+.nx-product__vertical-divider {
+  align-self: center;
```
rpokorny, line 60: Rather than separately specifying `align-self` on the flex items you could specify `align-items` on the parent

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -39,61 +40,41 @@ $main-header-product-version-row-height: 1fr;
...
-  align-self: start;
-  font-size: 11px;
-  grid-area: version;
-  -ms-grid-column-align: left;
-  justify-self: left;
-  line-height: 1.1;
+  align-self: center;
+  color: #545454;
+  font-size: 12px;
+  margin-top: 10px; // move version down to match product name baseline
```
rpokorny, line 78: In zeplin their baselines aren't aligned

lib/src/base-styles/_nx-page-header.scss
glenhunter: > Do we have a copy of the image that either fits this height exactly, or is scalable? The current image is natively 42px tall, which at least is exactly double so the scaling shouldn't be ugly, but I'd think we'd want one that's pre-scaled

I got the image directly from @dturner32 I expect the larger size has something to do with how Sketch exports files.

lib/src/base-styles/_nx-page-header.scss
glenhunter: > Rather than separately specifying `align-self` on the flex items you could specify `align-items` on the parent

I did that here: https://github.com/sonatype/sonatype-react-shared-components/pull/30/commits/b5967e477de0d203e96ecf12ddb8baeb42123e1b

lib/src/base-styles/_nx-page-header.scss
glenhunter: > In zeplin their baselines aren't aligned

https://github.com/sonatype/sonatype-react-shared-components/pull/30/commits/ff2a27a1893f6060ef3056c61cb2527edd740ca5

lib/src/base-styles/_nx-page-header.scss
rpokorny: > I got the image directly from @dturner32 I expect the larger size has something to do with how Sketch exports files.

I'm not sure what you're trying to say here

gallery/src/PageHeader.tsx
glenhunter: > > This worked first try and seemed simpler than applying a border to the box, shrinking the box to match the required height, and then potentially shifting the box around to match the required positioning
> 
> I guess that's alright styling-wise but semantically I'm not a huge fan. Could you move the pipe character into a `::before` so it isn't in the DOM?

Sure: https://github.com/sonatype/sonatype-react-shared-components/pull/30/commits/a9177cf8d66cab40995ef54af80cb59724a78e31

lib/src/base-styles/_nx-page-header.scss
glenhunter: > > I got the image directly from @dturner32 I expect the larger size has something to do with how Sketch exports files.
> 
> I'm not sure what you're trying to say here

I shrank it.

PR #31, opened 2020-05-29
sidebar redesign - RSC-104 

glenhunter:
https://issues.sonatype.org/browse/RSC-104

Sidebar redesign from Derrek.


gallery/src/main.scss
```diff
@@ -9,41 +9,63 @@
 @import '~@sonatype/react-shared-components/scss-shared/nx-container-helpers';
 @import '~@sonatype/react-shared-components/scss-shared/nx-flexbox-helpers';
 
+.nx-page-sidebar.gallery-page-sidebar {
+  padding-right: $nx-spacing-md;
+  padding-left: $nx-spacing-md;
```
rpokorny, line 14: Why are we deviating from the usual padding here?  Or should the usual padding be updated?

gallery/src/main.scss
```diff
@@ -9,41 +9,63 @@
...
+  line-height: 32px;
+  margin-right: 0;
   margin-bottom: 4px;
   margin-left: 0;
 }
 
 .gallery-link {
+  border-radius: 6px;
+  box-sizing: border-box;
+  color: $nx-text-color-light;
```
rpokorny, line 44: This isn't actually doing anything; it's getting overridden on the same style on `.nx-tree-view__child a`

gallery/src/main.scss
```diff
@@ -9,41 +9,63 @@
...
-    color: $nx-text-color;
+    background-color: #1a2eac;
+    color: #fff;
     cursor: default;
     text-decoration: none;
+
+    &:hover {
+      background-color: #1a2eac;
+      color: #fff;
+    }
```
rpokorny, line 68: The amount of custom tree view styles here surprises me.  Is it because we don't have selectable items in the normal tree views at all?  If we did, would they follow these styles?

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -66,7 +66,8 @@ $nx-fail-color-hover: #5d0000;
 // Structural
 $nx-site-background-color: $nx-grey-1;
 $nx-page-background-color: $nx-grey-1;
-$nx-sidebar-background-color: $nx-grey-1;
+$nx-sidebar-background-color: #f1f3fa;
+$nx-sidebar-selected: #d1d6f0;
```
rpokorny, line 70: Unused I think

gallery/src/main.scss
glenhunter: > Why are we deviating from the usual padding here? Or should the usual padding be updated?

The usual padding should be updated. I was hesitating to change the `nx-sidebar` default padding because of downstream effects (IQ uses it in one place). But I think that was a mistake and I should have.

gallery/src/main.scss
glenhunter: > The amount of custom tree view styles here surprises me. Is it because we don't have selectable items in the normal tree views at all? If we did, would they follow these styles?

There are :hover states but no selected states. The second part is a question for @dturner32

gallery/src/pageConfig.ts
```diff
@@ -109,6 +84,31 @@ const pageConfig: PageConfig = {
...
+    'nx-text-input': NxTextInputStylesPage,
+    'nx-threat-number': NxThreatNumberPage,
+    'nx-tile': NxTilePage
+  },
+  'Styles - Mixins': {
+    'nx-container-helpers': NxContainerHelpersPage
+  },
+  'Layout Examples': {
+    'Form Layout Styles': NxFormLayoutPage,
+    'Page Layout': PageLayoutPage
```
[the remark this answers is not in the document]
glenhunter: > I'm assuming this was the agreed-upon order? I thought I heard you say the HTML Elements would be last.

Agreed upon is really just do you and I agree. I was thinking that we'd have React Components, then meta-React Components (the guidelines), HTML Styles, HTML Helpers, meta-HTML styles (the layout guides).

lib/src/scss-shared/_nx-colors.scss
glenhunter: > Unused I think

https://github.com/sonatype/sonatype-react-shared-components/pull/31/commits/6855d2021c7345f89bbd5293bf5a6c9c47094b0e

gallery/src/main.scss
glenhunter: > This isn't actually doing anything; it's getting overridden on the same style on `.nx-tree-view__child a`

https://github.com/sonatype/sonatype-react-shared-components/pull/31/commits/6855d2021c7345f89bbd5293bf5a6c9c47094b0e

gallery/src/main.scss
rpokorny: > There are :hover states but no selected states. The second part is a question for @dturner32

What about all the sizing changes?

rpokorny: > I'm uncomfortable applying these changes to the nx-tree-view component directly until we have a full design document from @dturner32

So is restyling the tree views now a 1.0 task? Do we have a ticket for that?

glenhunter: > > I'm uncomfortable applying these changes to the nx-tree-view component directly until we have a full design document from @dturner32
> 
> So is restyling the tree views now a 1.0 task? Do we have a ticket for that?

I have asked @dturner32 what he wants to do.

glenhunter: > > > I'm uncomfortable applying these changes to the nx-tree-view component directly until we have a full design document from @dturner32
> > 
> > 
> > So is restyling the tree views now a 1.0 task? Do we have a ticket for that?
> 
> I have asked @dturner32 what he wants to do.

From @dturner32 "so the custom style for the side-bar was intended to be used only for that instance. As for the current treeview, those styles will remain the same"

glenhunter: I'm ok with this since we're using the tree view explicitly for navigation which is a new use. I would like to see us develop this into general nav styles if we intend to use tree views in this way elsewhere

PR #32, opened 2020-06-02
NxCheckbox, NxRadio keyboard interaction - RSC-85 RSC-86

rpokorny:
https://issues.sonatype.org/browse/RSC-85
https://issues.sonatype.org/browse/RSC-86


glenhunter:
I think that the NxCheckbox and NxRadio examples need to be fixed so that they can be tabbed through and interacted with correctly. Right now the NxRadio examples only tab to the first radio in a group but not the second. NxCheckbox also has many weird behaviours.

Blacktiger:
Not sure what's happening, but when I'm trying to tab through the checkbox example page I end up in some weird part of the page below and can't scroll back without tabbing back to a normally visible element.

Screen Shot 2020-06-03 at 9 08 35 AM

Blacktiger:
I also can't tab through the first example on the radio page. (The other examples are fine)

Blacktiger:
The active checkbox state doesn't match the picture on https://issues.sonatype.org/browse/RSC-85.

rpokorny:
> Not sure what's happening, but when I'm trying to tab through the checkbox example page I end up in some weird part of the page below and can't scroll back without tabbing back to a normally visible element.

I saw that in IE yesterday but hadn't been seeing it in other browsers, but both you and @glenhunter are reporting it now so I will investigate.

rpokorny:
> I also can't tab through the first example on the radio page. (The other examples are fine)

Right, having the already-selected radio be disabled seems to confuse the browsers.  I'll adjust the example

rpokorny:
> The active checkbox state doesn't match the picture on https://issues.sonatype.org/browse/RSC-85.

That picture refers to the planned redesign of the checkbox look and feel, something which is out of scope for this ticket.  It was only attached so that I'd know what to shoot for in the meantime with the focus style

rpokorny:
> I saw that in IE yesterday but hadn't been seeing it in other browsers, but both you and @glenhunter are reporting it now so I will investigate.

3bd43fef

rpokorny:
> Right, having the already-selected radio be disabled seems to confuse the browsers. I'll adjust the example

e2eb94f

rpokorny:
> Right now the NxRadio examples only tab to the first radio in a group but not the second

That is the browser-default behavior.  You tab to a radio group, and then change the selection with the arrow keys.

glenhunter:
If I click on the label of an unselected radio button it will focus and fill that button. But if I click on the empty disc it gets the focus, but doesn’t fill. Clicking again keeps the focus and fills the disc. If I arrow through all of the it always focuses and fills the radio button I have arrowed to

rpokorny:
> In IE if I click on the label of an unselected radio button it will focus and fill that button. But if I click on the empty disc it gets the focus, but doesn’t fill. Clicking again keeps the focus and fills the disc. If I arrow through all of the it always focuses and fills the radio button I have arrowed to

7f6a74f

PR #34, opened 2020-06-04
nx-tile redesign RSC-126 

glenhunter:
https://issues.sonatype.org/browse/RSC-126

In addition to the nx-tile changes I added temporary variables for the body and header font colours which Derrek says are now locked in.

Alert tiles are on hold until the nx-alert styles are done, I put a note in https://issues.sonatype.org/browse/RSC-111 to that effect

### DIFF
--- a/gallery/src/styles/NxTile/NxTileWithActionsExample.tsx
+++ b/gallery/src/styles/NxTile/NxTileWithActionsExample.tsx
@@ -7,32 +7,6 @@
 import React from 'react';
 
 const NxTileWithActionsExample = () =>
-  <>
-  <div className="nx-tile">
-    <div className="nx-tile-header">
-      <div className="nx-tile-header__title">
-        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
-      </div>
-      <div className="nx-tile-header__subtitle">The Subtitle</div>
-    </div>
-    <div className="nx-tile-content">
-      Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
-    </div>
-  </div>
-  <div className="nx-tile">
-    <div className="nx-tile-header">
-      <div className="nx-tile-header__title">
-        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
-      </div>
-      <div className="nx-tile-header__subtitle">The Subtitle</div>
-      <div className="nx-tile__actions">
-        <button className="nx-btn nx-btn--tertiary">Action 1</button>
-      </div>
-    </div>
-    <div className="nx-tile-content">
-      Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
-    </div>
-  </div>
   <div className="nx-tile">
     <div className="nx-tile-header">
       <div className="nx-tile-header__title">
@@ -48,39 +22,6 @@ const NxTileWithActionsExample = () =>
       Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
     </div>
   </div>
-  <div className="nx-tile">
-    <div className="nx-tile-header">
-      <div className="nx-tile-header__title">
-        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
-      </div>
-      <div className="nx-tile-header__subtitle">The Subtitle</div>
-      <div className="nx-tile__actions">
-        <button className="nx-btn nx-btn--tertiary">Action 1</button>
-        <button className="nx-btn nx-btn--tertiary">Action 2</button>
-        <button className="nx-btn nx-btn--tertiary">Action 3</button>
-      </div>
-    </div>
-    <div className="nx-tile-content">
-      Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
-    </div>
-  </div>
-  <div className="nx-tile">
-    <div className="nx-tile-header">
-      <div className="nx-tile-header__title">
-        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
-      </div>
-      <div className="nx-tile-header__subtitle">The Subtitle</div>
-      <div className="nx-tile__actions">
-        <button className="nx-btn nx-btn--tertiary">Action 1</button>
-        <button className="nx-btn nx-btn--tertiary">Action 2</button>
-        <button className="nx-btn nx-btn--tertiary">Action 3</button>
-        <button className="nx-btn nx-btn--tertiary">Action 4</button>
-      </div>
-    </div>
-    <div className="nx-tile-content">
-      Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
-    </div>
-  </div>
-</>;
+;
 
 export default NxTileWithActionsExample;
--- a/gallery/src/styles/NxTile/NxTileWithActionsExample.tsx
+++ b/gallery/src/styles/NxTile/NxTileWithActionsExample.tsx
@@ -7,6 +7,32 @@
 import React from 'react';
 
 const NxTileWithActionsExample = () =>
+  <>
+  <div className="nx-tile">
+    <div className="nx-tile-header">
+      <div className="nx-tile-header__title">
+        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
+      </div>
+      <div className="nx-tile-header__subtitle">The Subtitle</div>
+    </div>
+    <div className="nx-tile-content">
+      Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
+    </div>
+  </div>
+  <div className="nx-tile">
+    <div className="nx-tile-header">
+      <div className="nx-tile-header__title">
+        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
+      </div>
+      <div className="nx-tile-header__subtitle">The Subtitle</div>
+      <div className="nx-tile__actions">
+        <button className="nx-btn nx-btn--tertiary">Action 1</button>
+      </div>
+    </div>
+    <div className="nx-tile-content">
+      Vehicle motion artisanal corrupted warehouse marketing jeans physical drone neon
+    </div>
+  </div>
   <div className="nx-tile">
     <div className="nx-tile-header">
       <div className="nx-tile-header__title">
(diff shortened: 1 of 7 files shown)


rpokorny:
Zeplin shows the gap between the header and the content as 21px.  In the impl it appears to be a margin of only 15px

rpokorny:
> Zeplin shows the gap between the header and the content as 21px. In the impl it appears to be a margin of only 15px

And yet, when there's a subtitle the gap between the subtitle and the content in zeplin is 25px :thinking:

rpokorny:
Gap between bottom of content and footer border is 17px in zeplin, 32px in impl

rpokorny:
Subtitle should be 16px bold

rpokorny:
Header appears to be semi-bold in zeplin (font-weight 600).  Fully bold here.

rpokorny:
I don't think the "headline" style is supposed to be in a tile

gallery/src/styles/NxTile/NxTileWithActionsExample.tsx
```diff
@@ -8,26 +8,20 @@ import React from 'react';
...
-    <div className="nx-tile__actions">
-      <button className="nx-btn nx-btn--tertiary">Action 1</button>
-      <button className="nx-btn nx-btn--tertiary">Action 2</button>
-    </div>
     <div className="nx-tile-header">
-      <div className="nx-tile-header__title"><h2 className="nx-h2">NX Tile with Actions</h2></div>
+      <div className="nx-tile-header__title">
+        <h2 className="nx-h2">NX Tile with Actions - and this line is meant to wrap so that we can test truncation</h2>
+      </div>
+      <div className="nx-tile-header__subtitle">The Subtitle</div>
```
rpokorny, line 15: Would probably be good to have a longer subtitle in one example to prove they wrap

lib/src/base-styles/_nx-tile.scss
```diff
@@ -95,19 +95,28 @@ $nx-tile-background-color: $nx-white;
...
+  grid-column-start: 1;
+  grid-column-end: 1;
+  grid-row-start: 1;
+  grid-row-end: 1;
+  -ms-grid-row: 1;
+  -ms-grid-column: 1;
 
   h1, h2 {
-    display: inline-block;
+    @extend .nx-truncate-ellipsis;
```
rpokorny, line 118: Do you need this mixin both here and on the parent?

lib/src/base-styles/_nx-tile.scss
```diff
@@ -117,6 +126,7 @@ $nx-tile-background-color: $nx-white;
     .nx-tile-header__description {
```
rpokorny, line 126: What is this class for? I don't see it demonstrated on the nx-tile page

lib/src/base-styles/_nx-tile.scss
```diff
@@ -185,10 +206,10 @@ $nx-tile-background-color: $nx-white;
...
   @include container-horizontal;
 
-  border: 0 solid $nx-grey-4;
-  border-top-width: 1px;
-  margin-top: 20px;
-  padding: 12px 4px 0 4px;
+  border-top: 1px solid #e9e9e9;
+  margin-top: $nx-spacing-xl;
+  padding: $nx-spacing-xl 0;
+  text-align: right;
```
rpokorny, line 212: Is this always the case?  Is the footer always going to be just buttons?  Why are these button to the right when footer buttons everywhere else are to the left (or is that changing globally)?  If it is changing globally, maybe we want to put this style on `.nx-btn-bar` instead of here.

glenhunter:
> Subtitle should be 16px semibold

https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/34b408466d47ea18dada72b93351c9a2be232fc5

lib/src/base-styles/_nx-tile.scss
glenhunter: > Do you need this mixin both here and on the parent?

Frustratingly Yes. If it's only on `.nx-tile-header__title` it truncates but doesn't show the ellipsis. If it's only on the `<h#>` it doesn't truncate. If it's on both it truncates and shows the ellipsis.

lib/src/base-styles/_nx-tile.scss
glenhunter: > Is this always the case? Is the footer always going to be just buttons? Why are these button to the right when footer buttons everywhere else are to the left (or is that changing globally)? If it is changing globally, maybe we want to put this style on `.nx-btn-bar` instead of here.

As far as I understand it's global @dturner32 ?

We would want to add it to `.nx-btn-bar` for consistency but we don't nest `.nx-btn-bar` inside of footers so I don't see the connection WRT to this PR. `.nx-btn-bar` is for when there isn't a container like a footer to hold the buttons.

gallery/src/styles/NxTile/NxTileWithActionsExample.tsx
glenhunter: > Would probably be good to have a longer subtitle in one example to prove they wrap

Added to the subtitle example: https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/0d4cc745e574428f292fb47a85f4fc77c75b41cf

glenhunter:
> I don't think the "headline" style is supposed to be in a tile

I actually think we should remove it. I think we replaced all instances of this with `.nx-page-title` in IQ, and I can't think of anywhere we want to use this that `.nx-page-title` would apply. @dturner32 ?

lib/src/base-styles/_nx-tile.scss
glenhunter: > What is this class for? I don't see it demonstrated on the nx-tile page

It's a holdover from IQ and like the title only tile I think it should be removed.

glenhunter: > > What is this class for? I don't see it demonstrated on the nx-tile page
> 
> It's a holdover from IQ and like the title only tile I think it should be removed.

Which I have done here: https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/d7fc4b491b94534ea2784871f82a8374bb6897af

glenhunter:
> Header appears to be semi-bold in zeplin (font-weight 600). Fully bold here.
> 
> Also looks like the font-size is supposed to be 22px

I'm going to say that the font-size is an error in Zeplin. @dturner32?

glenhunter:
> Gap between bottom of content and footer border is 17px in zeplin, 32px in impl

https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/7342069b75f91f85424f382ad3c442f2da3a9a7d

glenhunter:
> > I don't think the "headline" style is supposed to be in a tile
> 
> I actually think we should remove it. I think we replaced all instances of this with `.nx-page-title` in IQ, and I can't think of anywhere we want to use this that `.nx-page-title` would apply. @dturner32 ?

I removed it here: https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/187f059916da60128f2ac27c36a5486da8b52490

dturner32:
> Subtitle should be 16px semibold

correct, 'subtitle' should be 16 and semibold. I have it defined as an H3 in the type scale. H3 will always be 16px and has two font weight variations: semibold, in this case used as a subtitle following an H2, and Bold used for subsection headers and labels

lib/src/base-styles/_nx-tile.scss
dturner32: Yes, the buttons are being moved to the right as a global change. This helps with visual flow as a user is reading from left to right. Their eye more naturally falls on the button placed in the bottom right.

dturner32:
> > I don't think the "headline" style is supposed to be in a tile
> 
> I actually think we should remove it. I think we replaced all instances of this with `.nx-page-title` in IQ, and I can't think of anywhere we want to use this that `.nx-page-title` would apply. @dturner32 ?

to Glen's point, nx-page-title would be what is applied

dturner32:
> > Header appears to be semi-bold in zeplin (font-weight 600). Fully bold here.
> > Also looks like the font-size is supposed to be 22px
> 
> I'm going to say that the font-size is an error in Zeplin. @dturner32?

the tile headers should always be styled as H2s (22px; Semibold) if they are part of the first tile/section on the page. Any tiles/sections following that one would then use H3s (16px; bold) as subsection headers. If there is an instance of a bolded H2 in a different font size then thats an error. I'll keep my eye out for any discrepancies in Zeplin

rpokorny:
> > Subtitle should be 16px semibold
>
> 34b4084

I'm seeing 14px

rpokorny:
> correct, 'subtitle' should be 16 and semibold. I have it defined as an H3 in the type scale.

We can style it that way, but that's not what `<h3>` is for.  The subtitle is not a separate header.

glenhunter:
> > > Subtitle should be 16px semibold
> > 
> > 
> > [34b4084](https://github.com/sonatype/sonatype-react-shared-components/commit/34b408466d47ea18dada72b93351c9a2be232fc5)
> 
> I'm seeing 14px

I have set it to: $nx-font-size-heading-level-3 which is 18px right now, but when RSC-106 is merged it will become 16px

rpokorny:
Header font size is still 20px when it needs to be 22px

glenhunter:
> Header font size is still 20px when it needs to be 22px
> 
> Edit: I guess this is supposed to be addressed in RSC-106

Yes, this and the sub-title font-size are addressed in RSC-106

rpokorny:
> the tile headers should always be styled as H2s (22px; Semibold) if they are part of the first tile/section on the page. Any tiles/sections following that one would then use H3s (16px; bold) as subsection headers. If there is an instance of a bolded H2 in a different font size then thats an error. I'll keep my eye out for any discrepancies in Zeplin

Per discussion in wall walk, tile headers should always be H2s.

lib/src/base-styles/_nx-tile.scss
```diff
@@ -93,30 +60,28 @@ $nx-tile-background-color: $nx-white;
...
+  grid-column-start: 1;
+  grid-column-end: 1;
+  grid-row-start: 1;
+  grid-row-end: 1;
+  -ms-grid-row: 1;
+  -ms-grid-column: 1;
 
   h1, h2 {
-    display: inline-block;
+    @extend .nx-truncate-ellipsis;
```
rpokorny, line 83: Don't we usually use the mixin for this?

```suggestion
    @include truncate-ellipsis;
```

lib/src/base-styles/_nx-tile.scss
```diff
@@ -93,30 +60,28 @@ $nx-tile-background-color: $nx-white;
...
+  -ms-grid-columns: 1fr auto;
   min-height: $nx-tile-header-height;
-  margin: 8px 0 15px 0;
+  margin-bottom: $nx-spacing-md;
   overflow: hidden;
 }
 
 .nx-tile-header__title {
-  display: inline-block;
+  @extend .nx-truncate-ellipsis;
```
rpokorny, line 74:
```suggestion
  @include truncate-ellipsis;
```

lib/src/base-styles/_nx-tile.scss
```diff
@@ -169,8 +144,8 @@ $nx-tile-background-color: $nx-white;
```
rpokorny, line 144: Doesn't look like this class is demonstrated anywhere.

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -72,8 +72,8 @@ $nx-header-border: #f1f3fa;
 
 //  Text
 $nx-heading-color: #2C2C2C;
-$nx-text-color: #474747;
-$nx-text-color-light: $nx-slate;
+$nx-text-color: #2c2c2c;
```
rpokorny, line 75: Does that usage of #474747 that we switched to `$nx-text-color` in the other PR need to be updated to `$nx-text-color-light` here?

glenhunter: > Does that usage of #474747 that we switched to `$nx-text-color` in the other PR need to be updated to `$nx-text-color-light` here?

Yes: https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/c9d9a4482410d10bfc33ccca51aa2ab3d80066cd

lib/src/base-styles/_nx-tile.scss
glenhunter: > Doesn't look like this class is demonstrated anywhere.

Doesn't apply anymore since we moved to the grid layout: https://github.com/sonatype/sonatype-react-shared-components/pull/34/commits/8f266140382cfa27762566dce6c8c9cce5c2b98c

rpokorny: > Doesn't apply anymore since we moved to the grid layout: 8f26614

The class I'm commenting on here is still in the code. That commit appears to be unrelated.

PR #37, opened 2020-06-05
Publish only for master

jorlina:
I think I had originally meant for publish to be in the same deploy block that only publishes to npmjs from master, but then changed my mind and moved it to the onSuccess block.  Anyway, this PR add the check to publish only from master and here's a Jenkins job that ran successfully for my PR branch but did NOT kick off a publish job:

https://jenkins.ci.sonatype.dev/job/uxui/job/sonatype-react-shared-components/job/publish-only-for-master/2/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -92,9 +92,11 @@ dockerizedBuildPipeline(
   testResults: ['lib/junit.xml'],
   onSuccess: {
     githubStatusUpdate('success')
-    build job:'/uxui/publish-gallery-to-s3', propagate: false, wait: false, parameters: [
-      run(name: 'Producer', runId: "${currentBuild.fullProjectName}${currentBuild.displayName}")
-    ]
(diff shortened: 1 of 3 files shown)


rpokorny:
> I think I had originally meant for publish to be in the same deploy block that only publishes to npmjs from master, but then changed my mind and moved it to the onSuccess block.

Why'd you change your mind?  This feels like deployment to me.

jorlina:
Because I thought it might be useful to be able to publish to the s3 bucket when building branches as well.  Of course, I didn't carry that train of thought through and figure out a way to keep publishes to branches from clobbering the stuff from master.  This is yet another instance of the half-baked-ness that is the hallmark of all Joel-ware ;)

jorlina:
Really, I should be able to publish to different sub-directories in the bucket inherit the branch name ... the part I haven't figured out is how to configure things in Cloudfront so that those sub-directories are accessible alongside master in a way that all the relative links resolve.

PR #39, opened 2020-06-08
Redesign nx-btn RSC-107

glenhunter:
https://issues.sonatype.org/browse/RSC-107

New buttons styles based on designs from Derrek. I also split out `nx-btn--icon-only` into its own example since the dimensions have changed on it.

### DIFF
--- a/lib/src/base-styles/_nx-btn.scss
+++ b/lib/src/base-styles/_nx-btn.scss
@@ -11,87 +11,83 @@
   @include semi-bold();
   @include container-horizontal;
 
-  background-color: $nx-grey-1;
-  border: $nx-border-dark;
-  border-radius: 4px;
-  color: $nx-text-color-light;
+  background-color: $nx-white;
+  border: 1px solid #045dd1;
+  border-radius: 6px;
+  color: #045dd1;
   cursor: pointer;
   display: inline-block;
-  font-size: 14px;
-  height: 32px;
+  font-size: 16px;
+  height: 40px;
   line-height: 16px;
-  margin: 0 8px;
-  padding: 0 12px;
+  margin: 0 $nx-spacing-md;
+  padding: 0 $nx-spacing-l;
   text-align: center;
   vertical-align: middle;
   min-width: 100px;
 
-  + .nx-btn {
-    // buttons should be 8px from adjacent items, except other buttons which they should only be 4px from
-    margin-left: -4px;
+  &:hover {
+    background-color: #c5deff;
   }
 
-  &:hover, &:focus, &:active {
-    border-color: $nx-slate;
-    color: $nx-text-color;
+  &:focus {
+    border: 1px solid #43cbf5;
+    outline: 0;
   }
 
   &:active {
-    background-color: $nx-grey-2;
+    background-color: #e8f2ff;
   }
 
   &.disabled, &[disabled] {
-    background-color: $nx-grey-1;
-    border-color: $nx-grey-2;
-    color: $nx-grey-4;
+    background-color: #e9e9e9;
+    border-color: #949494;
+    color: #949494;
     cursor: default;
-
-    &:focus {
-      outline: 0;
-    }
   }
 }
 
 .nx-btn--primary {
-  background-color: $nx-background-color-interactive;
-  border-color: transparent;
-  border-width: 2px;
-  color: $nx-off-white;
+  background-color: #045dd1;
+  border: 1px solid #045dd1;
+  color: $nx-white;
   padding: 0 11px;
 
-  &:hover, &:focus, &:active {
-    border-color: $nx-background-color-interactive-active;
-    color: $nx-off-white;
+  &:hover {
+    background-color: #034396;
   }
 
   &:active {
-    background: $nx-background-color-interactive-active;
-    border-color: $nx-background-color-interactive-active;
+    background-color:  #2783fb;
   }
 
   &.disabled, &[disabled] {
-    background-color: $nx-background-color-interactive-disabled;
-    border-color: transparent;
-    color: $nx-off-white;
+    background-color: #e9e9e9;
+    border-color: #949494;
+    color: #949494;
+    cursor: default;
   }
 }
 
 .nx-btn--tertiary {
-  background-color: transparent;
-  border-color: transparent;
+  border-color: #d1d6f0;
+  color: #2c2c2c;
 
-  &:hover, &:focus {
-    border-color: $nx-grey-4;
+  &:hover {
+    background-color: $nx-white;
+    border-color: #2c2c2c;
   }
 
   &:active {
-    background-color: transparent;
-    border-color: $nx-slate;
+    background-color: #f1f3fa;
+    border-color: #2c2c2c;
   }
 
   &.disabled, &[disabled] {
-    background-color: transparent;
-    border-color: transparent;
+    background-color: #e9e9e9;
+    border-color: #949494;
+    color: #949494;
+    cursor: default;
   }
 }
 
--- a/lib/src/base-styles/_nx-btn.scss
+++ b/lib/src/base-styles/_nx-btn.scss
@@ -92,22 +92,29 @@
 }
 
 .nx-btn--error {
-  background-color: $nx-fail-color;
-  border-color: $nx-fail-color;
+  background-color: #cd0028;
+  border: solid 1px #cd0028;
   color: $nx-off-white;
 
   &:hover, &:focus, &:active {
-    background-color: $nx-fail-color;
-    border-color: $nx-fail-color-active;
+    background-color: #a80222;
+    border: solid 1px #a80222;
     color: $nx-off-white;
   }
 
+  &:focus {
+    background-color: #cd0028;
+    border: solid 1px #43cbf5;
+  }
+
   &:active {
-    background-color: $nx-fail-color-active;
+    background-color: #fc496c;
+    border: solid 1px #fc496c;
   }
 }
 
 .nx-btn--icon-only {
+  height: 32px;
   padding: 0;
   width: 32px;
   min-width: 0;
--- a/lib/src/base-styles/_nx-btn.scss
+++ b/lib/src/base-styles/_nx-btn.scss
@@ -20,7 +20,7 @@
   font-size: 16px;
   height: 40px;
   line-height: 16px;
-  margin: 0 $nx-spacing-md;
+  margin: 0 $nx-spacing-xs;
   padding: 0 $nx-spacing-l;
   text-align: center;
   vertical-align: middle;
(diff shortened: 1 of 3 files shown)


lib/src/base-styles/_nx-btn.scss
```diff
@@ -11,107 +11,110 @@
...
+  background-color: $nx-white;
+  border: 1px solid #045dd1;
+  border-radius: 6px;
+  color: #045dd1;
   cursor: pointer;
   display: inline-block;
-  font-size: 14px;
-  height: 32px;
+  font-size: 16px;
+  height: 40px;
```
rpokorny, line 21: The fact that these are 40px tall is reminiscient of my recent work on the NxAlerts.  Should we try to be consistent about our approaches between the two?  With the alerts, I avoided using an explicit height and rather relied on the default line-height of the text and some standard-variable-based top and bottom padding.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -11,107 +11,110 @@
...
-    margin-left: -4px;
+  &:hover {
+    background-color: #c5deff;
   }
 
-  &:hover, &:focus, &:active {
-    border-color: $nx-slate;
-    color: $nx-text-color;
+  &:focus {
+    border: 1px solid #43cbf5;
```
rpokorny, line 34: FYI this border inevitably appears on click as well, since the click focuses the button.  Makes sense to me but strictly speaking it's not what's showing in the mockup.

rpokorny:
Are all these colors going to become variables at some point?

lib/src/base-styles/_nx-btn.scss
```diff
@@ -11,107 +11,110 @@
...
+    border-color: #2c2c2c;
   }
 
   &.disabled, &[disabled] {
-    background-color: transparent;
-    border-color: transparent;
+    background-color: #e9e9e9;
+    border-color: #949494;
+    color: #949494;
+    cursor: default;
```
rpokorny, line 90: The disabled styles are repeated throughout this file.  That repetition could be decreased by using a mixin or possibly by just rearranging the selectors.

rpokorny:
Disabled error button has a hover state:
image

glenhunter:
> Disabled error button has a hover state:
> image

Not anymore: https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/8fbf3a0cb44202c1c53a9eaa0882a5f8f797a4e8

lib/src/base-styles/_nx-btn.scss
glenhunter: > The disabled styles are repeated throughout this file. That repetition could be decreased by using a mixin or possibly by just rearranging the selectors.

I put them into a placeholder class which I @extend. However this messed with the cascade so I had to add a bit more specificity: https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/8fbf3a0cb44202c1c53a9eaa0882a5f8f797a4e8

glenhunter:
> Are all these colors going to become variables at some point?

There is a whole colour palette task, I'm just assuming that all of this colour refactoring will take place there

lib/src/base-styles/_nx-btn.scss
glenhunter: > The fact that these are 40px tall is reminiscient of my recent work on the NxAlerts. Should we try to be consistent about our approaches between the two? With the alerts, I avoided using an explicit height and rather relied on the default line-height of the text and some standard-variable-based top and bottom padding.

I get to 42px with default values, I think that's near enough: https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/90a22b5e75730be70b37ebf362972aea477b3b7a

rpokorny:
Primary buttons are broken: 
image

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
+    border: 1px solid #43cbf5;
+    outline: 0;
   }
 
   &:active {
-    background-color: $nx-grey-2;
+    background-color: #e8f2ff;
   }
 
   &.disabled, &[disabled] {
```
rpokorny, line 50: See https://github.com/sonatype/sonatype-react-shared-components/commit/68ca7ea49ada98929d132a41cfbf49c531ba17a3 for a slight improvement to the pattern you have here

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
-  height: 32px;
+  font-size: 16px;
   line-height: 16px;
-  margin: 0 8px;
-  padding: 0 12px;
+  margin: 0 $nx-spacing-xs;
+  padding: $nx-spacing-sm $nx-spacing-l;
   text-align: center;
   vertical-align: middle;
   min-width: 100px;
```
rpokorny, line 35: I don't think any of @dturner32 's mockups have text short enough to demonstrate a min-width, so we should probably check with him that this is still desired.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
-  font-size: 14px;
-  height: 32px;
+  font-size: 16px;
   line-height: 16px;
-  margin: 0 8px;
-  padding: 0 12px;
+  margin: 0 $nx-spacing-xs;
+  padding: $nx-spacing-sm $nx-spacing-l;
   text-align: center;
   vertical-align: middle;
```
rpokorny, line 34: Awhile ago I noticed that this style might not be having a desirable effect.  Without it, the text within the button is baseline-aligned with any other text that it is inline with, which I would think would be what we would want.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
+  background-color: $nx-white;
+  border: 1px solid #045dd1;
+  border-radius: 6px;
+  color: #045dd1;
   cursor: pointer;
   display: inline-block;
-  font-size: 14px;
-  height: 32px;
+  font-size: 16px;
   line-height: 16px;
```
rpokorny, line 30: This should probably be gotten rid of and the vertical padding decreased to 8px

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
   }
 }
 
 .nx-btn--primary {
-  background-color: $nx-background-color-interactive;
-  border-color: transparent;
-  border-width: 2px;
-  color: $nx-off-white;
+  background-color: #045dd1;
+  border: 1px solid #045dd1;
```
rpokorny, line 57:
```suggestion
  border-color: #045dd1;
```

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
-    border-color: transparent;
+    @extend %nx-btn--disabled
   }
 }
 
 .nx-btn--error {
-  background-color: $nx-fail-color;
-  border-color: $nx-fail-color;
+  background-color: #cd0028;
+  border: solid 1px #cd0028;
```
rpokorny, line 95:
```suggestion
  border-color: #cd0028;
```

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
   color: $nx-off-white;
 
   &:hover, &:focus, &:active {
-    background-color: $nx-fail-color;
-    border-color: $nx-fail-color-active;
+    background-color: #a80222;
+    border: solid 1px #a80222;
```
rpokorny, line 100:
```suggestion
    border-color: #a80222;
```

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
     color: $nx-off-white;
   }
 
+  &:focus {
+    background-color: #cd0028;
+    border: solid 1px #43cbf5;
```
rpokorny, line 106:
```suggestion
    border-color: #43cbf5;
```

lib/src/base-styles/_nx-btn.scss
```diff
@@ -7,111 +7,122 @@
...
+  }
+
   &:active {
-    background-color: $nx-fail-color-active;
+    background-color: #fc496c;
+    border: solid 1px #fc496c;
```
rpokorny, line 111:
```suggestion
    border-color: #fc496c;
```

lib/src/base-styles/_nx-btn.scss
glenhunter: > I don't think any of @dturner32 's mockups have text short enough to demonstrate a min-width, so we should probably check with him that this is still desired.

With the 24px of l/r padding I'm going to assume that we can remove this since any text will create a large enough click target

glenhunter:
> Primary buttons are broken:
> image

Weird. Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/2ad7df6461cea164809db9d15c650f6d6c81cf29

lib/src/base-styles/_nx-btn.scss
glenhunter: > This should probably be gotten rid of and the vertical padding decreased to 8px

Interesting side-effect, button is now 40px tall: https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/fff7ae93ed8ea9dfefb1e1b544413219a37e8032

lib/src/base-styles/_nx-btn.scss
glenhunter: > Awhile ago I noticed that this style might not be having a desirable effect. Without it, the text within the button is baseline-aligned with any other text that it is inline with, which I would think would be what we would want.

Subtle, but yeah, https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/fff7ae93ed8ea9dfefb1e1b544413219a37e8032

lib/src/base-styles/_nx-btn.scss
glenhunter: > > I don't think any of @dturner32 's mockups have text short enough to demonstrate a min-width, so we should probably check with him that this is still desired.
> 
> With the 24px of l/r padding I'm going to assume that we can remove this since any text will create a large enough click target

https://github.com/sonatype/sonatype-react-shared-components/pull/39/commits/fff7ae93ed8ea9dfefb1e1b544413219a37e8032

lib/src/base-styles/_nx-btn.scss
glenhunter: > See [68ca7ea](https://github.com/sonatype/sonatype-react-shared-components/commit/68ca7ea49ada98929d132a41cfbf49c531ba17a3) for a slight improvement to the pattern you have here

Nice, thanks.

PR #48, opened 2020-06-11
Standard example tile - RSC-125

rpokorny:
https://issues.sonatype.org/browse/RSC-125

This PR creates a more sophisticated structure for the `GalleryExampleTile`.  All examples that used `GalleryExampleTile` have been updated accordingly.  Other examples should be updated in follow-on tickets.


gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -36,13 +46,28 @@ export const GalleryTile: FunctionComponent<GalleryTileProps> =
...
+          const props = typeof example === 'string' ? { content: example } : example;
+          return <CodeExample key={idx} { ...props } />;
+        });
+
+    return (
+      <GalleryTile title={title} className={className}>
+        <p className="nx-p">{description}</p>
+
+        <h3 className="nx-h3 nx-tile__section-header">Example:</h3>
+        <div>{children}</div>
```
glenhunter, line 68: I think this should have the className of `.gallery-example`, which you have on the `<section>` element below.

gallery/src/CodeExample.tsx
```diff
@@ -28,12 +28,20 @@ const removeLicense = (content: string) =>
...
   function CodeExample({ content, language }): ReactElement<Props> {
-    const licenseStrippedContent = removeLicense(content);
+    const licenseStrippedContent = removeLicense(content),
+        headerLanguagePart = language ? ` (${language.toUpperCase()})` : '';
 
     return (
-      <SyntaxHighlighter language={language || 'tsx'} style={atomDark}>
-        {licenseStrippedContent}
-      </SyntaxHighlighter>
+      <section className="gallery-example">
```
glenhunter, line 35: I think this class should be `.gallery-example-code`.

glenhunter:
One thing that continues to bother me about this is that both the Description and Example tiles are both just `.nx-tile`. I would much prefer it if we built in classes to each (even if we don't use them right now) as a hedge against future changes. I suggest `.gallery-tile--description` & `.gallery-tile--example`

gallery/src/gallery-components/GalleryTiles.tsx
rpokorny: > I think this should have the className of .gallery-example, which you have on the <section> element below.

Why? All `.gallery-example` does is set the font size in the code snippets.  This `<div>` does not contain code snippets.

rpokorny: > Why? All .gallery-example does is set the font size in the code snippets. This <div> does not contain code snippets.

I guess I had originally moved it because I thought it was making the description paragraph text small.  But it's not the cause there.  So I moved it back up to the tile itself in fa1d62f9eb and 7865b104c3

rpokorny:
> One thing that continues to bother me about this is that both the Description and Example tiles are both just .nx-tile. I would much prefer it if we built in classes to each (even if we don't use them right now) as a hedge against future changes. I suggest .gallery-tile--description & .gallery-tile--example

With my most recent changes the example tile itself now (once again) has the `gallery-example` class

gallery/src/CodeExample.tsx
rpokorny: > I think this class should be .gallery-example-code.

What would be the purpose of this class?

glenhunter: > > I think this class should be .gallery-example-code.
> 
> What would be the purpose of this class?

I'm just trying to think ahead and anticipate future design changes. If we put these CSS classes in now they give us more flexibility in the future. Not having them might mean that we require the assistance of a JS dev to make a comparatively simple design change.

PR #50, opened 2020-06-11
NxLoadError retry styling - RSC-137

rpokorny:
https://issues.sonatype.org/browse/RSC-137


rpokorny:
Note that the alignment of the icon is a little bit off in IE11 due to lack of support for `align-items: first baseline`.  I'm willing to live with that

lib/src/components/NxLoadError/NxLoadError.scss
```diff
@@ -0,0 +1,21 @@
+.nx-load-error__message {
+  // TODO no magic numbers
+  max-width: calc(100% - 32px);
```
glenhunter, line 3: I noticed that in cases where the button wraps down (very clever solution BTW), but the text is long enough to read the right edge of the box it doesn't look very good.
Screen Shot 2020-06-11 at 2 32 19 PM
However if you add 8px of margin-bottom to `.nx-load-error__message` it adds some breathing space but doesn't effect non-wrapping instances.
Screen Shot 2020-06-11 at 2 46 07 PM

glenhunter, line 3:
```suggestion
  max-width: calc(100% - 32px);
```
  margin-bottom: $nx-spacing-sm;
  max-width: calc(100% - 32px);

rpokorny: > However if you add 8px of margin-bottom to .nx-load-error__message it adds some breathing space but doesn't effect non-wrapping instances.

631fb79

PR #51, opened 2020-06-12
No jira add in version bump script

timlevett:
Super trivial script to bump the version easily and keep gallery and lib in sync.


rpokorny:
We already basically have this.  When you bump the version in the lib, the version in the gallery gets bumped automatically to match.

timlevett:
>We already basically have this. When you bump the version in the lib, the version in the gallery gets bumped automatically to match.

shut the front door, how do I get that to work?!

PR #52, opened 2020-06-12
Close Button - RSC-127

rpokorny:
https://issues.sonatype.org/browse/RSC-127


lib/src/components/NxCloseButton/NxCloseButton.scss
```diff
@@ -0,0 +1,12 @@
...
+@import '../../scss-shared/nx-colors';
+@import '../../scss-shared/nx-container-helpers';
+
+.nx-btn--close {
+  border-color: transparent;
+
+  .nx-icon {
+    height: $nx-spacing-md;
+    width: $nx-spacing-md;
+  }
```
glenhunter, line 11: We should not use spacing variables for height/width. They should be kept separate.

lib/src/icons/Close.scss
```diff
@@ -0,0 +1,3 @@
+.nx-icon--close {
+  vertical-align: -0.125em;
```
glenhunter, line 2: Why use `em` here? We almost never use `em`.

rpokorny: > Why use em here? We almost never use em.

1. This style precisely matches how fontawesome aligns their icons
2. This would continue to work if the close button size was changed.

lib/src/components/NxCloseButton/NxCloseButton.scss
rpokorny: > We should not use spacing variables for height/width. They should be kept separate.

What variables should I be using here?

rpokorny: > What variables should I be using here?

Actually, I guess what I should really do is set the height to `1em` and then just use the font size, which is already conveniently the 16px we want.

lib/src/icons/Close.scss
```diff
@@ -0,0 +1,3 @@
+.nx-icon--close {
```
glenhunter, line 1: Does this file need the license header?

lib/src/components/NxCloseButton/NxCloseButton.scss
glenhunter: > > What variables should I be using here?
> 
> Actually, I guess what I should really do is set the height to `1em` and then just use the font size, which is already conveniently the 16px we want.

Or I was going to say that you should use `$nx-font-size`

rpokorny: > Actually, I guess what I should really do is set the height to 1em and then just use the font size, which is already conveniently the 16px we want.

25270990

lib/src/icons/Close.scss
rpokorny: > Does this file need the license header?

6443593f

PR #54, opened 2020-06-15
Fix gallery-nav-link alignment/cut-off issue - RSC-156

rpokorny:
https://issues.sonatype.org/browse/RSC-156

This PR makes the font size of the container equal to that of its contents, removing the problematic baseline alignment problems that come when it isn't

### DIFF
--- a/gallery/src/main.scss
+++ b/gallery/src/main.scss
@@ -25,7 +25,6 @@
 }
 
 .nx-tree-view__child.gallery-nav-link {
-  font-size: 14px;
   height: 32px;
   line-height: 32px;
   margin-right: 0;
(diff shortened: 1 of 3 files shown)


gallery/src/main.scss
```diff
@@ -25,7 +25,7 @@
 }
 
 .nx-tree-view__child.gallery-nav-link {
-  font-size: 16px;
+  font-size: 14px;
```
glenhunter, line 28: The mockup I have shows the text as 16px, which suggests that it's actually `.gallery-link` that's incorrect.

rpokorny: > The mockup I have shows the text as 16px, which suggests that it's actually .gallery-link that's incorrect.

8cd4e334

PR #57, opened 2020-06-17
nxTextInput redesign RSC-109 

glenhunter:
https://issues.sonatype.org/browse/RSC-109


rpokorny:
Valid icon is appearing on disabled inputs
image

gallery/src/components/NxTextInput/NxTextInputDisabledExample.tsx
```diff
@@ -10,6 +10,6 @@ import { NxTextInput } from '@sonatype/react-shared-components';
 
 export default function NxTextInputDisabledExample() {
   return (
-    <NxTextInput placeholder="Disabled input" value="" isPristine={true} disabled/>
+    <NxTextInput placeholder="Disabled input" value="" isPristine={false} disabled/>
```
rpokorny, line 13: Why this change?

lib/src/scss-shared/_nx-variables.scss
```diff
@@ -47,9 +47,9 @@ $nx-font-size-heading-level-4: 14px;
 $nx-line-height: 1.5;
 
 // Form variables
-$nx-form-element-height: 32px;
-$nx-form-element-width-normal: 219px;
-$nx-form-element-width-wide: 395px;
+$nx-form-element-height: 40px;
```
rpokorny, line 50: Now that buttons are 42px tall, should this be 42px?

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -5,33 +5,38 @@
...
-  border: $nx-border-dark;
-  border-left-width: 4px;
-  border-radius: 0;
-  box-shadow: none;
+.nx-text-input__input {
+  border: 1px solid #d1d6f0;
+  border-radius: $nx-border-radius;
+  color: $nx-text-color-light;
+  font-size: $nx-font-size;
   height: $nx-form-element-height;
```
rpokorny, line 13: Is there a reason this is explicit rather than relying on content line-height and padding?

rpokorny:
In zeplin it looks like the placeholder is supposed to be `#949494`.  Should be doable using the `::placeholder` pseudo-element, though it won't work in IE

rpokorny:
Zeplin question: what's the conceptual difference between the Complete state and the Valid state? 
image

rpokorny:
textareas are getting a monospace font for me: 
image

rpokorny:
When does the Entry State as shown in Zeplin take effect?  As soon as you start typing, validation kicks in and you get either the Valid State or the Invalid State.  Also, I doubt it would be possible/meaningful to differentiate between the Entry State and the Focus State.

rpokorny:
What's the reasoning behind putting the valid icon on the right but the invalid icon underneath?  To me it would make sense for them to show up in the same place.  On a related note, are we sure we want the overall height of this component to expand when there is a validation error (or, in my idea a validation non-error as well)?  We could build that space into the component always, and just have it initially empty.

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -5,33 +5,38 @@
...
-  padding: 0 8px;
+  padding: 0 $nx-spacing-md;
   width: $nx-form-element-width-normal;
 
+  &:hover {
+    border-color: #2c2c2c;
```
rpokorny, line 18: This doesn't seem to do anything

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -5,33 +5,38 @@
...
+  }
+
   &:focus {
+    border-color: #43cbf5;
```
rpokorny, line 22: This also doesn't seem to do anything

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -5,33 +5,38 @@
...
     outline: none;
   }
 
   &:disabled {
-    background: $nx-off-white;
-    border-color: $nx-grey-3;
-    border-left: 4px double $nx-grey-3;
+    background-color: #e9e9e9;
+    border: solid 1px #949494;
```
rpokorny, line 28:
```suggestion
    border-color: #949494;
```

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -5,33 +5,38 @@
...
 }
 
-textarea.nx-text-input {
-  min-height: 64px;
-  padding: 8px;
+textarea.nx-text-input__input {
+  min-height: 220px;
+  padding: $nx-spacing-xs $nx-spacing-md;
+  max-width: 700px;
 }
```
rpokorny, line 40: `textarea`s are getting `line-height: normal`, need explicit overridding

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -16,18 +16,6 @@ describe('NxTextInput', function() {
       },
       getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxTextInput, minimalProps);
 
-  it('renders a text input wrapped in a NxTooltip by default', function() {
```
rpokorny, line 19: The non-tooltip aspects of these two tests need to be restored

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -54,27 +42,6 @@ describe('NxTextInput', function() {
     expect(getShallowComponent({ isPristine: true }).find('input')).toHaveClassName('nx-text-input--pristine');
   });
 
-  it('sets the nx-tooltip--validation-error class on the tooltip', function() {
-    expect(getShallowComponent()).toHaveClassName('nx-tooltip--validation-error');
-  });
-
-  it('sets the tooltips `open` prop to false and title to null if there are no validation errors', function() {
```
rpokorny, line 61: Tests for the new validation display will need to be added

lib/src/icons/validationcheck.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-icon--validation-check {
+  vertical-align: -0.125em;
```
rpokorny, line 8: I'm wondering if we should just add this to `.nx-icon`

lib/src/icons/validationcheck.tsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+import React from 'react';
+
+import './validationcheck.scss';
+
+const ValidationCheck = () =>
+  <svg viewBox="0 0 16 16" className="nx-icon nx-icon--validation-check">
+    <path fill="currentColor"
+          d="M13.7142784,4 L16,6.25 L6.09520643,16 L3.80948484,13.75 L3.81,13.749 L-5.32907052e-14,9.99999135
+             L2.28572159,7.74999135 L6.095,11.5 L13.7142784,4 Z">
```
rpokorny, line 15: The FA check wasn't good?

rpokorny:
Is `lib/src/assets/img/Validation.png` being used anywhere?

lib/src/icons/validationcheck.tsx
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+import './validationcheck.scss';
```
rpokorny, line 9: PascalCase filenames

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,56 @@
...
+    border-color: #d1d6f0;
+    color: $nx-text-color-light;
+  }
+}
+
+.nx-text-input__valid-message, .nx-text-input__invalid-message {
+  display: none;
+}
+
+.nx-icon--validation-check {
```
rpokorny, line 41: Shouldn't this block be in the dedicated scss file for that icon?

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,56 @@
...
+
+.nx-text-input__invalid-message-text {
+  color: #cd0028;
+  font-size: $nx-font-size-xs;
+}
+
+.nx-icon--invalid {
+  color: #cd0028;
+  height: 1em;
+}
```
rpokorny, line 60: MIssing final newline

gallery/src/components/NxTextInput/NxTextInputDisabledExample.tsx
glenhunter: > Why this change?

It seemed odd to me that a disabled button started in a pristine state (though I agree it's no less odd than the other way), so I was curious if it made a difference to the styling, then I forgot that I did it.

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: > Shouldn't this block be in the dedicated scss file for that icon?

Moved here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/d0e175eb0abc97c26d1ce9f69d6d05aeecd1a92e

lib/src/icons/validationcheck.scss
glenhunter: > I'm wondering if we should just add this to `.nx-icon`

It seems to pop up a lot so I moved it here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/d0e175eb0abc97c26d1ce9f69d6d05aeecd1a92e

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
glenhunter: > The non-tooltip aspects of these two tests need to be restored

I think I restored them correctly here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/4f062184bfa8471193a4e5b93197713ba09b58ba

lib/src/base-styles/_nx-text-input.scss
glenhunter: > `textarea`s are getting `line-height: normal`, need explicit overridding

Both inputs had lost their fonts and were pulling the line-height from _user agent_. I restored both settings here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/7f3a465a2bea91a07b2f52357e48f986d4d19806

glenhunter:
> textareas are getting a monospace font for me:
> image

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/7f3a465a2bea91a07b2f52357e48f986d4d19806

glenhunter:
> In zeplin it looks like the placeholder is supposed to be `#949494`. Should be doable using the `::placeholder` pseudo-element, though it won't work in IE

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/a0287f5ebbab7dc638d7ca8b2dd8fa01f1b852fb

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: > MIssing final newline

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/f0255974e1eba719a051474b81d4edf2cc33af4e

lib/src/icons/validationcheck.tsx
glenhunter: > PascalCase filenames

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/84ab2f4651110e054033acc7a8ef8e17635c5d72

dturner32:
> Zeplin question: what's the conceptual difference between the Complete state and the Valid state?
> image

complete state is basically the state for when a user has completed and entry. Valid state would be for cases where a specific entry is required in order to proceed (something like a password or username field for example).

lib/src/base-styles/_nx-text-input.scss
glenhunter: > Is there a reason this is explicit rather than relying on content line-height and padding?

Just because that's how it was done, but I removed the height (and associated variables and it's consistent with other elements now: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/9e5655f8bd6cf083b80038a89ff85fa94fa8941d

lib/src/scss-shared/_nx-variables.scss
glenhunter: > Now that buttons are 42px tall, should this be 42px?

Now that layout is padding + font it's 42px: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/9e5655f8bd6cf083b80038a89ff85fa94fa8941d

gallery/src/components/NxTextInput/NxTextInputDisabledExample.tsx
glenhunter: > > Why this change?
> 
> It seemed odd to me that a disabled button started in a pristine state (though I agree it's no less odd than the other way), so I was curious if it made a difference to the styling, then I forgot that I did it.

Reverted here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/a592c64696532f2e463a287af8e6805e72ef465c

glenhunter:
> Valid icon is appearing on disabled inputs
> image

Reverting the other change to the disabled example fixed this: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/a592c64696532f2e463a287af8e6805e72ef465c

dturner32:
> When does the Entry State as shown in Zeplin take effect? As soon as you start typing, validation kicks in and you get either the Valid State or the Invalid State. Also, I doubt it would be possible/meaningful to differentiate between the Entry State and the Focus State.
> 
> image

entry state should take place as soon as a user clicks into a field. Valid or error would only occur after the user hits enter or tabs to the next field. Unless the field is meant to trigger an error or valid state, then the field would transition from 'entry' to 'complete' 

The reason for the focus state is intended for user guidance. Specifically for cases when a user only has access to a keyboard and not a mouse. The idea is to have a visual indicator for when they tabbing to the next thing

rpokorny:
> Reverting the other change to the disabled example fixed this: a592c64

In that case, the question is what should it look like when the field is in fact not pristine and disabled.  i.e., when the field was previously not disabled, and the user typed something into it, and then something happened that caused it to get disabled while containing that dirty input.  In that scenario, do we want the valid check to appear or not?  What about any invalid state?

rpokorny:
> Unless the field is meant to trigger an error or valid state, then the field would transition from 'entry' to 'complete'

Currently, every field is meant to trigger one or either of those states (error or valid).  Any field that has no validators attached is simply considered to always be valid (once dirty).  Are you saying that you only want the valid state to appear on fields that have a possibility of becoming _invalid_, i.e. fields that have a validator function attached?

rpokorny:
> The reason for the focus state is intended for user guidance. Specifically for cases when a user only has access to a keyboard and not a mouse. The idea is to have a visual indicator for when they tabbing to the next thing

I'm not sure it's feasible or useful to distinguish the styles of a field that the user arrived at by mouse click vs by keyboard navigation.  Either way, the state in question is that the field has focus.

dturner32:
> > complete state is basically the state for when a user has completed and entry. Valid state would be for cases where a specific entry is required in order to proceed (something like a password or username field for example).
> 
> > Unless the field is meant to trigger an error or valid state, then the field would transition from 'entry' to 'complete'
> 
> Currently, every field is meant to trigger one or either of those states (error or valid). Any field that has no validators attached is simply considered to always be valid (once dirty). Are you saying that you only want the valid state to appear on fields that have a possibility of becoming _invalid_, i.e. fields that have a validator function attached?

yes, to me it makes sense that if a field doesn't have any validation conditions tied to it then it should just show as complete. Otherwise yes, the valid state should only be tied to fields that could potentially become invalid from incorrect entry.

dturner32:
> What's the reasoning behind putting the valid icon on the right but the invalid icon underneath? To me it would make sense for them to show up in the same place. On a related note, are we sure we want the overall height of this component to expand when there is a validation error (or, in my idea a validation non-error as well)? We could build that space into the component always, and just have it initially empty.

so Glen and I actually discussed this in another convo as well. Originally I had the check mark for the valid state inside the entry field. My understanding is that it is possible to build in the blank space in the field to account for the validation icon, but potentially trickier to implement. Alternatively, I would be in favor of positioning the error icon to the right of the entry field as well so that it is consistent with the valid state. The error message itself would stay aligned to the bottom left of the field.

lib/src/base-styles/_nx-text-input.scss
dturner32: > This doesn't seem to do anything

could you expand upon this a little more?

lib/src/icons/validationcheck.tsx
dturner32: > The FA check wasn't good?

FA check is good to use, when I adjust the dimensions of the icons however, I consider them custom at that point. I guess the question is, does it matter if I tweak the height/width. Reason being is that some of the icons appear smaller/larger than others and I would like to try and keep them all as uniform as possible

glenhunter:
> > Reverting the other change to the disabled example fixed this: [a592c64](https://github.com/sonatype/sonatype-react-shared-components/commit/a592c64696532f2e463a287af8e6805e72ef465c)
> 
> In that case, the question is what should it look like when the field is in fact not pristine and disabled. i.e., when the field was previously not disabled, and the user typed something into it, and then something happened that caused it to get disabled while containing that dirty input. In that scenario, do we want the valid check to appear or not? What about any invalid state?

To solve this properly we would need to the wrapping `<div>` to know when the `<input>` it contains is in a disabled state.

dturner32:
> > Reverting the other change to the disabled example fixed this: [a592c64](https://github.com/sonatype/sonatype-react-shared-components/commit/a592c64696532f2e463a287af8e6805e72ef465c)
> 
> In that case, the question is what should it look like when the field is in fact not pristine and disabled. i.e., when the field was previously not disabled, and the user typed something into it, and then something happened that caused it to get disabled while containing that dirty input. In that scenario, do we want the valid check to appear or not? What about any invalid state?

I have bit of confusion around this one. In the example above I see a disabled field with a check mark next to it which feels weird to me. The second part of that is I don't really have a general understanding of what a 'pristine' state is and the use case for it. Up until I arrived at sonatype, I had never heard of a pristine state used among typical UI patterns.

lib/src/base-styles/_nx-text-input.scss
rpokorny: > could you expand upon this a little more?

That comment was more for @glenhunter than for you.  Currently, despite this style, I was never able to get the hover styles to appear.  I believe they were always being overridden by other styles.  Same with the focus style below.

lib/src/icons/validationcheck.tsx
rpokorny: > I guess the question is, does it matter if I tweak the height/width.

We use SVG for all icons, so scaling is not considered a separate icon.  Even if you are talking about changing the aspect ratio, If it's still ultimately based on the FA icon I'd rather use said FA icon along with a `transform` rather than created a completely custom SVG that essentially duplicates the FA one.

rpokorny:
> I have bit of confusion around this one. In the example above I see a disabled field with a check mark next to it which feels weird to me. The second part of that is I don't really have a general understanding of what a 'pristine' state is and the use case for it. Up until I arrived at sonatype, I had never heard of a pristine state used among typical UI patterns.

The pristine state is the state where the user has not yet modified the value of the field since the form was first rendered.  I believe in your mockups you refer to it as the Default state.  Once the user modifies the field, its always in one of the other states, correct?

To rephrase the question about the disabled field a different way:  let's say there's a field that's currently in the Valid State - the user has touched it and entered some valid text.  Then, some button gets clicked or something that causes that field to become disabled.  Once that happens should the valid checkmark on that field still be visible or not?

rpokorny:
> To solve this properly we would need to the wrapping <div> to know when the <input> it contains is in a disabled state.

It wouldn't be hard to add a `nx-text-input--disabled` class

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -6,32 +6,52 @@
...
+}
+
+.nx-text-input__input {
+  background-color: transparent;
+  border: none;
+  color: $nx-text-color-light;
+  flex-grow: 1;
+  font-size: $nx-font-size;
+  line-height: $nx-line-height;
+  padding: $nx-spacing-xs $nx-spacing-md;
```
rpokorny, line 26: I feel like this padding should probably be on the box; with the input itself having none

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -6,17 +6,18 @@
  */
 @import '../../scss-shared/nx-variables';
 @import '../../scss-shared/_nx-form-validation.scss';
+@import '../../scss-shared/_nx-container-helpers.scss';
 
-.nx-text-input:not(:disabled) {
-  &.nx-text-input--valid {
+.nx-text-input {
+  &.valid:not(.pristine) {
```
rpokorny, line 12: This `:not` shouldn't be necessary, depending on how the `pristine` mixin is defined

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -6,17 +6,18 @@
...
     @include valid();
```
rpokorny, line 13: Note that the expected green border on valid fields is currently missing

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -6,17 +6,18 @@
...
   }
 
-  &.nx-text-input--invalid {
+  &.invalid {
     @include invalid();
```
rpokorny, line 17: Red border is missing

lib/src/base-styles/_nx-forms.scss
```diff
@@ -6,10 +6,6 @@
  */
 @import 'nx-text-input';
```
rpokorny, line 7: Just noticed this import is causing the nx-text-input styles to be present twice.  It should be removed

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -6,32 +6,52 @@
...
-  border-radius: 0;
-  box-shadow: none;
-  height: $nx-form-element-height;
-  padding: 0 8px;
   width: $nx-form-element-width-normal;
+}
+
+.nx-text-input__box {
+  align-items: center;
+  border: 1px solid #d1d6f0;
```
rpokorny, line 14: I believe this border color is only supposed to apply in the pristine state

rpokorny:
Form Layout Page is very broken: 
image

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -6,17 +6,18 @@
  */
 @import '../../scss-shared/nx-variables';
 @import '../../scss-shared/_nx-form-validation.scss';
+@import '../../scss-shared/_nx-container-helpers.scss';
```
rpokorny, line 9: Is this import still needed here?

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -7,10 +7,11 @@
 import React, { forwardRef, FormEvent, KeyboardEvent } from 'react';
 import classnames from 'classnames';
 import { omit } from 'ramda';
-
+import { faExclamationCircle, faCheck } from '@fortawesome/free-solid-svg-icons';
```
rpokorny, line 10:
```suggestion
import { faExclamationCircle, faCheck } from '@fortawesome/free-solid-svg-icons';

```

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -61,19 +63,26 @@ const NxTextInput = forwardRef<HTMLInputElement | HTMLTextAreaElement, Props>(
...
+              className: "nx-text-input__input",
+              onKeyPress: inputOnKeyPress
+            })}
+            <span className="nx-text-input__validation-icon">
+              <NxFontAwesomeIcon icon={faCheck} className="nx-icon--valid"/>
+              <NxFontAwesomeIcon icon={faExclamationCircle} className="nx-icon--invalid"/>
+            </span>
+          </div>
+          <div className="nx-text-input__invalid-message">
+            <span className="nx-text-input__invalid-message-text">{firstValidationError} &nbsp;</span>
```
rpokorny, line 83: I think this should work:
```suggestion
            <span className="nx-text-input__invalid-message-text">{firstValidationError || '&nbsp;'}</span>
```

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -16,78 +15,53 @@ describe('NxTextInput', function() {
...
 
   it('renders a textarea if type is "textarea"', function() {
-    expect(getShallowComponent({ type: 'textarea' })).toMatchSelector(NxTooltip);
     expect(getShallowComponent({ type: 'textarea' }).find('textarea')).toMatchSelector('textarea');
     expect(getShallowComponent({ type: 'textarea' }).find('textarea')).toHaveProp('type', undefined);
   });
 
   it('sets the value as specified', function() {
-    expect(getShallowComponent({ value: 'foo' }).find('input')).toHaveProp('value', 'foo');
+    expect(getShallowComponent({ value: 'foo' }).find('textarea')).toHaveProp('value', 'foo');
```
rpokorny, line 34: This shouldn't work

lib/src/icons/ValidationCheck.scss
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ .nx-icon--validation-check {
```
rpokorny, line 7: Still used?

lib/src/icons/ValidationCheck.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+import './ValidationCheck.scss';
+
+const ValidationCheck = () =>
```
rpokorny, line 11: Still used?

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -61,19 +63,26 @@ const NxTextInput = forwardRef<HTMLInputElement | HTMLTextAreaElement, Props>(
...
+            {React.createElement(element, {
+              ...newProps,
+              disabled,
+              ref,
+              type: typeAttr,
+              onChange: inputOnChange,
+              className: "nx-text-input__input",
+              onKeyPress: inputOnKeyPress
+            })}
+            <span className="nx-text-input__validation-icon">
```
rpokorny, line 77: What's the purpose of this wrapper?

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,70 @@
...
+  &, .nx-icon {
+    display: none;
+  }
+}
+
+.nx-text-input__invalid-message {
+  color: #cd0028;
+  font-size: $nx-font-size-xs;
+  margin-top: $nx-spacing-xs;
+  visibility: hidden;
```
rpokorny, line 59: Unnecessary I believe (along with the `visibility: visible` above)

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,70 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 @import 'nx-variables';
+@import 'nx-container-helpers';
 
 @mixin valid() {
-  border-left-color: $nx-valid-color;
-  border-left-style: solid;
+  .nx-text-input__input:not(:disabled) {
```
rpokorny, line 11: I'll continue to crusade against `:not`.  I suspect there's a way to avoid using it for this.

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,70 @@
...
+.nx-text-input__invalid-message {
+  color: #cd0028;
+  font-size: $nx-font-size-xs;
+  margin-top: $nx-spacing-xs;
+  visibility: hidden;
+}
+
+.nx-icon--invalid {
+  color: #cd0028;
+  height: 1em;
```
rpokorny, line 64: Redundant (here and below)

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,70 @@
...
+  margin-left: 0;
+  margin-right: $nx-spacing-md;
```
rpokorny: Tieing into my comment above: "I feel like this padding should probably be on the box; with the input itself having none" this margin goes against our standard margin and padding rules.  `nx-text-input__box` should use the `container-horizontal` mixin to negate undesireable default margins on the icons.

lib/src/components/NxTextInput/NxTextInput.tsx
glenhunter: > What's the purpose of this wrapper?

Because we're using flexbox the icons are seen as two separate columns if they aren't wrapped

lib/src/components/NxTextInput/NxTextInput.scss
glenhunter: > Red border is missing

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/0da768a36b206b39dff5be5afbac78162d38d69a

lib/src/components/NxTextInput/NxTextInput.scss
glenhunter: > Note that the expected green border on valid fields is currently missing

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/0da768a36b206b39dff5be5afbac78162d38d69a

lib/src/base-styles/_nx-forms.scss
glenhunter: > Just noticed this import is causing the nx-text-input styles to be present twice. It should be removed

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/0eda4995d66fc3c02b64aea0a476fbdf5dffdc79

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
glenhunter: > This shouldn't work

I like that your comment leaves it ambiguous about whether it does, or does not in fact work.

lib/src/base-styles/_nx-text-input.scss
glenhunter: > I feel like this padding should probably be on the box; with the input itself having none

The commit diff looks weird: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/be4cce51f3c368f60b3e7089c9fbb628121170eb

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
rpokorny: > I like that your comment leaves it ambiguous about whether it does, or does not in fact work.

Hey it's not my job to run your tests :grin:

glenhunter:
> Form Layout Page is very broken:
> image

Because it uses the bare HTML styles which at this point are so far out of whack with the component that I don't think they can ever come back. I will create a ticket to recreate this page with the proper components (and validation) and I will pull the page from the nav for now.

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: The problem with that is that because they are inside a container themselves the first icon gets stripped of its left margin, but keeps its right, and the opposite happens with the second icon, but we don't actually want either icon to have any l/r margin

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: > Redundant (here and below)

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/df7ed84e680e1b2c2d1c699d939d738fec3357b3

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: > Unnecessary I believe (along with the `visibility: visible` above)

The purpose of this is to hide/show the invalid message

lib/src/base-styles/_nx-text-input.scss
glenhunter: > I believe this border color is only supposed to apply in the pristine state

In Zeplin it shows that color for the default state.

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: > I'll continue to crusade against `:not`. I suspect there's a way to avoid using it for this.

While historically speaking crusades have had mixed results in this case we have achieved victory: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/600837d63c526088a976a7de4384b2426cb07531

lib/src/components/NxTextInput/NxTextInput.scss
glenhunter: > This `:not` shouldn't be necessary, depending on how the `pristine` mixin is defined

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/600837d63c526088a976a7de4384b2426cb07531

lib/src/base-styles/_nx-text-input.scss
glenhunter: > This doesn't seem to do anything

fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/600837d63c526088a976a7de4384b2426cb07531

lib/src/components/NxTextInput/NxTextInput.scss
glenhunter: > Is this import still needed here?

No, and neither was `nx-variables`: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/08e04017637c8308f996adc15728b6b7f0016ad1

lib/src/scss-shared/_nx-form-validation.scss
glenhunter: > > Unnecessary I believe (along with the `visibility: visible` above)
> 
> The purpose of this is to hide/show the invalid message

But you improved it here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/2e6ac1d6ba0a1dfa0cd45198f8e49831fc35c81d

lib/src/icons/ValidationCheck.tsx
glenhunter: > Still used?

Nope: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/2f4e5e38dead3a723c7732866beb25e848b3b3ad

lib/src/icons/ValidationCheck.scss
glenhunter: > Still used?

Nope: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/2f4e5e38dead3a723c7732866beb25e848b3b3ad

glenhunter:
> When does the Entry State as shown in Zeplin take effect? As soon as you start typing, validation kicks in and you get either the Valid State or the Invalid State. Also, I doubt it would be possible/meaningful to differentiate between the Entry State and the Focus State.
> 
> image

So just to be clear @dturner32 this isn't really within our control. The browser doesn't recognize an "entry" state. There's `:hover`, `:focus`, `disabled`, and validation (pristine, valid, invalid).

glenhunter:
> > When does the Entry State as shown in Zeplin take effect? As soon as you start typing, validation kicks in and you get either the Valid State or the Invalid State. Also, I doubt it would be possible/meaningful to differentiate between the Entry State and the Focus State.
> > image
> 
> entry state should take place as soon as a user clicks into a field. Valid or error would only occur after the user hits enter or tabs to the next field. Unless the field is meant to trigger an error or valid state, then the field would transition from 'entry' to 'complete'
> 
> The reason for the focus state is intended for user guidance. Specifically for cases when a user only has access to a keyboard and not a mouse. The idea is to have a visual indicator for when they tabbing to the next thing

So let's imagine this (*very* common) use case: the user is entering a new password in a field with validation. The validation is whether the password is of a minimum length and contains certain special characters which we can check for (20 characters, one symbol, one number). They enter a 19 character password and we provide no validation. The user clicks away and the field shows invalid. Now they must click (or tab back) try and get it right, but again they won't know for sure until the click away again.

That is not a good user experience.

lib/src/scss-shared/_nx-form-validation.scss
```diff
@@ -5,20 +5,53 @@
...
+    border-color: #d1d6f0;
+    color: $nx-text-color-light;
+  }
+
+  .nx-icon--invalid, .nx-icon--valid {
+    display: none;
+  }
+}
+
+.nx-icon--invalid, .nx-icon--valid {
```
rpokorny, line 41: Everything from here down in this file is emitting, so cannot be in this file since this is in the `scss-shared` dir.

glenhunter: > Everything from here down in this file is emitting, so cannot be in this file since this is in the `scss-shared` dir.

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/d734699347f1f2d14bfce7f33d24ff38a7d6d3ea

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -4,19 +4,32 @@
...
+  color: #cd0028;
+  font-size: $nx-font-size-xs;
+  margin-top: $nx-spacing-xs;
+}
+
+.nx-icon--invalid {
+  color: #cd0028;
+}
+
+.nx-icon--valid {
```
rpokorny, line 33: These two should either be in the base styles or should be scoped to this component (e.g. `.nx-text-input .nx-icon--valid`.  The way it is now, the styles of unrelated `.nx-icon--valid` instances could change based on whether or not there is an `NxTextInput` present in the app

glenhunter: > These two should either be in the base styles or should be scoped to this component (e.g. `.nx-text-input .nx-icon--valid`. The way it is now, the styles of unrelated `.nx-icon--valid` instances could change based on whether or not there is an `NxTextInput` present in the app

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/7b2daa2af25f66c685960a1ae8f68aa6a63099d3

lib/src/scss-shared/_nx-form-validation.scss
rpokorny: > Fixed here: d734699

Missed the one this comment is on

glenhunter: > > Fixed here: [d734699](https://github.com/sonatype/sonatype-react-shared-components/commit/d734699347f1f2d14bfce7f33d24ff38a7d6d3ea)
> 
> Missed the one this comment is on

https://github.com/sonatype/sonatype-react-shared-components/pull/57/commits/351e88cf94fb709cad84ec6f4724eeb5694ee72a

PR #58, opened 2020-06-17
Document JS/TS utilities - RSC-159

rpokorny:
https://issues.sonatype.org/browse/RSC-159


gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -6,24 +6,27 @@
...
 
+// Props for GalleryTile
+type GalleryTileProps = PropsWithRequiredChildren & GalleryBaseProps;
+
 type StringOrCodeExampleProps = string | CodeExampleProps;
 
-interface GalleryExampleTileProps extends GalleryTileProps {
-  description: string;
+interface GalleryExampleTileProps extends GalleryBaseProps {
+  children?: ReactNode;
```
rpokorny, line 28: These utilities have code examples that don't render JSX.  They are only code to display, no live example.  Therefore the `GalleryExampleTile` children now needs to be optional

gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -6,24 +6,27 @@
...
+  description: ReactNode;
```
rpokorny, line 29: Realized while writing these that the description often would benefit from having HTML elements in it, particularly `<code>` elements.

gallery/src/util/jsUtil.ts
```diff
@@ -0,0 +1,14 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+/**
+ * If an array is passed in, returns an equivalent array.
+ * If a value that is not an array is passed in, returns an array containing only that value
+ */
+export function ensureArray<T>(items: T | T[]): T[] {
```
rpokorny, line 12: This function was actually only added for a single use in the gallery itself, so I moved it out of the lib jsUtil and to here.

gallery/src/jsUtilPages/PolicyThreatLevelUtils/CategoryByPolicyThreatLevelExample.ts
```diff
@@ -0,0 +1,21 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import { categoryByPolicyThreatLevel } from '@sonatype/react-shared-components';
+
+/* eslint-disable no-console */
+console.log('CategoryByPolicyThreatLevelExample');
+
+console.log(categoryByPolicyThreatLevel[5]); // 'severe'
```
Blacktiger, line 12: If we want to make this more visual we could adjust the example to just drop the output into the dom. Maybe a `nx-list` with the number next to the output of the function? (same with the other examples that just do a console log)

rpokorny: > If we want to make this more visual we could adjust the example to just drop the output into the dom. Maybe a nx-list with the number next to the output of the function? (same with the other examples that just do a console log)

I considered that, but was worried it would result in more clutter than content.

PR #59, opened 2020-06-18
nxDropdown redesign fixes RSC-123

glenhunter:
https://issues.sonatype.org/browse/RSC-123

Fixes 3 issues found in testing RSC-123
1. dropdown height was hard coded
2. <button> tags looked awful
3. dropdown items weren't truncating when necessary

### DIFF
--- a/lib/src/base-styles/_nx-typography.scss
+++ b/lib/src/base-styles/_nx-typography.scss
@@ -98,10 +98,3 @@
   line-height: 18px;
   margin-bottom: 0;
 }
-
-button {
-  @include regular();
-
-  font-size: $nx-font-size;
-  line-height: $nx-line-height;
-}
--- a/lib/src/base-styles/_nx-typography.scss
+++ b/lib/src/base-styles/_nx-typography.scss
@@ -98,3 +98,10 @@
   line-height: 18px;
   margin-bottom: 0;
 }
+
+button {
+  @include regular();
+
+  font-size: $nx-font-size;
+  line-height: $nx-line-height;
+}
--- a/lib/src/components/NxDropdown/NxDropdown.scss
+++ b/lib/src/components/NxDropdown/NxDropdown.scss
@@ -28,7 +28,6 @@ $nx-dropdown-width: 250px;
   .nx-dropdown__toggle-label {
     @include truncate-ellipsis();
     flex-grow: 1;
-    max-width: $nx-dropdown-width;
   }
 
   .nx-icon {
@@ -80,7 +79,6 @@ $nx-dropdown-width: 250px;
   line-height: $nx-line-height;
   padding: $nx-spacing-xxs $nx-spacing-md;
   text-decoration: none;
-  max-width: $nx-dropdown-width;
 }
 
 .nx-dropdown-button {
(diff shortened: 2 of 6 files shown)


lib/src/base-styles/_nx-typography.scss
```diff
@@ -98,3 +98,10 @@
   line-height: 18px;
   margin-bottom: 0;
 }
+
+button {
```
rpokorny, line 102: Global bare-element selectors are forbidden to allow compatibility with existing projects

rpokorny:
You fixed a different truncation issue than I noted in my testing notes.  I was referring to long text within the toggle button, not the menu buttons.  Though if the menu buttons had problems that's good to fix as well

lib/src/base-styles/_nx-typography.scss
glenhunter: > Global bare-element selectors are forbidden to allow compatibility with existing projects

https://github.com/sonatype/sonatype-react-shared-components/pull/59/commits/67a7617c53ee049ea6c5cd4d1c407c69e4170039

glenhunter:
> You fixed a different truncation issue than I noted in my testing notes. I was referring to long text within the toggle button, not the menu buttons. Though if the menu buttons had problems that's good to fix as well

https://github.com/sonatype/sonatype-react-shared-components/pull/59/commits/c579a9a88ebc2e410d2c8fce8feb1a5786951e2d

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -26,7 +26,9 @@ $nx-dropdown-width: 250px;
   width: $nx-dropdown-width;
 
   .nx-dropdown__toggle-label {
+    @include truncate-ellipsis();
     flex-grow: 1;
+    max-width: $nx-dropdown-width;
```
rpokorny, line 31: This doesn't seem necessary or correct.  The parent already has this width, and contains another element (the caret) in addition to this element, so the apparent max here is actually 198px.  But it doesn't seem like that needs to be made explicit, it works without it.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -69,13 +71,16 @@ $nx-dropdown-width: 250px;
...
   cursor: pointer;
   display: block;
-  height: 32px;
-  line-height: 32px;
-  padding: 0 $nx-spacing-md;
+  font-size: $nx-font-size;
+  line-height: $nx-line-height;
+  padding: $nx-spacing-xxs $nx-spacing-md;
   text-decoration: none;
+  max-width: $nx-dropdown-width;
```
rpokorny, line 83: Unnecessary

glenhunter: > Unnecessary

https://github.com/sonatype/sonatype-react-shared-components/pull/59/commits/addb08d80953eeea53853937acaf642e93314baa

lib/src/components/NxDropdown/NxDropdown.scss
glenhunter: > This doesn't seem necessary or correct. The parent already has this width, and contains another element (the caret) in addition to this element, so the apparent max here is actually 198px. But it doesn't seem like that needs to be made explicit, it works without it.

https://github.com/sonatype/sonatype-react-shared-components/pull/59/commits/addb08d80953eeea53853937acaf642e93314baa

PR #61, opened 2020-06-18
Non-validatable NxTextInput - RSC-109

rpokorny:
https://issues.sonatype.org/browse/RSC-109

Per designer guidance on RSC-109, NxTextInputs which don't have any validation logic should no longer display validation styles.  This turned out to be a non-trivial change however because NxTextInput itself is completely divorced from the validation logic - it only knows whether there were validation errors or not, not whether there ever _could_ be validation errors.  This PR addresses that by creating a distinction between `null` and `undefined` for the `validationErrors` prop of `NxTextInput`.  See the updates to the documentation page for details.

Note: there are test failures on the target branch which also affect this branch, but I believe I've made all relevant test updates here.

### DIFF
--- a/lib/src/util/validationUtil.ts
+++ b/lib/src/util/validationUtil.ts
@@ -11,7 +11,7 @@ export { ValidationErrors };
 /**
  * @return whether this ValidationErrors object contains an error
  */
-export const hasValidationErrors = (validationErrors: ValidationErrors | undefined) => {
+export const hasValidationErrors = (validationErrors: ValidationErrors) => {
   if (validationErrors == null) {
     return false;
   }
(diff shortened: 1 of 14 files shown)


lib/src/util/validationUtil.ts
```diff
@@ -11,7 +11,7 @@ export { ValidationErrors };
 /**
  * @return whether this ValidationErrors object contains an error
  */
-export const hasValidationErrors = (validationErrors: ValidationErrors | undefined) => {
+export const hasValidationErrors = (validationErrors: ValidationErrors) => {
```
Blacktiger, line 14: I suppose someone calling this from js instead of ts just needs to be aware that it will return true for undefined? Should the js doc spell out the accepted values for validation errors and/or be more explicit about the return type?

rpokorny:
Change of approach after talking with @glenhunter: whether or not the input is validatable is now a separate prop, and `validationErrors` is simply ignored if it is not.  I suspect this is simpler, as evidenced by the reduction in the length of the documentation

lib/src/util/validationUtil.ts
rpokorny: > I suppose someone calling this from js instead of ts just needs to be aware that it will return true for undefined? Should the js doc spell out the accepted values for validation errors and/or be more explicit about the return type?

The change here has now been reverted, but at any rate, this function returns `false` for undefined.

PR #62, opened 2020-06-22
NxFilterInput redesign RSC-124

glenhunter:
https://issues.sonatype.org/browse/RSC-124

This ticket just covers the basic styling changes from Derrek. In the course of the work we discovered that `NxFilterInput` isn't built on top of `NxTextInput` and that it probably should be. That's a much larger task and so it was broken out into RSC-165.

Because `NxFilterInput` cannot show all of the user states specified in the design we have opted (after discussing with Derrek) to use a default border colour of `#2c2c2c`.

### DIFF
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -34,10 +34,5 @@
     color: $nx-text-color-light;
     margin-left: 0;
     margin-right: 8px;
-
-    &.fa-times {
-      cursor: pointer;
-      width: 1em;
-    }
   }
 }
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -11,56 +11,42 @@
   @include container-horizontal;
 
   display: inline-block;
-  vertical-align: middle;
   white-space: nowrap;
 
-  .nx-filter-text-input {
-    border-left: 0;
-    box-sizing: content-box;
-    margin-left: 0;
+  &:focus {
+    outline: 0;
   }
 
-  .nx-filter-input__add-on {
-    border: $nx-border;
-    border-right: none;
-    box-sizing: content-box;
-    display: inline-block;
-    padding: 4px 4px 4px 2px;
-    vertical-align: top;
-    width: 20px;
+  &:focus-within {
+    .nx-text-input__box {
+      border-color: #43cbf5;
+    }
+  }
 
-    .nx-icon {
-      color: $nx-text-color-light;
-      height: 10px;
-      margin-top: 5px;
-      margin-left: 5px;
+  &:hover {
+    .nx-text-input__box {
+      border-color: #2c2c2c;
+    }
+  }
 
-      &.fa-times {
-        cursor: pointer;
+  &.nx-filter-input--disabled {
+    .nx-text-input__box {
+      background-color: #e9e9e9;
+      border-color: #949494;
 
-        height: 14px;
-        margin-top: 4px;
-        margin-left: 6px;
+      .nx-icon--filter-icons {
+        color: #949494;
       }
     }
   }
 
-  .nx-filter-input--disabled {
-    .nx-filter-input__add-on {
-      background-color: $nx-off-white;
+  .nx-icon--filter-icons {
+    color: $nx-text-color-light;
+    height: 1em;
+    margin-right: 8px;
 
-      .nx-icon {
-        color: $nx-text-color-disabled;
-      }
-    }
-
-    .nx-filter-text-input {
-      background-color: $nx-off-white;
-      border-color: $nx-grey-3;
-
-      &::placeholder {
-        color: $nx-text-color-disabled;
-      }
+    &.fa-times {
+      cursor: pointer;
     }
   }
 }
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -19,12 +19,6 @@
     }
   }
 
-  &:hover {
-    .nx-text-input__box {
-      border-color: #2c2c2c;
-    }
-  }
-
   &.nx-filter-input--disabled {
     .nx-text-input__box {
       background-color: #e9e9e9;
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -13,10 +13,6 @@
   display: inline-block;
   white-space: nowrap;
 
-  &:focus {
-    outline: 0;
-  }
-
   &:focus-within {
     .nx-text-input__box {
       border-color: #43cbf5;
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -42,11 +42,12 @@
 
   .nx-icon--filter-icons {
     color: $nx-text-color-light;
-    height: 1em;
+    margin-left: 0;
     margin-right: 8px;
 
     &.fa-times {
       cursor: pointer;
+      width: 1em;
     }
   }
 }
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -7,8 +7,6 @@
 @import '../../scss-shared/nx-variables';
 @import '../../scss-shared/nx-container-helpers';
 
-$nx-filter-input-height: 20px;
-
 .nx-filter-input {
   @include container-horizontal;
 
@@ -17,28 +15,19 @@ $nx-filter-input-height: 20px;
   white-space: nowrap;
 
   .nx-filter-text-input {
-    border: $nx-border;
     border-left: 0;
-    font-size: 14px;
-    height: $nx-filter-input-height;
-    margin-left: 0;
-    padding: 4px 6px;
     box-sizing: content-box;
-
-    &:focus {
-      outline: none;
-    }
+    margin-left: 0;
   }
 
   .nx-filter-input__add-on {
     border: $nx-border;
     border-right: none;
+    box-sizing: content-box;
     display: inline-block;
-    height: $nx-filter-input-height;
     padding: 4px 4px 4px 2px;
     vertical-align: top;
     width: 20px;
-    box-sizing: content-box;
 
     .nx-icon {
       color: $nx-text-color-light;
--- a/lib/src/components/NxFilterInput/NxFilterInput.tsx
+++ b/lib/src/components/NxFilterInput/NxFilterInput.tsx
@@ -39,7 +39,7 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
                  value={value}
                  onChange={inputOnChange}
                  placeholder={placeholder || undefined}
-                 className="nx-filter-text-input"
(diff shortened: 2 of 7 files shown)


rpokorny:
Spacing to the left and right of the icon appears to be about 25px and 13px respectively, should be 16px and 8px.

rpokorny:
The icon feels like it shifts to the left as it switches from the filter to the x.  I think this is because the x is narrower, and the icon is left-justified within its box.

rpokorny:
I know we already debated the idea of "pristine" state for this box, and left it out of this PR for the sake of time.  But now, seeing the Close button interaction again, I'm wondering if it might make more sense for this text box if the "Default" (light) border in the design was used for "empty" and not actually "pristine".  That way the changing of the border would match the changing of the icon/button.

rpokorny:
The mockups don't seem to include the close button functionality.  Should we remove it?

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -7,71 +7,46 @@
...
-  .nx-filter-text-input {
-    border: $nx-border;
-    border-left: 0;
-    font-size: 14px;
-    height: $nx-filter-input-height;
-    margin-left: 0;
-    padding: 4px 6px;
-    box-sizing: content-box;
+  &:focus {
+    outline: 0;
```
rpokorny, line 17: I'm fairly sure this doesn't do anything; the div isn't going to be getting focus

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -7,71 +7,46 @@
...
-      &.fa-times {
-        cursor: pointer;
-
-        height: 14px;
-        margin-top: 4px;
-        margin-left: 6px;
-      }
+  &:hover {
+    .nx-text-input__box {
+      border-color: #2c2c2c;
```
rpokorny, line 28: Redundant

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -7,71 +7,46 @@
...
       }
     }
+  }
 
-    .nx-filter-text-input {
-      background-color: $nx-off-white;
-      border-color: $nx-grey-3;
+  .nx-icon--filter-icons {
+    color: $nx-text-color-light;
+    height: 1em;
```
rpokorny, line 45: Redundant

rpokorny:
I think I've found a small styling bug that affects both `NxTextInput` and `NxFilterInput`: The `<input>` itself has a little bit of default padding, making the overall box 44px tall instead of 42px.

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -29,18 +30,19 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
-                 autoComplete="off"
-                 id={inputId || undefined}
-                 value={value}
-                 onChange={inputOnChange}
-                 placeholder={placeholder || undefined}
-                 className="nx-filter-text-input"
-                 disabled={disabled || undefined} />
+          <div className="nx-text-input__box">
+            <span className="nx-filter-input__add-on" onClick={value && onClear || undefined}>
+              <NxFontAwesomeIcon icon={icon} className="nx-icon--filter-icons" />
```
rpokorny, line 35: For the clear icon do we want to use our custom Close icon instead of FA?

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -29,18 +30,19 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
-          <input type="text"
-                 autoComplete="off"
-                 id={inputId || undefined}
-                 value={value}
-                 onChange={inputOnChange}
-                 placeholder={placeholder || undefined}
-                 className="nx-filter-text-input"
-                 disabled={disabled || undefined} />
+          <div className="nx-text-input__box">
+            <span className="nx-filter-input__add-on" onClick={value && onClear || undefined}>
```
rpokorny, line 34: Is there any purpose to this span, or could the icon be placed directly into the `.nx-text-input__box`?

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -69,7 +69,7 @@ describe('NxFilterInput', function() {
 
   it('passes unknown props to the div element', function() {
     const props = { className: 'extra-class', id: 'some-id' },
-        componentDiv = shallowComponent(props).find('div');
+        componentDiv = shallowComponent(props).find('div.nx-filter-input');
```
rpokorny, line 72: This find shouldn't even be necessary.
```suggestion
        componentDiv = shallowComponent(props);
```

glenhunter:
> The mockups don't seem to include the close button functionality. Should we remove it?

I asked @dturner32 and he said that it should be there

lib/src/components/NxFilterInput/NxFilterInput.tsx
glenhunter: > Is there any purpose to this span, or could the icon be placed directly into the `.nx-text-input__box`?

The icon doesn't support `onClick`.

lib/src/components/NxFilterInput/NxFilterInput.tsx
glenhunter: > For the clear icon do we want to use our custom Close icon instead of FA?

@dturner32 ?

lib/src/components/NxFilterInput/NxFilterInput.scss
glenhunter: > Redundant

Removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/1ad28c9c89f21527b3080aff6a60d3da34aaeefd

lib/src/components/NxFilterInput/NxFilterInput.scss
glenhunter: > I'm fairly sure this doesn't do anything; the div isn't going to be getting focus

https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/0c354249c5b6f29decf837b6f576ff3ddf5068cc

glenhunter:
> The icon feels like it shifts to the left as it switches from the filter to the x. I think this is because the x is narrower, and the icon is left-justified within its box. Correspondingly the left edge of the typing area shifts left once a character is entered.

There was another issue where the two icons have different widths and that was causing the box to resize, I have adjusted that here and it looks much better now.

glenhunter:
> Spacing to the left and right of the icon appears to be about 25px and 13px respectively, should be 16px and 8px.

Adjusted here: https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/1ad28c9c89f21527b3080aff6a60d3da34aaeefd

lib/src/components/NxFilterInput/NxFilterInput.tsx
glenhunter: > > For the clear icon do we want to use our custom Close icon instead of FA?
> 
> @dturner32 ?

Note that it is not a "close' button, but a "clear" icon.

dturner32:
> > The icon feels like it shifts to the left as it switches from the filter to the x. I think this is because the x is narrower, and the icon is left-justified within its box. Correspondingly the left edge of the typing area shifts left once a character is entered.
> 
> There was another issue where the two icons have different widths and that was causing the box to resize, I have adjusted that here and it looks much better now.

glenhunter:
> I know we already debated the idea of "pristine" state for this box, and left it out of this PR for the sake of time. But now, seeing the Close button interaction again, I'm wondering if it might make more sense for this text box if the "Default" (light) border in the design was used for "empty" and not actually "pristine". That way the changing of the border would match the changing of the icon/button.

I thought of that but the problem is that the border is on `nx-text-input__box` not the input itself. And `nx-text-input__box` is never empty because it contains the input. 

But otherwise I agree with the broader styling suggestion for when we're using `NxTextInput` within `NxFilterInput`

lib/src/components/NxFilterInput/NxFilterInput.scss
glenhunter: > Redundant

Removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/a7035dcb4307486eb67f8ac8f7e8023ff3459b7c

glenhunter:
Per direction from @dturner32 I have removed the clear icon and its functionality here: https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/5822d6775bad9b5ee29e118e61fc6e91b9e4f346

rpokorny:
> I thought of that but the problem is that the border is on nx-text-input__box not the input itself. And nx-text-input__box is never empty because it contains the input.

Well sure, to do that would require some JavaScript support but it wouldn't be hard.

> But otherwise I agree with the broader styling suggestion for when we're using NxTextInput within NxFilterInput

I'm not sure, are you saying here that you agree with the empty vs pristine styling suggestion or not?

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -7,71 +7,37 @@
...
-      border-color: $nx-grey-3;
+  .nx-icon--filter-icons {
+    color: $nx-text-color-light;
+    margin-left: 0;
+    margin-right: 8px;
 
-      &::placeholder {
-        color: $nx-text-color-disabled;
-      }
+    &.fa-times {
```
rpokorny, line 38: No longer used

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -19,44 +19,12 @@ describe('NxFilterInput', function() {
...
-            onClear
-          }).find('.nx-filter-input__add-on');
-
-      expect(iconWrapper).toExist();
-      iconWrapper.simulate('click');
-      expect(onClear).toHaveBeenCalled();
-    });
-  });
-
   describe('when value prop is an empty string', function() {
```
rpokorny, line 22: This `describe` wrapper should probably removed, since the `it` that it contains is always supposed to be true now

lib/src/components/NxFilterInput/NxFilterInput.scss
glenhunter: > No longer used

https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/2a0562f9ba6c1e2c95eca6bede04a30b2c9b8ff1

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
glenhunter: > This `describe` wrapper should probably removed, since the `it` that it contains is always supposed to be true now

https://github.com/sonatype/sonatype-react-shared-components/pull/62/commits/ef38611f63239b49c2eed85e6985cf2941a3ab19

glenhunter:
> > I thought of that but the problem is that the border is on nx-text-input__box not the input itself. And nx-text-input__box is never empty because it contains the input.
> 
> Well sure, to do that would require some JavaScript support but it wouldn't be hard.
> 
> > But otherwise I agree with the broader styling suggestion for when we're using NxTextInput within NxFilterInput
> 
> I'm not sure, are you saying here that you agree with the empty vs pristine styling suggestion or not?

I agree with this statement:
> it might make more sense for this text box if the "Default" (light) border in the design was used for "empty" and not actually "pristine"

PR #63, opened 2020-06-22
React Component for Page Header - RSC-129

rpokorny:
https://issues.sonatype.org/browse/RSC-129

### DIFF
--- a/lib/src/components/NxPageHeader/NxPageHeader.tsx
+++ b/lib/src/components/NxPageHeader/NxPageHeader.tsx
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+import { Props, propTypes } from './types';
+
+export { Props };
+
+const logoImg = require('../../assets/img/sonatype-logo-with-hexagon.png');
+
+export default function NxPageHeader({ links, homeLink, productInfo, children }: Props) {
+  const logoEl = <img src={logoImg} className="nx-product__wordmark" alt="Sonatype" />,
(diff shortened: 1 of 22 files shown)


lib/src/components/NxPageHeader/NxPageHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <span className="nx-page-header__link-sizing-text">{name}</span>
+    </a>
+  );
+}
+
+// visible for testing
+export { HeaderLinkProps, HeaderLink };
+
+export default function NxPageHeader({ links, homeLink, productInfo, children }: Props) {
+  const logoEl = <img src={logoImg} className="nx-product__wordmark" alt="⬡ Sonatype"/>,
```
glenhunter, line 34: Clever, but I don't think the hexagon should be in the alt text

rpokorny: > Clever, but I don't think the hexagon should be in the alt text

I don't feel too strongly about it, but after double checking the purpose of the alt text I felt like adding this was most faithful to it:

> Content associated with embedded content, to be used when the embedded content cannot itself be used (for example, when the embedded content is of an unsupported format).

glenhunter:
I pushed a branch that shows a method of dealing with the normal/semi-bold sizing issue in the navigation here: https://github.com/sonatype/sonatype-react-shared-components/tree/RSC-129-page-header-component-glen

It doesn't require duplicate `<span>` tags and IMO is much cleaner.

lib/src/components/NxPageHeader/NxPageHeader.tsx
glenhunter: Ok

rpokorny:
> I pushed a branch that shows a method of dealing with the normal/semi-bold sizing issue in the navigation here: https://github.com/sonatype/sonatype-react-shared-components/tree/RSC-129-page-header-component-glen
>
> It doesn't require duplicate <span> tags and IMO is much cleaner.

Accepted.

PR #64, opened 2020-06-23
 nx-tile subsections have dividing lines RSC-155

glenhunter:
https://issues.sonatype.org/browse/RSC-155

Used :first-of-type, increased default top margin to match Zeplin

added extra subsections to example to demonstrate

### DIFF
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -146,8 +146,10 @@ $nx-tile-background-color: $nx-white;
     padding-top: $nx-spacing-l;
   }
 
-  header {
+  .nx-tile-subsection__header {
     @include container-vertical;
+
+    margin-bottom: $nx-spacing-md;
   }
 }
 
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -136,17 +136,18 @@ $nx-tile-background-color: $nx-white;
 
 // On pages with tiles with multiple sections these are lighter weight titles
 .nx-tile-subsection {
-  .nx-h3--tile-subsection {
-    margin-top: $nx-spacing-xl;
-    margin-bottom: $nx-spacing-md;
-  }
+  @include container-vertical;
+  margin-top: $nx-spacing-xl;
+  margin-bottom: $nx-spacing-md;
 
   &:first-of-type {
-    .nx-h3--tile-subsection {
-      border-top: 1px solid #b7bfdb;
-      margin-top: $nx-spacing-l;
-      padding-top: $nx-spacing-l;
-    }
+    border-top: 1px solid #b7bfdb;
+    margin-top: $nx-spacing-l;
+    padding-top: $nx-spacing-l;
+  }
+
+  header {
+    @include container-vertical;
   }
 }
 
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -135,15 +135,18 @@ $nx-tile-background-color: $nx-white;
 }
 
 // On pages with tiles with multiple sections these are lighter weight titles
-.nx-tile__section-header {
-  border-top: 1px solid #b7bfdb;
-  margin-top: $nx-spacing-l;
-  margin-bottom: $nx-spacing-md;
-  padding-top: $nx-spacing-l;
-
-  &.nx-tile__section-header--no-border {
-    border: none;
+.nx-tile-subsection {
+  .nx-h3--tile-subsection {
     margin-top: $nx-spacing-xl;
+    margin-bottom: $nx-spacing-md;
+  }
+
+  &:first-of-type {
+    .nx-h3--tile-subsection {
+      border-top: 1px solid #b7bfdb;
+      margin-top: $nx-spacing-l;
+      padding-top: $nx-spacing-l;
+    }
   }
 }
 
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -136,13 +136,14 @@ $nx-tile-background-color: $nx-white;
 
 // On pages with tiles with multiple sections these are lighter weight titles
 .nx-tile__section-header {
-  margin-top: $nx-spacing-xl;
+  border-top: 1px solid #b7bfdb;
+  margin-top: $nx-spacing-l;
   margin-bottom: $nx-spacing-md;
+  padding-top: $nx-spacing-l;
 
-  &:first-of-type {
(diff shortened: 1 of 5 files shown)


lib/src/base-styles/_nx-tile.scss
```diff
@@ -136,10 +136,14 @@ $nx-tile-background-color: $nx-white;
 
 // On pages with tiles with multiple sections these are lighter weight titles
 .nx-tile__section-header {
-  border-top: 1px solid #b7bfdb;
-  margin-top: $nx-spacing-l;
+  margin-top: $nx-spacing-xl;
   margin-bottom: $nx-spacing-md;
-  padding-top: $nx-spacing-l;
+
+  &:first-of-type {
```
rpokorny, line 142: This approach presents a problem in conjunction with any interleaving elements, such as the `section.gallery-example-code` elements that `GalleryExampleTile` produces.  See the `NxFontAwesomeIcon` page for an example a a multi-section `GalleryExampleTile`

glenhunter: > This approach presents a problem in conjunction with any interleaving elements, such as the `section.gallery-example-code` elements that `GalleryExampleTile` produces. See the `NxFontAwesomeIcon` page for an example a a multi-section `GalleryExampleTile`

After discussion I think this works well. https://github.com/sonatype/sonatype-react-shared-components/pull/64/commits/6a793c18e18b08d1a354e5b78125af7b73c1c654

Is it worth refactoring the other `nx-tile` examples to use the more semantic `<section>`, `<header>`, and `<footer>` tags?

rpokorny: > Is it worth refactoring the other nx-tile examples to use the more semantic <section>, <header>, and <footer> tags?

Maybe the ones on nx-tile documentation page, sure.  I don't know that we need to refactor the whole gallery though.  I'd consider the use of those semantic elements to be optional but recommended — not required.

rpokorny:
It makes me a little nervous that the `<section>`s don't have margins of their own and are relying on their child margins to collapse out of them.  I think they should be given standard margins and use `container-vertical` like we do elsewhere.

lib/src/base-styles/_nx-tile.scss
```diff
@@ -135,11 +135,19 @@ $nx-tile-background-color: $nx-white;
 }
 
 // On pages with tiles with multiple sections these are lighter weight titles
-.nx-tile__section-header {
-  border-top: 1px solid #b7bfdb;
-  margin-top: $nx-spacing-l;
-  margin-bottom: $nx-spacing-md;
-  padding-top: $nx-spacing-l;
+.nx-tile-subsection {
+  .nx-h3--tile-subsection {
```
rpokorny, line 139: Why the change of this classname?

lib/src/base-styles/_nx-tile.scss
```diff
@@ -135,11 +135,19 @@ $nx-tile-background-color: $nx-white;
...
+    margin-top: $nx-spacing-xl;
+    margin-bottom: $nx-spacing-md;
+  }
+
+  &:first-of-type {
+    .nx-h3--tile-subsection {
```
rpokorny, line 145: Would it make more sense to put the border on the `<section>`?

glenhunter: > Would it make more sense to put the border on the `<section>`?

I moved all the spacing and the border to the section: https://github.com/sonatype/sonatype-react-shared-components/pull/64/commits/0d21ac22dfb016139ac73ffb51c347c60e3e2eee

lib/src/base-styles/_nx-tile.scss
glenhunter: > > Why the change of this classname?
> 
> My thinking was that the changes expressed in the SCSS were modifications to `nx-h3` so the class should reflect that. Though this may be moot with the suggested changes to `<section>`

As I suspected there's no need for the custom classes anymore: https://github.com/sonatype/sonatype-react-shared-components/pull/64/commits/d0be78e77757fc08057408e161bfdde678a3e8d0

glenhunter:
> It makes me a little nervous that the `<section>`s don't have margins of their own and are relying on their child margins to collapse out of them. I think they should be given standard margins and use `container-vertical` like we do elsewhere.

Yes, it's much better with the spacing and border on the `<section>`: https://github.com/sonatype/sonatype-react-shared-components/pull/64/commits/0d21ac22dfb016139ac73ffb51c347c60e3e2eee

lib/src/base-styles/_nx-tile.scss
```diff
@@ -135,11 +135,20 @@ $nx-tile-background-color: $nx-white;
...
-  padding-top: $nx-spacing-l;
+
+  &:first-of-type {
+    border-top: 1px solid #b7bfdb;
+    margin-top: $nx-spacing-l;
+    padding-top: $nx-spacing-l;
+  }
+
+  header {
+    @include container-vertical;
```
rpokorny, line 150: Needs 16px bottom margin

lib/src/base-styles/_nx-tile.scss
```diff
@@ -135,11 +135,20 @@ $nx-tile-background-color: $nx-white;
 }
 
 // On pages with tiles with multiple sections these are lighter weight titles
-.nx-tile__section-header {
-  border-top: 1px solid #b7bfdb;
-  margin-top: $nx-spacing-l;
+.nx-tile-subsection {
+  @include container-vertical;
+  margin-top: $nx-spacing-xl;
   margin-bottom: $nx-spacing-md;
```
rpokorny, line 141: What's the purpose of this smaller bottom margin? It just collapses into the larger top margin of the next section

glenhunter: > What's the purpose of this smaller bottom margin? It just collapses into the larger top margin of the next section

It should be safe to remove: https://github.com/sonatype/sonatype-react-shared-components/pull/64/commits/8808c4962dcbee77e2175930c6705301822aacc1

lib/src/base-styles/_nx-tile.scss
glenhunter: > Needs 16px bottom margin

Adjusted here: https://github.com/sonatype/sonatype-react-shared-components/pull/64/commits/746bf45b649ae99397db34263618b70ed91daff2

lib/src/base-styles/_nx-tile.scss
rpokorny: > It should be safe to remove: 8808c49

Safest probably to set it equal to the top margin

rpokorny:
Now that the Styling Components page is in the branch, it appears to have some styling problems.

PR #66, opened 2020-06-23
nx-truncate-ellipsis RSC-115

glenhunter:

Added an example page for the `.nx-truncation-ellipsis` mixin. 

As part of this I added the `.nx-` prefix to the mixin to make it consistent with our other styles. I then refactored existing instances of the mixin in the RSC.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "0.49.0",
+  "version": "0.49.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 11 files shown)


rpokorny:
> As part of this I added the .nx- prefix to the mixin to make it consistent with our other styles. I then refactored existing instances of the mixin in the RSC.

None of our other mixins except for `nx-font-family` start with `nx`.  Should they?  In particular the container mixins, which are already documented in the gallery?

glenhunter:
> > As part of this I added the .nx- prefix to the mixin to make it consistent with our other styles. I then refactored existing instances of the mixin in the RSC.
> 
> None of our other mixins except for `nx-font-family` start with `nx`. Should they? In particular the container mixins, which are already documented in the gallery?

Probably. I changed this one mostly because I was *constantly* typing `nx-truncate-ellipsis`.

PR #67, opened 2020-06-23
 nx-clickable RSC-115

glenhunter:

Simple example page for `.nx-clickable`


gallery/src/styles/NxClickable/NxClickableExample.scss
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ @import '~@sonatype/react-shared-components/scss-shared/nx-text-helpers';
+
+.nx-truncate-example1 {
```
rpokorny, line 9: This file seems wrong and also like it shouldn't exist

gallery/src/styles/NxClickable/NxClickablePage.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
+
+import NxClickableExample from './NxClickableExample';
+
+const nxClickableExampleCode = require('!!raw-loader!./NxClickableExample').default;
+
+const NxClickablePage = () => {
+  const codeExamples = [
+    nxClickableExampleCode
+  ];
```
rpokorny, line 18: Since it's just the one, you can pass `nxClickableExampleCode` directly as the `codeExamples` prop on the tile, and remove this separate variable.

glenhunter: > Since it's just the one, you can pass `nxClickableExampleCode` directly as the `codeExamples` prop on the tile, and remove this separate variable.

Ah, that's easier: https://github.com/sonatype/sonatype-react-shared-components/pull/67/commits/095a7c8802e9560e4579013ed4326161b2083901

gallery/src/styles/NxClickable/NxClickableExample.scss
glenhunter: > This file seems wrong and also like it shouldn't exist in this PR

Yeah that should have been deleted: https://github.com/sonatype/sonatype-react-shared-components/pull/67/commits/095a7c8802e9560e4579013ed4326161b2083901

PR #68, opened 2020-06-24
Applitools test setup - RSC-167

rpokorny:
https://issues.sonatype.org/browse/RSC-167


gallery/wdio.conf.js
```diff
@@ -69,7 +72,7 @@ exports.config = {
     // Define all options that are relevant for the WebdriverIO instance here
     //
     // Level of logging verbosity: trace | debug | info | warn | error | silent
-    logLevel: 'info',
+    logLevel: 'warn',
```
regoarrarr, line 75: Did you mean to update this?

rpokorny: > Did you mean to update this?

Yes.  It spits out what I consider to be debug-level details (full HTTP headers and bodies) of every single call to the selenium server at the INFO level.  It's a large amount of output that is in no way useful to me.

rpokorny:
> It looks reasonable to me though I was not involved in setting up Applitools for IQ so I don't have a ton of knowledge about this. This is an open source project right so might be worth seeing if Brandon would be interested in reviewing it?

We'd like to release 1.0.0 today, and ideally have this be in that release, so in the interest of time I'll forgo that.

PR #69, opened 2020-06-24
nx-code RSC-115

glenhunter:

Just the simple `<code className="nx-code">` version here though I do think there is value in adding a "block" version in v1.x

### DIFF
--- a/gallery/src/styles/NxCode/NxCodeExample.tsx
+++ b/gallery/src/styles/NxCode/NxCodeExample.tsx
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import './NxCodeExample.scss';
+
+const NxCodeExample = () =>
+  <>
+    <p className="nx-p"><code className="nx-code">&lt;HTML&gt;</code></p>
+    <p className="nx-p"><code className="nx-code">.scss-class</code></p>
+  </>
+export default NxCodeExample;
--- a/gallery/src/styles/NxCode/NxCodeExample.tsx
+++ b/gallery/src/styles/NxCode/NxCodeExample.tsx
@@ -5,11 +5,11 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React from 'react';
-import './NxCodeExample.scss';
 
 const NxCodeExample = () =>
   <>
     <p className="nx-p"><code className="nx-code">&lt;HTML&gt;</code></p>
     <p className="nx-p"><code className="nx-code">.scss-class</code></p>
-  </>
+  </>;
+
 export default NxCodeExample;
--- a/gallery/src/styles/NxCode/NxCodeExample.tsx
+++ b/gallery/src/styles/NxCode/NxCodeExample.tsx
@@ -8,8 +8,10 @@ import React from 'react';
 
 const NxCodeExample = () =>
   <>
-    <p className="nx-p"><code className="nx-code">&lt;HTML&gt;</code></p>
-    <p className="nx-p"><code className="nx-code">.scss-class</code></p>
+    <p className="nx-p">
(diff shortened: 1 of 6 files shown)


rpokorny:
Shouldn't this really be in the HTML elements section of the nav.  It's not really a helper class in that it isn't meant to be added onto some element that already primarily has another class

gallery/src/styles/NxCode/NxCodePage.tsx
```diff
@@ -0,0 +1,36 @@
...
+          Using the &lt;code&gt; HTML tag with the <code className="nx-code">.nx-code</code> className applies a
+          monospace font and other styling to make your code snippets stand out.
+        </p>
+        <p className="nx-p">
+          <code className="nx-code">.nx-code</code> is used extensively throughout the RSC Gallery.
+        </p>
+      </GalleryDescriptionTile>
+      <GalleryExampleTile title="Styling code snippets"
+                          codeExamples={nxCodeExampleCode}
+                          description="">
```
rpokorny, line 29: My understanding was that the description is mandatory, which is why it's a required prop

gallery/src/styles/NxCode/NxCodeExample.tsx
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+const NxCodeExample = () =>
+  <>
+    <p className="nx-p"><code className="nx-code">&lt;HTML&gt;</code></p>
```
rpokorny, line 11: Maybe include one of these amidst some other text to show that they are inline

gallery/src/styles/NxCode/NxCodePage.tsx
```diff
@@ -0,0 +1,36 @@
...
+import NxCodeExample from './NxCodeExample';
+
+const nxCodeExampleCode = require('!!raw-loader!./NxCodeExample').default;
+
+const NxCodePage = () => {
+  return (
+    <>
+      <GalleryDescriptionTile>
+        <p className="nx-p">
+          Using the &lt;code&gt; HTML tag with the <code className="nx-code">.nx-code</code> className applies a
```
rpokorny, line 20:
```suggestion
          Using the <code className="nx-code">&lt;code&gt;</code> HTML tag with the <code className="nx-code">.nx-code</code> className applies a
```

gallery/src/styles/NxCode/NxCodePage.tsx
```diff
@@ -0,0 +1,36 @@
...
+          monospace font and other styling to make your code snippets stand out.
```
rpokorny, line 21: `.nx-code` doesn't actually apply a monospace font.  User-agent styles for `<code>` do that.

gallery/src/styles/NxCode/NxCodePage.tsx
glenhunter: > My understanding was that the description is mandatory, which is why it's a required prop

https://github.com/sonatype/sonatype-react-shared-components/pull/69/commits/87a8dcdf82f0a003c806f9ab3f2f0b881ff290a0

gallery/src/styles/NxCode/NxCodeExample.tsx
glenhunter: > Maybe include one of these amidst some other text to show that they are inline

https://github.com/sonatype/sonatype-react-shared-components/pull/69/commits/23ba6961af8a23886a45e8d3577f9f4ad69b299a

glenhunter:
> Shouldn't this really be in the HTML elements section of the nav. It's not really a helper class in that it isn't meant to be added onto some element that already primarily has another class

https://github.com/sonatype/sonatype-react-shared-components/pull/69/commits/4ccd3fd154e828041aa736909e60d4fc6aa4489c

gallery/src/styles/NxCode/NxCodePage.tsx
glenhunter: > `.nx-code` doesn't actually apply a monospace font. User-agent styles for `<code>` do that.

Now `nx-code does it: https://github.com/sonatype/sonatype-react-shared-components/pull/69/commits/a1c5d4c9375d633896b0a21584c35816865a8388

PR #71, opened 2020-06-24
Adjust GalleryExampleTile API - RSC-168

rpokorny:
https://issues.sonatype.org/browse/RSC-168

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "0.50.4",
+  "version": "0.50.5",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "0.50.5",
+  "version": "0.50.6",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
+++ b/gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
@@ -79,7 +79,7 @@ const NxCheckboxPage = () =>
 
     <GalleryExampleTile title="General Example"
                         codeExamples={exampleCode}
-                          liveExample={NxCheckboxExample}>
+                        liveExample={NxCheckboxExample}>
       This example shows a series of checkboxes in a typical vertical layout with
       different label content. Note that one of the checkboxes is disabled. Another has no label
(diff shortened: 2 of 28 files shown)


glenhunter:
Is the description optional now?

rpokorny:
> Is the description optional now?

No, note that the optionalness of the `children` has been removed.

glenhunter:
> > Is the description optional now?
> 
> No, note that the optionalness of the `children` has been removed.

Can we make the description optional? I don't think every example necessarily needs it looking at NxLoadWrapper for example it's hard to see what info could be added to each example that isn't supplied by the title..

rpokorny:
> Can we make the description optional? I don't think every example necessarily needs it. Looking at NxLoadWrapper (for example) it's hard to see what info could be added to each example that isn't supplied by the title.

IIRC when I first implemented the new GalleryExampleTile I asked @dturner32 if the description was optional and he said no.  Before that my instinct was to make it optional as well.

dturner32:
> > Can we make the description optional? I don't think every example necessarily needs it. Looking at NxLoadWrapper (for example) it's hard to see what info could be added to each example that isn't supplied by the title.
> 
> IIRC when I first implemented the new GalleryExampleTile I asked @dturner32 if the description was optional and he said no. Before that my instinct was to make it optional as well.

So I originally made this decision on the basis that there would a top level description giving a brief overall explanation of the components in each category. As long as there is general context around what each component and their functions are within the example tiles then I am fine with making the top level description tile optional

glenhunter:
> > > Can we make the description optional? I don't think every example necessarily needs it. Looking at NxLoadWrapper (for example) it's hard to see what info could be added to each example that isn't supplied by the title.
> > 
> > 
> > IIRC when I first implemented the new GalleryExampleTile I asked @dturner32 if the description was optional and he said no. Before that my instinct was to make it optional as well.
> 
> So I originally made this decision on the basis that there would a top level description giving a brief overall explanation of the components in each category. As long as there is general context around what each component and their functions are within the example tiles then I am fine with making the top level description tile optional

So we're not talking about the top level description, that's mandatory. We're talking about the (much briefer) description text that is at the beginning of each specific example tile.

dturner32:
> > > > Can we make the description optional? I don't think every example necessarily needs it. Looking at NxLoadWrapper (for example) it's hard to see what info could be added to each example that isn't supplied by the title.
> > > 
> > > 
> > > IIRC when I first implemented the new GalleryExampleTile I asked @dturner32 if the description was optional and he said no. Before that my instinct was to make it optional as well.
> > 
> > 
> > So I originally made this decision on the basis that there would a top level description giving a brief overall explanation of the components in each category. As long as there is general context around what each component and their functions are within the example tiles then I am fine with making the top level description tile optional
> 
> So we're not talking about the top level description, that's mandatory. We're talking about the (much briefer) description text that is at the beginning of each specific example tile.
> 
> Screen Shot 2020-06-25 at 10 12 36 AM

so in this case the description is more for guidelines and usage of the specific component. I would still like to have something here at all times just to give added context around when, where and how the component is used if that makes sense. These same guidelines would go into the design language as well. This would all be a part of the documentation I am putting together

PR #75, opened 2020-06-25
nx modal error styles RSC-150

glenhunter:
https://issues.sonatype.org/browse/RSC-150

Removed the .nx-error styles from the modal footer. 
Used `NxLoadError` component for the error. 
Fixed spacing to match Zeplin. 
Refactored this example to use NxTextInput components.
Changed `nx-btn-bar` to right justify the buttons as this is the new globe style.


gallery/src/components/NxModal/NxModalFormErrorExample.tsx
```diff
@@ -29,32 +40,21 @@ export default function NxModalFormErrorExample() {
...
                 </label>
               </div>
             </div>
-            <footer className="nx-modal-footer nx-error">
-              <div id="login-error" className="nx-alert nx-alert--error">
-                <NxFontAwesomeIcon icon={faAngry} />
-                <span>You really messed something up!</span>
-              </div>
+            <footer className="nx-modal-footer">
+              <NxLoadError { ...({ error, retryHandler }) } />
```
rpokorny, line 54: It seems a little weird to have a load error, as if the form had been saved or something, but then to not also have any sort of Save button, either before or after the retry succeeds.  As it is now, after pressing retry, you're left with only a Cancel button.  Which wouldn't realistically make much sense.

lib/src/base-styles/_nx-btn-bar.scss
```diff
@@ -8,5 +8,6 @@
 .nx-btn-bar {
   @include container-horizontal;
 
-  margin: 8px 0;
+  margin: $nx-spacing-sm 0;
+  text-align: right;
```
rpokorny, line 12: This is a pretty big change, do we really want to be sneaking this in like this?

lib/src/components/NxModal/NxModal.scss
```diff
@@ -34,21 +34,12 @@ $nx-modal-content-max-height: calc(100vh - #{$nx-modal-top-spacing} - #{$nx-moda
...
-  &.nx-error {
-    background-color: $nx-fail-color-bg;
-    border: $nx-border-error;
-    color: $nx-text-color-error;
-    margin-right: -1px;
-    margin-bottom: -1px;
-    margin-left: -1px;
+  background-color: #fff;
+  border-top: 1px solid #e9e9e9;
+  padding: $nx-spacing-l $nx-spacing-xl $nx-spacing-l $nx-spacing-l;
```
rpokorny, line 39: Did you mean:

```suggestion
  padding: $nx-spacing-l $nx-spacing-l $nx-spacing-xl $nx-spacing-l;
```

lib/src/components/NxModal/NxModal.scss
```diff
@@ -34,21 +34,12 @@ $nx-modal-content-max-height: calc(100vh - #{$nx-modal-top-spacing} - #{$nx-moda
...
 
-    > .nx-alert--error {
-      border: none;
-    }
+  .nx-alert {
+    margin-bottom: $nx-spacing-xl;
```
rpokorny, line 42: Is this size of bottom margin supposed to apply only here or to all nx-alerts?

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
glenhunter: > It seems a little weird to have a load error, as if the form had been saved or something, but then to not also have any sort of Save button, either before or after the retry succeeds. As it is now, after pressing retry, you're left with only a Cancel button. Which wouldn't realistically make much sense.

I find it weird that we have a retry button in the alert rather than adjacent to the Cancel button, @dturner32 ?

lib/src/base-styles/_nx-btn-bar.scss
glenhunter: > This is a pretty big change, do we really want to be sneaking this in like this?

I thought about that, it's first time that it has appeared in our refactoring so that's why I added it. I'm ok with pulling this out to a separate story and just accepting that in this case our work will not match the Zeplin file.

lib/src/components/NxModal/NxModal.scss
glenhunter: > Is this size of bottom margin supposed to apply only here or to all nx-alerts?

I assumed not, I took the values from the Modal Error Zeplin file

lib/src/base-styles/_nx-btn-bar.scss
rpokorny: > I thought about that, it's first time that it has appeared in our refactoring so that's why I added it. I'm ok with pulling this out to a separate story and just accepting that in this case our work will not match the Zeplin file.

The bigger question to me is whether we actually want this to be universally styled this way.  For instance it makes the `NxButton` examples arguably look weird

lib/src/components/NxModal/NxModal.scss
rpokorny: > I assumed not, I took the values from the Modal Error Zeplin file

The individual component mockups don't really convey external spacing, so I don't think we've had any guidance here one way or the other before this.  @dturner32 ?

lib/src/base-styles/_nx-btn-bar.scss
glenhunter: > > This is a pretty big change, do we really want to be sneaking this in like this?
> 
> I thought about that, it's first time that it has appeared in our refactoring so that's why I added it. I'm ok with pulling this out to a separate story and just accepting that in this case our work will not match the Zeplin file.

I have reverted it here: https://github.com/sonatype/sonatype-react-shared-components/pull/75/commits/98ef407d49b7b11e031a02d8bc56aef15c1be9af

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
glenhunter: > It seems a little weird to have a load error, as if the form had been saved or something, but then to not also have any sort of Save button, either before or after the retry succeeds. As it is now, after pressing retry, you're left with only a Cancel button. Which wouldn't realistically make much sense.

The Retry button now closes the modal: https://github.com/sonatype/sonatype-react-shared-components/pull/75/commits/d0b9587a5d0daaee57666433d2e46ac006e37919

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
```diff
@@ -6,12 +6,23 @@
...
 export default function NxModalFormErrorExample() {
   const [showModal, setShowModal] = useState(false);
   const modalCloseHandler = () => setShowModal(false);
+  const [state, setState] = useState(initialState(''));
+  const [error, setError] = useState<string | null>('');
+
+  function retryHandler() {
+    // lets say the retried action succeeded this time
+    setError(null);
+  }
```
rpokorny, line 22: That `onClick` handler isn't the right way to do this, and if you notice, clicking _anywhere_ on the alert will close the modal at the moment.  Remove that `onClick` handler and try this:

```suggestion
  function retryHandler() {
    // lets say the retried action succeeded this time
    setShowModal(false);
  }
```

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
```diff
@@ -6,12 +6,23 @@
...
 
-import {NxModal, NxFontAwesomeIcon} from '@sonatype/react-shared-components';
+import {NxModal, NxFontAwesomeIcon, NxTextInput, NxLoadError} from '@sonatype/react-shared-components';
 import {faAngry} from '@fortawesome/free-solid-svg-icons';
+import { initialState, userInput } from '@sonatype/react-shared-components/components/NxTextInput/stateHelpers';
 
 export default function NxModalFormErrorExample() {
   const [showModal, setShowModal] = useState(false);
   const modalCloseHandler = () => setShowModal(false);
+  const [state, setState] = useState(initialState(''));
```
rpokorny, line 16:
```suggestion
  const [textFieldState, setTextFieldState] = useState(initialState(''));
```

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
dturner32: > > It seems a little weird to have a load error, as if the form had been saved or something, but then to not also have any sort of Save button, either before or after the retry succeeds. As it is now, after pressing retry, you're left with only a Cancel button. Which wouldn't realistically make much sense.
> 
> I find it weird that we have a retry button in the alert rather than adjacent to the Cancel button, @dturner32 ?

I included the retry button in the error alert because I was under the intention that that was the desired action a user would take in this case. I agree that it makes more sense to include the x/close instead. If a username or password is entered incorrectly the alert message would reflect that and the user could go back and retry be re-entering their credentials. To Ross' point I think a save or submit button should be included as well.

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
```diff
@@ -6,12 +6,21 @@
...
 export default function NxModalFormErrorExample() {
   const [showModal, setShowModal] = useState(false);
   const modalCloseHandler = () => setShowModal(false);
+  const [textFieldState, setTextFieldState] = useState(initialState(''));
+  function retryHandler() {
+    // lets say the retried action succeeded this time
+    setShowModal(false);
+  }
+  function onChange(val: string) {
+    setState(userInput(null, val));
```
rpokorny, line 22:
```suggestion
    setTextFieldState(userInput(null, val));
```

PR #78, opened 2020-06-26
nx-code font size RSC-174

glenhunter:
https://issues.sonatype.org/browse/RSC-174

Brought the `nx-code` font styles in line with the default be removing the font-size declaration. This had the benefit of making the code examples easier to read so I decreased the top/bottom padding by 1px.

I also changed the border radius to our new default value

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "0.50.11",
+  "version": "0.50.12",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


rpokorny:
_On my system_, the `nx-code` elements in this PR are 1px taller than their surrounding inline elements.  I'm not sure if that matters that much, as they're still smaller than our line-box height.  I do worry however that the height of the `nx-code` elements will be system dependent since they use the system monospace font.  It is possible that on some systems they might exceed the line box height.

glenhunter:
> _On my system_, the `nx-code` elements in this PR are 1px taller than their surrounding inline elements. I'm not sure if that matters that much, as they're still smaller than our line-box height. I do worry however that the height of the `nx-code` elements will be system dependent since they use the system monospace font. It is possible that on some systems they might exceed the line box height.

I have created a ticket to select a common OSS monospace font for RSC: https://issues.sonatype.org/browse/RSC-176

PR #83, opened 2020-06-29
Add some real Applitools Tests - RSC-167

rpokorny:
https://issues.sonatype.org/browse/RSC-167

While #68 got the test infrastructure set up, this PR is geared towards getting some realistic tests in place to establish the patterns involved in writing them


Jenkinsfile
```diff
@@ -16,7 +16,7 @@ dockerizedBuildPipeline(
   prepare: {
     githubStatusUpdate('pending')
 
-    sh "docker run --name selenium-chrome -d -p 4444:4444 ${seleniumDockerImage}:${seleniumDockerVersion}"
+    sh "docker run --name selenium-chrome -d -p 4444:4444 -v /dev/shm:/dev/shm ${seleniumDockerImage}:${seleniumDockerVersion}"
```
rpokorny, line 19: This param is recommended in the docs for this docker image.  It turns out we were able to get by without it when running one test at a time but once I turned up `maxInstances` I got errors without this

gallery/visualtests/testUtils.js
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+const { Target } = require('@applitools/eyes-webdriverio');
+
+module.exports = {
+  simpleTest(selector) {
+    return async () => {
```
raveneyex, line 11: Can we use `async` and `await` now? 😮

gallery/visualtests/testUtils.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+
+  hoverTest(elementSelector, hoverSelector = elementSelector) {
+    return async () => {
+      const [targetElement, hoverElement] = await Promise.all([browser.$(elementSelector), browser.$(hoverSelector)]);
+
+      await targetElement.scrollIntoView({ block: 'center' });
+      await hoverElement.moveTo();
+
+      await browser.saveScreenshot('/tmp/before.png');
```
raveneyex, line 26: Where can I read about the difference between `browser.saveScreenshot` and `browser.eyesRegionSnapshot()`?

What do we do with the screenshots?

gallery/visualtests/nxTextInput.js
```diff
@@ -0,0 +1,139 @@
...
+  function getInputElementSelector(componentSelector, inputType = 'input') {
+    return `${componentSelector} ${inputType}`;
+  }
+
+  beforeEach(async function() {
+    await browser.url('#/pages/NxTextInput');
+  });
+
+  describe('Simple NxTextInput', function() {
+    it('has a light border when pristine', simpleTest(simpleComponentSelector));
```
raveneyex, line 29: I'm probably missing something but I don't see how `simpleTest(selector)` is testing that the element has a light border.

gallery/visualtests/nxTextInput.js
```diff
@@ -0,0 +1,139 @@
...
+
+    it('has a dark border when hovered',
+        hoverTest(simpleComponentSelector, getInputElementSelector(simpleComponentSelector)));
```
raveneyex, line 32: Basically the same concern as above: how is `hoverTest(selector, hoverSelector)` testing for a "dark border when hovered"?

gallery/visualtests/testUtils.js
rpokorny: > Can we use async and await now? open_mouth

I'm pretty sure we've been able to for some time.  That said, these files are run directly in node, not put through webpack or anything, so the ability to use async/await here doesn't imply anything about whether it is supported in application code.

I hadn't found async/await particularly necessary in our application code, as we usually have fairly succinct asychronous chains in any given method.  In these tests though, virtually every operation is async so it really helps. (Also, the example code that I started from was written in this style which is really what prompted me to use it 🙂).

gallery/visualtests/testUtils.js
rpokorny: > Where can I read about the difference between browser.saveScreenshot and browser.eyesRegionSnapshot()?

`saveScreenshot` saves the screenshot to a local file, and has nothing to do with applitools.  The calls to that were actually for debugging and I need to remove them; thanks for reminding me that they're still there.

`eyesRegionSnapshot` is one of two custom commands that I added to the `browser` object in `wdio.conf.js`.

rpokorny: > The calls to that were actually for debugging and I need to remove them; thanks for reminding me that they're still there.

fb6f499c

gallery/visualtests/nxTextInput.js
rpokorny: > I'm probably missing something but I don't see how simpleTest(selector) is testing that the element has a light border.

It's up to the writer of the test to then go into applitools and verify that the initial screenshot of a newly written test matches the condition described in the test name.

gallery/visualtests/nxTextInput.js
rpokorny: > Basically the same concern as above: how is hoverTest(selector, hoverSelector) testing for a "dark border when hovered"?

> It's up to the writer of the test to then go into applitools and verify that the initial screenshot of a newly written test matches the condition described in the test name.

PR #84, opened 2020-06-29
update sidebar RSC-151

glenhunter:
https://issues.sonatype.org/browse/RSC-151

There were some discrepancies between what was described in the ticket and what was shown in the Zeplin file which I noted. I used the Zeplin file as the source of truth.

Current styling is in line with the Zeplin file now, however there is a new caret icon which is a change that should be attached to `NxTreeView` not the sidebar.

### DIFF
--- a/gallery/src/main.scss
+++ b/gallery/src/main.scss
@@ -15,6 +15,8 @@
 
 .gallery-nav {
   .nx-tree-view__trigger {
+    padding-left: $nx-spacing-xs;
+
     .nx-tree-view__text {
       @include semi-bold();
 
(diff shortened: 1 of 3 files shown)


gallery/src/main.scss
```diff
@@ -15,30 +15,31 @@
...
 .gallery-link {
+  @include regular();
+
   border-radius: 6px;
   box-sizing: border-box;
   display: inline-block;
-  height: 32px;
-  margin-bottom: 4px;
-  padding-left: 45px;
+  padding: $nx-spacing-xxs $nx-spacing-xs $nx-spacing-xxs 36px;
```
rpokorny, line 42: What's the motivation for the non-standard number here?

glenhunter: > What's the motivation for the non-standard number here?

With the sidebar spacing, and the indentation of the tree view this was the remainder to get to the indentation number specified in the designs

rpokorny: > With the sidebar spacing, and the indentation of the tree view this was the remainder to get to the indentation number specified in the designs

Yes but why was a non-standard number specified in the designs?

glenhunter: > > With the sidebar spacing, and the indentation of the tree view this was the remainder to get to the indentation number specified in the designs
> 
> Yes but why was a non-standard number specified in the designs?

From @dturner32's comment in the ticket, "In other words, it should be 52px from the left of the side bar container" 52 is not divisible by 8. As to why maybe @dturner can chime in?

PR #86, opened 2020-07-03
NxRadio redesign RSC-120

glenhunter:
https://issues.sonatype.org/browse/RSC-120


rpokorny:
Zeplin appears to be specifying a non-standard 14px of vertical space between each radio (measuring from text to text).  The implementation doesn't appear to match

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -10,52 +10,66 @@
...
+  height: 24px;
+  margin-right: $nx-spacing-md;
+  width: 16px;
+  vertical-align: top;
 
-  .nx-radio__inner-circle {
-    fill: currentColor;
+  &:last-child {
+    // due to specificity issues the `container-horizontal` mixin on nx-radio isn't successfully setting this
+    margin-right: 0;
```
rpokorny, line 51: Doesn't appear to be needed anymore

lib/src/components/NxRadio/NxRadio.tsx
```diff
@@ -30,8 +30,8 @@ const NxRadio = forwardRef<HTMLLabelElement, Props>(
                  onChange={() => onChange && onChange(value)}
                  readOnly={!onChange}/>
           <svg className="nx-radio__circle" viewBox="-8 -8 16 16" focusable={false}>
-            <circle r="7" strokeWidth="1" fill="none" className="nx-radio__outer-circle"/>
-            { isChecked && <circle r="5" stroke="none" className="nx-radio__inner-circle"/> }
+            <circle r="7" strokeWidth="1" className="nx-radio__outer-circle"/>
+            { isChecked && <circle r="5" strokeWidth="3" className="nx-radio__inner-circle"/> }
```
rpokorny, line 34: With this new design, constructing the radio of two circles leads to a small visual issue, mostly noticeable when zoomed in as seem below.  It would probably be cleaner to just use one circle for the new design

lib/src/components/NxRadio/NxRadio.tsx
```diff
@@ -30,8 +30,8 @@ const NxRadio = forwardRef<HTMLLabelElement, Props>(
                  onChange={() => onChange && onChange(value)}
                  readOnly={!onChange}/>
           <svg className="nx-radio__circle" viewBox="-8 -8 16 16" focusable={false}>
-            <circle r="7" strokeWidth="1" fill="none" className="nx-radio__outer-circle"/>
-            { isChecked && <circle r="5" stroke="none" className="nx-radio__inner-circle"/> }
+            <circle r="7" strokeWidth="1" className="nx-radio__outer-circle"/>
```
rpokorny, line 33: SVG stroke width emanates equally from both sides of the stroke.  Therefore, to get a circle with a stroke width of 1 and an _outer_ radius of 8 (which I believe is what zeplin is specifying), set the radius to 7.5.
```suggestion
            <circle r="7.5" strokeWidth="1" className="nx-radio__outer-circle"/>
```

rpokorny:
Spacing in the inline example needs work: 
image

rpokorny:
I think we need a focused+checked visual state.  Right now when tabbing into a radio button group the checked button in that group is focused, but gets no visual distinction from when it was checked but not focused.

On a related note, I think we need an example where none of the radios are initial checked, so that we can verify the focused+nonchecked visual state specified in zeplin.

rpokorny:
In zeplin, the radio buttons don't seem to be quite vertically centered relative to their text.  They _are_ centered in the implementation, and I would think they ought to be.  Is zeplin wrong here?

glenhunter:
> In zeplin, the radio buttons don't seem to be quite vertically centered relative to their text. They _are_ centered in the implementation, and I would think they ought to be. Is zeplin wrong here?

@dturner32 ^^

I certainly assumed so.

glenhunter:
> On a related note, I think we need an example where none of the radios are initial checked, so that we can verify the focused+nonchecked visual state specified in zeplin.

I changed the first example to not have a default: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/5fce1073dde8b01ed177a0477ac035e6594835cb

lib/src/components/NxRadio/NxRadio.tsx
glenhunter: > With this new design, constructing the radio of two circles leads to a small visual issue, mostly noticeable when zoomed in as seem below. It would probably be cleaner to just use one circle for the new design
> image

This was really simple to fix with SCSS: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/e58b3889d67bac7d1aa35ac2eb1e670c33143afa

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > Doesn't appear to be needed anymore

Removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/044c3d73d644c420c1a85954d25fd38f2c3829c1

glenhunter:
> I think we need a focused+checked visual state. Right now when tabbing into a radio button group the checked button in that group is focused, but gets no visual distinction from when it was checked but not focused.

@dturner32 WDYT?

rpokorny:
> I changed the first example to not have a default: 5fce107

`null` would be more correct than `''`.

rpokorny:
Is it just me or do the default styles look disabled and the disabled styles look like the older browser-native styles?
image

glenhunter:
> Is it just me or do the default styles look disabled and the disabled styles look like the older browser-native styles?
> image

It's not just you. @dturner32 ?

dturner32:
> > In zeplin, the radio buttons don't seem to be quite vertically centered relative to their text. They _are_ centered in the implementation, and I would think they ought to be. Is zeplin wrong here?
> 
> @dturner32 ^^
> 
> I certainly assumed so.

not quite certain what you mean here Ross, yes they should be centered vertically so if there is a discrepancy in Zeplin then I can fix that

dturner32:
> > Is it just me or do the default styles look disabled and the disabled styles look like the older browser-native styles?
> > image
> 
> It's not just you. @dturner32 ?

the default and disabled styles here should mimic the styling for the form fields. Not sure what you are asking here...

dturner32:
> > I think we need a focused+checked visual state. Right now when tabbing into a radio button group the checked button in that group is focused, but gets no visual distinction from when it was checked but not focused.
> 
> @dturner32 WDYT?

I get what you're saying here Ross, I think this is something I'd like to clarify a little more in standup this morning before making a final decision

rpokorny:
> the default and disabled styles here should mimic the styling for the form fields. Not sure what you are asking here...

Just that at first glance looking at that radio group, I'd think the 1st, 3rd, and 4th options are disabled and the 3rd one is enabled - the opposite of the truth.  I guess it's not that big a deal though since the hover styles would help make it clear and typically one radio will already be selected anyway.

lib/src/components/NxRadio/NxRadio.tsx
rpokorny: > This was really simple to fix with SCSS: e58b388

That was a nicely minimal change from what was already there but it still results in SVG that's needlessly complicated for the current appearance.  How about this: https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-120-NxRadio-redesign...RSC-120-NxRadio-redesign_ross

rpokorny:
> not quite certain what you mean here Ross, yes they should be centered vertically so if there is a discrepancy in Zeplin then I can fix that

Note the 4px vs 2px here: 
image

lib/src/components/NxRadio/NxRadio.tsx
rpokorny: > That was a nicely minimal change from what was already there but it still results in SVG that's needlessly complicated for the current appearance. How about this: RSC-120-NxRadio-redesign...RSC-120-NxRadio-redesign_ross

Oh a related question there: I noticed you set the interior to explicitly be white whereas it's just see-through in my impl.  Which should it be?

glenhunter:
> > I changed the first example to not have a default: [5fce107](https://github.com/sonatype/sonatype-react-shared-components/commit/5fce1073dde8b01ed177a0477ac035e6594835cb)
> 
> `null` would be more correct than `''`.

https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/39f7812e4f2c684388f4c545c323612a9711f9d5

gallery/src/components/NxRadio/NxRadioExample.tsx
```diff
@@ -9,7 +9,7 @@ import React, { useState } from 'react';
 import { NxRadio } from '@sonatype/react-shared-components';
 
 export default function NxRadioExample() {
-  const [color, setColor] = useState<string | null>('green');
+  const [color, setColor] = useState<string | null>('null');
```
rpokorny, line 12:
```suggestion
  const [color, setColor] = useState<string | null>(null);
```

lib/src/components/NxRadio/NxRadio.tsx
glenhunter: > > This was really simple to fix with SCSS: [e58b388](https://github.com/sonatype/sonatype-react-shared-components/commit/e58b3889d67bac7d1aa35ac2eb1e670c33143afa)
> 
> That was a nicely minimal change from what was already there but it still results in SVG that's needlessly complicated for the current appearance. How about this: [RSC-120-NxRadio-redesign...RSC-120-NxRadio-redesign_ross](https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-120-NxRadio-redesign...RSC-120-NxRadio-redesign_ross)

Merged with thanks.

glenhunter: > > That was a nicely minimal change from what was already there but it still results in SVG that's needlessly complicated for the current appearance. How about this: RSC-120-NxRadio-redesign...RSC-120-NxRadio-redesign_ross
> 
> Oh a related question there: I noticed you set the interior to explicitly be white whereas it's just see-through in my impl. Which should it be?

I needed to make it white for my code to work, but there's no reason it cannot be transparent given your implementation

glenhunter:
> Zeplin appears to be specifying a non-standard 14px of vertical space between each radio (measuring from text to text). The implementation doesn't appear to match

Zeplin now shows 12px for the space between the radio buttons, but I thought we were working in multiples of 8px. Should it be 8px or 16px @dturner32 ?

Screen Shot 2020-07-08 at 1 49 46 PM

dturner32:
> > Zeplin appears to be specifying a non-standard 14px of vertical space between each radio (measuring from text to text). The implementation doesn't appear to match
> 
> Zeplin now shows 12px for the space between the radio buttons, but I thought we were working in multiples of 8px. Should it be 8px or 16px @dturner32 ?
> 
> Screen Shot 2020-07-08 at 1 49 46 PM

it should be 8px

glenhunter:
> > > Zeplin appears to be specifying a non-standard 14px of vertical space between each radio (measuring from text to text). The implementation doesn't appear to match
> > 
> > 
> > Zeplin now shows 12px for the space between the radio buttons, but I thought we were working in multiples of 8px. Should it be 8px or 16px @dturner32 ?
> > Screen Shot 2020-07-08 at 1 49 46 PM
> 
> it should be 8px

Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/b250c12389ebb704c9127765911f2f0bfda487c5

glenhunter:
> Spacing in the inline example needs work:
> image

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/07c6b19906d19eb23766ef01c7631332b66e7993

glenhunter:
> > > I think we need a focused+checked visual state. Right now when tabbing into a radio button group the checked button in that group is focused, but gets no visual distinction from when it was checked but not focused.
> > 
> > 
> > @dturner32 WDYT?
> 
> I get what you're saying here Ross, I think this is something I'd like to clarify a little more in standup this morning before making a final decision

`:focused:checked` state added here: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/dc9772ef4f62dea0755307ac0bb35a96b53181ed

rpokorny:
> :focused:checked state added here: dc9772e

That doesn't implement @dturner32 's latest styles which have a box-shadow

rpokorny:
Some more SVG tweaks for you; getting back to the see-through donut: https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-120-NxRadio-redesign...RSC-120-NxRadio-redesign_ross

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.1.1",
+  "version": "1.2.1",
```
rpokorny, line 3:
```suggestion
  "version": "1.2.0",
```

lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "1.1.1",
+  "version": "1.2.1",
```
rpokorny, line 3:
```suggestion
  "version": "1.2.0",
```

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -10,52 +10,70 @@
...
+  cursor: pointer;
+  fill: none;
+  height: 24px;
+  margin-right: $nx-spacing-md;
+  width: 16px;
+  vertical-align: top;
+}
+
+.nx-radio__outer-circle {
+  stroke: #d1d6f0;
```
rpokorny, line 57: Now that the inner circle has a white fill (when visible), I feel like the outer circle should also have a white fill so that we're consistent on whether or not you can see through the middle.  Note of course that the outer circle should only have that white fill when it is _not_ selected - otherwise that fill would cover up the inner circle.

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -10,52 +10,70 @@
...
+}
+
+.nx-radio__inner-circle {
+  fill: #fff;
+  stroke: #045dd1;
+}
+
+.nx-radio__input {
+  margin: 0;
```
rpokorny, line 66: What's this for?

rpokorny:
The circle appears to get clipped a little bit in IE.  I don't think this is worth investigating.

image

rpokorny:
Shouldn't the focus glow apply to focused non-checked radios as well?

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > What's this for?

Overrides browser agent styling

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > Now that the inner circle has a white fill (when visible), I feel like the outer circle should also have a white fill so that we're consistent on whether or not you can see through the middle. Note of course that the outer circle should only have that white fill when it is _not_ selected - otherwise that fill would cover up the inner circle.

Tweaked here: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/3f81f224ad4f5af9bcddacd04d168d4c20b81afe

glenhunter:
> Shouldn't the focus glow apply to focused non-checked radios as well?

Ah, I missed that: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/426dcb32cd1918ba156fc56b83f56cb298704dc2

lib/src/components/NxRadio/NxRadio.scss
rpokorny: > Overrides browser agent styling

But this component uses absolute positioning, so I'm not finding a way in which it's margin matters

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -10,52 +10,72 @@
...
+      fill: #e9e9e9;
+      stroke: #949494;
+    }
   }
+}
+
+.nx-radio__circle {
+  border-radius: 8px;
+  cursor: pointer;
+  fill: none;
```
rpokorny, line 49: Unnecessary now

rpokorny:
Should the hover style remove the focus glow? 
image

rpokorny:
I've noticed that the hover style (and clickability) is activated if anywhere on the `<label>` is hovered - which can be far out to the right when in a fieldset, since fieldsets display the `<label>`s as block elements.  This is fine, but what's inconsistent is that you get the default cursor here.  In fact even hovering over the label text you get the default cursor; only hovering over the SVG results in a pointer cursor.

image

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -17,13 +17,25 @@
   }
 
   &:focus-within {
+    .nx-radio__circle {
+      filter: drop-shadow(0 0 3px rgba(2, 166, 217, 0.4));
```
rpokorny, line 21: If this drop-shadow is going to be used across the board it should probably be moved to a variable or something.  Maybe the entire `filter` property should be moved to a mixin.

glenhunter:
> Should the hover style remove the focus glow?
> image

I'm going to go ahead and assume that if we remove the :focus colour on :hover, we also remove the glow: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/1c07a789fefe8a31b0adff34b3ecc0c9f6c75f91

lib/src/scss-shared/_nx-input-helpers.scss
rpokorny, line 21: When designing the color system don't forget to include this non-opaque color.

glenhunter: > When designing the color system don't forget to include this non-opaque color.

@dturner32 ^^

glenhunter:
> I've noticed that the hover style (and clickability) is activated if anywhere on the `<label>` is hovered - which can be far out to the right when in a fieldset, since fieldsets display the `<label>`s as block elements. This is fine, but what's inconsistent is that you get the default cursor here. In fact even hovering over the label text you get the default cursor; only hovering over the SVG results in a pointer cursor.
> 
> image

After talking with @dturner32 we decided to go with a width of 100px for NxRadio and NxCheckbox. Because the text can't wrap longer text will push the right edge out. 100px gives us a slightly larger click target on very short text, but longer text only has a click target as long as the label: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/4c01575f631a127edf402dd1f452794a5f0cde57

lib/src/scss-shared/_nx-input-helpers.scss
glenhunter: > If this drop-shadow is going to be used across the board it should probably be moved to a variable or something. Maybe the entire `filter` property should be moved to a mixin.

I think that's a good idea, ironically the best place for it is `input-helpers.scss` where it is referenced in the very next mixin: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/69446e5e5883d8d68d0b1ac6f603cc94c2a9b966

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -10,52 +10,78 @@
...
-  $line-height: 20px;
-
   @include container-horizontal;
   @include regular();
   @include focusable-toggle-input(".nx-radio__outer-circle");
 
-  display: inline-block;
-  font-size: 14px;
-  line-height: $line-height;
+  display: block;
```
rpokorny, line 17: With this change they're _always_ going to be block.  I was expecting the width adjustment to be done in the context of a fieldset.

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -10,52 +10,78 @@
...
+  margin-bottom: $nx-spacing-xs;
   position: relative;
   white-space: nowrap;
+  max-width: 100px;
```
rpokorny, line 21: Did you mean `min-width`?
image

glenhunter: > Did you mean `min-width`?
> image

No, if it's `min-width` then the element will still expand to full width which is what I'm trying to avoid. What you see above in your screenshot is expected. Either `width` or `max-width` give the results we want.

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > With this change they're _always_ going to be block. I was expecting the width adjustment to be done in the context of a fieldset.

Yes, I think they should always be block. There are exceedingly few instances where we would want `NxRadio` to be inline, so I think it's better to make them block by default and if there's a case where they are inline that should be custom.

glenhunter: > > With this change they're _always_ going to be block. I was expecting the width adjustment to be done in the context of a fieldset.
> 
> Yes, I think they should always be block. There are exceedingly few instances where we would want `NxRadio` to be inline, so I think it's better to make them block by default and if there's a case where they are inline that should be custom.

The same will be true of `NxCheckbox`.

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > > Did you mean `min-width`?
> > image
> 
> No, if it's `min-width` then the element will still expand to full width which is what I'm trying to avoid. What you see above in your screenshot is expected. Either `width` or `max-width` give the results we want.

Note that despite the fact that the `max-width` is only 100px the entire length of the label triggers the cursor change on hover, but _only_ the label. I have confirmed this behaviour in IE

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > > Overrides browser agent styling
> 
> But this component uses absolute positioning, so I'm not finding a way in which its margin matters

There was something I was doing where the margin was an issue, but even if I'm mistaken I see no harm in removing the user agent margin

lib/src/components/NxRadio/NxRadio.scss
rpokorny: > Yes, I think they should always be block. There are few instances where we would want NxRadio to be inline, so I think it's better to make them block by default and if there's a case where they are inline that should be custom.

> The same will be true of NxCheckbox.

Be aware that my original reason for making them inline was for consistency with natively styled radios and checkboxes.

lib/src/components/NxRadio/NxRadio.scss
rpokorny: > Either width or max-width give the results we want.

So the result that you want is that if it's less than 100px wide you can click to the right of it, up to the 100px line, but if it's more than 100px wide you can only click on the text?

rpokorny: While looking at this I found that it doesn't handle overflow properly.  In this case it appears to cut off at the edge of `<main>`

image

glenhunter: > > Either width or max-width give the results we want.
> 
> So the result that you want is that if it's less than 100px wide you can click to the right of it, up to the 100px line, but if it's more than 100px wide you can only click on the text?

Yes

glenhunter: > While looking at this I found that it doesn't handle overflow properly. In this case it appears to cut off at the edge of `<main>`
> 
> image

After speaking with @dturner32 we decided on using ellipsis truncation on form labels, and so I have created a new variable for form label width. https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/223eb5b462605d3af1066750992960559edc03fe

glenhunter: > > > Either width or max-width give the results we want.
> > 
> > 
> > So the result that you want is that if it's less than 100px wide you can click to the right of it, up to the 100px line, but if it's more than 100px wide you can only click on the text?
> 
> Yes

Solved this problem by using flex box for the layout: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/ad7a2c1bab2dff105b449cbdd3309372d8bd5173

glenhunter: > > > > Either width or max-width give the results we want.
> > > 
> > > 
> > > So the result that you want is that if it's less than 100px wide you can click to the right of it, up to the 100px line, but if it's more than 100px wide you can only click on the text?
> > 
> > 
> > Yes
> 
> Solved this problem by using flex box for the layout: [ad7a2c1](https://github.com/sonatype/sonatype-react-shared-components/commit/ad7a2c1bab2dff105b449cbdd3309372d8bd5173)

Didn't need the flex-shrink values: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/ef865a5c5920b11d0b03f1253d58bcd12d2fc708

rpokorny: > Didn't need the flex-shrink values: ef865a5

Doesn't look like the flex is taking effect to me: 
image

rpokorny:
You may recall we had an alternative focus indicator in IE11 due to it not supporting `:focus-within`.  That indicator seems to have become mis-aligned: 
image

lib/src/components/NxRadio/NxRadio.scss
glenhunter: > > Didn't need the flex-shrink values: [ef865a5](https://github.com/sonatype/sonatype-react-shared-components/commit/ef865a5c5920b11d0b03f1253d58bcd12d2fc708)
> 
> Doesn't look like the flex is taking effect to me:
> image

Fixed here with `max-content`: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/354c7030d195004b246fd5f513f6b0b8a78d1b59

glenhunter:
> You may recall we had an alternative focus indicator in IE11 due to it not supporting `:focus-within`. That indicator seems to have become mis-aligned:
> image

Well a bunch of the attributes in there didn't seem to do anything so I removed them, but the solution required a magic number, I'm open to other ideas: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/674882aa58f053a9f0b65288d412f21a5b168958

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -3,19 +3,16 @@
...
 @mixin focusable-toggle-input($focus-border-selector) {
-  position: relative;
-
   .nx-radio__input, .nx-checkbox__input {
-    position: absolute;
-    z-index: -1;
-
     &:focus {
       // less-nice IE11 focus style because IE11 doesn't support :focus-within
+      height: 100%;
```
rpokorny, line 9: Doesn't seem to be necessary

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -3,19 +3,16 @@
...
+      margin-top: 5px;
       margin-right: $nx-spacing-xxs;
       outline: 1px solid $nx-color-blue-44;
       opacity: 1;
       position: static;
-      height: 100%;
+      vertical-align: top;
```
rpokorny, line 15: Consider using a number for `vertical-align` (coincidentally 5px still seems to work) rather than this plus a separate margin.

rpokorny:
`NxCheckbox` page is now broken.  I suspect this is what those extra CSS properties you removed were for:
image

lib/src/scss-shared/_nx-input-helpers.scss
glenhunter: > Consider using a number for `vertical-align` (coincidentally 5px still seems to work) rather than this plus a separate margin.

I always forget I can use a value for vertical-align, thanks for the reminder: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/74308db5d3ddf7206ab167c431feba89dc373d74

lib/src/scss-shared/_nx-input-helpers.scss
glenhunter: > Doesn't seem to be necessary

Removed: https://github.com/sonatype/sonatype-react-shared-components/pull/86/commits/74308db5d3ddf7206ab167c431feba89dc373d74

glenhunter:
> `NxCheckbox` page is now broken. I suspect this is what those extra CSS properties you removed were for:
> image

I switched the base to be a feature branch, then we can complete NxCheckbox and merge both together

PR #88, opened 2020-07-07
Allow fuzzy filter config from props - RSC-217

raveneyex:
[RSC-217](https://issues.sonatype.org/browse/RSC-217)

### DIFF
--- a/gallery/src/components/NxStatefulTreeViewMultiSelect/NxStatefulTreeViewMultiSelectFuzzyFilterExample.tsx
+++ b/gallery/src/components/NxStatefulTreeViewMultiSelect/NxStatefulTreeViewMultiSelectFuzzyFilterExample.tsx
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { useState } from 'react';
+import { faCapsules } from '@fortawesome/free-solid-svg-icons';
+
+import {
+  NxStatefulTreeViewMultiSelect,
+  NxFontAwesomeIcon
+} from '@sonatype/react-shared-components';
+
+const NxStatefulTreeViewMultiSelectFuzzyFilterExample = () => {
+  const options = [
+    {
+      id: 'app1',
+      name: 'App With Many Spaces 1'
+    }, {
+      id: 'app2',
+      name: 'App With Many Spaces 2'
+    }, {
+      id: 'app3',
+      name: 'App With Many Spaces 3'
+    }, {
+      id: 'app4',
+      name: 'App With Many Spaces 4'
+    }, {
+      id: 'app5',
+      name: '- App With Many Spaces 5'
+    }, {
+      id: 'app6',
+      name: '-App With Many Spaces 6'
+    }, {
+      id: 'app7',
+      name: '_App With Many Spaces 7'
+    }, {
+      id: 'app8',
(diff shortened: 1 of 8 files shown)


lib/src/components/NxTreeViewSelect/NxTreeViewMultiSelect/stateful/NxStatefulTreeViewMultiSelect.tsx
```diff
@@ -7,20 +7,36 @@
...
+   *
+   * Defaults provided to handle our usual max-length for orgs/apps (200):
+   * According to https://fusejs.io/concepts/scoring-theory.html#scoring-theory
+   * an entry is considered a match if it is X characters from location,
+   * where:
+   *  X = distance * threshold,
+   *  location = 0 (default)
+   */
+  const filterConfigOptions: FuseOptions<Option> = {
+    keys: ['name'],
```
rpokorny, line 34: I don't think we want the caller to be able to override the `keys`

rpokorny:
Shouldn't the same changes be made in `NxTreeViewRadioSelect`?

rpokorny:
This work ultimately needs to be done on a version that IQ can easily upgrade to right? IQ is currently on 0.44.9. So what I'd recommend is to branch this branch off of the 0.44.9 tag, and bump the version here to 0.44.10.  Then after it's approved, Manually publish a release directly from this branch.  Then after that, add another commit to the branch bumping the version to 1.1.0, and merge _that_ into master.

gallery/src/components/NxStatefulTreeViewMultiSelect/NxStatefulTreeViewMultiSelectFuzzyFilterExample.tsx
```diff
@@ -0,0 +1,80 @@
...
+      name: '- App-With-Many-Dashes- 12'
+    }
+  ];
+
+  const [selection, onSelectionChange] = useState<Set<string | null>>(new Set(['app10']));
+
+  // See https://fusejs.io/api/options.html for more config options
+  const fuseConfig = {
+    threshold: 0.1,
+    distance: 100
```
rpokorny, line 61: Add an explanation of the effects of this config to the example description

lib/src/components/NxTreeViewSelect/NxTreeViewMultiSelect/stateful/NxStatefulTreeViewMultiSelect.tsx
```diff
@@ -7,20 +7,36 @@
...
+    threshold: 0.1,
+    distance: 2000,
```
rpokorny, line 36: These defaults that we provide should probably be documented, so that callers who only partially override them can know what to expect

rpokorny: > These defaults that we provide should probably be documented, so that callers who only partially override them can know what to expect

Or alternatively, maybe these defaults should only be used if the `fuzzyFilterConfig` isn't specified at all.

rpokorny:
Looking in the section of the Fuze docs that you reference in the code comment, I found an alternative approach that is probably cleaner:

> If you don't care where the pattern appears in the string, and you still want to consider it a match, set `ignoreLocation` to `true`.

rpokorny:
> Looking in the section of the Fuze docs that you reference in the code comment, I found an alternative  approach that is probably cleaner:
> 
> > If you don't care where the pattern appears in the string, and you still want to consider it a match, set ignoreLocation to true.

Consider how much smaller of a change that would be, I'd probably be in favor of discarding all the stuff here around making the Fuze config customizable.  No one is asking for that at the moment, so it feels a bit YAGNI.

PR #89, opened 2020-07-08
Create mixin for adding a custom default font-size to projects RSC-182

glenhunter:
https://issues.sonatype.org/browse/RSC-182

The mixin should be `@include`d in the project's `<body>` tag. I changed the default `<body>` font as well as all other classes that are set to that size in order to override browser agent settings.

Screenshot below shows the default font-size changed to `8px` using the mixin:

Screen Shot 2020-07-08 at 2 09 26 PM

I believe that we could expand upon this to resize the headings to proportionally match the custom size.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/src/scss-shared/_nx-text-helpers.scss
+++ b/lib/src/scss-shared/_nx-text-helpers.scss
@@ -45,7 +45,7 @@
   .nx-text-input__input,
   .nx-radio,
   .nx-checkbox {
-    font-size: #{$custom-font-size};
+    font-size: $custom-font-size;
   }
 }
 
--- a/lib/src/scss-shared/_nx-text-helpers.scss
+++ b/lib/src/scss-shared/_nx-text-helpers.scss
@@ -32,7 +32,7 @@
 }
 
 // for specificity reasons this should be called using a custom class on the <body> tag
-// e.g. `<body className="nx-body nx-body--custom-font-size">`
+// e.g. `<body className="nx-body nx-body--my-app">`
 @mixin font-size($custom-font-size) {
   &.nx-body,
 
--- a/lib/src/scss-shared/_nx-text-helpers.scss
+++ b/lib/src/scss-shared/_nx-text-helpers.scss
@@ -31,6 +31,24 @@
   white-space: nowrap;
 }
 
+// Should be called on a custom class attached to the <body> tag
+@mixin font-size($custom-font-size) {
+  &.nx-body,
+
+  // buttons
+  .nx-btn,
+  .nx-dropdown-link,
+  .nx-dropdown-button,
+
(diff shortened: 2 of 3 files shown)


gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.0.2",
+  "version": "1.0.3",
```
rpokorny, line 3: Should be 1.1.0; this is a new feature

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -31,6 +31,24 @@
   white-space: nowrap;
 }
 
+// Should be called on a custom class attached to the <body> tag
```
rpokorny, line 34: I'm not sure I follow what this means by "should be called on a custom class"

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -31,6 +31,24 @@
...
+@mixin font-size($custom-font-size) {
+  &.nx-body,
```
rpokorny, line 36: Is it possible to `@include` at the top level (not nested inside a block)? If so, this `&` could be removed.

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -31,6 +31,24 @@
...
+  .nx-btn,
+  .nx-dropdown-link,
+  .nx-dropdown-button,
+
+  // form elements
+  .nx-label,
+  .nx-text-input__input,
+  .nx-radio,
+  .nx-checkbox {
+    font-size: #{$custom-font-size};
```
rpokorny, line 48:
```suggestion
    font-size: $custom-font-size;
```

gallery/package.json
glenhunter: > Should be 1.1.0; this is a new feature

Adjusted here: https://github.com/sonatype/sonatype-react-shared-components/pull/89/commits/3dfbe6a21e17758446d391f0cc5c0d1bfcb30ea5

lib/src/scss-shared/_nx-text-helpers.scss
glenhunter: > I'm not sure I follow what this means by "should be called on a custom class"

Improved the comment here: https://github.com/sonatype/sonatype-react-shared-components/pull/89/commits/e7519e4dce13ee2b2ef0373a0c8edeb30c8bdedc

lib/src/scss-shared/_nx-text-helpers.scss
glenhunter: > Is it possible to `@include` at the top level (not nested inside a block)? If so, this `&` could be removed.

It's not, it has to be associated with an element. I've gone further and decided that it would be best to be proscriptive and state that it should be used specifically on `<body>`

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -31,6 +31,25 @@
   white-space: nowrap;
 }
 
+// for specificity reasons this should be called using a custom class on the <body> tag
+// e.g. `<body className="nx-body nx-body--custom-font-size">`
```
rpokorny, line 35: A more realistic suggestion I think - we don't want to imply that they need to make up a separate classname strictly pertaining to the font-size adjustment.
```suggestion
// e.g. `<body className="nx-body nx-body--my-app">`
```

PR #90, opened 2020-07-08
Updgrade and tweak fusejs - RSC-217

raveneyex:
[RSC-217](https://issues.sonatype.org/browse/RSC-217)

## The Problem
By default fuse.js is biased to matching terms at the beginning of the string and does not match terms at the tail unless configured to do so.
For example, given a string `"A sentence with many spaces"` and a search term `" spaces"`, fuse.js won't return any matches if the default configuration is used.
The way to change that is to leverage a config property called `ignoreLocation`, but this property is not present in the version that RSC was using, hence the need to update it.


lib/src/util/fuzzyFilter.ts
```diff
@@ -5,18 +5,20 @@
...
  */
 import Fuse from 'fuse.js';
+import { map, prop } from 'ramda';
 
-export default function fuzzyFilter<T>(input: T[], term: string, options: Fuse.FuseOptions<T>): T[] {
+export default function fuzzyFilter<T>(input: T[], term: string, options: Fuse.IFuseOptions<T>): T[] {
   // fuse.js defaults "shouldSort" option to true, which is counter intuitive and contradicts their usage example
   // fix it by setting shouldSort to false by default
   const normalizedOptions = {
+    ...options,
```
rpokorny, line 14: Is there a reason this moved?

rpokorny:
Needs version bump

lib/src/util/fuzzyFilter.ts
raveneyex: > Is there a reason this moved?

Yeah, I moved it so that it wouldn't end up overwriting the defaults below.

raveneyex:
> Needs version bump

Done in 1ba36052cf9467d2799af1ecb12f53f01a58f72d

lib/src/util/fuzzyFilter.ts
marushkevych: > Yeah, I moved it so that it wouldn't end up overwriting the defaults below.

I think that was intentional, we want to allow overwriting defaults.

raveneyex: > > Yeah, I moved it so that it wouldn't end up overwriting the defaults below.
> 
> I think that was intentional, we want to allow overwriting defaults.

Changed in f10386b29b7074ceaf74dd50a8df8d4087d1df48

PR #91, opened 2020-07-09
Improvement Day: enforce yarn.lock urls

rpokorny:
As this is an open-source project, the URLs in the yarn.lock files should point to publicly reachable servers - namely registry.npmjs.org.  However, most of us here at Sonatype have our systems configured to use repo.sonatype.com as an npm proxy, which results in new dependencies getting installed with repo.s.c URLs in the lockfiles.  This PR adds another shell script to Jenkins that checks for the existence of repo.s.c URLs and fails the build if they are present.  Additionally it fixes the repo.s.c URLs that had made there way in over the past few months.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+registry "https://registry.npmjs.org/"
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,1 +1,1 @@
-registry "https://registry.npmjs.org/"
+registry "https://registry.npmjs.org"
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+registry: "https://registry.npmjs.org/"
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,0 @@
-registry: "https://registry.npmjs.org/"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,7 +56,7 @@ dockerizedBuildPipeline(
       fi
     '''
 
-    // As this is an open source project, yarn.lock URLs should point to npmjs.com, not repo.sonatype.com
+    // As this is an open source project, yarn.lock URLs should point to npmjs.org, not repo.sonatype.com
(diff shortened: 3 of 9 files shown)


Blacktiger:
Should we also define a .yarnrc in the project to override the registry back to the public one? https://stackoverflow.com/questions/45998418/how-to-override-global-yarn-registry-for-particular-project

rpokorny:
> Should we also define a .yarnrc in the project to override the registry back to the public one?

Hmm, wouldn't that also override it for Jenkins, slowing down CI?

rpokorny:
> > Should we also define a .yarnrc in the project to override the registry back to the public one?
> 
> Hmm, wouldn't that also override it for Jenkins, slowing down CI?

Though I think we could get around that by invoking yarn with the `--registry` flag in Jenkins.  I'll ask CDI what URL that should point to, and if it's OK for that URL to be present in publicly visible code.

Blacktiger:
> > > Should we also define a .yarnrc in the project to override the registry back to the public one?
> > 
> > 
> > Hmm, wouldn't that also override it for Jenkins, slowing down CI?
> 
> Though I think we could get around that by invoking yarn with the `--registry` flag in Jenkins. I'll ask CDI what URL that should point to, and if it's OK for that URL to be present in publicly visible code.

Right, that way we can avoid devs having to change their registry whenever they want to add a new dependency. We might still want to keep the script around just-in-case, but we might also be able to just drop it which would be simpler.

rpokorny:
> Though I think we could get around that by invoking yarn with the --registry flag in Jenkins. I'll ask CDI what URL that should point to, and if it's OK for that URL to be present in publicly visible code.

It took a weird combination of npmrc and yarnrc files, but I got it to work as of 8feeb19

rpokorny:
> Right, that way we can avoid devs having to change their registry whenever they want to add a new dependency. We might still want to keep the script around just-in-case, but we might also be able to just drop it which would be simpler.

I'd prefer to keep it around.

PR #92, opened 2020-07-16
Example page for font size mixin RSC-225

glenhunter:
https://issues.sonatype.org/browse/RSC-225

Create an example page for the `nx-font-size` mixin that explains its purpose and use.

I also renamed it to `nx-font-size` to match other mixins. We probably want to make this consistent will all of them.

### DIFF
--- a/gallery/src/pageConfig.ts
+++ b/gallery/src/pageConfig.ts
@@ -122,9 +122,9 @@ const pageConfig: PageConfig = {
     'nx-tile': NxTilePage
   },
   'Styles - Mixins & Helpers': {
+    'Custom app font size': NxFontSizePage,
     'nx-clickable': NxClickablePage,
     'nx-container-helpers': NxContainerHelpersPage,
-    'Custom app font-size': NxFontSizePage,
     'nx-scrollable': NxScrollablePage,
     'nx-truncate-ellipsis': NxTruncatePage
   },
--- a/gallery/src/pageConfig.ts
+++ b/gallery/src/pageConfig.ts
@@ -64,6 +64,7 @@ import NxTruncatePage from './styles/NxTruncateEllipsis/NxTruncatePage';
 import NxCodePage from './styles/NxCode/NxCodePage';
 import StylingComponentsPage from './pages/StylingComponents';
 import AdditionalResourcePage from './pages/AdditionalResources';
+import NxFontSizePage from './styles/NxFontSize/NxFontSizePage';
 
 const pageConfig: PageConfig = {
   'React Components': {
@@ -123,6 +124,7 @@ const pageConfig: PageConfig = {
   'Styles - Mixins & Helpers': {
     'nx-clickable': NxClickablePage,
     'nx-container-helpers': NxContainerHelpersPage,
+    'nx-font-size': NxFontSizePage,
     'nx-scrollable': NxScrollablePage,
     'nx-truncate-ellipsis': NxTruncatePage
   },
--- a/gallery/src/pageConfig.ts
+++ b/gallery/src/pageConfig.ts
@@ -124,7 +124,7 @@ const pageConfig: PageConfig = {
   'Styles - Mixins & Helpers': {
     'nx-clickable': NxClickablePage,
     'nx-container-helpers': NxContainerHelpersPage,
-    'nx-font-size': NxFontSizePage,
+    'Custom app font-size': NxFontSizePage,
     'nx-scrollable': NxScrollablePage,
     'nx-truncate-ellipsis': NxTruncatePage
   },
--- a/gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
+++ b/gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
@@ -1,0 +1,10 @@
+<!--
+
+    Copyright (c) 2019-present Sonatype, Inc.
+    This program and the accompanying materials are made available under
+    the terms of the Eclipse Public License 2.0 which accompanies this
+    distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+
+-->
+<body class="nx-body project-body--custom-font-size">
+</body>
--- a/gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
+++ b/gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
@@ -9,7 +9,7 @@
 <div class="nx-body project-body--custom-font-size">
   <p class="nx-p">Normally this text would be 16px, but it should be 12px.</p>
   <p class="nx-p">
-    Note: for illustration purposes we have used the mixin to style a &lt;div&gt;, but it should be
+    Note: for illustration purposes we have used the mixin to style a &lt;div&gt;, but in actual usage it should be
     applied to the &lt;body&gt; tag.
   </p>
 </div>
--- a/gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
+++ b/gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
@@ -6,5 +6,10 @@
     distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
 
 -->
-<body class="nx-body project-body--custom-font-size">
-</body>
+<div class="nx-body project-body--custom-font-size">
+  <p class="nx-p">Normally this text would be 16px, but it should be 12px.</p>
+  <p class="nx-p">
+    Note: for illustration purposes we have used the mixin to style a &lt;div&gt;, but it should be
+    applied to the &lt;body&gt; tag.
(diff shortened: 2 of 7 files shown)


lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -33,7 +33,7 @@
 
 // for specificity reasons this should be called using a custom class on the <body> tag
 // e.g. `<body className="nx-body nx-body--my-app">`
-@mixin font-size($custom-font-size) {
+@mixin nx-font-size($custom-font-size) {
```
rpokorny, line 36: This is a breaking change.  I don't think we want to go to 2.0.0 over this.  Especially considering that less than half of our mixin names begin with `nx-`.

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
```diff
@@ -0,0 +1,48 @@
...
+        {' '}<code className="nx-code">&lt;body&gt;</code> tag of your project to adjust the default RSC font size to
+        better match your project. The default font size in RSC is 16px.
+      </p>
+      <p className="nx-p">
+        The mixin takes a single variable which is the desired font-size with unit. For example:
+        {' '}<code className="nx-code">@include nx-font-size(14px);</code> or
+        {' '}<code className="nx-code">@include nx-font-size(1.1em);</code>.
+      </p>
+      <p className="nx-p">
+        The mixin adjusts the default text size as well as several user agent set form element font sizes.
```
rpokorny, line 28: "user agent set form element font sizes" is pretty clunky; consider rewording.

lib/src/scss-shared/_nx-text-helpers.scss
glenhunter: > This is a breaking change. I don't think we want to go to 2.0.0 over this. Especially considering that less than half of our mixin names begin with `nx-`.

Right, reverted here: https://github.com/sonatype/sonatype-react-shared-components/pull/92/commits/f250f06da45d57fa575fddb0be981a3ea3eb2221

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
glenhunter: > "user agent set form element font sizes" is pretty clunky; consider rewording.

Yeah, that wasn't great, how's this: https://github.com/sonatype/sonatype-react-shared-components/pull/92/commits/b997b6b56d0d6584116ae9f5e611320992c82bd8

gallery/src/pageConfig.ts
```diff
@@ -123,6 +124,7 @@ const pageConfig: PageConfig = {
   'Styles - Mixins & Helpers': {
     'nx-clickable': NxClickablePage,
     'nx-container-helpers': NxContainerHelpersPage,
+    'nx-font-size': NxFontSizePage,
```
rpokorny, line 127: Should this also be adjusted to match its current name?

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
```diff
@@ -0,0 +1,49 @@
...
+        using it on a new project then you should confirm with your designer, or the design group that custom font
+        sizes are a design requirement.
+      </p>
+    </GalleryDescriptionTile>
+    <GalleryExampleTile title="Sample HTML"
+                        codeExamples={nxFontSizeHtmlExampleCode}
+                        htmlExample={nxFontSizeHtmlExampleCode}>
+      In the example below the &lt;body&gt; tag has a custom CSS class applied which is referenced in the SCSS.
+    </GalleryExampleTile>
+    <GalleryExampleTile title="Mixin usage"
```
rpokorny, line 42: These two examples should be one with two code examples

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
```diff
@@ -0,0 +1,49 @@
...
+                        codeExamples={nxFontSizeScssExampleCode}
```
rpokorny, line 43: Set the language correctly

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </p>
+      <p className="nx-p">
+        This mixin is primarily intended for use when RSC is incorporated into existing/legacy projects. If you're
+        using it on a new project then you should confirm with your designer, or the design group that custom font
+        sizes are a design requirement.
+      </p>
+    </GalleryDescriptionTile>
+    <GalleryExampleTile title="Sample HTML"
+                        codeExamples={nxFontSizeHtmlExampleCode}
+                        htmlExample={nxFontSizeHtmlExampleCode}>
```
rpokorny, line 39: I would have expected an example that actually renders something so the smaller font size can be seen.  You may need to adjust the example to use a `<div>` instead of `<body>`.  If so, add a note to the example explaining that you'd usually use it on `<body>` but for the example it has to be on `<div>`

gallery/src/pageConfig.ts
glenhunter: > Should this also be adjusted to match its current name?

"font-size" did not seem like a very helpful name so I changed it to "Custom app font size": https://github.com/sonatype/sonatype-react-shared-components/pull/92/commits/1feeffdc9a085094d55cbb5682c76fdd494a3c3a

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
glenhunter: > Set the language correctly

https://github.com/sonatype/sonatype-react-shared-components/pull/92/commits/2d6a168e17745955a4ec3f27a5bd9b19f58a6d49

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
glenhunter: > These two examples should be one with two code examples

https://github.com/sonatype/sonatype-react-shared-components/pull/92/commits/2d6a168e17745955a4ec3f27a5bd9b19f58a6d49

gallery/src/styles/NxFontSize/NxFontSizePage.tsx
glenhunter: > I would have expected an example that actually renders something so the smaller font size can be seen. You may need to adjust the example to use a `<div>` instead of `<body>`. If so, add a note to the example explaining that you'd usually use it on `<body>` but for the example it has to be on `<div>`

The mixin only works when applied to `<body className="nx-body">` by design since that is where the default font-size is set. I do not think that we should modify the mixin just for the sake of showing an example. Use of this mixin represents a pretty big change to the layout, if the user can't figure out how to use the mixin then I would say that they probably shouldn't be using it at all.

rpokorny: > The mixin only works when applied to <body className="nx-body">

Well, it only works when applied to `.nx-body`, at any rate.  Doesn't truly have to be a `<body>` tag.  If you don't actually want it to render anything then the `htmlExample` prop should be removed.  Right now I think it is trying to render that extra `<body>` element, but seems to be failing to actually add it to the DOM probably because it really doesn't want to have two `<body>` tags on the page.

glenhunter: > > The mixin only works when applied to
> 
> Well, it only works when applied to `.nx-body`, at any rate. Doesn't truly have to be a `<body>` tag. If you don't actually want this example to render anything then the `htmlExample` prop should be removed. Right now I think it is trying to render that extra `<body>` element, but seems to be failing to actually add it to the DOM probably because it really doesn't want to have two `<body>` tags on the page.

What do you think of this?
https://github.com/sonatype/sonatype-react-shared-components/pull/92/commits/ec790228f3511d08f8ea4055c3ce28a991e2c3b5

gallery/src/styles/NxFontSize/NxFontSizeHtmlExample.html
```diff
@@ -0,0 +1,15 @@
...
+    Copyright (c) 2019-present Sonatype, Inc.
+    This program and the accompanying materials are made available under
+    the terms of the Eclipse Public License 2.0 which accompanies this
+    distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+
+-->
+<div class="nx-body project-body--custom-font-size">
+  <p class="nx-p">Normally this text would be 16px, but it should be 12px.</p>
+  <p class="nx-p">
+    Note: for illustration purposes we have used the mixin to style a &lt;div&gt;, but it should be
```
rpokorny, line 12:
```suggestion
    Note: for illustration purposes we have used the mixin to style a &lt;div&gt;, but in actual usage it should be
```

PR #93, opened 2020-07-20
Expand input types RSC-234

glenhunter:
Draft PR to illustrate Number, Date, Email, and Range input types

### DIFF
--- a/gallery/src/components/NxTextInput/NxTextInputPage.tsx
+++ b/gallery/src/components/NxTextInput/NxTextInputPage.tsx
@@ -14,6 +14,7 @@ import NxTextInputPasswordExample from './NxTextInputPasswordExample';
 import NxTextInputTextAreaExample from './NxTextInputTextAreaExample';
 import NxTextInputTextAreaValidationExample from './NxTextInputTextAreaValidationExample';
 import NxTextInputDisabledExample from './NxTextInputDisabledExample';
+import NxTextInputTypesExample from './NxTextInputTypesExample';
 
 const simpleSourceCode = require('!!raw-loader!./NxTextInputSimpleExample').default;
 const validationSourceCode = require('!!raw-loader!./NxTextInputValidationExample').default;
@@ -21,6 +22,7 @@ const passwordSourceCode = require('!!raw-loader!./NxTextInputPasswordExample').
 const textAreaSourceCode = require('!!raw-loader!./NxTextInputTextAreaExample').default;
 const textAreaValidationSourceCode = require('!!raw-loader!./NxTextInputTextAreaValidationExample').default;
 const disabledSourceCode = require('!!raw-loader!./NxTextInputDisabledExample').default;
+const typesSourceCode = require('!!raw-loader!./NxTextInputTypesExample').default;
 
 const NxTextInputPage = () =>
   <>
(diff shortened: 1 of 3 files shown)


rpokorny:
Note that there are a significant number of additional input types beyond what is demonstrated here, including datetime, time, a color picker, etc.

glenhunter:
> Note that there are a significant number of additional input types beyond what is demonstrated here, including datetime, time, a color picker, etc.

Datetime is deprecated. There is also a URL and a Search input type, but I didn't think they'd be different enough to bother looking at right now. the other time based inputs seemed like they'd probably be too specific for our use cases.

rpokorny:
We need to figure out how (or if) we want to tie into the browser-native validation capabilities.  For some input types, such as `email`, this validation is the only purpose of the type.  The thing is, the native HTML 5 input validation capabilities represent a different architectural philosophy about where validation should be done - they do it in the UI layer, in the native elements themselves.  Our stateless components on the other hand leave the validation to be done outside the UI layer in order to be compatible with architectures like redux where most/all business logic is done outside the UI layer.  So these two approaches are never going to mesh well.  I'm probably inclined to say we should ignore or maybe even disallow browser-native validation in favor or our JS-oriented architecture.  Some particular nuances to figuring this out:

* Some attributes affect both native validation and other things.  For instance, the `min`, `max`, and `step` attributes on a number input affect its native validation but also the behavior of the up and down arrows.
* Email validation in particular is very challenging, and isn't something that most people recommend implementing yourself.  So one way or another I'd _like_ to rely on the browser's built-in email validation, but the only way to access it is in the UI layer via an email input.  I wish it was also exposed as a JS function that could be used in our own validator setup.

rpokorny:
> Datetime is deprecated

In favor of `datetime-local`

glenhunter:
> We need to figure out how (or if) we want to tie into the browser-native validation capabilities. For some input types, such as `email`, this validation is the only purpose of the type. The thing is, the native HTML 5 input validation capabilities represent a different architectural philosophy about where validation should be done - they do it in the UI layer, in the native elements themselves. Our stateless components on the other hand leave the validation to be done outside the UI layer in order to be compatible with architectures like redux where most/all business logic is done outside the UI layer. So these two approaches are never going to mesh well. I'm probably inclined to say we should ignore or maybe even disallow browser-native validation in favor or our JS-oriented architecture. Some particular nuances to figuring this out:
> 
> * Some attributes affect both native validation and other things.  For instance, the `min`, `max`, and `step` attributes on a number input affect its native validation but also the behavior of the up and down arrows.
> * Email validation in particular is very challenging, and isn't something that most people recommend implementing yourself.  So one way or another I'd _like_ to rely on the browser's built-in email validation, but the only way to access it is in the UI layer via an email input.  I wish it was also exposed as a JS function that could be used in our own validator setup.

Part of the purpose of this PR (IMO) was to stimulate discussion on which (if any) of these input types we should include. From that we can make separate tickets for each type and add them in as needed (or as we get to them).

I know that @Blacktiger was saying that he'd like to see the number type added.

Blacktiger:
Looks like the future way of handling mobile keyboards is to use inputmode instead (https://caniuse.com/#search=inputmode).

As @rpokorny pointed out, we don't have control over features like input sanitization and validation if we use these input types, but maybe we should leave that up to the user of this component to worry about. For example, the numeric type strips leading 0s which would be a problem for certain types of input (credit cards for example) but that doesn't apply to _all_ numeric fields.

The other alternative would be to create specialized components to handle cases like date selection, or the increment/decrement numeric fields.

rpokorny:
> The other alternative would be to create specialized components to handle cases like date selection, or the increment/decrement numeric fields.

FWIW we already plan to create JS components for date selection since IE11 doesn't support native date inputs

glenhunter:
Per discussion within Team Plaid, we're closing this PR. When we have need to add new input types we will add them on a case by case basis with individual stories.

PR #95, opened 2020-07-23
Iday testing library

Blacktiger:
This PR shows some tests based on the `@testing-library/react`. We've been using this library successfully in repository manager so I wanted to see how tests would look in the RSC. The idea behind this testing library is that you write tests to interact with your components in a similar fashion to how the user interacts with the components (think black-box testing). In particular, you are encouraged to use the `getByRole` query to select elements by their ARIA accessibility role (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques). This should result in writing components in a more accessible manner because you are using similar tools to identify the components as a screen reader would be.

In writing the tests for the NxAlert component I ended up adding the `alert` role and making the `icon` prop optional so that I could pass in an `NxFontAwesome` icon with an `aria-label` to identify the type of alert by the label for the icon. This should make that component more accessible.

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -33,6 +33,8 @@
   "author": "Sonatype, Inc.",
   "license": "EPL-2.0",
   "devDependencies": {
+    "@testing-library/jest-dom": "^5.11.1",
+    "@testing-library/react": "^10.4.7",
     "@types/enzyme": "=3.9.1",
     "@types/enzyme-adapter-react-16": "=1.0.5",
     "@types/jest": "=24.0.11",
--- a/lib/src/components/NxAlert/NxAlert.tsx
+++ b/lib/src/components/NxAlert/NxAlert.tsx
@@ -27,8 +27,8 @@ const NxAlert = forwardRef<HTMLDivElement, NxAlertProps>(
           classes = classnames('nx-alert', className);
 
       return (
-        <div { ...otherProps } ref={ref} className={classes}>
-          <NxFontAwesomeIcon icon={icon}/>
+        <div role="alert" { ...otherProps } ref={ref} className={classes}>
+          { icon && <NxFontAwesomeIcon icon={icon}/> }
           { ensureElement(children) }
         </div>
       );
@@ -43,10 +43,15 @@ export default NxAlert;
  * @param children VDOM nodes to be included after the icon.
  */
 export const NxErrorAlert = forwardRef<HTMLDivElement, Props>(
-    function NxErrorAlert(props, ref) {
+    function NxErrorAlert({children, ...props}, ref) {
       const classes = classnames('nx-alert--error', props.className);
 
-      return <NxAlert { ...props } ref={ref} className={classes} icon={faExclamationCircle} />;
+      return (
+        <NxAlert {...props} className={classes} ref={ref}>
+          <NxFontAwesomeIcon icon={faExclamationCircle} aria-label="Error"/>
+          {children}
+        </NxAlert>
+      );
     }
 );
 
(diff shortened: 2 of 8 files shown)


rpokorny:
This doesn't do shallow rendering I take it?

Blacktiger:
> This doesn't do shallow rendering I take it?

Correct, it should be a bit slower as a result but I didn't do any performance analysis. The reason they didn't implement shallow rendering is explained here: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering

rpokorny:
> Correct, it should be a bit slower as a result but I didn't do any performance analysis. The reason they didn't implement shallow rendering is explained here: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering

We use shallow rendering because it is the way to unit test react components.  If I have a simple component that wraps a complex component (quite common given our stateful/stateless pattern), I don't want to have to re-test all the behavior of the wrapped complex component when I'm testing the simple wrapper.  To me, this is the main point of unit testing in general.  I only want to test (at that point) that the wrapper is calling the wrappee correctly.

I disagree with that article on many points:
* When he says "The argument for mocking third party components I often hear is Testing composed components introduces new dependencies that might trigger an error while the unit itself might still work as intended.." I honestly don't know what he means.  That has never been the point of unit testing or mocking as far as I've ever known.
* His main contention against using shallow rendering for unit testing has to do with inner component classes that are local to the file - a pattern that I generally avoid (due to the way it complicates testing) - so not a big issue to me.
* The article says "The more your tests resemble the way your software is used, the more confidence they can give you. " - debatable, but even if true, keep in mind that RSC is a library.  In some senses, it is "used" at the programmatic way as much as it is in a user interface way.
* Part of what unit testing is about is testing the interface with other units.  In my mind one of those units is the styles that accompany a component, and the interface with that unit is the CSS class names.  In this PR you avoid testing the presence of CSS classes, which I think is important.

Blacktiger:
> > Correct, it should be a bit slower as a result but I didn't do any performance analysis. The reason they didn't implement shallow rendering is explained here: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering
> 
> We use shallow rendering because it is the way to unit test react components. If I have a simple component that wraps a complex component (quite common given our stateful/stateless pattern), I don't want to have to re-test all the behavior of the wrapped complex component when I'm testing the simple wrapper. To me, this is the main point of unit testing in general. I only want to test (at that point) that the wrapper is calling the wrappee correctly.
> 
> I disagree with that article on many points:
> 
> * When he says "The argument for mocking third party components I often hear is Testing composed components introduces new dependencies that might trigger an error while the unit itself might still work as intended.." I honestly don't know what he means.  That has never been the point of unit testing or mocking as far as I've ever known.
> * His main contention against using shallow rendering for unit testing has to do with inner component classes that are local to the file - a pattern that I generally avoid (due to the way it complicates testing) - so not a big issue to me.
> * The article says "The more your tests resemble the way your software is used, the more confidence they can give you. " - debatable, but even if true, keep in mind that RSC is a library.  In some senses, it is "used" at the programmatic way as much as it is in a user interface way.
> * Part of what unit testing is about is testing the interface with other units.  In my mind one of those units is the styles that accompany a component, and the interface with that unit is the CSS class names.  In this PR you avoid testing the presence of CSS classes, which I think is important.
> 
> Note that we do of course also want to have integration tests, which in the case of RSC will be in the form of applitools tests verifying that they appear correctly and respond to user interaction correctly. It's taken longer than we'd like to set that up, but we do understand that unit tests alone are not enough.

Note also that the testing-library folks suggest that rather than using shallow rendering, you can just mock any React components you don't want involved in the test. I didn't try that in any of these tests, so I'm unsure how well that works. So far, I think I was able to duplicate the functionality of all the existing unit tests that I tried converting. Ultimately if we need to dig deeper there is the option of using the "container" to query children of the react component as well.

Mostly I just wanted to see if this library or enzyme works better for the RSC. I think generally it works very well for the application-level testing but I wasn't sure if it was going to work well for the component library. We do use it for both in repository manager (since we have a repo-specific component library, though I think many of the components there will end up replaced by the RSC) and so far it's working well there. My main hope for the RSC was that it would help encourage us to make sure the components are more accessible, though that may be better accomplished in other ways.

PR #97, opened 2020-07-28
NxButton visual tests - RSC-186

rpokorny:


Jenkinsfile
```diff
@@ -16,7 +16,13 @@ dockerizedBuildPipeline(
   prepare: {
     githubStatusUpdate('pending')
 
-    sh "docker run --name selenium-chrome -d -p 4444:4444 -v /dev/shm:/dev/shm ${seleniumDockerImage}:${seleniumDockerVersion}"
+    sh """
+      docker run --name selenium-chrome -d \
+        -p 4444:4444 \
+        -v /dev/shm:/dev/shm \
+        -e NODE_MAX_INSTANCES=5 -e NODE_MAX_SESSION=5 -e START_XVFB=false \
```
rpokorny, line 23: It turns out the tests weren't correctly being run in parallel and the mouse moves were conflicting between tests.  The server needs to be configured for parallel runs and the mouse moves don't happen in true headless mode (which doesn't use XVFB)

gallery/wdio.conf.js
```diff
@@ -56,14 +57,17 @@ exports.config = {
...
     //
     // If you have trouble getting all important capabilities together, check out the
     // Sauce Labs platform configurator - a great tool to configure your capabilities:
     // https://docs.saucelabs.com/reference/platforms-configurator
     //
     capabilities: [{
-        browserName: 'chrome',
+      browserName: 'chrome',
+      'goog:chromeOptions': {
+        args: [ 'headless', 'font-render-hinting=none']
```
rpokorny, line 69: Run in headless mode.  Without the `font-render-hinting` setting, the fonts come out substantially differently.

Blacktiger: If we run in headless mode will the mouse events in your tests above not work correctly?

rpokorny: > If we run in headless mode will the mouse events in your tests above not work correctly?

They do seem to work correctly.  Headless mode seems to be able to simulate a mouse, and more importantly, each headless browser instance seems to simulate the mouse independently whereas with XVFB I think there's a single mouse globally which is why the XVFB approach couldn't successfully run mouse tests in parallel.

PR #113, opened 2020-08-10
NxRadio fixes RSC-120

glenhunter:
Solves the two problems found in testing RSC-120: https://issues.sonatype.org/browse/RSC-120

### DIFF
--- a/lib/src/scss-shared/_nx-input-helpers.scss
+++ b/lib/src/scss-shared/_nx-input-helpers.scss
@@ -4,6 +4,9 @@
 
 @mixin focusable-toggle-input($focus-border-selector) {
   .nx-radio__input, .nx-checkbox__input {
+    position: absolute;
+    z-index: -1;
+
     &:focus {
       // less-nice IE11 focus style because IE11 doesn't support :focus-within
       margin-right: $nx-spacing-xxs;
--- a/lib/src/scss-shared/_nx-input-helpers.scss
+++ b/lib/src/scss-shared/_nx-input-helpers.scss
@@ -35,6 +35,13 @@
           stroke: #02A6D9;
         }
       }
+      &:hover {
+        + .nx-radio__circle {
+          #{$focus-border-selector} {
+            stroke: #045dd1;
+          }
+        }
+      }
     }
 
     .nx-radio__input:focus, .nx-checkbox__input:focus {


lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -32,6 +35,13 @@
           stroke: #02A6D9;
         }
       }
+      &:hover {
+        + .nx-radio__circle {
+          #{$focus-border-selector} {
+            stroke: #045dd1;
```
glenhunter, line 41: I couldn't see why we'd want to use the black border when hovering on a selected radio button and @dturner32 wasn't here to ask so I made it the selected blue color. It's a trivial change if it's wrong.

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -4,6 +4,9 @@
 
 @mixin focusable-toggle-input($focus-border-selector) {
   .nx-radio__input, .nx-checkbox__input {
+    position: absolute;
+    z-index: -1;
+
```
glenhunter, line 9: Both of these attributes were removed (by me) because they didn't seem to do anything, it turns out they are necessary because FF ignores the `width: 0` attribute.

lib/src/scss-shared/_nx-input-helpers.scss
rpokorny: > I couldn't see why we'd want to use the black border when hovering on a selected radio button and @dturner32 wasn't here to ask so I made it the selected blue color. It's a trivial change if it's wrong.

That does mean that hovering over a non-focused, checked radio has no visual effect.  Which is probably ok as clicking on it would also have no effect since it's already selected.

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -32,6 +35,13 @@
           stroke: #02A6D9;
         }
       }
+      &:hover {
+        + .nx-radio__circle {
```
rpokorny, line 39: This shouldn't be in the mixin, it is specific to the radio.

lib/src/scss-shared/_nx-input-helpers.scss
glenhunter: > > I couldn't see why we'd want to use the black border when hovering on a selected radio button and @dturner32 wasn't here to ask so I made it the selected blue color. It's a trivial change if it's wrong.
> 
> That does mean that hovering over a non-focused, checked radio has no visual effect. Which is probably ok as clicking on it would also have no effect since it's already selected.

That was my thinking yes.

lib/src/scss-shared/_nx-input-helpers.scss
glenhunter: > This shouldn't be in the mixin, it is specific to the radio.

Everything within this portion of the mixin (under `&:focus-within {`) is specific to the radio.

lib/src/scss-shared/_nx-input-helpers.scss
dturner32: I am ok with this adjustment

lib/src/scss-shared/_nx-input-helpers.scss
rpokorny: > Everything within this portion of the mixin (under &:focus-within {) is specific to the radio.

That isn't exactly true, though I can see that the selector immediately above this one is specific to the radio. I guess that makes me also question why that selector is in the mixin.

PR #117, opened 2020-08-11
NxCheckbox redesign RSC-119

glenhunter:
https://issues.sonatype.org/browse/RSC-119


gallery/src/components/NxCheckbox/NxCheckboxExample.tsx
```diff
@@ -11,38 +11,36 @@ import { NxCheckbox } from '@sonatype/react-shared-components';
...
-  const [isSubscribed, setIsSubscribed] = useState(false),
-      onChange = () => setIsSubscribed(!isSubscribed);
+  const [isRed, setIsRed] = useState(false),
+      [isBlue, setIsBlue] = useState(false),
+      [isGreen, setIsGreen] = useState(false),
+      [isDisabled, setIsDisabled] = useState(false),
+      onChange1 = () => setIsRed(!isRed),
+      onChange2 = () => setIsBlue(!isBlue),
+      onChange3 = () => setIsGreen(!isGreen),
+      onChange4 = () => setIsDisabled(!isDisabled);
```
rpokorny, line 21:
```suggestion
      toggleRed = () => setIsRed(!isRed),
      toggleBlue = () => setIsBlue(!isBlue),
      toggleGreen = () => setIsGreen(!isGreen),
      toggleDisabled = () => setIsDisabled(!isDisabled);
```

rpokorny:
In zeplin, measuring from checkbox to checkbox, the rows are 18px apart.  Your implementation has them 24px apart

rpokorny:
Text should be #474747, not #2c2c2c

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -14,38 +14,76 @@
...
-  vertical-align: -2px;
-  width: 14px;
+  vertical-align: 1px;
+  width: 16px;
 
   .fa-check {
-    margin-bottom: 1.5px;
+    color: #fff;
+    font-size: $nx-font-size-s;
+    margin-bottom: 6px;
```
rpokorny, line 47: Since you're doing other vertical tweaks using `vertical-align`, you could stick to that here too.  This looks right to me:
```suggestion
    vertical-align: 0.28em;
```
Or `vertical-align: 4px;`, if you prefer

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -14,38 +14,76 @@
...
+  margin: 0;
+  width: 0;
+
+  &:checked {
+    + .nx-checkbox__box {
+        background-color: #045dd1;
+        border-color: #045dd1;
+
+      &:hover {
+        border-color: #045dd1;
```
rpokorny, line 77: I think this is redundant

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -34,6 +37,12 @@
       }
     }
 
+    .nx-checkbox__input:checked {
+      + .nx-checkbox__box {
```
rpokorny, line 41: This really goes against the point of this being a mixin

lib/src/scss-shared/_nx-input-helpers.scss
```diff
@@ -16,13 +19,13 @@
   }
 
   &:focus-within {
-    .nx-radio__circle {
+    .nx-radio__circle, .nx-checkbox__box {
```
rpokorny, line 22: Should these selectors be parameters to the mixin?

lib/src/components/NxCheckbox/NxCheckbox.scss
glenhunter: > Since you're doing other vertical tweaks using `vertical-align`, you could stick to that here too. This looks right to me:
> 
> Or `vertical-align: 4px;`, if you prefer

I changed it to to px just to keep the units consistent. https://github.com/sonatype/sonatype-react-shared-components/pull/117/commits/b3678a23321a00d8c3e830869c24ef2e17ad71a9

lib/src/scss-shared/_nx-input-helpers.scss
glenhunter: > This really goes against the point of this being a mixin

Similar to the NxRadio button mixin issues the interactions between `:focus`, `:hover`, and `:checked` states are now so complicated that this needs to be here because of specificity.

I like all of the `:focus` styles being collected in one place since (for the most part) they are shared so it is DRY-er. But obviously we have the NxRadio and NxCheckbox specific styles to deal with.

We could get rid of the mixin altogether, and just have some duplicated styles between NXRadio and NxCheckbox.

We could recreate the whole cascade in each component's SCSS to deal with the inheritance issues (I dislike that idea) and it negates the benefits of the mixin.

There is a little bit of refining that could be done after both NxRadio and NxCheckbox are merged to the feature branch (grouping both `:checked` styles into one group for example, but this won't eliminate the specific styles).

I'm open to suggestions.

lib/src/base-styles/_nx-toggle.scss
```diff
@@ -0,0 +1,129 @@
...
+      border-color: #2c2c2c;
+      filter: none;
+      stroke: #2c2c2c;
+    }
+  }
+
+  &:focus-within {
+    .nx-toggle__control {
+      border-color: #02a6d9;
+      stroke: #02A6D9;
```
rpokorny, line 33: I think we use lowercase colors everywhere

glenhunter:
> In zeplin, measuring from checkbox to checkbox, the rows are 18px apart. Your implementation has them 24px apart

I think they're 16px which is what I have changed them too here: https://github.com/sonatype/sonatype-react-shared-components/pull/117/commits/528e037534110402a9954bc0789752e3850ff196

lib/src/base-styles/_nx-toggle.scss
glenhunter: > I think we use lowercase colors everywhere

https://github.com/sonatype/sonatype-react-shared-components/pull/117/commits/502a52f1a486eadfa699634663a60861f10b0d28

glenhunter:
> Text should be #474747, not #2c2c2c

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/117/commits/c26b07e0ca36d115f478f394ee4f9f6727048ef6

rpokorny:
Looks like the cursor style on the disabled checkbox is now wrong

glenhunter:
> Looks like the cursor style on the disabled checkbox is now wrong

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/117/commits/37293c25c4b187d708a04d4ad18f2fa5c59988e2

Also cleaned up some duplicates

PR #121, opened 2020-08-14
Scrolling/Sticky Header Tables - RSC-249

rpokorny:
https://issues.sonatype.org/browse/RSC-249


glenhunter:
Screen Shot 2020-08-17 at 10 49 29 AM

Interesting glitch when the table is scrolled, and then the page is scrolled

glenhunter:
I think it looks good, I tried custom sizing some columns and it all held together

rpokorny:
Note that this PR adds the vertical-container mixin to nx-scrollable, which affected the existing applitools snapshot by removing the top margin from the first paragraph.  I'm assuming that's fine.

gallery/src/styles/NxScrollable/NxScrollableTableExample.html
```diff
@@ -0,0 +1,79 @@
+<!--
+
+    Copyright (c) 2019-present Sonatype, Inc.
+    This program and the accompanying materials are made available under
+    the terms of the Eclipse Public License 2.0 which accompanies this
+    distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+
+-->
+<div class="nx-scrollable" style="height: 150px">
+  <table class="nx-table nx-scrollable__table">
```
glenhunter, line 10: I find it weird that the class suggests that the table is an element of the scrolling container rather than a modification of a table, which makes more sense to me. I think this should be `nx-table--scrollable`.

gallery/src/styles/NxTable/NxTableStylePage.tsx
```diff
@@ -105,6 +101,14 @@ const NxTableStylePage = () =>
           </tr>
         </tbody>
       </table>
+
+      <p className="nx-p">
+        When a table which scrolls in of itself is desired, wrap the table in
+        a <code className="nx-code">.nx-scrollable</code> wrapper and give it a class
```
glenhunter, line 107:
```suggestion
        an <code className="nx-code">.nx-scrollable</code> wrapper and give it a class
```

lib/src/base-styles/_nx-globals.scss
```diff
@@ -49,9 +49,22 @@ $nx-scrollable-default-height: 400px;
 }
 
 .nx-scrollable {
+  @include container-vertical;
   @include vertically-scrollable($nx-scrollable-default-height);
 }
 
+.nx-scrollable__table {
+  // create a stacking context so the headers don't appear over other content outside the table
```
glenhunter, line 57: Nice

lib/src/base-styles/_nx-tables.scss
```diff
@@ -9,7 +9,7 @@
   #Tables
 */
 .nx-table {
-  border-collapse: collapse;
+  border-spacing: 0;
```
glenhunter, line 12: I'm curious about the why of this change.

gallery/src/styles/NxScrollable/NxScrollableTableExample.html
rpokorny: > I find it weird that the class suggests that the table is an element of the scrolling container rather than a modification of a table, which makes more sense to me. I think this should be nx-table--scrollable.

I see it the opposite way which is what led me to name the classes the way that I did.  The thing is, it's not really the table that scrolls, its the `nx-scrollable`.

lib/src/base-styles/_nx-tables.scss
rpokorny: > I'm curious about the why of this change.

IIRC with collapsed borders, the border separating the headers from the body rows would scroll instead of remaining stationary.  So I had to make changes to make it unambiguous that that border is part of the header and not the body.

gallery/src/styles/NxScrollable/NxScrollableTableExample.html
glenhunter: > > I find it weird that the class suggests that the table is an element of the scrolling container rather than a modification of a table, which makes more sense to me. I think this should be nx-table--scrollable.
> 
> I see it the opposite way which is what led me to name the classes the way that I did. The thing is, it's not really the table that scrolls, its the `nx-scrollable`.

The more that I think about it though the more I think that not only is the class incorrect, but that the example is in the wrong place. Regardless of the _why_ of the scrolling the difference to a consumer of this is in the table. One type of table scrolls, one doesn't. 

When I pulled the branch I couldn't find the example and had to go back to the diff to figure out where it was, it never occurred to me that it would be in with `NxScrollable`. I think that the class name needs to be changed and that the example should be moved to its own page.

rpokorny: > One type of table scrolls, one doesn't.

That's the whole disagreement.  Neither table scrolls (itself).  The nx-scrollable scrolls, and when you have an nx-scrollable that solely contains a table, that table is supposed to behave a certain way.

> the example is in the wrong place

I had it in nx-table originally, but very deliberately moved it to nx-scrollable as a result of my perspective described above.

> When I pulled the branch I couldn't find the example and had to go back to the diff to figure out where it was

I do mention and link to it from the nx-table page.

glenhunter: > That's the whole disagreement. Neither table scrolls (itself). The nx-scrollable scrolls, and when you have an nx-scrollable that solely contains a table, that table is supposed to behave a certain way.

Right, I get it, but I don't think that we should be basing the name or example on an implementation detail. If in 6 months we figure out some way to do this using some other technique will we rename and move it? The point is that this is a table whose contents scroll. Who cares why they scroll?

rpokorny: > Right, I get it, but I don't think that we should be basing the name or example on an implementation detail. If in 6 months we figure out some way to do this using some other technique will we rename and move it? The point is that this is a table whose contents scroll. Who cares why they scroll?

Thoughts from @sonatype/shared-components-group ?

marushkevych: > I find it weird that the class suggests that the table is an element of the scrolling container rather than a modification of a table, which makes more sense to me. I think this should be `nx-table--scrollable`.

I agree with this. To have it as **element** of `.nx-scrollable` is not intuitive. I find it most intuitive as **modifier** of `.nx-table`. We are modifying  the table to be used in `.nx-scrollable`

rpokorny: @glenhunter and @marushkevych where would you guys say the example for it should live?  It seems weird to put it on the `.nx-table` page when the top element of the example is not an `.nx-table`.  In fact it was when I was originally trying to answer this question myself that I came up with the perspective of it being a BEM element under `.nx-scrollable`.

rpokorny: Refactored to `nx-table--scrollable` in 1dfcb28a.  I decided to just put it on the nx-table page as it also seemed weird/bad to have a modifier of `nx-table` that isn't documented/demonstrated on that page.  Having an example there that doesn't have `.nx-table` as its top element seems like the least bad placement.

glenhunter (approved):

PR #122, opened 2020-08-14
refactor form layout example page  RSC-162

glenhunter:
https://issues.sonatype.org/browse/RSC-162

Refactor the examples on the Form Layout Examples page to match design direction from Derrek.

I chose to only make the NxCheckbox and NxRadio button examples functional in the top example, I didn't think it was necessary for them to work in both (or maybe at all).


lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
-.nx-form--modal-form {
-  padding-top: 0;
+  border-top: 1px solid #e9e9e9;
+  padding: $nx-spacing-xl $nx-spacing-l;
+  text-align: right;
 }
 
 .nx-form-group {
-  margin-bottom: 20px;
+  margin-bottom: 6px;
```
glenhunter, line 26: This kinda sucks, but the spacing between groups must take into account the validation message space below the text inputs.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
 }
 
 .nx-form-row {
   @include container-horizontal;
-  margin-bottom: 24px;
+  margin-bottom: 6px;
```
glenhunter, line 31: See note above about this 6px value.

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
 
 export default function NxFormLayoutExample() {
+  function validator(val: string) {
+    return val.length ? null : 'Must be non-empty';
+  }
+
   return (
     <form className="nx-form">
       <div className="nx-form-row">
         <div className="nx-form-group">
```
rpokorny, line 22: Looking through the DOM makes me ask, what exactly is the point of `.nx-form-group`?  Does it ever contain anything other than exactly one `.nx-label`?

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
-  padding-top: 8px;
-}
-
 // labels
 
 .nx-label, .nx-legend {
   @include semi-bold();
 
   color: $nx-text-color;
-  display: block;
```
rpokorny, line 64: This shouldn't have been removed; `.nx-label` contains block elements.

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
-            </span>
-            <input type="text" className="nx-text-input"/>
+            <span className="nx-label__text">Label</span>
+            <NxStatefulTextInput/>
           </label>
         </div>
         <div className="nx-form-group">
-          <label className="nx-label">
-            <span className="nx-label__text">Second column</span>
-            <span className="nx-sub-label">This is a sub-label.</span>
```
rpokorny, line 32: What if one of the labels in a form row has a sub-label, but the other doesn't?  Won't they end up being different heights? How should they align in that case?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
+        <NxCheckbox isChecked={false}>Checkbox 3</NxCheckbox>
+      </fieldset>
+      <fieldset className="nx-fieldset">
+        <legend className="nx-legend"><span className="nx-legend__text">Radio buttons</span></legend>
+        <NxRadio name="demo1" value="demo1" isChecked={true}>Radio Button 1</NxRadio>
+        <NxRadio name="demo2" value="demo2" isChecked={false}>Radio Button 2</NxRadio>
+        <NxRadio name="demo3" value="demo3" isChecked={false}>Radio Button 3</NxRadio>
+      </fieldset>
+      <footer className="nx-form-footer">
+        <NxButton>Cancel</NxButton>
```
rpokorny, line 48: This needs `type="button"` or `type="reset"`

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
-        <fieldset className="nx-fieldset">
-          <legend className="nx-label">Radio buttons</legend>
-          <NxRadio name="demo1" value="demo1" isChecked={true}>Radio Button 1</NxRadio>
-          <NxRadio name="demo2" value="demo2" isChecked={false}>Radio Button 2</NxRadio>
-          <NxRadio name="demo3" value="demo3" isChecked={true}>Radio Button 3</NxRadio>
-        </fieldset>
       </div>
+      <fieldset className="nx-fieldset">
+        <legend className="nx-legend nx-legend--optional"><span className="nx-legend__text">Checkboxes</span></legend>
+        <NxCheckbox isChecked={false}>Checkbox 1</NxCheckbox>
```
rpokorny, line 37: The fact that these are hardcoded should be mentioned in the description for this example.

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -4,61 +4,112 @@
...
         </label>
       </div>
       <fieldset className="nx-fieldset">
-        <legend className="nx-label">Checkboxes</legend>
-        <NxCheckbox isChecked={false}>Checkbox 1</NxCheckbox>
-        <NxCheckbox isChecked={true}>Checkbox 2</NxCheckbox>
-        <NxCheckbox isChecked={false}>Checkbox 3</NxCheckbox>
+        <legend className="nx-legend">
+          <span className="nx-legend__text">
+            Checkboxes ({isRed && ' Red'}{isBlue && ' Blue'}{isGreen && ' Green'} )
```
rpokorny, line 54:
```suggestion
            Checkboxes ({isRed && ' Red'}{isBlue && ' Blue'}{isGreen && ' Green'})
```

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -4,61 +4,112 @@
...
+          </span>
+        </legend>
+        <NxCheckbox checkboxId="subscribe-check" onChange={toggleRed} isChecked={isRed}>
+          Red
+        </NxCheckbox>
+        <NxCheckbox checkboxId="no-label-check" onChange={toggleBlue} isChecked={isBlue}>Blue</NxCheckbox>
```
rpokorny, line 60: This id doesn't make sense.  Overall I'd suggest simplifying the VDOM of this part of the example.  We already demonstrate all the edge-case capabilities on the `NxCheckbox` page itself, no need to have anything more than a straightforward set of NxCheckboxes (and NxRadios) on this page.

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
```diff
@@ -107,15 +106,19 @@ const NxFormLayoutPage = () =>
       </table>
     </GalleryDescriptionTile>
 
-    <GalleryTile title="NX Form Layout Example">
-      <NxFormLayoutExample />
-      <CodeExample content={NxFormLayoutCode}/>
-    </GalleryTile>
+    <GalleryExampleTile title="General Example"
+                        id="nx-checkbox-example"
```
rpokorny, line 110:
```suggestion
```

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
```diff
@@ -107,15 +106,19 @@ const NxFormLayoutPage = () =>
...
+                        codeExamples={NxFormLayoutCode}
+                        liveExample={NxFormLayoutExample}>
+      This example shows a stand vertical form layout with validation on some fields.
+    </GalleryExampleTile>
 
-    <GalleryTile title="NX Form Horizontal Layout Example">
-      <NxFormHorizontalLayoutExample />
-      <CodeExample content={NxFormHorizontalLayoutCode}/>
-    </GalleryTile>
+    <GalleryExampleTile title="Checkbox label should not wrap"
```
rpokorny, line 116: Needs correct title

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
```diff
@@ -107,15 +106,19 @@ const NxFormLayoutPage = () =>
...
 
-    <GalleryTile title="NX Form Layout Example">
-      <NxFormLayoutExample />
-      <CodeExample content={NxFormLayoutCode}/>
-    </GalleryTile>
+    <GalleryExampleTile title="General Example"
+                        id="nx-checkbox-example"
+                        codeExamples={NxFormLayoutCode}
+                        liveExample={NxFormLayoutExample}>
+      This example shows a stand vertical form layout with validation on some fields.
```
rpokorny, line 113:
```suggestion
      This example shows a standard vertical form layout with validation on some fields.
```

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
+import { NxStatefulTextInput } from '@sonatype/react-shared-components';
 
 export default function NxFormLayoutExample() {
+  function validator(val: string) {
+    return val.length ? null : 'Must be non-empty';
+  }
+
   return (
     <form className="nx-form">
       <div className="nx-form-row">
```
rpokorny, line 21: It seems a little unfortunate to me that we need a separate element for a strictly layout concern.  Would it not work to instead create `nx-form-group--inline` (or `nx-label--inline` if `nx-form-group` is removed)?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
+import { NxButton } from '@sonatype/react-shared-components';
+import { NxStatefulTextInput } from '@sonatype/react-shared-components';
 
 export default function NxFormLayoutExample() {
+  function validator(val: string) {
+    return val.length ? null : 'Must be non-empty';
+  }
+
   return (
     <form className="nx-form">
```
rpokorny, line 20:
```suggestion

  function onSubmit(evt) {
    evt.preventDefaul();
    alert('Submitted!');
  }

  return (
    <form className="nx-form" onSubmit={onSubmit}>
```

lib/src/base-styles/_nx-forms.scss
rpokorny: > This kinda sucks, but the spacing between groups must take into account the validation message space below the text inputs.

Or should the specs be changed such that this can become a more standard 8px?

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
 
 .nx-form {
   @include container-vertical;
-  border-top: $nx-border-dark;
   margin: 0;
```
rpokorny, line 13: Like most elements, the form should probably have some default spacing from whatever comes before and after it, vertically

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
-}
+.nx-form-footer {
+  @include container-horizontal;
 
-// for a form that is the entirety of a modal
-.nx-form--modal-form {
-  padding-top: 0;
+  border-top: 1px solid #e9e9e9;
+  padding: $nx-spacing-xl $nx-spacing-l;
+  text-align: right;
```
rpokorny, line 22: I thought we decided that the form footer should include an `.nx-btn-bar` and that `.nx-btn-bar` should get this text-align style.

rpokorny: > I thought we decided that the form footer should include an .nx-btn-bar and that .nx-btn-bar should get this text-align style.

Whatever we decide here may need to consider what the form footer is supposed to look like in case of a submit error.  @dturner32 ?

rpokorny, line 22: There's also the question of how the form and tile styles are supposed to commingle in the case of a tile that contains nothing but a form.  Do you use `.nx-form-footer` or `.nx-tile-footer`?  Or both on the same element?  If the latter, which element is the `<form>`? The whole `.nx-tile`?

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
-  margin-bottom: 20px;
+  margin-bottom: 6px;
 }
 
 .nx-form-row {
   @include container-horizontal;
-  margin-bottom: 24px;
+  margin-bottom: 6px;
 
   .nx-form-group, .nx-fieldset {
```
rpokorny, line 33: Isn't `.nx-fieldset` no longer supposed to be used within `.nx-form-row`?

lib/src/base-styles/_nx-forms.scss
dturner32: correct, nx-btn bar should be included in the footer of a form . As for the error message handling, similar to the mock up of the design for the modal view, the error would appear above the buttons. I will create a mock up for that and post to zeplin

dturner32: in the case of a tile containing just a form, it should use the nx-form footer

rpokorny: > As for the error message handling, similar to the mock up of the design for the modal view, the error would appear above the buttons. I will create a mock up for that and post to zeplin

Keep in mind that if both Retry and Submit are visible at the same time, they'd presumably do the same thing.  Or the user might wonder whether or not they do the same thing.  So you may consider hiding one or the other button in that scenario

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -8,49 +8,46 @@ import React from 'react';
...
           </label>
         </div>
         <div className="nx-form-group">
-          <label className="nx-label">
-            <span className="nx-label__text">Second column</span>
-            <span className="nx-sub-label">This is a sub-label.</span>
-            <input type="text" className="nx-text-input"/>
+          <label className="nx-label nx-label--optional">
+            <span className="nx-label__text">Label</span>
+            <NxStatefulTextInput validator={validator}/>
```
rpokorny, line 31: I know these are just examples, but it seems backwards that you've attached the non-empty validator to the optional text input

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
 
-.nx-btn-bar--forms {
-  border-top: $nx-border-dark;
-  padding-top: 8px;
-}
-
 // labels
 
 .nx-label, .nx-legend {
   @include semi-bold();
```
rpokorny, line 43: This is affecting the validation message

rpokorny, line 43: `_nx-typography.scss` contains the following which conflicts with this:
```
.nx-label {
  @include bold();

  cursor: default;
}
```

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
 
   color: $nx-text-color;
```
rpokorny, line 45: Seems redundant

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
-  display: block;
-  font-size: $nx-font-size;
-  margin-bottom: $nx-spacing-xs;
+  padding: 0;
```
rpokorny, line 46: Seems redundant

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
 
   .disabled {
     color: $nx-grey-4;
```
rpokorny, line 49: What does this do

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,74 @@
...
+  margin-bottom: $nx-spacing-xs;
+}
+
+.nx-label--optional, .nx-legend--optional {
+  .nx-label__text::after, .nx-legend__text::after {
     @include regular();
 
     color: $nx-text-color-light;
-    content: " - optional";
+    content: " optional";
```
rpokorny, line 67:
```suggestion
    content: "Optional";
```

lib/src/base-styles/_nx-forms.scss
rpokorny: > Or should the specs be changed such that this can become a more standard 8px?

Ah but then we couldn't use the standard 32px margin on the bottom of the fieldsets.  I think the root of the problem is that the validation message is 18px tall (12px font size * 1.5 line-height).  2px too tall to fit cleanly into our multiples-of-8 spacing scheme.  We could force the line-height to 16px there.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -104,5 +90,9 @@
   border: 0;
   margin: 0 0 $nx-spacing-xl 0;
   padding: 0;
+
+  .nx-label {
```
rpokorny, line 94: I'm not seeing any `.nx-label`s that are descendants of `.nx-fieldset`s

lib/src/base-styles/_nx-typography.scss
```diff
@@ -83,10 +83,9 @@
 }
 
 .nx-label {
-  @include semi-bold();
+  @include bold();
```
rpokorny, line 86: As I mentioned above, this is out of sync with the styles in `_nx-forms.scss`.  I also wonder why any `.nx-label` styles would be in this file at all given the existence of `_nx-forms.scss`

lib/src/base-styles/_nx-forms.scss
rpokorny: > in the case of a tile containing just a form, it should use the nx-form footer

Could you clarify what that means from a design perspective?  I was thinking this was more of an implementation question.

lib/src/base-styles/_nx-forms.scss
glenhunter: > This shouldn't have been removed; `.nx-label` contains block elements.

Put back here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/1c38ba52ad23e9a7a48948ca4119360923e9fc9a

lib/src/base-styles/_nx-forms.scss
glenhunter: > `_nx-typography.scss` contains the following which conflicts with this:
> 
> ```
> .nx-label {
>   @include bold();
> 
>   cursor: default;
> }
> ```

Resolved here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/1c38ba52ad23e9a7a48948ca4119360923e9fc9a

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > What if one of the labels in a form row has a sub-label, but the other doesn't? Won't they end up being different heights? How should they align in that case?

I mocked this up: 
Screen Shot 2020-08-17 at 2 58 07 PM

And @dturner32 gave his thumbs up (https://sonatype.slack.com/archives/CJUF9G7T2/p1597690790017000).

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > This needs `type="button"` or `type="reset"`

Here and all other buttons: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/f6ac79f95d8b4d44ac16dd8a24ee6888b8076a46

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > This id doesn't make sense. Overall I'd suggest simplifying the VDOM of this part of the example. We already demonstrate all the edge-case capabilities on the `NxCheckbox` page itself, no need to have anything more than a straightforward set of NxCheckboxes (and NxRadios) on this page.

Cleaned up and consolidated here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/bf7e0b1dc09a3fba78513c1d932c38a63fe9b941

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > The fact that these are hardcoded should be mentioned in the description for this example.

I replaced these with the cleaned up examples from the regular example: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/c877f23d3c7ba913e2d885f62328594770c6b519

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
glenhunter: > Needs correct title

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/4a870e25a6895b03fddc22f91c5b669ce4992587

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > It seems a little unfortunate to me that we need a separate element for a strictly layout concern. Would it not work to instead create `nx-form-group--inline` (or `nx-label--inline` if `nx-form-group` is removed)?

By enclosing the groups inside a row we control how many and which appear side-by-side. There are ways that we could do that and still drop the row, but that would add complexity elsewhere. I also prefer this because it mimics table layout and therefore should be familiar to most.

lib/src/base-styles/_nx-forms.scss
glenhunter: > Isn't `.nx-fieldset` no longer supposed to be used within `.nx-form-row`?

No, no it is not: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/c23bc1d427120b13235d50e6dea8e5544be14744

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > I know these are just examples, but it seems backwards that you've attached the non-empty validator to the optional text input

I cleaned up all of the examples here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/bf7e0b1dc09a3fba78513c1d932c38a63fe9b941

lib/src/base-styles/_nx-forms.scss
glenhunter: > This is affecting the validation message

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/9c9f846340607e957775377792757d836f512245

lib/src/base-styles/_nx-forms.scss
glenhunter: > I'm not seeing any `.nx-label`s that are descendants of `.nx-fieldset`s

Removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/1c38ba52ad23e9a7a48948ca4119360923e9fc9a

lib/src/base-styles/_nx-typography.scss
glenhunter: > As I mentioned above, this is out of sync with the styles in `_nx-forms.scss`. I also wonder why any `.nx-label` styles would be in this file at all given the existence of `_nx-forms.scss`

I don't think so, I consolidated it into `forms.scss`: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/1c38ba52ad23e9a7a48948ca4119360923e9fc9a

lib/src/base-styles/_nx-forms.scss
glenhunter: > Seems redundant

The browser agent adds 2px of padding to `<legend>`, I just removed it here for both for clarity.

lib/src/base-styles/_nx-forms.scss
glenhunter: > What does this do

The browser agent adds 2px of padding to <legend>, I just removed it here for both for clarity.

lib/src/base-styles/_nx-forms.scss
glenhunter: > Seems redundant

It was: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/eb676176ba43fcc00e26d8fa4ed571c6efc9488e

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > Looking through the DOM makes me ask, what exactly is the point of `.nx-form-group`? Does it ever contain anything other than exactly one `.nx-label`?

There are some instances in IQ where a form control doesn't have a label, and there may be others elsewhere.

lib/src/base-styles/_nx-forms.scss
glenhunter: > > Or should the specs be changed such that this can become a more standard 8px?
> 
> Ah but then we couldn't use the standard 32px margin on the bottom of the fieldsets. I think the root of the problem is that the validation message is 18px tall (12px font size * 1.5 line-height). 2px too tall to fit cleanly into our multiples-of-8 spacing scheme. We could force the line-height to 16px there.

The validation message uses a standard font' and margin:
```
.nx-text-input__invalid-message {
  @include regular();
  color: #cd0028;
  font-size: $nx-font-size-xs;
  margin-top: $nx-spacing-xs;
}
```
So if we customize that we're really just moving the magic number from here to there.

lib/src/base-styles/_nx-forms.scss
glenhunter: > I thought we decided that the form footer should include an `.nx-btn-bar` and that `.nx-btn-bar` should get this text-align style.

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/35beb45c12a1e0bc51400ada141840795012206b

lib/src/base-styles/_nx-forms.scss
glenhunter: > Like most elements, the form should probably have some default spacing from whatever comes before and after it, vertically

Spoke with @dturner32 and he would like 32px of bottom margin: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/c440080d4bcc0749d3df053adf38dbb4e0068cdc

lib/src/base-styles/_nx-forms.scss
glenhunter: > > in the case of a tile containing just a form, it should use the nx-form footer
> 
> Could you clarify what that means from a design perspective? I was thinking this was more of an implementation question.

One thing that makes this confusing is that the form footer and the tile footer are slightly different and I'm not sure why we would want that. I propose that they should have the same spacing, same top border, both should either use `nx-btn-bar` or not.

glenhunter: > There's also the question of how the form and tile styles are supposed to commingle in the case of a tile that contains nothing but a form. Do you use `.nx-form-footer` or `.nx-tile-footer`? Or both on the same element? If the latter, which element is the `<form>`? The whole `.nx-tile`?

We can write `<form className="nx-tile">` and it works aside from the discrepancies between the footers mentioned above. Remember that most tiles have a tile header, which (to me anyway) strongly suggests that we use `nx-tile-footer` rather than `nx-form-footer` for consistency.

i.e. this seems wrong to me:
```
<form className="nx-tile">
  <header className="nx-tileheader">Header</header>
  <div className="nx-tile-content">Form fields etc</div>
  <footer className="nx-form-footer"><div className="nx-btn-bar">Buttons</div></footer>
</form>
```

dturner32: > One thing that makes this confusing is that the form footer and the tile footer are slightly different and I'm not sure why we would want that. I propose that they should have the same spacing, same top border, both should either use `nx-btn-bar` or not.

both form and tile footers should use nx-btn bar going forward. To Glen's point, this should correct the inconsistencies between nx-tile footer and nx-form footer

glenhunter: > > One thing that makes this confusing is that the form footer and the tile footer are slightly different and I'm not sure why we would want that. I propose that they should have the same spacing, same top border, both should either use `nx-btn-bar` or not.
> 
> both form and tile footers should use nx-btn bar going forward. To Glen's point, this should correct the inconsistencies between nx-tile footer and nx-form footer

It does not clear up the inconsistencies though:
```
.nx-form-footer {
  @include container-vertical;
  border-top: 1px solid #e9e9e9;
  padding: $nx-spacing-l;
```

vs

```
.nx-tile-footer {
  border-top: 1px solid #e9e9e9;
  margin-top: $nx-spacing-md;
  padding: $nx-spacing-xl 0;
  text-align: right;
}
```

The primary concern is the different padding. The other discrepancies either make sense or are trivial to excise.

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,87 @@
...
+  toggleRed = () => setIsRed(!isRed),
+  toggleBlue = () => setIsBlue(!isBlue),
+  toggleGreen = () => setIsGreen(!isGreen);
+
+  const [color, setColor] = useState<string | null>(null);
+
+  // function onSubmit(evt) {
+  //   evt.preventDefaul();
+  //   alert('Submitted!');
+  // }
```
marushkevych, line 32: should this be removed?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,91 @@
...
-          <label className="nx-label">
-            <span className="nx-label__text">Horizontal layout</span>
-            <span className="nx-sub-label">
-              <NxFontAwesomeIcon icon={faCalendar}/>
-              <span>Sub-label with icon.</span>
-            </span>
-            <input type="text" className="nx-text-input"/>
-          </label>
+    <form className="nx-tile">
+      <div className="nx-tile-content">
```
marushkevych, line 36: should this be `form.nx-form` with no `div.nx-tile-content`?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,91 @@
...
+          </NxRadio>
+          <NxRadio name="color" value="green" onChange={setColor} isChecked={color === 'green'} radioId="color-green">
+            Green
+          </NxRadio>
+          <NxRadio name="color" value="blue" onChange={setColor} isChecked={color === 'blue'} radioId="color-blue">
+            Blue
+          </NxRadio>
         </fieldset>
       </div>
+      <footer className="nx-tile-footer">
```
marushkevych, line 85: `nx-form-footer`? Noticed becase the margins are not removed from `nx-btn-bar`. Also, should it remove margins from the alert?

lib/src/base-styles/_nx-forms.scss
marushkevych: > > `_nx-typography.scss` contains the following which conflicts with this:
> > ```
> > .nx-label {
> >   @include bold();
> > 
> >   cursor: default;
> > }
> > ```
> 
> Resolved here: [1c38ba5](https://github.com/sonatype/sonatype-react-shared-components/commit/1c38ba52ad23e9a7a48948ca4119360923e9fc9a)

Do we need to `@include bold();` here again?

lib/src/base-styles/_nx-forms.scss
marushkevych: > > What does this do
> 
> The browser agent adds 2px of padding to , I just removed it here for both for clarity.

I think the question was about
```
  .disabled {
    color: $nx-grey-4;
  }
```

lib/src/base-styles/_nx-forms.scss
marushkevych: > Do we need to `@include bold();` here again?

Ah, I see you removed it from _nx-typography.scss

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > should this be `form.nx-form` with no `div.nx-tile-content`?

The idea here was to create an example of a tile that consists of nothing but a form. In this instance we want the tile's look and feel hence the use of the tile header, content and footer. This example highlighted some discrepancies in the footers which need to be resolved.

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > `nx-form-footer`? Noticed becase the margins are not removed from `nx-btn-bar`. Also, should it remove margins from the alert?

I considered this but I think because we're demonstrating a form/tile hybrid and we use only tile styling elsewhere it should be the tile footer. I will be fixing the discrepancies between the various footers.

glenhunter: > > `nx-form-footer`? Noticed becase the margins are not removed from `nx-btn-bar`. Also, should it remove margins from the alert?
> 
> I considered this but I think because we're demonstrating a form/tile hybrid and we use only tile styling elsewhere it should be the tile footer. I will be fixing the discrepancies between the various footers.

I removed the whole form as tile example, I decided that it was too complicated: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/af777406b35c1158e36713e2935d2b2a6b76a5da

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > > should this be `form.nx-form` with no `div.nx-tile-content`?
> 
> The idea here was to create an example of a tile that consists of nothing but a form. In this instance we want the tile's look and feel hence the use of the tile header, content and footer. This example highlighted some discrepancies in the footers which need to be resolved.

I removed the example here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/af777406b35c1158e36713e2935d2b2a6b76a5da

lib/src/base-styles/_nx-forms.scss
glenhunter: > > > What does this do
> > 
> > 
> > The browser agent adds 2px of padding to , I just removed it here for both for clarity.
> 
> I think the question was about
> 
> ```
>   .disabled {
>     color: $nx-grey-4;
>   }
> ```

Ah, I was thrown off by the green highlight above it, yeah that value is not longer required, removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/019733681eb9b5b1f86776237262b0795ea47123

lib/src/base-styles/_nx-forms.scss
glenhunter: I created https://issues.sonatype.org/browse/RSC-253 to handle the changes to the tiles.

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > should this be removed?

So Ross asked that this be added here: https://github.com/sonatype/sonatype-react-shared-components/pull/122#discussion_r470777647

But when I did it wouldn't compile, so if you could hep me figure out why that would be cool.

glenhunter: > > should this be removed?
> 
> So Ross asked that this be added here: [#122 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/122#discussion_r470777647)
> 
> But when I did it wouldn't compile, so if you could hep me figure out why that would be cool.

As discussed removed entirely here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/885306b2ba998cb4d78a4f4b5074bad813adc2d7

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
rpokorny: > There are some instances in IQ where a form control doesn't have a label, and there may be others elsewhere.

Does RSC support that? How? `NxTextInput` always has a label and we no longer document how to create text inputs without the react component.

lib/src/base-styles/_nx-forms.scss
rpokorny: > So if we customize that we're really just moving the magic number from here to there.

A magic number will be necessary somewhere for this.  I think I prefer it being the line-height that is custom rather than the margin.

rpokorny:
Notice the alignment difference:

Zeplin:
image

This PR:
image

I'd say that the general `nx-form` and `nx-form-footer` containers should not have side padding but should instead rely on their parent container for that spacing.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
 
   .nx-text-input, .nx-textarea {
     display: block;
-    margin-top: $nx-spacing-xs;
   }
 }
 
-.nx-label--optional {
-  .nx-label__text::after {
+.nx-label__text, .nx-legend__text {
```
rpokorny, line 66: It seems to me that `.nx-legend` itself is the element most analogous to `.nx-label__text`.  I'm wondering if we should get rid of `.nx-legend__text` and apply these styles to `.nx-legend` instead.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
 }
 
 // labels
 
 .nx-label, .nx-legend {
-  @include semi-bold();
+  @include bold();
 
-  color: $nx-text-color;
+  cursor: default;
```
rpokorny, line 57: In conjunction with my comment below about getting rid of `.nx-legend__text`, I'm thinking this `cursor` property and also the `@include bold` should be moved to the presumptive `.nx-label__text, .nx-legend` block.  Once that is done, I think `.nx-legend` can be removed from the selector on this block.

rpokorny: > Once that is done, I think .nx-legend can be removed from the selector on this block.

Looks like there's also some padding to deal with

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,84 @@
...
+
   return (
     <form className="nx-form">
       <div className="nx-form-row">
         <div className="nx-form-group">
           <label className="nx-label">
-            <span className="nx-label__text">Horizontal layout</span>
-            <span className="nx-sub-label">
-              <NxFontAwesomeIcon icon={faCalendar}/>
-              <span>Sub-label with icon.</span>
```
rpokorny, line 24: Are sub-labels with icons no longer supported?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
rpokorny: > NxTextInput always has a label and we no longer document how to create text inputs without the react component.

Nevermind, I don't know where I got the idea that that was true, it isn't.

rpokorny: >Nevermind, I don't know where I got the idea that that was true, it isn't.

We should probably demonstrate how text fields without labels fit into form layouts

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,84 @@
...
+                 isChecked={color === 'purple'}
+                 radioId="color-purple">
+          Purple
+        </NxRadio>
+        <NxRadio name="color" value="green" onChange={setColor} isChecked={color === 'green'} radioId="color-green">
+          Green
+        </NxRadio>
+        <NxRadio name="color" value="blue" onChange={setColor} isChecked={color === 'blue'} radioId="color-blue">
+          Blue
+        </NxRadio>
```
rpokorny, line 76: `radioId` isn't important to this example
```suggestion
        <NxRadio name="color"
                 value="red"
                 onChange={setColor}
                 isChecked={color === 'red'}>
          Red
        </NxRadio>
        <NxRadio name="color"
                 value="purple"
                 onChange={setColor}
                 isChecked={color === 'purple'}>
          Purple
        </NxRadio>
        <NxRadio name="color" value="green" onChange={setColor} isChecked={color === 'green'}>
          Green
        </NxRadio>
        <NxRadio name="color" value="blue" onChange={setColor} isChecked={color === 'blue'}>
          Blue
        </NxRadio>
```

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,84 @@
...
+      </fieldset>
+      <footer className="nx-form-footer">
+        <NxErrorAlert>This is an example <strong>error</strong> message.</NxErrorAlert>
```
rpokorny, line 79: What scenario are we trying to emulate here, and what pattern are we trying to demonstrate for dealing with it?  In particular, an error like this would typically appear when there is an error saving the form.  I know @dturner32 likes to always have Retry buttons for things like that, so do we need one here?  Should we be demonstrating to the user that `NxLoadError` would actually be the most correct component for handling that situation, rather than a bare `NxErrorAlert`?  If there is a Retry button, should there also still be the Submit button?  If so would those two buttons do the same thing?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -4,53 +4,84 @@
...
+        <div className="nx-btn-bar">
+          <NxButton type="button">Cancel</NxButton>
+          <NxButton variant="primary" type="button">Submit</NxButton>
```
rpokorny, line 82: This should have the default type (`submit`), and then the form should have an `onSubmit` handler, which in this care would just be a function that calls `preventDefault()` on the event object.

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -4,61 +4,93 @@
...
+                 value="purple"
+                 onChange={setColor}
+                 isChecked={color === 'purple'}
+                 radioId="color-purple">
+          Purple
+        </NxRadio>
+        <NxRadio name="color" value="green" onChange={setColor} isChecked={color === 'green'} radioId="color-green">
+          Green
+        </NxRadio>
+        <NxRadio name="color" value="blue" onChange={setColor} isChecked={color === 'blue'} radioId="color-blue">
```
rpokorny, line 78:
```suggestion
        <NxRadio name="color"
                 value="red"
                 onChange={setColor}
                 isChecked={color === 'red'}>
          Red
        </NxRadio>
        <NxRadio name="color"
                 value="purple"
                 onChange={setColor}
                 isChecked={color === 'purple'}>
          Purple
        </NxRadio>
        <NxRadio name="color" value="green" onChange={setColor} isChecked={color === 'green'}>
          Green
        </NxRadio>
        <NxRadio name="color" value="blue" onChange={setColor} isChecked={color === 'blue'}>
```

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -4,61 +4,93 @@
...
         </label>
       </div>
-      <div className="nx-btn-bar nx-btn-bar--forms">
-        <button className="nx-btn nx-btn--primary">Submit</button>
-        <button className="nx-btn">Cancel</button>
-      </div>
+      <footer className="nx-form-footer">
+        <div className="nx-btn-bar">
+          <NxButton type="button">Cancel</NxButton>
+          <NxButton variant="primary" type="button">Submit</NxButton>
```
rpokorny, line 91: Like the other one, change it to have an `onSubmit` handler.

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
```diff
@@ -107,15 +106,19 @@ const NxFormLayoutPage = () =>
...
+    </GalleryExampleTile>
 
-    <GalleryTile title="NX Form Horizontal Layout Example">
-      <NxFormHorizontalLayoutExample />
-      <CodeExample content={NxFormHorizontalLayoutCode}/>
-    </GalleryTile>
+    <GalleryExampleTile title="Horizontal form layout"
+                        liveExample={NxFormHorizontalLayoutExample}
+                        codeExamples={NxFormHorizontalLayoutCode}>
+      This example demonstrates a form layout with horizontally placed text input fields. Note that the chcekbox and
```
rpokorny, line 118:
```suggestion
      This example demonstrates a form layout with horizontally placed text input fields. Note that the checkbox and
```

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
 
 .nx-form {
   @include container-vertical;
-  border-top: $nx-border-dark;
-  margin: 0;
-  padding-top: 20px;
+  margin: 0 0 $nx-spacing-xl 0;
```
rpokorny, line 13: I still think it should have top margin too

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
   position: relative; // to make a containing block for validation popovers
```
rpokorny, line 14: We don't have validation popovers anymore.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
 }
 
-.nx-form--simple {
-  border-top: none;
-  padding: 0;
-}
+.nx-form-footer {
+  @include container-horizontal;
```
rpokorny, line 18: Similar to my comment on the `nx-tile` footer PR, this should probably use `container-vertical` and _not_ `container-horizontal`.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
+
+  border-top: 1px solid #e9e9e9;
+  padding: $nx-spacing-l;
+
+  .nx-btn-bar {
+    margin: 0;
+    text-align: right;
```
rpokorny, line 25: This should be in `.nx-btn-bar`s styles in general (in the other PR I guess)

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
-  padding: 0;
-}
+.nx-form-footer {
+  @include container-horizontal;
+
+  border-top: 1px solid #e9e9e9;
+  padding: $nx-spacing-l;
+
+  .nx-btn-bar {
+    margin: 0;
```
rpokorny, line 24: Removable with `container-vertical` on the footer

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
+    text-align: right;
+  }
 
-// for a form that is the entirety of a modal
-.nx-form--modal-form {
-  padding-top: 0;
+  .nx-alert {
```
rpokorny, line 28: This whole block should be removable with `container-vertical` on the footer

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
 
-.nx-label--optional {
-  .nx-label__text::after {
+.nx-label__text, .nx-legend__text {
+  display: block;
+  margin-bottom: $nx-spacing-xs;
+}
+
+.nx-label--optional, .nx-legend--optional {
+  .nx-label__text::after, .nx-legend__text::after {
```
rpokorny, line 72: Should probably have an example of `.nx-legend--optional` to prove it works as expected

rpokorny: > Should probably have an example of .nx-legend--optional to prove it works as expected

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,83 @@
...
     @include regular();
 
     color: $nx-text-color-light;
-    content: " - optional";
+    content: "Optional";
+    font-size: $nx-font-size-xs;
     font-style: italic;
+    margin-left: 4px;
```
rpokorny, line 79:
```suggestion
    margin-left: $nx-spacing-xxs;
```

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -33,6 +34,7 @@
 }
 
 .nx-text-input__invalid-message {
+  @include regular();
```
rpokorny, line 37: This is only necessary because the `bold` on the label is being applied to the whole `nx-label` not just `nx-label__text` which is all we actually want it to affect.

lib/src/base-styles/_nx-forms.scss
glenhunter: > It seems to me that `.nx-legend` itself is the element most analogous to `.nx-label__text`. I'm wondering if we should get rid of `.nx-legend__text` and apply these styles to `.nx-legend` instead.

This doesn't make much sense to me, the purpose of the `__text` classes is to prevent bare text nodes which would cause issues if there is a sub-label, or an icon, etc.

glenhunter:
> Notice the alignment difference:
> 
> Zeplin:
> image
> 
> This PR:
> image
> 
> I'd say that the general `nx-form` and `nx-form-footer` containers should not have side padding but should instead rely on their parent container for that spacing.

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/3a8e8c4381ae05d8601a633bd46d0390f8cee417

lib/src/base-styles/_nx-forms.scss
rpokorny: > This doesn't make much sense to me, the purpose of the __text classes is to prevent bare text nodes which would cause issues if there is a sub-label, or an icon, etc.

I'm skeptical of that, considering that their implementation would conflict with that purpose.  They are `display: block`, so to give them an icon sibling would not have the desired effect - the icon would appear above the `__text` element.

rpokorny: > I'm skeptical of that, considering that their implementation would conflict with that purpose. They are display: block, so to give them an icon sibling would not have the desired effect - the icon would appear above the __text element.

Ok I do see your point regarding sublabels though.  I just think icons aren't the concern here.

lib/src/base-styles/_nx-forms.scss
glenhunter: > In conjunction with my comment below about getting rid of `.nx-legend__text`, I'm thinking this `cursor` property and also the `@include bold` should be moved to the presumptive `.nx-label__text, .nx-legend` block. Once that is done, I think `.nx-legend` can be removed from the selector on this block.

Moved here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/ff63dd0deb6f738e3f58f46a684a961d7ed5a903

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > Are sub-labels with icons no longer supported?

They are: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/269e73936381669e5546af3c31419782da4d1afa

I'm not sure that there's much benefit in demonstrating the addition of icons to things like this since the icons are now a component and robust enough to be "plug-and-play".

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > What scenario are we trying to emulate here, and what pattern are we trying to demonstrate for dealing with it? In particular, an error like this would typically appear when there is an error saving the form. I know @dturner32 likes to always have Retry buttons for things like that, so do we need one here? Should we be demonstrating to the user that `NxLoadError` would actually be the most correct component for handling that situation, rather than a bare `NxErrorAlert`? If there is a Retry button, should there also still be the Submit button? If so would those two buttons do the same thing?

Because this is supposed to be a layout example I was attempting to demonstrate how NxAlert lays out in a form footer. It doesn't have to be an error alert we could demonstrate a different kind of alert.

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > Like the other one, change it to have an `onSubmit` handler.

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/09a0b0871980983fb119a6202677911e21261a13

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
rpokorny: > It doesn't have to be an error alert we could demonstrate a different kind of alert.

Let's do that.  I wouldn't want people thinking this is the preferred way to communicate errors when it is missing pieces.  Then there's the question of whether Design thinks other sorts of alerts should go in the footer in the first place. @dturner32 ?

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
rpokorny: > There are some instances in IQ where a form control doesn't have a label, and there may be others elsewhere.

I guess I also question whether we really want that in our design language.  There are certainly places where a standalone text input makes sense (search boxes, table filters, etc), but I don't know that there's any case _in a form_ where you'd want an unlabeled text input.  Maybe if it's a form that consists of only a single text input and a submit button?

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -42,7 +44,10 @@ export default function NxFormLayoutExample() {
       <div className="nx-form-group">
         <label className="nx-label nx-label--optional">
           <span className="nx-label__text">Long field</span>
-          <span className="nx-sub-label">This is a sub-label. The field element below is wider than the default.</span>
+          <span className="nx-sub-label">
+            <NxFontAwesomeIcon icon={faCalendar}/> This is a sub-label. The field element below is wider than the
+            default.
+          </span>
```
rpokorny, line 55:
```suggestion
          <span className="nx-sub-label">
            <NxFontAwesomeIcon icon={faCalendar}/>
            <span>This is a sub-label. The field element below is wider than the default.</span>
          </span>
```

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
rpokorny: > I'm not sure that there's much benefit in demonstrating the addition of icons to things like this since the icons are now a component and robust enough to be "plug-and-play".

It proves the margins are working right

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -77,7 +82,7 @@ export default function NxFormLayoutExample() {
         <NxErrorAlert>This is an example <strong>error</strong> message.</NxErrorAlert>
         <div className="nx-btn-bar">
           <NxButton type="button">Cancel</NxButton>
-          <NxButton variant="primary" type="button">Submit</NxButton>
+          <NxButton variant="primary" type="submit">Submit</NxButton>
```
rpokorny, line 85: Btw since its the default, this `type` attr could have just been omitted.

lib/src/base-styles/_nx-forms.scss
glenhunter: > We don't have validation popovers anymore.

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/9b23a8ef7f13cbbd2845202a566be6930a420445

lib/src/base-styles/_nx-forms.scss
glenhunter: > I still think it should have top margin too

I have a feeling that this is going to result in us just removing the top margin in a bunch of places (like every tile): https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/f26df0a08f4ce9050772e14ec243743eb6eb9bb7

lib/src/base-styles/_nx-forms.scss
glenhunter: > Similar to my comment on the `nx-tile` footer PR, this should probably use `container-vertical` and _not_ `container-horizontal`.

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/2ac5a2b44a793efa167d05a98b550e0e9e8c5f92

lib/src/base-styles/_nx-forms.scss
glenhunter: > Removable with `container-vertical` on the footer

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/f8cf02dfed188d659f551d553375f363663f0a4b

lib/src/base-styles/_nx-forms.scss
glenhunter: > This whole block should be removable with `container-vertical` on the footer

The `margin: 0` can be removed but we still need to add the 24px of top margin to `nx-btn-bar` since that's custom.

lib/src/base-styles/_nx-forms.scss
glenhunter: > This should be in `.nx-btn-bar`s styles in general (in the other PR I guess)

I had already moved it here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/8d7de5adb14d6437832b48e3f73dcd28f491d9df

lib/src/base-styles/_nx-forms.scss
rpokorny: > I have a feeling that this is going to result in us just removing the top margin in a bunch of places (like every tile): f26df0a

That's what `container-vertical` is for

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > > It doesn't have to be an error alert we could demonstrate a different kind of alert.
> 
> Let's do that. I wouldn't want people thinking this is the preferred way to communicate errors when it is missing pieces. Then there's the question of whether Design thinks other sorts of alerts should go in the footer in the first place. @dturner32 ?

NxInfoAlert: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/085ddd03220d314b9bfd2ec3435f1f6e5745ef08

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > > There are some instances in IQ where a form control doesn't have a label, and there may be others elsewhere.
> 
> I guess I also question whether we really want that in our design language. There are certainly places where a standalone text input makes sense (search boxes, table filters, etc), but I don't know that there's any case _in a form_ where you'd want an unlabeled text input. Maybe if it's a form that consists of only a single text input and a submit button?

I'm not sure either, but the initial question was about whether we could dispense with the wrapping `.nx-form-group` since `<label>` appears to do the same job, but I think that we are better off with the extra wrapping `<div>` and the flexibility that gives us. I think asking `<label>` to also be our form group wrapper gives it too much responsibility within the DOM.

lib/src/base-styles/_nx-forms.scss
glenhunter: > > So if we customize that we're really just moving the magic number from here to there.
> 
> A magic number will be necessary somewhere for this. I think I prefer it being the line-height that is custom rather than the margin.

I think that makes more sense: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/4bd611669fd5e61a800a3b345de9c967e56274eb

lib/src/base-styles/_nx-forms.scss
glenhunter: > See note above about this 6px value.

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/765e91c924d0f35c452be938c152fabc6a306bf4

lib/src/base-styles/_nx-forms.scss
glenhunter: > Should probably have an example of `.nx-legend--optional` to prove it works as expected

Added here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/9367790a17224976136ab993c8d0457bd224f746

lib/src/components/NxTextInput/NxTextInput.scss
glenhunter: > This is only necessary because the `bold` on the label is being applied to the whole `nx-label` not just `nx-label__text` which is all we actually want it to affect.

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/d79d81ea38f5afcf06cb6bc37f753e43d8f8372d

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -79,7 +79,7 @@ export default function NxFormLayoutExample() {
         </NxRadio>
       </fieldset>
       <footer className="nx-form-footer">
-        <NxErrorAlert>This is an example <strong>error</strong> message.</NxErrorAlert>
+        <NxInfoAlert className="nx-alert--modifier">This is a sample alert message</NxInfoAlert>
```
rpokorny, line 82: I don't think this is doing anything.
```suggestion
        <NxInfoAlert>This is a sample alert message</NxInfoAlert>
```

lib/src/base-styles/_nx-forms.scss
rpokorny: > The margin: 0 can be removed but we still need to add the 24px of top margin to nx-btn-bar since that's custom. f8cf02d

I assume this discussion will be closely tied to the analogous one from the other PR: https://github.com/sonatype/sonatype-react-shared-components/pull/124#discussion_r476682973

rpokorny:
The sub-label on the legend doesn't feel great to me - the vertical spacing and text styles are identical to those of the radios themselves, making it almost look more like another radio option that's missing it's button, rather than part of the legend.  Maybe if it was closer to the main label.  @dturner32 what are your thoughts?
image

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,88 +10,75 @@
...
+    content: "Optional";
+    font-size: $nx-font-size-xs;
     font-style: italic;
+    margin-left: $nx-spacing-xxs;
   }
 }
 
 .nx-sub-label {
   @include container-horizontal;
   @include regular();
```
rpokorny, line 77: No longer required

glenhunter: > No longer required

https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/1a112ed02c02a4aaef8e99ffc5ebfb1fa980d7ab

dturner32:
> The sub-label on the legend doesn't feel great to me - the vertical spacing and text styles are identical to those of the radios themselves, making it almost look more like another radio option that's missing it's button, rather than part of the legend. Maybe if it was closer to the main label. @dturner32 what are your thoughts?
> image

I would allow for 16px of padding after the sub-label for the list of radios. Right now it looks like it's about at 8px from eyeballing it. The extra spacing should help add some separation

glenhunter:
> > The sub-label on the legend doesn't feel great to me - the vertical spacing and text styles are identical to those of the radios themselves, making it almost look more like another radio option that's missing it's button, rather than part of the legend. Maybe if it was closer to the main label. @dturner32 what are your thoughts?
> > image
> 
> I would allow for 16px of padding after the sub-label for the list of radios. Right now it looks like it's about at 8px from eyeballing it. The extra spacing should help add some separation

Screen Shot 2020-08-26 at 4 26 52 PM

That helped a bit, but I think the issue has more to do with the font/text. https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/b1b45064d7cb9317315fd2791d29fbb788a966e8

dturner32:
> That helped a bit, but I think the issue has more to do with the font/text.

In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that

glenhunter:
> > That helped a bit, but I think the issue has more to do with the font/text.
> 
> In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that

Ok, that sounds logical, should that be true of all form sub-labels?

glenhunter:
> > That helped a bit, but I think the issue has more to do with the font/text.
> 
> In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that

FYI the 4px after the label is tricky because the label text must have 8px for instances when the sub-label isn't there. 
Screen Shot 2020-08-26 at 5 09 33 PM

The screenshot shows 8px of spacing, 14px semi-bold font, and 16px of spacing

rpokorny:
> and 16px of spacing

I'm seeing 12px currently

dturner32:
> > > That helped a bit, but I think the issue has more to do with the font/text.
> > 
> > 
> > In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that
> 
> Ok, that sounds logical, should that be true of all form sub-labels?

correct. As for the 8px spacing after the label, you are also correct. And then there should always be 16px padding following the sub-label

dturner32:
> > > That helped a bit, but I think the issue has more to do with the font/text.
> > 
> > 
> > In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that
> 
> FYI the 4px after the label is tricky because the label text must have 8px for instances when the sub-label isn't there.
> Screen Shot 2020-08-26 at 5 09 33 PM
> 
> The screenshot shows 8px of spacing, 14px semi-bold font, and 16px of spacing

I'm noticing in the screenshot that the font color for the sub-label looks like the #474747 color instead of the #2c2c2c

glenhunter:
> > > > That helped a bit, but I think the issue has more to do with the font/text.
> > > 
> > > 
> > > In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that
> > 
> > 
> > FYI the 4px after the label is tricky because the label text must have 8px for instances when the sub-label isn't there.
> > Screen Shot 2020-08-26 at 5 09 33 PM
> > The screenshot shows 8px of spacing, 14px semi-bold font, and 16px of spacing
> 
> I'm noticing in the screenshot that the font color for the sub-label looks like the #474747 color instead of the #2c2c2c

Zeplin calls for #474747

Screen Shot 2020-08-27 at 10 04 39 AM

glenhunter:
> > > > That helped a bit, but I think the issue has more to do with the font/text.
> > > 
> > > 
> > > In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that
> > 
> > 
> > FYI the 4px after the label is tricky because the label text must have 8px for instances when the sub-label isn't there.
> > Screen Shot 2020-08-26 at 5 09 33 PM
> > The screenshot shows 8px of spacing, 14px semi-bold font, and 16px of spacing
> 
> I'm noticing in the screenshot that the font color for the sub-label looks like the #474747 color instead of the #2c2c2c

You need to update Zeplin with all this

glenhunter:
> > and 16px of spacing
> 
> I'm seeing 12px currently

Darn. Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/3c345736051f5a96e5376f0b9e4d114ce986eb9c

dturner32:
> > > > > That helped a bit, but I think the issue has more to do with the font/text.
> > > > 
> > > > 
> > > > In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that
> > > 
> > > 
> > > FYI the 4px after the label is tricky because the label text must have 8px for instances when the sub-label isn't there.
> > > Screen Shot 2020-08-26 at 5 09 33 PM
> > > The screenshot shows 8px of spacing, 14px semi-bold font, and 16px of spacing
> > 
> > 
> > I'm noticing in the screenshot that the font color for the sub-label looks like the #474747 color instead of the #2c2c2c
> 
> Zeplin calls for #474747
> 
> Screen Shot 2020-08-27 at 10 04 39 AM

I will fix this and update in Zeplin

dturner32:
> > > > > > That helped a bit, but I think the issue has more to do with the font/text.
> > > > > 
> > > > > 
> > > > > In the type scale I have an H4 set to follow an H3 (in this case the main label) as a label sub header. The sub-label here would technically count as an H4 and should be 14px;Semibold;#2c2c2c. I would then bump it down to 4px padding after the main label with the 16px padding following that
> > > > 
> > > > 
> > > > FYI the 4px after the label is tricky because the label text must have 8px for instances when the sub-label isn't there.
> > > > Screen Shot 2020-08-26 at 5 09 33 PM
> > > > The screenshot shows 8px of spacing, 14px semi-bold font, and 16px of spacing
> > > 
> > > 
> > > I'm noticing in the screenshot that the font color for the sub-label looks like the #474747 color instead of the #2c2c2c
> > 
> > 
> > Zeplin calls for #474747
> > Screen Shot 2020-08-27 at 10 04 39 AM
> 
> I will fix this and update in Zeplin

just updated this in Zeplin

lib/src/base-styles/_nx-forms.scss
glenhunter: > > The margin: 0 can be removed but we still need to add the 24px of top margin to nx-btn-bar since that's custom. [f8cf02d](https://github.com/sonatype/sonatype-react-shared-components/commit/f8cf02dfed188d659f551d553375f363663f0a4b)
> 
> I assume this discussion will be closely tied to the analogous one from the other PR: [#124 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/124#discussion_r476682973)

Ah, I see what you're saying now: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/9c2b33196d4e9225d698759b4cfbb6627f688d4f

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -5,6 +5,7 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 @import '../../scss-shared/_nx-form-validation.scss';
+@import '../../scss-shared/_nx-text-helpers.scss';
```
rpokorny, line 8: Is this import actually needed?

glenhunter: > Is this import actually needed?

No, apparently it is not: https://github.com/sonatype/sonatype-react-shared-components/pull/122/commits/bacb658dffb98e510909b16c3405b6963ede1c2b

PR #123, opened 2020-08-14
Fix Gallery in IE11

rpokorny:
I found today that the Gallery had been broken in IE11 since the query-string dependency was included.  Additionally the HTML element classname adjustments that were being in conjunction with query-string used a native API (the two-param classList.toggle function) that IE11 does not support, so that code had to be rewritten as well.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -60,6 +60,7 @@
     "@fortawesome/react-fontawesome": "=0.1.4",
     "@sonatype/react-shared-components": "link:../lib/dist",
     "classnames": "2.2.6",
+    "core-js": "3.6.5",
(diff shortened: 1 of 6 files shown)


gallery/package.json
```diff
@@ -60,6 +60,7 @@
     "@fortawesome/react-fontawesome": "=0.1.4",
     "@sonatype/react-shared-components": "link:../lib/dist",
     "classnames": "2.2.6",
+    "core-js": "3.6.5",
```
Blacktiger, line 63: Is this that library that always asks for money during npm install? I don't suppose there is something else we could use?

PR #124, opened 2020-08-19
nxTile footer changes RSC-253

glenhunter:
https://issues.sonatype.org/browse/RSC-253

Per changes to the designs update the nx-tile footers, add `nx-btn-bar` to the footers with buttons.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.4.3",
+  "version": "1.4.4",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -160,9 +160,7 @@ $nx-tile-background-color: $nx-white;
   padding: $nx-spacing-l 0;
 
   .nx-alert {
-    + .nx-btn-bar {
-      margin-top: $nx-spacing-l;
-    }
+    margin-bottom: $nx-spacing-l;
   }
 }
 
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -156,7 +156,7 @@ $nx-tile-background-color: $nx-white;
   @include container-vertical;
 
   border-top: 1px solid #e9e9e9;
-  margin-top: $nx-spacing-md;
+  margin-top: $nx-spacing-l;
   padding: $nx-spacing-l 0;
 
   .nx-alert {
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -157,8 +157,21 @@ $nx-tile-background-color: $nx-white;
 
   border-top: 1px solid #e9e9e9;
   margin-top: $nx-spacing-md;
-  padding: $nx-spacing-xl 0;
+  padding: $nx-spacing-l 0;
   text-align: right;
+
+  .nx-btn-bar {
+    margin: 0;
+    text-align: right;
+  }
+
+  .nx-alert {
+    margin: 0;
+
+    + .nx-btn-bar {
+      margin-top: $nx-spacing-l;
+    }
+  }
 }
 
 // A tile at the top of the page, usually contains a nav bar with scrolling buttons
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -153,21 +153,13 @@ $nx-tile-background-color: $nx-white;
 }
 
 .nx-tile-footer {
-  @include container-horizontal;
+  @include container-vertical;
 
   border-top: 1px solid #e9e9e9;
   margin-top: $nx-spacing-md;
   padding: $nx-spacing-l 0;
-  text-align: right;
-
-  .nx-btn-bar {
-    margin: 0;
-    text-align: right;
-  }
 
   .nx-alert {
-    margin: 0;
-
     + .nx-btn-bar {
       margin-top: $nx-spacing-l;
     }
--- a/gallery/src/styles/NxTile/NxSimpleTileExample.html
+++ b/gallery/src/styles/NxTile/NxSimpleTileExample.html
@@ -20,6 +20,6 @@
     meta- youtube warehouse Legba hotdog long-chain hydrocarbons carbon refrigerator saturation point.
   </div>
   <div class="nx-tile-footer">
-    <button class="nx-btn nx-btn--primary">Footer Button</button>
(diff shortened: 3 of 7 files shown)


lib/src/base-styles/_nx-tile.scss
```diff
@@ -157,8 +157,21 @@ $nx-tile-background-color: $nx-white;
 
   border-top: 1px solid #e9e9e9;
   margin-top: $nx-spacing-md;
-  padding: $nx-spacing-xl 0;
+  padding: $nx-spacing-l 0;
   text-align: right;
+
+  .nx-btn-bar {
+    margin: 0;
+    text-align: right;
```
rpokorny, line 165: My understanding is that _all_ `.nx-btn-bar`s should right-align their contents, not just ones in tile footers (or form footers).

lib/src/base-styles/_nx-tile.scss
```diff
@@ -157,8 +157,21 @@ $nx-tile-background-color: $nx-white;
 
   border-top: 1px solid #e9e9e9;
   margin-top: $nx-spacing-md;
-  padding: $nx-spacing-xl 0;
+  padding: $nx-spacing-l 0;
   text-align: right;
```
rpokorny, line 161: Should be removed now, unless there is an expectation that other footer content aside from a button bar should also be right-aligned.

lib/src/base-styles/_nx-tile.scss
```diff
@@ -157,8 +157,21 @@ $nx-tile-background-color: $nx-white;
 
   border-top: 1px solid #e9e9e9;
   margin-top: $nx-spacing-md;
```
rpokorny, line 159: Looking at the tile mocks in zeplin, this should be 24

lib/src/base-styles/_nx-tile.scss
```diff
@@ -157,8 +157,21 @@ $nx-tile-background-color: $nx-white;
```
rpokorny, line 157: Should this use container-vertical instead of container-horizontal?  That would allow you to get rid of the `margin: 0` properties below

rpokorny: > That would allow you to get rid of the margin: 0 properties below

And probably that sibling selector part too

lib/src/base-styles/_nx-tile.scss
```diff
@@ -157,8 +157,21 @@ $nx-tile-background-color: $nx-white;
...
-  padding: $nx-spacing-xl 0;
+  padding: $nx-spacing-l 0;
   text-align: right;
+
+  .nx-btn-bar {
+    margin: 0;
+    text-align: right;
+  }
+
+  .nx-alert {
```
rpokorny, line 168: Where can I see this scenario demonstrated?

lib/src/base-styles/_nx-tile.scss
glenhunter: > My understanding is that _all_ `.nx-btn-bar`s should right-align their contents, not just ones in tile footers (or form footers).

I changed this in the form layout PR

lib/src/base-styles/_nx-tile.scss
glenhunter: > Should be removed now, unless there is an expectation that other footer content aside from a button bar should also be right-aligned.

https://github.com/sonatype/sonatype-react-shared-components/pull/124/commits/6bc60547bcdc20d4f8eb364a2e584b0f091870e1

lib/src/base-styles/_nx-tile.scss
glenhunter: > Should this use container-vertical instead of container-horizontal? That would allow you to get rid of the `margin: 0` properties below

https://github.com/sonatype/sonatype-react-shared-components/pull/124/commits/6bc60547bcdc20d4f8eb364a2e584b0f091870e1

glenhunter: > > That would allow you to get rid of the margin: 0 properties below
> 
> And probably that sibling selector part too

The 24px top margin is still custom

lib/src/base-styles/_nx-tile.scss
glenhunter: > Where can I see this scenario demonstrated?

Form layout page

lib/src/base-styles/_nx-tile.scss
glenhunter: > Looking at the tile mocks in zeplin, this should be 24

https://github.com/sonatype/sonatype-react-shared-components/pull/124/commits/85709f44f4f557bcdf54b5fdc99486971fc1441c

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.4.3",
+  "version": "1.4.4",
```
rpokorny, line 3: Hmm, technically this is a breaking change in how the footers are supposed to be structured.

lib/src/base-styles/_nx-tile.scss
rpokorny: > The 24px top margin is still custom

Currently, in general NxAlert has 8px vertical margins, and nx-btn-bar has 12px vertical margins.  So am I to understand that in a footer the space between these two elements is non-standard? Or is one (or both) of those two margins I mentioned actually supposed to be 24px?

If we really do want this margin to be non-standard, it could probably be implemented more simply as a `margin-bottom` on `.nx-tile-footer .nx-alert` rather than using a sibling selector.

lib/src/base-styles/_nx-tile.scss
rpokorny: > Form layout page

That's a form footer not a tile footer

lib/src/base-styles/_nx-tile.scss
glenhunter: > If we really do want this margin to be non-standard, it could probably be implemented more simply as a `margin-bottom` on `.nx-tile-footer .nx-alert` rather than using a sibling selector.

I thought of that, but then the bottom margin is always on the `.nx-alert` so if there isn't any object below it then as it stands right now there's be 24px of bottom margin on the alert and 24px of padding on the footer which is 48px of space.

rpokorny: > I thought of that, but then the bottom margin is always on the .nx-alert so if there isn't any object below it then as it stands right now there's be 24px of bottom margin on the alert and 24px of padding on the footer which is 48px of space.

Again, this is what `container-vertical` is for.  In general every block component should declare the minimum distance it wants to be from siblings, and its up to the container to strip off that margin from its first and last children.

glenhunter: > > I thought of that, but then the bottom margin is always on the .nx-alert so if there isn't any object below it then as it stands right now there's be 24px of bottom margin on the alert and 24px of padding on the footer which is 48px of space.
> 
> Again, this is what `container-vertical` is for. In general every block component should declare the minimum distance it wants to be from siblings, and its up to the container to strip off that margin from its first and last children.

Right: https://github.com/sonatype/sonatype-react-shared-components/pull/124/commits/b2f1df5bd9c05c1dcd6e31676a065f9910e6734c

gallery/package.json
glenhunter: > Hmm, technically this is a breaking change in how the footers are supposed to be structured.

As discussed I have created a feature branch for V2 breaking changes

lib/src/base-styles/_nx-tile.scss
glenhunter: > > Form layout page
> 
> That's a form footer not a tile footer

I added an example here: https://github.com/sonatype/sonatype-react-shared-components/pull/124/commits/1c2eaf488ea5a9cca068f88e984929d2f8c4649d

PR #125, opened 2020-08-25
NxLoadError visual tests - RSC-191

rpokorny:
https://issues.sonatype.org/browse/RSC-191

This PR includes visual tests for NxLoadError and also a bug fix for the layout of NxLoadError: currently on master, when the error message is short and the Retry button is inline with it, the Retry button has too much space below it owing to the bottom margin on the error message.  Additionally, @dturner32 has made an adjustment to the expected spacing under that message which would have made the issue worse.  This PR fixes that by using the flexbox `row-gap` property instead of bottom margin.  Note that this property is only supported on the very latest versions of Chrome and Edge, and recent versions of Firefox.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@
 @Library(['private-pipeline-library', 'jenkins-shared']) _
 
 def seleniumDockerImage = 'selenium/standalone-chrome'
-def seleniumDockerVersion = '3.141.59-20200409'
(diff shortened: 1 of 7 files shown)


rpokorny:
> In IE11 the icon disappears in the long text version (this is not new, it happens in Master as well). Do we know why?

Fixed in 3f896d43

PR #128, opened 2020-08-27
DateRange POC (Not to be merged) - RSC-221

rpokorny:
https://issues.sonatype.org/browse/RSC-221

This PR demonstrates the best third-party daterange picker component that I found.  I just stuck it right into the gallery homepage for simplicity.  I recommend reviewing with white diffs turned off.

image

### DIFF
--- a/gallery/src/pages/Home.tsx
+++ b/gallery/src/pages/Home.tsx
@@ -4,79 +4,94 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React from 'react';
+import React, { useState } from 'react';
+import 'core-js/features/object/values';
+import 'core-js/features/number/is-nan';
+import 'core-js/features/array/find';
+import DateRangePicker from '@wojtekmaj/react-daterange-picker';
 import {GalleryTile} from '../gallery-components/GalleryTiles';
 
-const Home = () =>
-  <GalleryTile title="Welcome to the Sonatype React Shared Components Gallery">
-    <p className="nx-p">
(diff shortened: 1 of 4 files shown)


gallery/src/pages/Home.tsx
```diff
@@ -4,79 +4,94 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React from 'react';
+import React, { useState } from 'react';
+import 'core-js/features/object/values';
+import 'core-js/features/number/is-nan';
+import 'core-js/features/array/find';
```
rpokorny, line 10: These core-js imports are necessary to get it to work in IE11

gallery/src/react-daterange-picker.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@wojtekmaj/react-daterange-picker';
```
rpokorny, line 1: This daterange picker lib does not have any typescript typings.  This line enables typescript to accept it without errors (though of course also without type safety)

glenhunter:
I walked @dturner32 through it and we experimented with the functionality and the styling. It works as expected so that's fine, and we had success customizing the design. The only issue I see is that this is a date range picker, but we also need a date picker, and I wasn't sure that this could do double duty. If it can't then we need to make sure that the date picker can be styled to match whatever we have for this so that they don't look completely different.

rpokorny:
> The only issue I see is that this is a date range picker, but we also need a date picker, and I wasn't sure that this could do double duty.

I thought so far the range picker was all that was being requested.  But at any rate, the same person who created this library also has a single-date library based on much of the same shared code.  So I don't think it'll be a problem.

PR #130, opened 2020-09-03
Unified footer - Improvement Day

rpokorny:
While simultaneously restyling both tile footers and form footers @glenhunter and I found that they have the same styles and the same concerns.  Once the NxModal styles are updated, the modal footers will fall into the same boat.  Rather than have 2-3 different classes that do the same thing, this PR creates a single `.nx-footer` class.


raveneyex (approved):
LGTM

jlopez-sonatype (approved):
Awesome to simplify the use of styles, less things to remember when setting up new pages

glenhunter (approved):
Much simpler!

PR #131, opened 2020-09-03
Applitools Upgrade - RSC-266

rpokorny:
https://issues.sonatype.org/browse/RSC-266

This is expected to fix the issue where all of the applitools screenshot regions are off by 1px.


glenhunter (approved):
It makes me very happy to see that fixed

PR #132, opened 2020-09-03
Example of how a tile containing only a form should be structured - Improvement Day

rpokorny:
I found it ambiguous determining how a tile containing only a form should be structured.  After doing more research into how `header` and `footer` are supposed to work, I determined that the form and the tile cannot be the same element, because `form` is not a [Sectioning Content](https://www.w3.org/TR/2011/WD-html5-20110525/content-models.html#sectioning-content) element - that is, it does not create a semantic container for the `<header>` and `<footer>`, like `<section>` does.  This Gallery example proves that the styles work if you nest the `<form>` inside the` <section class="nx-tile">`, and asserts that that is the correct way to do it.

The other age-old question about this scenario – whether to use `nx-tile-footer` or `nx-form-footer` – is moot in this PR as it is based on my other branch where I merged them both into simply `nx-footer`.

### DIFF
--- a/gallery/src/styles/NxTile/NxTilesExamples.tsx
+++ b/gallery/src/styles/NxTile/NxTilesExamples.tsx
@@ -7,11 +7,15 @@
 import React from 'react';
 import { GalleryExampleTile } from '../../gallery-components/GalleryTiles';
 
+import NxTileFormExample from './NxTileFormExample';
+
 const NxSimpleTileCode = require('!!raw-loader!./NxSimpleTileExample.html').default,
     NxTileWithActionsCode = require('!!raw-loader!./NxTileWithActionsExample.html').default,
     NxTileWithSubtitleCode = require('!!raw-loader!./NxTileWithSubtitleExample.html').default,
     NxTileSubsectionCode = require('!!raw-loader!./NxTileSubsectionExample.html').default,
-    NxTileWithHorizontalRuleCode = require('!!raw-loader!./NxTileWithHorizontalRuleExample.html').default;
+    NxTileWithHorizontalRuleCode = require('!!raw-loader!./NxTileWithHorizontalRuleExample.html').default,
+    NxTileFormCode = require('!!raw-loader!./NxTileFormExample.tsx').default;
+
 
 const NxTilesExamples = () =>
   <>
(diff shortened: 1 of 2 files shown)


gallery/src/styles/NxTile/NxTilesExamples.tsx
```diff
@@ -7,11 +7,15 @@
...
+import NxTileFormExample from './NxTileFormExample';
+
 const NxSimpleTileCode = require('!!raw-loader!./NxSimpleTileExample.html').default,
     NxTileWithActionsCode = require('!!raw-loader!./NxTileWithActionsExample.html').default,
     NxTileWithSubtitleCode = require('!!raw-loader!./NxTileWithSubtitleExample.html').default,
     NxTileSubsectionCode = require('!!raw-loader!./NxTileSubsectionExample.html').default,
-    NxTileWithHorizontalRuleCode = require('!!raw-loader!./NxTileWithHorizontalRuleExample.html').default;
+    NxTileWithHorizontalRuleCode = require('!!raw-loader!./NxTileWithHorizontalRuleExample.html').default,
+    NxTileFormCode = require('!!raw-loader!./NxTileFormExample.tsx').default;
+
```
raveneyex, line 18: These consecutive empty-lines are triggering the linter.

raveneyex:
Shouldn't we have another example where a Form shares a tile with some other content? after all that's a scenario we've already seen.

rpokorny:
> after all that's a scenario we've already seen.

Where have we already seen that?

gallery/src/styles/NxTile/NxTilesExamples.tsx
rpokorny: > These consecutive empty-lines are triggering the linter.

321020a3

raveneyex:
> > after all that's a scenario we've already seen.
> 
> Where have we already seen that?

The new add waiver page is all within a single tile and has things that are not inputs.

rpokorny:
> The new add waiver page is all within a single tile and has things that are not inputs.

It has things that are not inputs but if IIRC, the entire tile is semantically considered a form.  Recall this exchange: https://sonatype.slack.com/archives/CCAPS4KL5/p1596042575286300?thread_ts=1596041650.285200&cid=CCAPS4KL5.

I've just gone back and checked the DOM, and the entire tile contents is a form: 
image

raveneyex:
> > The new add waiver page is all within a single tile and has things that are not inputs.
> 
> It has things that are not inputs but if IIRC, the entire tile is semantically considered a form. Recall this exchange: https://sonatype.slack.com/archives/CCAPS4KL5/p1596042575286300?thread_ts=1596041650.285200&cid=CCAPS4KL5.
> 
> I've just gone back and checked the DOM, and the entire tile contents is a form:
> image

Oh, ok.
I was under the impression we did that as a workaround for not having defined the whole form layout.

raveneyex (approved):

jlopez-sonatype (approved):

glenhunter (approved):
Interesting about a `<form>` not being a sectioning block. I did not know that.

PR #133, opened 2020-09-03
Experiment: add a checkered background to all live examples - Improvement Day

rpokorny:
In order to make the edges and backgrounds of our elements totally clear, I thought I'd experiment with adding a checkered background to all live examples.  This is particularly helpful with the `nx-tile` examples, where the rounded corners and padding of the tiles was not otherwise obvious.  It does however make many of our other examples quite ugly, as we have so far been assuming most things will be displayed on a white container such as a tile.  Whether or not that's a good assumption, and whether some of these components should have their own white background or truly be transparent to the background of whatever the container is, I am not sure.


gallery/src/main.scss
```diff
@@ -77,6 +77,24 @@
...
+  $checker-layer-3-positioning: -#{$checker-size / 2} -#{$checker-size / 2}/$checker-size $checker-size;
+
+  padding: $nx-spacing-l;
+
+  // create a checkerbox pattern for the background so that the edges and background of the example are clear
+  background: repeat $checker-layer-3-positioning linear-gradient(45deg, transparent 75%, $checker-color 75%),
+              repeat $checker-layer-2-positioning linear-gradient(45deg, $checker-color 25%, transparent 25%),
+              repeat $checker-layer-1-positioning
+                  linear-gradient(45deg, $checker-color 25%, transparent 25% 75%, $checker-color 75%),
+              $checker-bg-color;
```
rpokorny, line 95: The way that this works is admittedly non-obvious; I largely did it like this as a learning experience.  I'd be willing to replace it with a png or svg background if desired.

Blacktiger:
> In order to make the edges and backgrounds of our elements totally clear, I thought I'd experiment with adding a checkered background to all live examples. This is particularly helpful with the `nx-tile` examples, where the rounded corners and padding of the tiles was not otherwise obvious. It does however make many of our other examples quite ugly, as we have so far been assuming most things will be displayed on a white container such as a tile. Whether or not that's a good assumption, and whether some of these components should have their own white background or truly be transparent to the background of whatever the container is, I am not sure.

We could maybe use a query parameter or something to have this be something we can turn on during development.

rpokorny:
> We could maybe use a query parameter or something to have this be something we can turn on during development.

I think it's just as useful, if not more, for consumers of RSC as it is for the RSC developers.  One idea that I had was to have the checkered background on some examples but not all.  I think part of answering this, however, is asking which components should have their own background vs being transparent.  For instance, I suspect that the lack of background on our text inputs as seen below is incorrect. @dturner32 what are your thoughts?

image

glenhunter:
I found that the checkered background made some UI elements like dividing lines difficult to see so I tweaked the values a little bit to reduce the contrast of the checker pattern and I think it comes out a bit better, but still not great so I agree with @Blacktiger that it might be nice to make it a toggled feature.

Screen Shot 2020-09-08 at 12 36 14 PM
Screen Shot 2020-09-08 at 12 35 56 PM

Overall though (with the lower contrast colours) I like the effect. I find the higher contrast colours distracting.

rpokorny:
> I found that the checkered background made some UI elements like dividing lines difficult to see

I think that goes back to my point that some of these elements probably should have their _own_ background and not be transparent.  If `.nx-list` has dividing lines that only show up correctly on a white background, maybe it should have it's own white background?  As an aside, this experiment has helped to point out that `.nx-list` currently has some borders which are intended to be invisible but which are actually white.  If it _is_ meant to be usable on varying backgrounds, that should probably be fixed in your RSC-178 work, @glenhunter .

rpokorny:
> Overall though (with the lower contrast colours) I like the effect. I find the higher contrast colours distracting.

Contrast decreased in 4c1d37f8, though not as far you decreased it.

dturner32:
> > We could maybe use a query parameter or something to have this be something we can turn on during development.
> 
> I think it's just as useful, if not more, for consumers of RSC as it is for the RSC developers. One idea that I had was to have the checkered background on some examples but not all. I think part of answering this, however, is asking which components should have their own background vs being transparent. For instance, I suspect that the lack of background on our text inputs as seen below is incorrect. @dturner32 what are your thoughts?
> 
> image

Correct, in the case of text fields and inputs there should always be a white background present. For Buttons, secondary should always have a white background as well. Tertiary buttons however, would have transparent backgrounds.

dturner32:
> I found that the checkered background made some UI elements like dividing lines difficult to see so I tweaked the values a little bit to reduce the contrast of the checker pattern and I think it comes out a bit better, but still not great so I agree with @Blacktiger that it might be nice to make it a toggled feature.
> 
> Screen Shot 2020-09-08 at 12 36 14 PM
> 
> Screen Shot 2020-09-08 at 12 35 56 PM
> 
> Overall though (with the lower contrast colours) I like the effect. I find the higher contrast colours distracting.

I guess the question I would have here is in what instance would components like nx-list appear on a transparent/checkered background? I would assume it would always appear in some sort of container element, i.e. tile, accordion, etc. Unless the list appears as part of a treeview in a left/rail navigation rail such as in the RSC site, in which case it just shows up on the light grey page background

rpokorny:
> I guess the question I would have here is in what instance would components like nx-list appear on a transparent/checkered background

It would never realistically be shown on checkers, the checkers are only to make sure the edges and background of the component are unambiguous in the gallery.  It is however very possible that it could be used in, as you said, a sidebar.  In that scenario what should it look like? I guess we wouldn't want it to have it's own white background there, so the answer is that we need to make sure the border lines and text are visible enough on whatever background colors we expect there (which I would think they are).

dturner32:
> > I guess the question I would have here is in what instance would components like nx-list appear on a transparent/checkered background
> 
> It would never realistically be shown on checkers, the checkers are only to make sure the edges and background of the component are unambiguous in the gallery. It is however very possible that it could be used in, as you said, a sidebar. In that scenario what should it look like? I guess we wouldn't want it to have it's own white background there, so the answer is that we need to make sure the border lines and text are visible enough on whatever background colors we expect there (which I would think they are).

correct, in the case of a list appearing in a sidebar there would be no white background. The text should be dark enough to pass contrast on lighter backgrounds. As for the borders, #2c2c2c and #474747 should be dark enough as well. For dividers, #D1D6F0 is visible enough but #e9e9e9 isn't

rpokorny:
> For dividers, #D1D6F0 is visible enough but #e9e9e9 isn't

Do we need a bug ticket for that then?

gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -60,28 +65,48 @@ export const GalleryDescriptionTile: FunctionComponent<PropsWithRequiredChildren
...
         codeExampleElements = ensureArray(codeExamples)
             .map((example, idx) => {
               const props = typeof example === 'string' ? { content: example } : example;
               return <CodeExample key={idx} { ...props } />;
-            });
+            }),
+        tileActions = (
+          <NxCheckbox isChecked={checkeredBackground}
+                      onChange={toggleCheckeredBackground}>
+            Show Checkered Background on example
```
glenhunter, line 98:
```suggestion
            Show checkered background on example
```

glenhunter: Single the toggle is non-destructive I think we can shorten the label

glenhunter:
As discussed in Slack there's an issue where the padding isn't working as intended.

rpokorny:
> As discussed in Slack there's an issue where the padding isn't working as intended.

Fixed. 333b373

gallery/src/main.scss
glenhunter: > The way that this works is admittedly non-obvious; I largely did it like this as a learning experience. I'd be willing to replace it with a png or svg background if desired.

It's complicated but all of the variables presented allow so much latitude for customization that I think it's worth it.

glenhunter (approved):

PR #137, opened 2020-09-16
Long, user-resizable textareas - RSC-271

rpokorny:
https://issues.sonatype.org/browse/RSC-271

Adjusts the styles for textarea NxTextInputs so that the `--long` modifier works ~~on them while still allowing them to be resizable~~.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.6.1",
+  "version": "1.6.2",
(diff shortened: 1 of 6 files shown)


Blacktiger (approved):

glenhunter:
Per conversation in Wall Walk Sept 16, 2020 we just need to have vertical resizing which can be achieved by removing `width: auto;` from `.nx-text-input--textarea`.

glenhunter (approved):

Blacktiger (approved):

PR #138, opened 2020-09-16
Modify nx btn  icon only style RSC-281

glenhunter:
https://issues.sonatype.org/browse/RSC-281

Moved close btn classes to `--icon-only` and made the necessary modifications to `NxCloseButton`


lib/src/base-styles/_nx-btn.scss
```diff
@@ -110,8 +110,14 @@
...
 .nx-btn--icon-only {
+  border-color: transparent;
   height: 32px;
   padding: 0;
   width: 32px;
   min-width: 0;
+
+  .nx-icon {
+    height: 1em;
+    width: 1em;
```
rpokorny, line 121: I'm not sure this width is right universally, it might've only been for the X icon

lib/src/components/NxButton/NxButton.tsx
```diff
@@ -14,7 +14,7 @@ const NxButton = forwardRef<HTMLButtonElement, Props>(
       const classNames = classnames('nx-btn', className, {
         // secondary is the default, its styles are directly on `nx-btn`
         [`nx-btn--${variant}`]: variant && variant !== 'secondary',
-        'nx-btn--icon-only': iconOnly,
+        'nx-btn--tertiary nx-btn--icon-only': iconOnly,
```
rpokorny, line 17: If something like this is constructed: `<NxButton iconOnly variant="primary"/>`, both the `.nx-btn--primary` and `.nx-btn-tertiary` classes would end up on the `<button>`.  Whereas previously we wanted to allow icon-only buttons of any variant, and thus had `iconOnly` as a separate prop, I think now what we want is for `iconOnly` to be one of the `variant` values and to get rid of the separate prop.  This would prevent the caller from using `iconOnly` and a different, incompatible `variant` at the same time.

lib/src/components/NxCloseButton/NxCloseButton.tsx
```diff
@@ -11,14 +11,12 @@ import classnames from 'classnames';
 import Close from '../../icons/Close';
 import NxButton from '../NxButton/NxButton';
 
-import './NxCloseButton.scss';
-
 const NxCloseButton = forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HTMLButtonElement>>(
     function NxCloseButton({ className, ...otherProps }, ref) {
-      const btnClasses = classnames('nx-btn--close', className);
+      const btnClasses = classnames('nx-btn--icon-only', className);
```
rpokorny, line 16: This is redundant, the `iconOnly` prop to `NxButton` adds this already.

lib/src/base-styles/_nx-btn.scss
glenhunter: > I'm not sure this width is right universally, it might've only been for the X icon

I checked the Zeplin examples and they show icon sizes at 16px so I think that's universal

rpokorny: > I checked the Zeplin examples and they show icon sizes at 16px so I think that's universal

But I'm pretty certain that applitools icons aren't all the same aspect ratio.  So if they're all 1em tall they're not (all) also 1em wide

glenhunter: > > I checked the Zeplin examples and they show icon sizes at 16px so I think that's universal
> 
> But I'm pretty certain that applitools icons aren't all the same aspect ratio. So if they're all 1em tall they're not (all) also 1em wide

Ah: https://github.com/sonatype/sonatype-react-shared-components/pull/138/commits/253f6db96df88c119282b5409272f37d5c3ac185

lib/src/components/NxButton/NxButton.tsx
glenhunter: > If something like this is constructed: `<NxButton iconOnly variant="primary"/>`, both the `.nx-btn--primary` and `.nx-btn-tertiary` classes would end up on the `<button>`. Whereas previously we wanted to allow icon-only buttons of any variant, and thus had `iconOnly` as a separate prop, I think now what we want is for `iconOnly` to be one of the `variant` values and to get rid of the separate prop. This would prevent the caller from using `iconOnly` and a different, incompatible `variant` at the same time.

I knew there was something fishy there but needed your comment to fully grok it: https://github.com/sonatype/sonatype-react-shared-components/pull/138/commits/b1733248a4de3895fc82da9fdad044f2cc2dcd4c

lib/src/components/NxCloseButton/NxCloseButton.tsx
glenhunter: > This is redundant, the `iconOnly` prop to `NxButton` adds this already.

Right: https://github.com/sonatype/sonatype-react-shared-components/pull/138/commits/f912877a9dea22625b93362c0f3a59997b4267cd

rpokorny (approved):

PR #140, opened 2020-09-17
Improvement day: NxForm component

rpokorny:
This component encapsulates typical skeleton structure and behavior on an RSC-based form.  It is stateless, but manages the display of typical form components such as the loading wrapper, submit mask, validation tooltip on the submit button, etc.

So far I have only ported the basic form layout page to use this component, as a POC.  When bringing this work up to production quality I'd imagine `NxForm` would get it's own page with its own examples instead, which may or may not replace the two form pages we already have.

Note that I've also currently got the example contrived so that it has slow "loading" and intermittent "errors" in order to demonstrate the various states


lib/src/base-styles/_nx-forms.scss
```diff
@@ -11,6 +11,8 @@
 .nx-form {
   @include container-vertical;
   margin: $nx-spacing-xl 0;
+
+  position: relative;
```
glenhunter, line 15: What does this do? Removing it didn't seem to make a difference.

glenhunter:
I like the idea a lot and the execution seems very good.

I think that somebody with more React experience than I have needs to take a look as well.

PR #143, opened 2020-09-18
Update primary text color - RSC-276

rpokorny:
https://issues.sonatype.org/browse/RSC-276

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "1.6.5",
+  "version": "1.6.8",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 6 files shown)


glenhunter:
It seems to me that most of the instances of `color: $nx-text-color;` are now redundant and could be removed since the elements in question will pick up the color from `.nx-body`.

rpokorny:
> It seems to me that most of the instances of color: $nx-text-color; are now redundant and could be removed since the elements in question will pick up the color from .nx-body.

I did that, but I'll make another pass at it if you think I missed some

glenhunter:
> > It seems to me that most of the instances of color: $nx-text-color; are now redundant and could be removed since the elements in question will pick up the color from .nx-body.
> 
> I did that, but I'll make another pass at it if you think I missed some

I see lots. Basically anywhere that `$nx-text-color-light` changed to `$nx-text-color` I think could be removed

rpokorny:
> I see lots. Basically anywhere that $nx-text-color-light changed to $nx-text-color I think could be removed

The majority of them are necessary, but I found a few that could be removed: 7ba643b7

rpokorny:
> The majority of them are necessary, but I found a few that could be removed: 7ba643b

To clarify, I also didn't bother removing redundancy in components that have their own yet-to-be-completed refactorings: counter, list, and table

glenhunter (approved):

PR #147, opened 2020-09-22
nx-table updates - RSC-177

rpokorny:
https://issues.sonatype.org/browse/RSC-177


Blacktiger (approved):
I take it `metaInfo` is an internal prop and doesn't need any documentation? What's the purpose of that prop?

rpokorny:
> I take it metaInfo is an internal prop and doesn't need any documentation? What's the purpose of that prop?

That's how I intended it but I was thinking last night maybe it actually should be documented.  It just adds the `.nx-cell--meta-info` class, which is now the class that should be used for "special" cells that stretch all the way across.  The three current use cases being the empty, loading, and error states.  But I realized consumers might have their own additional scenarios requiring a meta-row like that, so I think I'll add docs for it.

glenhunter:
I'd like to see a table with fewer columns but longer text similar to what Derrek has in Zeplin, just to make sure that the text wraps and aligns correctly

glenhunter:
Derrek also shows wrapping and truncation in table cells so it would be nice to see examples of both

rpokorny:
> That's how I intended it but I was thinking last night maybe it actually should be documented. It just adds the .nx-cell--meta-info class, which is now the class that should be used for "special" cells that stretch all the way across. The three current use cases being the empty, loading, and error states. But I realized consumers might have their own additional scenarios requiring a meta-row like that, so I think I'll add docs for it.

3e79296

lib/src/base-styles/_nx-alert.scss
```diff
@@ -19,6 +19,7 @@
   display: flex;
   margin: 8px 0;
   padding: $nx-spacing-xs $nx-spacing-md $nx-spacing-xs $nx-spacing-md;
+  text-align: initial;
```
rpokorny, line 22: Needed for when a generic `NxAlert` is used within an `.nx-cell--meta-info`

Blacktiger (approved):

rpokorny:
> Derrek also shows wrapping and truncation in table cells so it would be nice to see examples of both

e613c06 and also note a5c06d4

lib/src/base-styles/_nx-tables.scss
```diff
@@ -10,141 +10,145 @@
...
 }
 
-// Used when table contents are empty
-.nx-cell--empty {
-  @include regular();
-
-  color: $nx-text-color;
+// Used for table cells containiner meta-information such as the fact that the table is empty, or
+// that the data is loading. These cells are expected to take up the entire row using colspan attributes
+.nx-cell--meta-info {
```
glenhunter, line 160: Zeplin shows the error, empty, and alert cells as having 16px of top spacing

rpokorny: > Zeplin shows the error, empty, and alert cells as having 16px of top spacing

78c2c7f0

rpokorny:
Just noticed scrolling table styles have broken, looking into it.

glenhunter:
@dturner32 I think that we should have a `selected` state for table rows.

glenhunter:
@dturner32 it seems od to me that in footers and lists the error alerts are the full width of the container but in tables they are constrained in width

lib/src/base-styles/_nx-tables.scss
```diff
@@ -10,141 +10,173 @@
...
-.nx-cell--label {
-  @include semi-bold();
-
-  color: $nx-text-color;
+  white-space: nowrap;
 }
 
 // columns whose content is numerical
 .nx-cell--num {
   text-align: center;
```
glenhunter, line 170: The Zeplin files show numbers left aligned, but I thought there had been a conversation about numbers being centred. @dturner32 ?

rpokorny:
> Ah, I see that Zeplin shows them as constrained but the examples here are full width - so which is correct?

@dturner32 and I discussed this while I was implementing.  Apparently, all `nx-alert`s are supposed to have a 950px max width.  That was never implemented and we need to put in a bug about it.  That said, I believe the examples here fall under that max width anyway.

rpokorny:
> sp: "container"

0e463fa2

lib/src/base-styles/_nx-tables.scss
dturner32: > The Zeplin files show numbers left aligned, but I thought there had been a conversation about numbers being centred. @dturner32 ?

yes they should be centered, I just forgot to update the mocks in zeplin. I will fix that

dturner32:
> @dturner32 I think that we should have a `selected` state for table rows.

I can add that in

dturner32:
> > @dturner32 I think that we should have a `selected` state for table rows.
> 
> I can add that in

just added selected states for table rows and lists to zeplin

lib/src/base-styles/_nx-tables.scss
dturner32: > > The Zeplin files show numbers left aligned, but I thought there had been a conversation about numbers being centred. @dturner32 ?
> 
> yes they should be centered, I just forgot to update the mocks in zeplin. I will fix that

just updated zeplin

lib/src/base-styles/_nx-tables.scss
```diff
@@ -10,141 +10,180 @@
...
-
-.nx-cell--icon {
-  color: $nx-text-color;
-  padding: 0 $nx-spacing-sm;
-  text-align: right;
-  width: 14px;
-  max-width: 32px;
-  min-width: 14px;
+  // seems there's still no cleaner way to get a cell to shrink to fit its contents
+  width: 1px;
```
glenhunter, line 186: If you add multiple icons they appear vertically. This doesn't seem optimal. Are you thinking that in that case each icon would have a separate cell?
Screen Shot 2020-09-29 at 4 38 27 PM

glenhunter: Or better yet could we have it adapt to multiple icons?

rpokorny: > Or better yet could we have it adapt to multiple icons?

I've separated out the styles for the chevron column vs general icon columns, allowing general icon columns to now contain multiple icons. 9cd195e0

glenhunter (approved):

PR #148, opened 2020-09-22
update link colour variables RSC-278

glenhunter:
https://issues.sonatype.org/browse/RSC-278

Updated the colours in the variables, checked for occurrences within the code that could also be updated with the variables

### DIFF
--- a/lib/src/scss-shared/_nx-colors.scss
+++ b/lib/src/scss-shared/_nx-colors.scss
@@ -77,10 +77,10 @@ $nx-text-color: #474747;
 $nx-text-color-disabled: #949494;
 $nx-text-color-error: $nx-fail-color;
 
-$nx-link-color: $nx-color-blue-44;
-$nx-link-color-active: #0046b1;
-$nx-link-color-hover: $nx-color-blue-42;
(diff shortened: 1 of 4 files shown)


lib/src/scss-shared/_nx-colors.scss
```diff
@@ -77,10 +77,10 @@ $nx-text-color: #474747;
...
 $nx-text-color-error: $nx-fail-color;
 
-$nx-link-color: $nx-color-blue-44;
-$nx-link-color-active: #0046b1;
-$nx-link-color-hover: $nx-color-blue-42;
-$nx-link-color-disabled: #97cbee;
+$nx-link-color: #045dd1;
+$nx-link-color-active: #2783fb;
+$nx-link-color-hover: #034396;
+$nx-link-color-disabled: #949494;
```
rpokorny, line 83: Do we use the disabled color anywhere?

glenhunter: > Do we use the disabled color anywhere?

Not yet, it's added for completeness

rpokorny (approved):

PR #141, opened 2020-09-17
nx-list redesign RSC-178

glenhunter:
https://issues.sonatype.org/browse/RSC-178

I used CSS grid as the default layout for list items and their contents, but back it out for the euro, alert and bulleted examples

PR #150, opened 2020-09-22
Fix Gallery Home Page Headings

rpokorny:
Re: this conversation from the nx-list PR https://github.com/sonatype/sonatype-react-shared-components/pull/141#discussion_r490578132

image


glenhunter:
I get what you have done and it makes sense semantically but I really do not like the way it looks. IMO the legibility of the text is much worse, I find the `h4`'s very inscrutable between the very similar `p` text and where with the old styles I felt a connection between the `h3`'s and the tree view sections they referenced I find that completely absent here. I'm +1 on the HTML, but 0 borderline -1 on the appearance. I would very much like @dturner32 to look this over and give his feedback/approval.

dturner32:
> I get what you have done and it makes sense semantically but I really do not like the way it looks. IMO the legibility of the text is much worse, I find the `h4`'s very inscrutable between the very similar `p` text and where with the old styles I felt a connection between the `h3`'s and the tree view sections they referenced I find that completely absent here. I'm +1 on the HTML, but 0 borderline -1 on the appearance. I would very much like @dturner32 to look this over and give his feedback/approval.

solution: change the style for h4's from semibold to bold

glenhunter (approved):

PR #151, opened 2020-09-23
NxLoadWrapper mandatory retryHandler - RSC-269

rpokorny:
https://issues.sonatype.org/browse/RSC-269


glenhunter:
Just to make sure that I understand what we're doing here (the JIRA ticket wasn't especially helpful), all `NxLoadWrapper`s must have a Retry button from now on?

rpokorny:
> Just to make sure that I understand what we're doing here (the JIRA ticket wasn't especially helpful), all NxLoadWrappers must have a Retry button from now on?

Yes, and in a softly enforced sense, that's been the case since 1.0.  This just makes it more firmly enforced.

glenhunter (approved):

PR #153, opened 2020-09-23
update tab styles RSC-153

glenhunter:
https://issues.sonatype.org/browse/RSC-153


lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,32 @@
  */
  @import '../../scss-shared/nx-colors';
  @import '../../scss-shared/nx-variables';
+ @import '../../scss-shared/nx-text-helpers';
 
 .nx-tab {
-  border-bottom: 4px solid transparent;
+  border-bottom: 2px solid transparent;
```
rpokorny, line 12: Looks like it's 1px except when active, then it's 2

rpokorny: > Looks like it's supposed to be 1px except when active, then it's 2

Oh I see, you've got the active border on top of the tar bar border.  It looks like it's supposed to overlap it.

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,32 @@
...
   cursor: pointer;
   display: inline-block;
   list-style: none;
-  padding: $nx-spacing-xs $nx-spacing-l;
+  margin-right: $nx-spacing-md;
+  // subtract border thickness for consistent spacing
+  padding: $nx-spacing-md 0 ($nx-spacing-md - 2px) 0;
+  text-align: center;
+  min-width: 125px;
```
rpokorny, line 20: What's the basis for this number?

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,32 @@
...
 .nx-tab {
-  border-bottom: 4px solid transparent;
+  border-bottom: 2px solid transparent;
   cursor: pointer;
   display: inline-block;
   list-style: none;
-  padding: $nx-spacing-xs $nx-spacing-l;
+  margin-right: $nx-spacing-md;
+  // subtract border thickness for consistent spacing
+  padding: $nx-spacing-md 0 ($nx-spacing-md - 2px) 0;
```
rpokorny, line 18: Not 24px on the sides?

rpokorny:
Long tabs (above the min-width) change width on hover, pushing the others over.  I seem to remember doing some crazy stuff to fix something like that a few months ago.  Ah right, it was on the links in NxPageHeader.

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,32 @@
...
+    border-bottom-color: #045dd1;
+    color: #045dd1;
+  }
+
+  &:hover {
+    @include semi-bold();
+  }
+
+  &:focus {
+    box-shadow: 0 0 3px 1px transparentize(#02a6d9, 0.7);
```
rpokorny, line 34: Supposed to have a border as well

rpokorny:
It seems like the padding on `.nx-tab-panel` needs to be adjusted - none on the sides, 24 above. (0 below?)

rpokorny:
`.nx-tabs` should probably have typical top and bottom margin

rpokorny:
`.nx-tab-panel` probably should use `container-vertical`

rpokorny:
This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing.  If so, we'd want an example on the nx-tile page demonstrating/proving that.

image

lib/src/components/NxTab/NxTab.scss
glenhunter: > What's the basis for this number?

It's the width of the underline borders in the Zeplin file

glenhunter:
> This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing. If so, we'd want an example on the nx-tile page demonstrating/proving that.
> 
> image

@dturner32 ?

glenhunter:
> `.nx-tab-panel` probably should use `container-vertical`

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/4430fc5bac08f24e0a5c477ecb69fb56bdd67850

glenhunter:
> It seems like the padding on `.nx-tab-panel` needs to be adjusted - none on the sides, 24 above. (0 below?)

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/4430fc5bac08f24e0a5c477ecb69fb56bdd67850 and I have asked about the bottom

lib/src/components/NxTab/NxTab.scss
glenhunter: > Supposed to have a border as well

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/43bb743a46d56793c00fbc5a3058e7efdcb2e077

glenhunter:
> `.nx-tabs` should probably have typical top and bottom margin

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/4430fc5bac08f24e0a5c477ecb69fb56bdd67850

lib/src/components/NxTab/NxTab.scss
glenhunter: > > Looks like it's supposed to be 1px except when active, then it's 2
> 
> Oh I see, you've got the active border on top of the tab bar border. It looks like it's supposed to overlap it.

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/4430fc5bac08f24e0a5c477ecb69fb56bdd67850

lib/src/components/NxTab/NxTab.scss
glenhunter: > Supposed to have a border as well

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/43bb743a46d56793c00fbc5a3058e7efdcb2e077

dturner32:
> This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing. If so, we'd want an example on the nx-tile page demonstrating/proving that.
> 
> image

I will add this as a mock up in the tiles page in zeplin

rpokorny:
> I have asked about the bottom

I'm thinking, given that the tab panel has no bottom border, that any space it defines at the bottom _should_ be margin, per our margin and padding guidelines.  That's why I suggested 0 for the padding.

dturner32:
> `.nx-tabs` should probably have typical top and bottom margin

tab names have a general top and bottom padding of 16px within the tab bar. However, when a tab is selected the 2px high blue bar underneath will cut the bottom padding of tab names to 14px. I think that addresses this comment, let me know if I'm off base

glenhunter:
> > `.nx-tabs` should probably have typical top and bottom margin
> 
> tab names have a general top and bottom padding of 16px within the tab bar. However, when a tab is selected the 2px high blue bar underneath will cut the bottom padding of tab names to 14px. I think that addresses this comment, let me know if I'm off base

We were talking about the spacing above and below the whole tab component which includes the tab bar (which contains the tabs), and the tab panel. I applied our stock 16px above, 24px below

glenhunter:
> Long tabs (above the min-width) change width on hover, pushing the others over. I seem to remember doing some crazy stuff to fix something like that a few months ago. Ah right, it was on the links in NxPageHeader.

The solution from page-header used a `data-text` attribute which is not present here. I have replicated it as a starting point for discussion. I tried a few other solutions that I found but none worked as well. https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/77560b647f77595231c70b7ad81908d6df300871

rpokorny:
> The solution from page-header used a data-text attribute which is not present here. I have replicated it as a starting point for discussion. I tried a few other solutions that I found but none worked as well. 77560b6

The right way to replicate it would have been to implement the setting of `data-text` within `NxTab`.  However, after writing my previous comment about this I realized that this approach has a very real shortcoming here: the tab content could include additional HTML (notably icons) and not just text.  `data-text` won't be able to handle that.  What we could do is construct the entire children DOM twice within `NxTab` - once visible and hoverable, and once invisible, always semi-bold, and in such a way that it affects the width of the tab (similarly to how that `::before` element works in the `data-text` approach).

rpokorny:
The focus border currently overrides the bottom selection border.  Should it? @dturner32

rpokorny:
Using the transparent borders all the way around causes the selection border to become trapezoidal.  Consider these changes where I avoid the transparent borders by using `outline` for the focus style: https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-153-update-tab-styles...RSC-153-update-tab-styles_ross

rpokorny:
> > This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing. If so, we'd want an example on the nx-tile page demonstrating/proving that.
>
> I will add this as a mock up in the tiles page in zeplin

@dturner32 it wasn't really a question about where it should be in zeplin.  It was about where it should be in the gallery, and more importantly to point out that the current mockup shows tabs in a place that we currently don't allow arbitrary content, or really any content: in an `nx-tile` above the tile header.  So that increases the scope of this ticket a fair amount and we wanted to make sure you really meant it to be that way, and to understand what the use case for such an arrangement would be.

dturner32:
> The focus border currently overrides the bottom selection border. Should it? @dturner32

yes it should, similar to the way the bar under a selected tab overlays the bottom border

dturner32:
> > > `.nx-tabs` should probably have typical top and bottom margin
> > 
> > 
> > tab names have a general top and bottom padding of 16px within the tab bar. However, when a tab is selected the 2px high blue bar underneath will cut the bottom padding of tab names to 14px. I think that addresses this comment, let me know if I'm off base
> 
> We were talking about the spacing above and below the whole tab component which includes the tab bar (which contains the tabs), and the tab panel. I applied our stock 16px above, 24px below

Ah, I see what you mean. Yes 16/24 would be correct then

dturner32:
> > > This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing. If so, we'd want an example on the nx-tile page demonstrating/proving that.
> > 
> > 
> > I will add this as a mock up in the tiles page in zeplin
> 
> @dturner32 it wasn't really a question about where it should be in zeplin. It was about where it should be in the gallery, and more importantly to point out that the current mockup shows tabs in a place that we currently don't allow arbitrary content, or really any content: in an `nx-tile` above the tile header. So that increases the scope of this ticket a fair amount and we wanted to make sure you really meant it to be that way, and to understand what the use case for such an arrangement would be.

thanks for clarifying this, I must've misunderstood. Tabs would be their own component independent of tiles so a tile wouldn't necessarily have it's own tab bar. The tab bar would have panels with content in them (i.e. Tab header, tab content) as discussed in our slack thread.

rpokorny:
> > The focus border currently overrides the bottom selection border. Should it? @dturner32
> 
> yes it should, similar to the way the bar under a selected tab overlays the bottom border

Currently though it doesn't overlay it, it overrides it.  Meaning, a focused, selected tab looks like this: 
image

That is incorrect, right?

dturner32:
> > > The focus border currently overrides the bottom selection border. Should it? @dturner32
> > 
> > 
> > yes it should, similar to the way the bar under a selected tab overlays the bottom border
> 
> Currently though it doesn't overlay it, it overrides it. Meaning, a focused, selected tab looks like this:
> image
> 
> That is incorrect, right?

I see what you mean now, yes that it is incorrect. I added an updated example of a 'focused/selected' state in zeplin. The focus state now has a 2px top and bottom padding so that it doesn't override the blue underline

glenhunter:
> > > > This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing. If so, we'd want an example on the nx-tile page demonstrating/proving that.
> > > 
> > > 
> > > I will add this as a mock up in the tiles page in zeplin
> > 
> > 
> > @dturner32 it wasn't really a question about where it should be in zeplin. It was about where it should be in the gallery, and more importantly to point out that the current mockup shows tabs in a place that we currently don't allow arbitrary content, or really any content: in an `nx-tile` above the tile header. So that increases the scope of this ticket a fair amount and we wanted to make sure you really meant it to be that way, and to understand what the use case for such an arrangement would be.
> 
> thanks for clarifying this, I must've misunderstood. Tabs would be their own component independent of tiles so a tile wouldn't necessarily have it's own tab bar. The tab bar would have panels with content in them (i.e. Tab header, tab content) as discussed in our slack thread.

I don't understand why we would want to make `nx-tabs` a component that would be used instead of an `nx-tile`. The issues of spacing `nx-tabs` within an `nx-tile` are trivial and the drawbacks of having it as a separate component are significant. For example, what happens if a design calls for text or a title above the tab bar? There are two examples of this in IQ, the CIP (in a modal) and LDAP config (in a tile).

Screen Shot 2020-09-25 at 9 53 15 AM

Screen Shot 2020-09-25 at 9 52 57 AM

How would we deal with this?

The issue where `nx-tabs` appears in a tile without a title is simple to solve, just omit the tile header.

`nx-tab-panel` can have its own `<header>`, content, and `<footer>` sections.

lib/src/components/NxTab/NxTab.scss
glenhunter: > Not 24px on the sides?

After clarification from Derrek: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/44a4cdae7cb8121108a56d03c97a62433c7e8074

glenhunter:
> > > > The focus border currently overrides the bottom selection border. Should it? @dturner32
> > > 
> > > 
> > > yes it should, similar to the way the bar under a selected tab overlays the bottom border
> > 
> > 
> > Currently though it doesn't overlay it, it overrides it. Meaning, a focused, selected tab looks like this:
> > image
> > That is incorrect, right?
> 
> I see what you mean now, yes that it is incorrect. I added an updated example of a 'focused/selected' state in zeplin. The focus state now has a 2px top and bottom padding so that it doesn't override the blue underline

Ok, so what is shown in the Zeplin mockup is not `padding`, it's `margin` this is a *very* important distinction. Padding happens inside the box, then moving from inside out there's border, then outline, then margin. 

https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/The_box_model

By putting 2px of margin around the entire tab for the focus state we need to shrink the tab by 2px in all dimensions in a way that doesn't cause the text to shift which may not be impossible but is needlessly complicated for such an ephemeral effect.

In addition what you describe in the Zeplin mockups for the selected + focused state is impossible because it shows the `:focus` outline _inside_ the border.

glenhunter:
Closing until design issues are worked out.

glenhunter:
Reopening, a `:focus` solution has been proposed and accepted, and `nx-tabs` will appear within the `content` sections of tiles and modals, examples of each with and without header will be created

lib/src/components/NxTab/NxTab.scss
glenhunter: > > What's the basis for this number?
> 
> It's the width of the underline borders in the Zeplin file

Per changes in Zeplin the min-width is now 76px: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/66e91ffe5e2498415d620ebac48d13c521b24d0c

glenhunter:
> > > > > The focus border currently overrides the bottom selection border. Should it? @dturner32
> > > > 
> > > > 
> > > > yes it should, similar to the way the bar under a selected tab overlays the bottom border
> > > 
> > > 
> > > Currently though it doesn't overlay it, it overrides it. Meaning, a focused, selected tab looks like this:
> > > image
> > > That is incorrect, right?
> > 
> > 
> > I see what you mean now, yes that it is incorrect. I added an updated example of a 'focused/selected' state in zeplin. The focus state now has a 2px top and bottom padding so that it doesn't override the blue underline
> 
> Ok, so what is shown in the Zeplin mockup is not `padding`, it's `margin` this is a _very_ important distinction. Padding happens inside the box, then moving from inside out there's border, then outline, then margin.
> 
> https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/The_box_model
> 
> By putting 2px of margin around the entire tab for the focus state we need to shrink the tab by 2px in all dimensions in a way that doesn't cause the text to shift which may not be impossible but is needlessly complicated for such an ephemeral effect.
> 
> In addition what you describe in the Zeplin mockups for the selected + focused state is impossible because it shows the `:focus` outline _inside_ the border.

I merged Ross' example branch with the solution: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/f9e26dd692a2e16c7c593f2cd41f1843204f0b91

glenhunter:
> > The solution from page-header used a data-text attribute which is not present here. I have replicated it as a starting point for discussion. I tried a few other solutions that I found but none worked as well. [77560b6](https://github.com/sonatype/sonatype-react-shared-components/commit/77560b647f77595231c70b7ad81908d6df300871)
> 
> The right way to replicate it would have been to implement the setting of `data-text` within `NxTab`. However, after writing my previous comment about this I realized that this approach has a very real shortcoming here: the tab content could include additional HTML (notably icons) and not just text. `data-text` won't be able to handle that. What we could do is construct the entire children DOM twice within `NxTab` - once visible and hoverable, and once invisible, always semi-bold, and in such a way that it affects the width of the tab (similarly to how that `::before` element works in the `data-text` approach).

Implemented here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/59b5183b2570e3d984a9d101c8ae2b3083f3ea09

glenhunter:
> > > > > This zeplin mockup makes it look like a tab bar at the very top of a tile, above the header, is supposed to be a thing. If so, we'd want an example on the nx-tile page demonstrating/proving that.
> > > > 
> > > > 
> > > > I will add this as a mock up in the tiles page in zeplin
> > > 
> > > 
> > > @dturner32 it wasn't really a question about where it should be in zeplin. It was about where it should be in the gallery, and more importantly to point out that the current mockup shows tabs in a place that we currently don't allow arbitrary content, or really any content: in an `nx-tile` above the tile header. So that increases the scope of this ticket a fair amount and we wanted to make sure you really meant it to be that way, and to understand what the use case for such an arrangement would be.
> > 
> > 
> > thanks for clarifying this, I must've misunderstood. Tabs would be their own component independent of tiles so a tile wouldn't necessarily have it's own tab bar. The tab bar would have panels with content in them (i.e. Tab header, tab content) as discussed in our slack thread.
> 
> I don't understand why we would want to make `nx-tabs` a component that would be used instead of an `nx-tile`. The issues of spacing `nx-tabs` within an `nx-tile` are trivial and the drawbacks of having it as a separate component are significant. For example, what happens if a design calls for text or a title above the tab bar? There are two examples of this in IQ, the CIP (in a modal) and LDAP config (in a tile).
> 
> Screen Shot 2020-09-25 at 9 53 15 AM
> 
> Screen Shot 2020-09-25 at 9 52 57 AM
> 
> How would we deal with this?
> 
> The issue where `nx-tabs` appears in a tile without a title is simple to solve, just omit the tile header.
> 
> `nx-tab-panel` can have its own `<header>`, content, and `<footer>` sections.
> 
> Turns out the only `.nx-tabs` example in NXRM has them inside a panel as well:
> 
> 83927806-3ca42e80-a74b-11ea-8a85-d1d2849c0116

Per discussion I have implemented examples of tabs within tiles and modals with and without headers: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/c99f6e873c69cc910546876c7400d799b223e121

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.0.0",
+  "version": "1.6.17",
```
rpokorny, line 3: Bad merge

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+  NxFontAwesomeIcon
+} from '@sonatype/react-shared-components';
+
+import {faAngry} from '@fortawesome/free-solid-svg-icons';
+
+const NxTabsModalExample = () => {
+  const [showModal, setShowModal] = useState(false);
+  const modalCloseHandler = () => setShowModal(false);
+
+  const [activeTabId, setActiveTabId] = useState<number>(0);
```
rpokorny, line 24: Explicit type parameter not needed here
```suggestion
  const [activeTabId, setActiveTabId] = useState(0);
```

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  function handleTabSelect(index: number) {
+    setActiveTabId(index);
+  }
+
+  return (
+    <>
+      <button onClick={() => setShowModal(true)} className="nx-btn">Open Modal</button>
```
rpokorny, line 32: I'd think, when in react component examples, we should use `NxButton` instead of `<button>`.  The other modal examples have already been updated in that way.

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+      { showModal &&
+        <NxModal id="nx-modal-simple-example" onClose={modalCloseHandler} className="nx-modal--wide">
```
rpokorny, line 34: Bad id

rpokorny, line 34: Unnecessary class, I don't think this really needs to be a wide modal

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <header className="nx-modal-header">
+            <h2 className="nx-h2">
+              <NxFontAwesomeIcon icon={faAngry} />
+              <span>Modal header</span>
+            </h2>
+          </header>
+          <div className="nx-modal-content">
+            <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect}>
```
rpokorny, line 42: I'd check with @dturner32 about the way this example ended up looking.  Feels odd to me

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+              <NxTabList aria-label="Simple Tabs">
+                <NxTab data-text="Tab">Tab</NxTab>
```
rpokorny, line 44: We aren't using the `data-text` thing anymore, right?

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <NxModal id="nx-modal-simple-example" onClose={modalCloseHandler} className="nx-modal--wide">
+          <header className="nx-modal-header">
+            <h2 className="nx-h2">
+              <NxFontAwesomeIcon icon={faAngry} />
+              <span>Modal header</span>
+            </h2>
+          </header>
+          <div className="nx-modal-content">
+            <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect}>
+              <NxTabList aria-label="Simple Tabs">
```
rpokorny, line 43: Does this aria label need to be updated?

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+                <NxTab data-text="Tab">Tab</NxTab>
+                <NxTab data-text="Tab with longer name">Tab with longer name</NxTab>
+                <NxTab data-text="Another Tab 3">Another Tab 3</NxTab>
+                <NxTab data-text="Forth Tab">Forth Tab</NxTab>
+              </NxTabList>
+              <NxTabPanel>Tab 1</NxTabPanel>
```
rpokorny, line 49: Adding larger content would probably be good.  How should the tab panel interact with the modal scrolling?

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+              <NxTabPanel>Tab 2</NxTabPanel>
+              <NxTabPanel>Tab 3</NxTabPanel>
+              <NxTabPanel>Tab 4</NxTabPanel>
+            </NxTabs>
+          </div>
+          <footer className="nx-footer">
+            <div className="nx-btn-bar">
+              <button type="button" onClick={modalCloseHandler} className="nx-btn">Close</button>
```
rpokorny, line 57: NxButton

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+  NxTab,
+  NxTabPanel,
+  NxModal
+} from '@sonatype/react-shared-components';
+
+const NxTabsModalNoHeaderExample = () => {
+  const [showModal, setShowModal] = useState(false);
+  const modalCloseHandler = () => setShowModal(false);
+
+  const [activeTabId, setActiveTabId] = useState<number>(0);
```
rpokorny, line 21:
```suggestion
  const [activeTabId, setActiveTabId] = useState(0);
```

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+              <span>Modal header</span>
+            </h2>
+          </header>
+          <div className="nx-modal-content">
+            <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect}>
+              <NxTabList aria-label="Simple Tabs">
+                <NxTab data-text="Tab">Tab</NxTab>
+                <NxTab data-text="Tab with longer name">Tab with longer name</NxTab>
+                <NxTab data-text="Another Tab 3">Another Tab 3</NxTab>
+                <NxTab data-text="Forth Tab">Forth Tab</NxTab>
```
rpokorny, line 47: Spelling
```suggestion
                <NxTab data-text="Forth Tab">Fourth Tab</NxTab>
```

rpokorny:
How should an excess of tabs be handled?
image

rpokorny:
How should tabs with excessively long names be handled? 
image

gallery/src/components/NxTabs/NxTabsTileHeaderExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  function handleTabSelect(index: number) {
+    setActiveTabId(index);
+  }
+
+  return (
+    <div className="nx-tile">
+      <div className="nx-tile-header">
+        <div className="nx-tile-header__title"><h2 className="nx-h2">Tile Header</h2></div>
+      </div>
```
rpokorny, line 27: Going forward I'm trying to use the semantic elements in all tile examples.  We should get around to updating the old examples at some point too.
```suggestion
    <section className="nx-tile">
      <header className="nx-tile-header">
        <div className="nx-tile-header__title"><h2 className="nx-h2">Tile Header</h2></div>
      </header>
```

gallery/src/components/NxTabs/NxTabsTileHeaderExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <NxTab data-text="Tab with longer name">Tab with longer name</NxTab>
+            <NxTab data-text="Another Tab 3">Another Tab 3</NxTab>
+            <NxTab data-text="Forth Tab">Forth Tab</NxTab>
+          </NxTabList>
+          <NxTabPanel>Tab 1</NxTabPanel>
+          <NxTabPanel>Tab 2</NxTabPanel>
+          <NxTabPanel>Tab 3</NxTabPanel>
+          <NxTabPanel>Tab 4</NxTabPanel>
+        </NxTabs>
+      </div>
```
rpokorny, line 41: Should probably have a footer in at least one of the tile examples.

gallery/src/components/NxTabs/NxTabsTileHeaderExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, {useState} from 'react';
+
+import {
+  NxTabs,
+  NxTabList,
+  NxTab,
+  NxTabPanel
+} from '@sonatype/react-shared-components';
+
+const NxTabsSimpleExample = () => {
```
rpokorny, line 16:
```suggestion
const NxTabsTileHeaderExample = () => {
```

lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "2.0.0",
+  "version": "1.6.17",
```
rpokorny, line 3: Bad merge

gallery/src/styles/NxCounter/NxCounterExample.html
```diff
@@ -6,10 +6,12 @@
...
 
 -->
-<div class="gallery-counter-example-container"><span>Default</span> <span class="nx-counter">42</span></div>
+<div class="gallery-counter-example-container">
+  <span>Default</span><span class="nx-counter">42</span>
+</div>
 
 <div class="gallery-counter-example-container">
-  <span>Active</span> <span class="nx-counter nx-counter--active">2 of 2</span>
+  <span>Active</span><span class="nx-counter nx-counter--active">2 of 2</span>
```
rpokorny, line 14: I made this change on master.  Did you merge master directly into this branch after merging the v2 feature branch into this branch?  Now that this branch is based off of the feature branch, you should merge from the feature branch to get upstream changes, not directly from master.  To get this out of the PR diff, merge master into the feature branch, and then the feature branch into this branch.

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,51 @@
  */
  @import '../../scss-shared/nx-colors';
  @import '../../scss-shared/nx-variables';
+ @import '../../scss-shared/nx-text-helpers';
+
+.nx-tabs {
+  margin: $nx-spacing-md 0 $nx-spacing-l 0;
+}
```
rpokorny, line 13: I know you're not a fan of the separate files per component, but given that we have them, I think this should go into the NxTabs.scss file.  As it stands now, it's not even in the right directory.  Also, if a consumer were to, for some reason, use `NxTabs` without also using `NxTab`, as it stands now they wouldn't get this style (only styles for components that are actually being used get bundled in a properly configured build).

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,51 @@
...
+    outline: solid 1px #02a6d9;
+  }
+
+  a:hover, a:focus {
     font-weight: bold;
+    letter-spacing: 0
+  }
+
+  // a hidden copy of the text, always at semi-bold weight, to enfore consistent element width
+  .hidden-children {
```
rpokorny, line 49: When I suggested this classname in slack I meant it only as pseudo-code.  For the actual impl I'd say it should be something like `.nx-tab__hidden-children`

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,51 @@
...
+  }
+
+  &:focus {
+    border-bottom-width: 1px;
+    box-shadow: 0 0 3px 1px transparentize(#02a6d9, 0.7);
+    outline: solid 1px #02a6d9;
+  }
+
+  a:hover, a:focus {
     font-weight: bold;
```
rpokorny, line 44: Any reason the mixin isn't being used here? (personally I think we should get rid of the mixins and just use `font-weight` everywhere)

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,51 @@
...
+  }
+
+  &:hover {
+    @include semi-bold();
+  }
+
+  &:focus {
+    border-bottom-width: 1px;
+    box-shadow: 0 0 3px 1px transparentize(#02a6d9, 0.7);
+    outline: solid 1px #02a6d9;
```
rpokorny, line 40: I would think that undoing the -1px bottom margin would be necessary here.  But for reasons I can't understand, it seems to make no difference.  I think it would be safer to add that change though.

lib/src/components/NxTabList/NxTabList.scss
```diff
@@ -5,9 +5,13 @@
...
+@import '../../scss-shared/nx-variables';
+@import '../../scss-shared/nx-container-helpers';
 
 .nx-tab-list {
-  border-bottom: $nx-border-row-dark;
+  @include container-horizontal;
+
+  border-bottom: 1px solid #d1d6f0;
+  margin: 0;
   padding: 0;
```
rpokorny, line 16: These margin and padding properties seem redundant

lib/src/components/NxTabs/types.ts
```diff
@@ -8,8 +8,8 @@ import { HTMLAttributes, ReactNode } from 'react';
 import PropTypes from 'prop-types';
 
 export type Props = HTMLAttributes<HTMLDivElement> & {
-  activeTab?: number | null ;
-  onTabSelect: ((index: number) => void);
+  activeTab?: number | null;
+  onTabSelect?: ((index: number) => void) | null;
```
rpokorny, line 12: Why'd you make this optional?  I don't think it should be.

glenhunter: > Why'd you make this optional? I don't think it should be.

I didn't. This was merged from your branch with the new `:focus` styles. I assumed that because it was in your branch you had a reason.

rpokorny: > I didn't. This was merged from your branch with the new :focus styles. I assumed that because it was in your branch you had a reason.

:thinking: no, I don't think so.  Does it still compile without it?

gallery/src/styles/NxCounter/NxCounterExample.html
glenhunter: > I made this change on master. Did you merge master directly into this branch after merging the v2 feature branch into this branch? Now that this branch is based off of the feature branch, you should merge from the feature branch to get upstream changes, not directly from master. To get this out of the PR diff, merge master into the feature branch, and then the feature branch into this branch.

This morning I merged Master into this branch, then we decided that we were going to put this branch against V2, so this afternoon I merged Master in V2, then V2 into this branch, so I'm not sure why it's showing up.

lib/src/components/NxTabs/types.ts
glenhunter: > > I didn't. This was merged from your branch with the new :focus styles. I assumed that because it was in your branch you had a reason.
> 
> 🤔 no, I don't think so. Does it still compile without it?

It does, so now it's gone again: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/e279fc6239b2fbd3b4ad9bce5219f8bb75abc358

lib/src/components/NxTabList/NxTabList.scss
glenhunter: > These margin and padding properties seem redundant

They aren't. They override the default browser values for a `<ul>`

lib/src/components/NxTab/NxTab.scss
glenhunter: > Any reason the mixin isn't being used here? (personally I think we should get rid of the mixins and just use `font-weight` everywhere)

No good reason: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/fe3e79d3cd8a31b9d93f46ae6dafdecee940e3fa

I think there is a historical reason why the mixing were needed, that might not be true anymore

lib/src/components/NxTab/NxTab.scss
glenhunter: > When I suggested this classname in slack I meant it only as pseudo-code. For the actual impl I'd say it should be something like `.nx-tab__hidden-children`

I waffled on that, but you're right it should reference `tab` just to make it specific. https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/3a74a316a9f1948537389aed91b2d849dfb6ae8a

lib/src/components/NxTab/NxTab.scss
glenhunter: > I know you're not a fan of the separate files per component, but given that we have them, I think this should go into the NxTabs.scss file. As it stands now, it's not even in the right directory. Also, if a consumer were to, for some reason, use `NxTabs` without also using `NxTab`, as it stands now they wouldn't get this style (only styles for components that are actually being used get bundled in a properly configured build).

I'm 100% in favor of a separate SCSS file for each component, I'm not very keen on cutting a single component up into four parts, that seems unnecessarily fussy. But yes this code was in the wrong place, fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/a24a17bb4ebe7479f1dbcc5823a98e7ddb316e22

gallery/src/styles/NxCounter/NxCounterExample.html
rpokorny: > I merged Master in V2, then V2 into this branch

I just did this part again, should fix it.  I also fixed the version numbers while at it.

glenhunter:
> How should an excess of tabs be handled?
> image

@dturner32 ?

glenhunter:
> How should tabs with excessively long names be handled?
> image

@dturner32 ?

I think a `max-width` and truncation would probably be the best solution, but what `max-width`?

lib/src/components/NxTab/NxTab.scss
rpokorny: > I'm 100% in favor of a separate SCSS file for each component, I'm not very keen on cutting a single component up into four parts, that seems unnecessarily fussy. But yes this code was in the wrong place, fixed here: a24a17b

Right.  There's nothing stopping someone from writing multiple closely-related React components in the same tsx file, and then all their styles in a single corresponding scss file.  That's probably what we should try to do in this situation in the future.

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > Unnecessary class, I don't think this really needs to be a wide modal

If it isn't then the four tabs can't display - which goes to your comment below and depending on what @dturner32 says I might make this modal normal to show the behavior

glenhunter: > Bad id

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/f33e0cfd25b5244821e92a8b17797abbdb6d5756

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > I'd think, when in react component examples, we should use `NxButton` instead of `<button>`. The other modal examples have already been updated in that way.

I agree, bad copy/pasta on my part: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/cda6e6e96355d6fbfe70cdbc3722c0e6835f934a

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > I'd check with @dturner32 about the way this example ended up looking. Feels odd to me

It is respecting the correct `24px` internal padding of `nx-modal-content`, I think it just looks a little off because of the space above the tabs that appears empty, but isn't because that's where the `:focus` rectangle goes.
Screen Shot 2020-09-29 at 2 42 08 PM
Screen Shot 2020-09-29 at 2 40 22 PM

lib/src/components/NxTabs/types.ts
rpokorny: > It does, so now it's gone again: e279fc6

There were still other remaining changes there, mainly the `| null` part.  I've reverted it to what's on master. e2fca1cc

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > We aren't using the `data-text` thing anymore, right?

Nope, I deleted the original and forgot the others: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/3aee5949b382944a42188aace36ed9c5aada9ab3

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > Does this aria label need to be updated?

All of them updated here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/3aee5949b382944a42188aace36ed9c5aada9ab3

gallery/src/components/NxTabs/NxTabsModalExample.tsx
rpokorny: > It is respecting the correct 24px internal padding of nx-modal-content, I think it just looks a little off because of the space above the tabs that appears empty, but isn't because that's where the :focus rectangle goes.

Right, it's certainly respecting the rules we have so far, it's just a question of whether the result looks "good", especially when none of the tabs are focused.
image

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > Adding larger content would probably be good. How should the tab panel interact with the modal scrolling?

I'll bet this isn't what we want:

Screen Shot 2020-09-29 at 2 50 04 PM

@dturner32 ^^

gallery/src/components/NxTabs/NxTabsTileHeaderExample.tsx
glenhunter: > Should probably have a footer in at least one of the tile examples.

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/cda6e6e96355d6fbfe70cdbc3722c0e6835f934a

gallery/src/components/NxTabs/NxTabsModalExample.tsx
rpokorny: > All of them updated here: 3aee594

They all appear unchanged to me

lib/src/components/NxTab/NxTab.scss
glenhunter: > I would think that undoing the -1px bottom margin would be necessary here. But for reasons I can't understand, it seems to make no difference. I think it would be safer to add that change though.

Yeah that's really weird, I reversed it here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/fd3679f6304ea30dc3ba1bbf58d806e063e779ab

gallery/src/components/NxTabs/NxTabsModalExample.tsx
rpokorny: > I'll bet this isn't what we want:

My thinking is that when an `NxTabs` is an only child of `.nx-modal-content`, it might get `flex-shrink: 1` such that it grows only to the size of the container and no higher.  Then within, it would use flexbox to manage the height of the tab panel, which would then have `overflow-y: auto`. This should institute a scrollbar within the tabe panel when necessary.

rpokorny:
> I think a max-width and truncation would probably be the best solution, but what max-width?

An alternative idea, which would address both excess number of tabs and excess tab content: use flex to force them to shrink to fit in one row (with truncation for overflow)

dturner32:
> > How should an excess of tabs be handled?
> > image
> 
> @dturner32 ?

tabs should never exceed past one row. If the team implementing this component has too many tabs to fit in one row then it becomes a question of being overly complex and a sign that they need to simplify. I updated zeplin to include a mock up of what a reasonable amount of tabs would look like in the tab bar. I also agree with Ross in using flex to force them shrink to fit everything in one row

rpokorny:
> becomes a question of being overly complex and a sign that they need to simplify

That's reasonable, except for situations where the tabs aren't hard-coded.  For instance, if the user can open an arbitrary number of tabs, or if the tab labels are based on dynamic data.  So it sounds like you're saying for those situations, we should shrink the tabs to fit in one row, correct?

dturner32:
> > How should tabs with excessively long names be handled?
> > image
> 
> @dturner32 ?
> 
> I think a `max-width` and truncation would probably be the best solution, but what `max-width`?

this is a very extreme edge case and I would hope no one would have a tab name that long in general. Tabs will have a max width of '240px' and any tab name beyond that would be truncated. I added a mockup in zeplin to reflect that

dturner32:
> > becomes a question of being overly complex and a sign that they need to simplify
> 
> That's reasonable, except for situations where the tabs aren't hard-coded. For instance, if the user can open an arbitrary number of tabs, or if the tab labels are based on dynamic data. So it sounds like you're saying for those situations, we should shrink the tabs to fit in one row, correct?

correct, as much as possible I'd like to keep the number of tabs to one row even if things have to shrink to fit (similar to tabs browser tabs). I'd rather shrink them than to stack them

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,49 @@
...
+  border-bottom: 2px solid transparent;
   cursor: pointer;
   display: inline-block;
   list-style: none;
-  padding: $nx-spacing-xs $nx-spacing-l;
+  margin: 0 $nx-spacing-md -1px 0;
+  // subtract border thickness for consistent spacing
+  padding: $nx-spacing-md $nx-spacing-l ($nx-spacing-md - 2px) $nx-spacing-l;
+  text-align: center;
+  min-width: 76px;
```
rpokorny, line 20: Needs `box-sizing: border-box` in order to do what was desired.

gallery/src/components/NxTabs/NxTabsModalExample.tsx
dturner32: for this one I was thinking that the tab bar would become the topmost element in the modal, an instance where the modal header would be replaced, and then within the tab panel there would be a tab panel header. This way only the content in the tab panel itself scrolls but the tab bar doesn't

lib/src/components/NxTab/NxTab.scss
glenhunter: > Needs `box-sizing: border-box` in order to do what was desired.

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/2d9f2cb8b2c0a7a0799718bbc70082fc0160d4e4

glenhunter:
> > > becomes a question of being overly complex and a sign that they need to simplify
> > 
> > 
> > That's reasonable, except for situations where the tabs aren't hard-coded. For instance, if the user can open an arbitrary number of tabs, or if the tab labels are based on dynamic data. So it sounds like you're saying for those situations, we should shrink the tabs to fit in one row, correct?
> 
> correct, as much as possible I'd like to keep the number of tabs to one row even if things have to shrink to fit (similar to browser tabs). I'd rather shrink them than to stack them

Per Ross' suggestion I converted the tabs to use `flex`, this solves both the "too many tabs" and "too much tab text" issues. I also increased the minimum tab size to `80px` as this allows the word "Tab" to appear without truncation ("Wab" requires `90px`)

rpokorny:
> Per Ross' suggestion I converted the tabs to use flex, this solves both the "too many tabs" and "too much tab text" issues. I also increased the minimum tab size to 80px as this allows the word "Tab" to appear without truncation ("Wab" requires 90px)

FWIW if you add so many tabs that they need to shrink down to the min width, and then keep adding more, they now overflow and become inaccessible.  I think this is acceptable though; the app just shouldn't let there be this many tabs.

image

gallery/src/components/NxTabs/NxTabsModalExample.tsx
rpokorny: > I agree, bad copy/pasta on my part: cda6e6e

You missed some, including the one this comment is on

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const [activeTabId, setActiveTabId] = useState(0);
+
+  function handleTabSelect(index: number) {
+    setActiveTabId(index);
+  }
+
+  return (
+    <>
+      <button onClick={() => setShowModal(true)} className="nx-btn">Open Modal</button>
```
rpokorny, line 29: NxButton

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+              </NxTabList>
+              <NxTabPanel>Tab 1</NxTabPanel>
+              <NxTabPanel>Tab 2</NxTabPanel>
+              <NxTabPanel>Tab 3</NxTabPanel>
+              <NxTabPanel>Tab 4</NxTabPanel>
+            </NxTabs>
+          </div>
+          <footer className="nx-footer">
+            <div className="nx-btn-bar">
+              <button type="button" onClick={modalCloseHandler} className="nx-btn">Close</button>
```
rpokorny, line 48: NxButton

gallery/src/components/NxTabs/NxTabsModalExample.tsx
rpokorny: > They all appear unchanged to me

Seems now that you got to them at some point.

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <button onClick={() => setShowModal(true)} className="nx-btn">Open Modal</button>
+      { showModal &&
+        <NxModal onClose={modalCloseHandler} className="nx-modal--wide">
+          <div className="nx-modal-content">
+            <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect}>
+              <NxTabList aria-label="Tabs in a modal with no header">
+                <NxTab>Tab</NxTab>
+                <NxTab>Tab with longer name</NxTab>
+                <NxTab>Another Tab 3</NxTab>
+                <NxTab>Forth Tab</NxTab>
```
rpokorny, line 38: Still misspelled in most examples

gallery/src/components/NxTabs/NxTabsTileNoHeaderExample.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <div className="nx-tile">
+      <div className="nx-tile-content">
+        <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect} id="test-tab-list">
+          <NxTabList aria-label="Tabs in a tile with no header">
+            <NxTab data-text="Tab">Tab</NxTab>
+            <NxTab data-text="Tab with longer name">Tab with longer name</NxTab>
+            <NxTab data-text="Another Tab 3">Another Tab 3</NxTab>
+            <NxTab data-text="Forth Tab">Forth Tab</NxTab>
```
rpokorny, line 31:
```suggestion
            <NxTab>Tab</NxTab>
            <NxTab>Tab with longer name</NxTab>
            <NxTab>Another Tab 3</NxTab>
            <NxTab>Forth Tab</NxTab>
```

gallery/src/components/NxTabs/NxTabsTileNoHeaderExample.tsx
```diff
@@ -0,0 +1,43 @@
...
+import React, {useState} from 'react';
+
+import {
+  NxTabs,
+  NxTabList,
+  NxTab,
+  NxTabPanel
+} from '@sonatype/react-shared-components';
+
+const NxTabsSimpleExample = () => {
```
rpokorny, line 16:
```suggestion
const NxTabsTileNoHeaderExample = () => {
```

gallery/src/components/NxTabs/NxTabsTileNoHeaderExample.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [activeTabId, setActiveTabId] = useState<number>(0);
```
rpokorny, line 17:
```suggestion
  const [activeTabId, setActiveTabId] = useState(0);
```

gallery/src/components/NxTabs/NxTabsTileHeaderExample.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {
+  NxTabs,
+  NxTabList,
+  NxTab,
+  NxTabPanel,
+  NxButton
+} from '@sonatype/react-shared-components';
+
+const NxTabsTileHeaderExample = () => {
+  const [activeTabId, setActiveTabId] = useState<number>(0);
```
rpokorny, line 18:
```suggestion
  const [activeTabId, setActiveTabId] = useState(0);
```

gallery/src/components/NxTabs/NxTabsModalExample.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+import {faAngry} from '@fortawesome/free-solid-svg-icons';
+
+const NxTabsModalExample = () => {
+  const [showModal, setShowModal] = useState(false);
+  const modalCloseHandler = () => setShowModal(false);
+
+  const [activeTabId, setActiveTabId] = useState(0);
+
+  function handleTabSelect(index: number) {
```
rpokorny, line 27: It occurs to me that all of these `handleTabSelect` functions are redundant.  They could be gotten rid of and `setActiveTabId` used directly in their place.

gallery/src/components/NxTabs/NxTabsTileNoHeaderExample.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  function handleTabSelect(index: number) {
+    setActiveTabId(index);
+  }
+
+  return (
+    <div className="nx-tile">
+      <div className="nx-tile-content">
+        <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect} id="test-tab-list">
```
rpokorny, line 26: Is this id needed?
```suggestion
        <NxTabs activeTab={activeTabId} onTabSelect={handleTabSelect}>
```

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
glenhunter: > Still misspelled in most examples

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/151565b015812ebb22b5fb2fcc53ec98daec903b

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
glenhunter: > NxButton

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/44260b957a56660669f7dbf8e4b6dfec8dcf2a23

gallery/src/components/NxTabs/NxTabsModalNoHeaderExample.tsx
glenhunter: > NxButton

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/d9517f5cb8a28d16b2b68bc41feaecfdb57bf8ab

gallery/src/components/NxTabs/NxTabsTileNoHeaderExample.tsx
glenhunter: > Suggested change 
>   const [activeTabId, setActiveTabId] = useState<number>(0);
>   const [activeTabId, setActiveTabId] = useState(0);

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/8a7f0e409818dee543dc08f90b9355bc80682b80

gallery/src/components/NxTabs/NxTabsTileNoHeaderExample.tsx
glenhunter: > Suggested change 
> const NxTabsSimpleExample = () => {
> const NxTabsTileNoHeaderExample = () => {

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/c9278395e500fa2cbbf537a7ae90ffd8e26be0f8

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > It occurs to me that all of these `handleTabSelect` functions are redundant. They could be gotten rid of and `setActiveTabId` used directly in their place.

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/f7fe532764bc5ebd4b64a1f6baf66f8bbb6b1f99

gallery/visualtests/nxTabs.js
```diff
@@ -11,7 +11,52 @@ describe('NxTabs', function() {
...
+      const openModalBtn = await browser.$(openModalBtnSelector);
+
+      await openModalBtn.scrollIntoView({ block: 'center' });
+      await openModalBtn.click();
+
+      const closeModalBtn = await browser.$(closeModalBtnSelector);
+
+      try {
+        // take image of entire viewport in order to capture the backdrop color
+        await browser.eyesSnapshot(null);
```
rpokorny, line 47: For this test you probably don't really need the entire viewport

lib/src/components/NxTab/NxTab.scss
```diff
@@ -6,16 +6,54 @@
...
+    box-shadow: 0 0 3px 1px transparentize(#02a6d9, 0.7);
+    margin-bottom: 0;
+    outline: solid 1px #02a6d9;
+  }
+
+  a:hover, a:focus {
+    @include bold();
+
+    letter-spacing: 0
+  }
```
rpokorny, line 49: What is this block? We aren't using anchor tags here, and I don't see why `letter-spacing` would need to be customized.

gallery/visualtests/nxTabs.js
glenhunter: > For this test you probably don't really need the entire viewport

https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/b9ae5a5e965c2e97e7cfc67c07f1bb9550864ba7

lib/src/components/NxTab/NxTab.scss
glenhunter: > What is this block? We aren't using anchor tags here, and I don't see why `letter-spacing` would need to be customized.

I was part of a failed attempt to solve the text shift issue, it should have been removed: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/86df685593587a13a03259835ba70ed4bdd6b29b

rpokorny: > I was part of a failed attempt to solve the text shift issue, it should have been removed: 86df685

Shouldn't the whole block be removed?

gallery/visualtests/nxTabs.js
rpokorny: > b9ae5a5

Pretty sure that needs to be `browser.eyesRegionSnapshot`

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > > I'll bet this isn't what we want:
> 
> My thinking is that when an `NxTabs` is an only child of `.nx-modal-content`, it might get `flex-shrink: 1` such that it grows only to the size of the container and no higher. Then within, it would use flexbox to manage the height of the tab panel, which would then have `overflow-y: auto`. This should institute a scrollbar within the tabe panel when necessary.

Used `flex` to solve the problem here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/9cb5091bae572c01d746a3e788bbda830dc616af

lib/src/components/NxTab/NxTab.scss
glenhunter: > > I was part of a failed attempt to solve the text shift issue, it should have been removed: [86df685](https://github.com/sonatype/sonatype-react-shared-components/commit/86df685593587a13a03259835ba70ed4bdd6b29b)
> 
> Shouldn't the whole block be removed?

Oh yeah: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/be57bcf2d4e48df8773383a536152955cb3a6072

gallery/visualtests/nxTabs.js
glenhunter: > > [b9ae5a5](https://github.com/sonatype/sonatype-react-shared-components/commit/b9ae5a5e965c2e97e7cfc67c07f1bb9550864ba7)
> 
> Pretty sure that needs to be `browser.eyesRegionSnapshot`

Finger crossed! https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/cc742e19e8a9ba3e202868da89a06f3d763d39fa

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -0,0 +1,34 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-variables';
+
+.nx-tabs {
+  margin: $nx-spacing-md 0 $nx-spacing-l 0;
+}
+
+.nx-modal-content--tabs {
```
rpokorny, line 13: This class should be documented somewhere.  And should it be defined here or in `NxModal.scss`?

gallery/src/components/NxTabs/NxTabsModalExample.tsx
glenhunter: > > Unnecessary class, I don't think this really needs to be a wide modal
> 
> If it isn't then the four tabs can't display - which goes to your comment below and depending on what @dturner32 says I might make this modal normal to show the behavior

Now with flex we don't need to worry as much about the modal width, removed `--wide` here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/e8558e498d966f6300fc18dfd12eb12c04172191

lib/src/components/NxTabs/NxTabs.scss
glenhunter: > This class should be documented somewhere. And should it be defined here or in `NxModal.scss`?

Documented here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/7109584281b08533c26c5c541b17e319dc56a5bc

I moved the modifier over to `nx-modal.scss` mostly because I think that's the most consistent with what we do elsewhere: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/c9b9b82706dd0ab322d28c31d084e138fe7f3e61

gallery/src/components/NxTabs/NxTabsPage.tsx
```diff
@@ -65,6 +71,16 @@ export default function NxTabsPage() {
                   Called with the index of the newly selected tab when the currently selected tab changes.
                 </NxTableCell>
               </NxTableRow>
+              <NxTableRow>
+                <NxTableCell>.nx-modal-content--tabs</NxTableCell>
```
rpokorny, line 75: This is the table of _props_ for `NxTabs`, so it's really not the right place for this class to be documented.  I'd document it on the `NxModal` page (in the classes table), though I think it's alright to still demonstrate it here.

gallery/src/components/NxTabs/NxTabsPage.tsx
```diff
@@ -65,6 +71,16 @@ export default function NxTabsPage() {
...
+                <NxTableCell>Modifier</NxTableCell>
+                <NxTableCell></NxTableCell>
+                <NxTableCell>
+                  A modifier class that must be added to <code className="nx-code">.nx-modal-content</code> when
+                  you want to use tabs within an <code className="nx-code">NxModal</code>. The modifier makes
+                  keeps the tabs "sticky" while allowing the tab content to scroll.
```
rpokorny, line 81:
```suggestion
                  you want to use tabs within an <code className="nx-code">NxModal</code>. The modifier
                  keeps the tabs "sticky" while allowing the tab content to scroll.
```

gallery/src/components/NxTabs/NxTabsPage.tsx
```diff
@@ -65,6 +71,16 @@ export default function NxTabsPage() {
                   Called with the index of the newly selected tab when the currently selected tab changes.
                 </NxTableCell>
               </NxTableRow>
+              <NxTableRow>
+                <NxTableCell>.nx-modal-content--tabs</NxTableCell>
+                <NxTableCell>Modifier</NxTableCell>
+                <NxTableCell></NxTableCell>
+                <NxTableCell>
+                  A modifier class that must be added to <code className="nx-code">.nx-modal-content</code> when
+                  you want to use tabs within an <code className="nx-code">NxModal</code>. The modifier makes
```
rpokorny, line 80:
```suggestion
                  you want to use an <code className="nx-code">NxTabs</code> as the sole contents of the modal body. The modifier makes
```

lib/src/components/NxTabs/NxTabs.scss
glenhunter: > This class should be documented somewhere. And should it be defined here or in `NxModal.scss`?

Documented here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/7109584281b08533c26c5c541b17e319dc56a5bc

And moved here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/c9b9b82706dd0ab322d28c31d084e138fe7f3e61

gallery/src/components/NxTabs/NxTabsPage.tsx
glenhunter: > This is the table of _props_ for `NxTabs`, so it's really not the right place for this class to be documented. I'd document it on the `NxModal` page (in the classes table), though I think it's alright to still demonstrate it here.

Changed here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/07abe89ddbbeb360e4b1b948735fd7eee23402a2 and text tweaked here: https://github.com/sonatype/sonatype-react-shared-components/pull/153/commits/6efdf8bd3af9065bdb7e501109679b1a4651d600

rpokorny (approved):

PR #154, opened 2020-09-24
Update NxThreatBar as NxThreatIndicator - RSC-149

rpokorny:
https://issues.sonatype.org/browse/RSC-149


gallery/src/RawHtmlExample.tsx
```diff
@@ -11,6 +11,6 @@ interface Props {
 }
 
 const RawHtmlExample = ({ html }: Props) =>
-  <div className="gallery-raw-html-example" dangerouslySetInnerHTML={{ __html: html }} />;
+  <div className="gallery-example-live gallery-example-live--raw-html" dangerouslySetInnerHTML={{ __html: html }} />;
```
rpokorny, line 14: This and similar changes were cherry-picked in from the checkered-background improvement day branch, in order to get the `.gallery-example-live` class on _all_ live examples (not just HTML examples).  This helps with the visual test selectors.

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -101,12 +101,12 @@ $nx-border-row-disabled: 1px solid $nx-grey-1;
...
-$nx-threat-color-critical: $nx-dark-red;
+$nx-threat-color-unspecified: #d1d6f0;
+$nx-threat-color-none: #c5deff;
+$nx-threat-color-low: #045dd1;
+$nx-threat-color-moderate: #ffbf27;
+$nx-threat-color-severe: #ff8100;
+$nx-threat-color-critical: #cd0028;
 
 // Threat colors - light variant
 $nx-threat-color-none-light: #d9ecfb;
```
rpokorny, line 112: Do these need to be updated too?  I think they're only used by the NxPolicyThreatSlider, so I guess if there's a ticket to update that component, these variables would be adjusted there.

gallery/src/styles/NxThreatNumber/NxThreatNumberPage.tsx
```diff
@@ -8,38 +8,38 @@ import React from 'react';
...
+    <GalleryExampleTile title="Basic Example"
+                        id="nx-threat-number-basic-example"
+                        liveExample={NxThreatNumberBasicExample}
+                        codeExamples={nxThreatNumberBasicExampleCode}>
+      An example of <code className="nx-code">.nx-threat-number</code> used
+      adjacent to an <code className="nx-code">NxThreatIndicator</code> in a simple inline layout. This how how the
+      two would be used together within a table cell, for instance.
+    </GalleryExampleTile>
+
     <GalleryExampleTile title="List Example"
```
rpokorny, line 35: I'm questioning whether we still even need `.nx-threat-number`.  In this example, it actually doesn't do anything since the list text already has the same styles.  In tables, there's an `nx-cell--label` class that also can be used to apply those same styles.  So I'm wondering if we should get rid of `.nx-threat-number` and just rely on those classes in tables and lists. @glenhunter ?

glenhunter: > I'm questioning whether we still even need `.nx-threat-number`. In this example, it actually doesn't do anything since the list text already has the same styles. In tables, there's an `nx-cell--label` class that also can be used to apply those same styles. So I'm wondering if we should get rid of `.nx-threat-number` and just rely on those classes in tables and lists. @glenhunter ?

I think we should probably keep the class, but I don't see any reason that it shouldn't be folded into this example as an option.

lib/src/scss-shared/_nx-colors.scss
glenhunter: > Do these need to be updated too? I think they're only used by the NxPolicyThreatSlider, so I guess if there's a ticket to update that component, these variables would be adjusted there.

It would be nice to see them consolidated if possible, but I think that's a task for another ticket

glenhunter:
Why did you delete the table example? The lists and tables are the primary places the indicators are used so it would be nice to see that they lay out correctly in both places.

rpokorny:
> Why did you delete the table example? The lists and tables are the primary places the indicators are used so it would be nice to see that they lay out correctly in both places.

My thinking was that since it no longer requires any special considerations for being in a table cell, there's nothing special to show there - it's just one of the countless containers that the inline example could be within.  Since it is one that we specifically intend though, I guess I'll add it back

rpokorny:
> > Why did you delete the table example? The lists and tables are the primary places the indicators are used so it would be nice to see that they lay out correctly in both places.
>
> My thinking was that since it no longer requires any special considerations for being in a table cell, there's nothing special to show there - it's just one of the countless containers that the inline example could be within. Since it is one that we specifically intend though, I guess I'll add it back

2b0eddc

glenhunter (approved):

PR #157, opened 2020-09-25
nx-list notes from testing RSC-178 

glenhunter:
https://issues.sonatype.org/browse/RSC-178

Issues addressed:

- nx-list--bulleted needs to use padding for its left spacing rather than margin. 
- nx-list--definition-list is still used on the "Policy Threat Level Utils" and "font-size and line-height"
- Truncated text can go right up to the edge of the action buttons; we probably want some margin there  
- Another issue: the chevron is not vertically centred on lists with subtext

The issue with the spacing between the bullet and the text in Mozilla proved more difficult and in the end I halved the negative margin which made the text acceptable in all browsers but which no longer matches the mockup 100%


gallery/src/guidelines/FontSize/FontSizePage.tsx
```diff
@@ -45,21 +45,23 @@ const FontSizePage = () =>
...
-        </dd>
-        <dt className="nx-list__item nx-list__item--label">Descender</dt>
-        <dd className="nx-list__item">
-          Many glyphs, such as <q>q</q>, contain parts that drop below the baseline. Fonts and line layouts
-          must take this into account
-        </dd>
-      </dl>
+      <ul className="nx-list">
+        <li className="nx-list__item">
+          <span className="nx-list__text"><strong>Glyph:</strong> The visible shape of a character.</span></li>
```
rpokorny, line 50: I feel like this looks pretty weird.  I wonder how it would be with the term as the `nx-list__text` and the definition as the subtext.

rpokorny: > I wonder how it would be with the term as the nx-list__text and the definition as the subtext.

I think I like it that way. 
image

gallery/src/jsUtilPages/PolicyThreatLevelUtils/PolicyThreatLevelUtilsPage.tsx
```diff
@@ -21,27 +21,25 @@ const PolicyThreatLevelUtilsPage = () =>
         objects and TypeScript types that provide code definitions and mappings between these two concepts.
       </p>
 
-      <dl className="nx-list nx-list--definition-list">
-        <dt className="nx-list__item nx-list__item--label">Policy Threat Level</dt>
-        <dd className="nx-list__item">
-          An integer between 0 and 10, inclusive, that describes the severity of a threat.  Zero represents no
+      <ul className="nx-list nx-list--bulleted">
+        <li className="nx-list__item"><strong>Policy Threat Level:</strong>
+          {' '}An integer between 0 and 10, inclusive, that describes the severity of a threat.  Zero represents no
```
rpokorny, line 26:
```suggestion
        <li className="nx-list__item">
          <strong>Policy Threat Level:</strong> An integer between 0 and 10, inclusive, that describes
          the severity of a threat.  Zero represents no
```

lib/src/base-styles/_nx-lists.scss
```diff
@@ -180,10 +181,9 @@
     > .nx-chevron {
       align-self: center;
       color: #2c2c2c;
-      -ms-grid-row: 1;
+      -ms-grid-row: 1 / 2;
       -ms-grid-column: 2;
-      grid-area: actions;
-
+      grid-row: actions / blank;
```
rpokorny, line 186: Is `blank` still a good name for that area?

rpokorny:
> The issue with the spacing between the bullet and the text in Mozilla proved more difficult and in the end I halved the negative margin which made the text acceptable in all browsers but which no longer matches the mockup 100%

That seems like it's probably the best option to me.  I guess if you make it perfect in Chrome, they're really too close in FF, and if you make it perfect in FF, they're noticeably too far in Chrome.

lib/src/base-styles/_nx-lists.scss
```diff
@@ -103,13 +104,13 @@
 
     .nx-list__text {
       display: inline-block;
-      margin-left: -6px;
+      margin-left: -3px;
```
rpokorny, line 107: This should probably have a comment explaining why it's not to spec and that browsers are inconsistent about the bullet spacing.

gallery/src/guidelines/FontSize/FontSizePage.tsx
glenhunter: > > I wonder how it would be with the term as the nx-list__text and the definition as the subtext.
> 
> I think I like it that way.
> image

https://github.com/sonatype/sonatype-react-shared-components/pull/157/commits/45fe2e30628b86fdcdccae1269e4827b4d934d22

lib/src/base-styles/_nx-lists.scss
glenhunter: > This should probably have a comment explaining why it's not to spec and that browsers are inconsistent about the bullet spacing.

Sure: https://github.com/sonatype/sonatype-react-shared-components/pull/157/commits/c4276f520a49e4a58ffc96fb24256663d00fdc67

lib/src/base-styles/_nx-lists.scss
glenhunter: > Is `blank` still a good name for that area?

The way that it's working now we really don't need to specify that grid cell as unique, we can just call it `actions` so the second column is a single area: https://github.com/sonatype/sonatype-react-shared-components/pull/157/commits/07382b2985964bed8a748eaecabd24565e7ddb75

gallery/src/guidelines/FontSize/FontSizePage.tsx
rpokorny: > 45fe2e3

I would've omitted the colons, but it's not a huge deal.

rpokorny:
Oh one more issue, that's only a thing now that RSC-285 is merged.  I assume this should be centered (the sidebar is hidden in order to get enough width):

image

glenhunter:
> Oh one more issue, that's only a thing now that RSC-285 is merged. I assume this should be centered (the sidebar is hidden in order to get enough width):
> 
> image

https://github.com/sonatype/sonatype-react-shared-components/pull/157/commits/1a9b94b3094e772108579bdfb8ea8678d0029193

rpokorny (approved):

PR #159, opened 2020-09-28
nx-counter redesign RSC-218

glenhunter:
https://issues.sonatype.org/browse/RSC-218

### DIFF
--- a/lib/src/base-styles/_nx-counter.scss
+++ b/lib/src/base-styles/_nx-counter.scss
@@ -14,7 +14,7 @@
   color: $nx-text-color-dark;
   display: inline-block;
   font-size: 14px;
-  line-height: 1.5;
+  line-height: $nx-line-height;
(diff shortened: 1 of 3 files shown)


rpokorny:
Something about the rounding looks noticeably janky on my not-so-high-DPI monitor: 
image

rpokorny (approved):
I don't have a suggestion to improve the smoothness issue I pointed out.  It seems better in firefox than in chrome.  No code problems though, so +1

lib/src/base-styles/_nx-counter.scss
```diff
@@ -14,6 +14,7 @@
   color: $nx-text-color-dark;
   display: inline-block;
   font-size: 14px;
+  line-height: 1.5;
```
rpokorny, line 17:
```suggestion
  line-height: $nx-line-height;
```

PR #162, opened 2020-10-02
Fix page description line height RSC-277 

glenhunter:
While working on this I noticed that the page title itself had a custom font-size that it shouldn't have so I changed that as well.

Before:
Screen Shot 2020-10-02 at 4 06 04 PM
After:
Screen Shot 2020-10-02 at 4 05 02 PM

### DIFF
--- a/lib/src/base-styles/_nx-page.scss
+++ b/lib/src/base-styles/_nx-page.scss
@@ -22,7 +22,6 @@
 
 .nx-page-title__page-icon {
   display: inline-block;
-  font-size: 32px;
   vertical-align: middle;
 }
 
--- a/lib/src/base-styles/_nx-page.scss
+++ b/lib/src/base-styles/_nx-page.scss
@@ -22,7 +22,7 @@
 
 .nx-page-title__page-icon {
   display: inline-block;
-  font-size: 26px;
+  font-size: 32px;
   vertical-align: middle;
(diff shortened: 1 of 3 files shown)


rpokorny:
`.nx-page-title__page-icon` font size needs to be adjusted as well.  That's why the circle looks misaligned

glenhunter:
> `.nx-page-title__page-icon` font size needs to be adjusted as well. That's why the circle looks misaligned

Ahh right: https://github.com/sonatype/sonatype-react-shared-components/pull/162/commits/73e8acaef36cad3818fbef73090c508700aadf3f

lib/src/base-styles/_nx-page.scss
```diff
@@ -23,13 +22,12 @@
 
 .nx-page-title__page-icon {
   display: inline-block;
-  font-size: 26px;
+  font-size: 32px;
```
rpokorny, line 25: Couldn't this just be removed?

glenhunter: > Couldn't this just be removed?

Yes it can: https://github.com/sonatype/sonatype-react-shared-components/pull/162/commits/83899fab8464f5dabce9f2c20e57244a11366456

rpokorny (approved):

PR #163, opened 2020-10-06
nxTextInput text length/focus bug RSC-273

glenhunter:
https://issues.sonatype.org/browse/RSC-273

I could not reproduce the issue with the error text not wrapping, but with Ross' help we fixed the focus issue.


rpokorny (approved):
I'm the one that wrote these changes so idk how valid my review is, but +1

Blacktiger (approved):

PR #164, opened 2020-10-06
Inline form example RSC-292 

glenhunter:
https://issues.sonatype.org/browse/RSC-292

Screen Shot 2020-10-06 at 12 56 21 PM

### DIFF
--- a/gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
+++ b/gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
@@ -5,9 +5,11 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React, { FormEvent } from 'react';
+import {faPlus, faMinus} from '@fortawesome/free-solid-svg-icons';
 
 import { NxButton } from '@sonatype/react-shared-components';
 import { NxStatefulTextInput } from '@sonatype/react-shared-components';
+import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
 
 export default function NxFormLayoutExample() {
   function validator(val: string) {
@@ -21,12 +23,29 @@ export default function NxFormLayoutExample() {
 
   return (
     <form className="nx-form" onSubmit={onSubmit}>
-      <div className="nx-form-group nx-form-group--inline">
-        <label className="nx-label">
-          <span className="nx-label__text">Label</span>
-          <NxStatefulTextInput validator={validator}/>
+      <div className="nx-form-row--inline-form">
+        <div className="nx-form-group">
+          <label className="nx-label">
+            <span className="nx-label__text">Label</span>
+            <NxStatefulTextInput validator={validator}/>
+          </label>
+        </div>
+        <div className="nx-form-group nx-form-group--button-only">
           <NxButton variant="primary" type="submit">Submit</NxButton>
-        </label>
+        </div>
+      </div>
+      <div className="nx-form-row--inline-form">
+        <div className="nx-form-group">
+          <label className="nx-label">
+            <span className="nx-label__text">Label</span>
+            <span className="nx-sub-label">This is a sub-label.</span>
+            <NxStatefulTextInput validator={validator}/>
+          </label>
+        </div>
+        <div className="nx-form-group nx-form-group--button-only">
+          <NxButton><NxFontAwesomeIcon icon={faPlus}/></NxButton>
+          <NxButton><NxFontAwesomeIcon icon={faMinus}/></NxButton>
+        </div>
       </div>
     </form>
   );
--- a/gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
+++ b/gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { FormEvent } from 'react';
+
+import { NxButton } from '@sonatype/react-shared-components';
+import { NxStatefulTextInput } from '@sonatype/react-shared-components';
+
+export default function NxFormLayoutExample() {
+  function validator(val: string) {
+    return val.length ? null : 'Must be non-empty';
+  }
+
+  function onSubmit(evt: FormEvent) {
+    evt.preventDefault();
+    alert('Submitted!');
+  }
+
+  return (
+    <form className="nx-form" onSubmit={onSubmit}>
+      <div className="nx-form-group nx-form-group--inline">
+        <label className="nx-label">
+          <span className="nx-label__text">Label</span>
+          <NxStatefulTextInput validator={validator}/>
+          <NxButton variant="primary" type="submit">Submit</NxButton>
+        </label>
+      </div>
+    </form>
+  );
+}
--- a/gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
+++ b/gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
@@ -23,18 +23,18 @@ export default function NxFormLayoutExample() {
 
   return (
     <form className="nx-form" onSubmit={onSubmit}>
-      <div className="nx-form-row--inline-form">
+      <div className="nx-form-row">
         <div className="nx-form-group">
           <label className="nx-label">
             <span className="nx-label__text">Label</span>
             <NxStatefulTextInput validator={validator}/>
           </label>
         </div>
-        <div className="nx-form-group nx-form-group--button-only">
+        <div className="nx-btn-bar">
           <NxButton variant="primary" type="submit">Submit</NxButton>
         </div>
       </div>
-      <div className="nx-form-row--inline-form">
+      <div className="nx-form-row">
         <div className="nx-form-group">
           <label className="nx-label">
             <span className="nx-label__text">Label</span>
@@ -42,7 +42,7 @@ export default function NxFormLayoutExample() {
             <NxStatefulTextInput validator={validator}/>
           </label>
         </div>
-        <div className="nx-form-group nx-form-group--button-only">
+        <div className="nx-btn-bar">
           <NxButton><NxFontAwesomeIcon icon={faPlus}/></NxButton>
           <NxButton><NxFontAwesomeIcon icon={faMinus}/></NxButton>
         </div>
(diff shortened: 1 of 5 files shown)


gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
```diff
@@ -0,0 +1,33 @@
...
+    alert('Submitted!');
+  }
+
+  return (
+    <form className="nx-form" onSubmit={onSubmit}>
+      <div className="nx-form-group nx-form-group--inline">
+        <label className="nx-label">
+          <span className="nx-label__text">Label</span>
+          <NxStatefulTextInput validator={validator}/>
+          <NxButton variant="primary" type="submit">Submit</NxButton>
```
rpokorny, line 28: You have the button inside the same `<label>` as the input?  Is that allowed/does it make sense?  Reading the standard, I don't think it is.

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
```diff
@@ -10,9 +10,13 @@ import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-compon
 
 import NxFormLayoutExample from './NxFormLayoutExample';
 import NxFormHorizontalLayoutExample from './NxFormHorizontalLayoutExample';
+import NxFormInlineLayoutExample from './NxFormInlineLayoutExample';
 
 const NxFormLayoutCode = require('!!raw-loader!./NxFormLayoutExample').default;
 const NxFormHorizontalLayoutCode = require('!!raw-loader!./NxFormHorizontalLayoutExample').default;
+const NxFormInlineLayoutCode = require('!!raw-loader!./NxFormInlineLayoutExample').default;
+
+
```
rpokorny, line 19: Too many blank lines

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,18 @@
   margin: 0 0 $nx-spacing-xl 0;
   padding: 0;
 }
+
+// Inline form group
+.nx-form-group--inline {
```
rpokorny, line 86: Per my other comment, I don't think putting in in the form group makes sense. But if it did, I'd ask whether the modifier is really necessary as this is the only case we have so far where multiple things are in the same form group.

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
glenhunter: > Too many blank lines

https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/43f39c241667c908c92b52989983f40248909845

gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
glenhunter: > You have the button inside the same `<label>` as the input? Is that allowed/does it make sense? Reading the standard, I don't think it is.

Yeah, that wasn't allowed at all, I replaced it with a `flex box` based solution here and added an example that shows it working with a sub-title: https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/43f39c241667c908c92b52989983f40248909845

lib/src/base-styles/_nx-forms.scss
glenhunter: > Per my other comment, I don't think putting in in the form group makes sense. But if it did, I'd ask whether the modifier is really necessary as this is the only case we have so far where multiple things are in the same form group.

https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/43f39c241667c908c92b52989983f40248909845

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,20 @@
   margin: 0 0 $nx-spacing-xl 0;
   padding: 0;
 }
+
+// Inline form group
+.nx-form-row--inline-form {
```
rpokorny, line 86: Could flex just be applied to `.nx-form-row` and the modifier class be avoided.  I tried this with the other row example and it didn't hurt, I think it actually improved it slightly by removing a height discrepancy between the form groups and the row.  If you're worried about wrapping you could always add `flex-wrap`

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,20 @@
...
+  display: flex;
+
+  .nx-form-group--button-only {
+    @include container-horizontal;
+
+    align-items: flex-end;
+    display: flex;
+    flex-shrink: 1;
```
rpokorny, line 94: What's this for?

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,20 @@
...
 }
+
+// Inline form group
+.nx-form-row--inline-form {
+  display: flex;
+
+  .nx-form-group--button-only {
+    @include container-horizontal;
+
+    align-items: flex-end;
```
rpokorny, line 92: If this property is moved to the form row, the form group doesn't have to be a flex container.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,20 @@
   margin: 0 0 $nx-spacing-xl 0;
   padding: 0;
 }
+
+// Inline form group
+.nx-form-row--inline-form {
+  display: flex;
+
+  .nx-form-group--button-only {
```
rpokorny, line 89: What would you think of just using `nx-btn-bar` here instead of creating a new class?

gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  function onSubmit(evt: FormEvent) {
+    evt.preventDefault();
+    alert('Submitted!');
+  }
+
+  return (
+    <form className="nx-form" onSubmit={onSubmit}>
+      <div className="nx-form-row--inline-form">
```
rpokorny, line 26: Missing nx-form-row itself

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,20 @@
...
+    @include container-horizontal;
+
+    align-items: flex-end;
+    display: flex;
+    flex-shrink: 1;
+    margin-left: $nx-spacing-l;
```
rpokorny, line 95: If the row had `.nx-form-row` itself (and not only the modifier) the text input form group would have right margin making this unnecessary

lib/src/base-styles/_nx-forms.scss
```diff
@@ -81,3 +81,20 @@
...
+    // bottom padding to compensate for the form field's validation message
+    padding-bottom: $nx-spacing-l;
```
rpokorny, line 97: Any particular reason this is padding instead of margin?

glenhunter: > Any particular reason this is padding instead of margin?

Padding kept the adjacent boxes the same height

glenhunter: > > Any particular reason this is padding instead of margin?
> 
> Padding kept the adjacent boxes the same height

But with the group being replaced by `nx-btn-bar` margin makes more sense: https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

lib/src/base-styles/_nx-forms.scss
glenhunter: > If the row had `.nx-form-row` itself (and not only the modifier) the text input form group would have right margin making this unnecessary

https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
glenhunter: > Missing nx-form-row itself

https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

lib/src/base-styles/_nx-forms.scss
glenhunter: > What would you think of just using `nx-btn-bar` here instead of creating a new class?

That's much better: https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

lib/src/base-styles/_nx-forms.scss
glenhunter: > If this property is moved to the form row, the form group doesn't have to be a flex container.

https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

lib/src/base-styles/_nx-forms.scss
glenhunter: > What's this for?

It's gone now: https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

lib/src/base-styles/_nx-forms.scss
glenhunter: > Could flex just be applied to `.nx-form-row` and the modifier class be avoided. I tried this with the other row example and it didn't hurt, I think it actually improved it slightly by removing a height discrepancy between the form groups and the row. If you're worried about wrapping you could always add `flex-wrap`

It can be and it's a much more elegant solution because it doesn't require the use of any custom classes when it's used with `nx-btn-bar`: https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/768f86743bae3a6b994fe0f05ae2f3ea359b6d52

lib/src/base-styles/_nx-forms.scss
```diff
@@ -19,13 +19,18 @@
...
   margin-bottom: $nx-spacing-xs;
 
   .nx-form-group {
-    display: inline-block;
     margin-right: $nx-spacing-l;
     margin-bottom: 0;
   }
+
+  .nx-btn-bar {
+    margin-bottom: $nx-spacing-l;
```
rpokorny, line 32: `.nx-btn-bar` comes with a top margin which you'll probably want to clear out here.  It doesn't make a difference in your examples, but I imagine it would if the adjacent text input didn't have any label text.

rpokorny (approved):
+1 aside from latest comment

lib/src/base-styles/_nx-forms.scss
glenhunter: > `.nx-btn-bar` comes with a top margin which you'll probably want to clear out here. It doesn't make a difference in your examples, but I imagine it would if the adjacent text input didn't have any label text.

https://github.com/sonatype/sonatype-react-shared-components/pull/164/commits/a0b67d2c8c575c77fab8731458c7017943d4e0ec

PR #170, opened 2020-10-14
NxCard RSC-222

glenhunter:
https://issues.sonatype.org/browse/RSC-222

https://app.zeplin.io/project/5ed14ea1b07f00b42032c55f/screen/5f7aa43c5c6b5892940a4818/

Screen Shot 2020-10-14 at 1 38 18 PM
Screen Shot 2020-10-14 at 1 38 04 PM

### DIFF
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -12,7 +12,7 @@ import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
 export default function NxCardVerticalExample() {
   return (
     <div className="nx-card-container nx-card-container--column">
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3">Header</h3>
         </header>
@@ -26,7 +26,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Link</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3">Card header</h3>
         </header>
@@ -40,7 +40,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Link</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3">Card header</h3>
         </header>
@@ -54,7 +54,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Link</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
         </header>
@@ -68,7 +68,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Link</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3 nx-h3--card-header">Chiba advert hacker hotdog shoes voodoo god 3D-printed</h3>
         </header>
@@ -83,7 +83,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Artisanal denim</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3 nx-h3--card-header">Render-farm dolphin beef noodles</h3>
         </header>
@@ -97,7 +97,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Film sprawl wristwatch math-engine</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3 nx-h3--card-header">Weathered plastic saturation point realism</h3>
         </header>
@@ -111,7 +111,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">Jabberwocky</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3 nx-h3--card-header">Augmented reality grenade</h3>
         </header>
@@ -126,7 +126,7 @@ export default function NxCardVerticalExample() {
           <a href="#" className="nx-text-link">uplink numino</a>
         </footer>
       </div>
-      <div className="nx-card nx-card--horizontal">
+      <div className="nx-card">
         <header className="nx-card-header">
           <h3 className="nx-h3 nx-h3--card-header">
             Fluidity futurity order-flow city neon pre-DIY wristwatch-space bicycle digital geodesic realism sensory
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -14,37 +14,85 @@ export default function NxCardVerticalExample() {
     <div className="nx-card-container nx-card-container--column">
       <div className="nx-card nx-card--horizontal">
         <header className="nx-card-header">
-          <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
+          <h3 className="nx-h3">Header</h3>
         </header>
-        <div className="nx-card__call-out">
-          <span className="nx-call__call-out-text">XXXXX%</span>
-        </div>
-        <div className="nx-card__text">
-          Descriptive text
+        <div className="nx-card-content">
+          <div className="nx-card__call-out">
+            <span className="nx-call__call-out-text">42</span>
+          </div>
+          <div className="nx-card__text">
+            Text
+          </div>
         </div>
         <footer className="nx-card-footer"><a href="nx-link">Link</a></footer>
       </div>
       <div className="nx-card nx-card--horizontal">
         <header className="nx-card-header">
-          <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
+          <h3 className="nx-h3">Card header</h3>
         </header>
-        <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes}/>
-        </div>
-        <div className="nx-card__text">
-          Descriptive text
+        <div className="nx-card-content">
+          <div className="nx-card__call-out">
+            <NxFontAwesomeIcon icon={faShapes}/>
+          </div>
+          <div className="nx-card__text">
+            Data point details
+          </div>
         </div>
         <footer className="nx-card-footer"><a href="nx-link">Link</a></footer>
       </div>
       <div className="nx-card nx-card--horizontal">
         <header className="nx-card-header">
-          <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
+          <h3 className="nx-h3">Card header</h3>
         </header>
-        <div className="nx-card__call-out">
-          <span className="nx-call__call-out-text">XXXXX%</span>
+        <div className="nx-card-content">
+          <div className="nx-card__call-out">
+            <span className="nx-call__call-out-text">XX%</span>
+          </div>
+          <div className="nx-card__text">
+            Descriptive text
+          </div>
         </div>
-        <div className="nx-card__text">
-          Descriptive text
+        <footer className="nx-card-footer"><a href="nx-link">Link</a></footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3">Card header</h3>
+        </header>
+        <div className="nx-card-content">
+          <div className="nx-card__call-out">
+            <span className="nx-call__call-out-text">XXX%</span>
+          </div>
+          <div className="nx-card__text">
+            Descriptive text
+          </div>
+        </div>
+        <footer className="nx-card-footer"><a href="nx-link">Link</a></footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3">Card header</h3>
+        </header>
+        <div className="nx-card-content">
+          <div className="nx-card__call-out">
+            <span className="nx-call__call-out-text">XXXX%</span>
+          </div>
+          <div className="nx-card__text">
+            Descriptive text
+          </div>
+        </div>
+        <footer className="nx-card-footer"><a href="nx-link">Link</a></footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3">Card header</h3>
+        </header>
+        <div className="nx-card-content">
+          <div className="nx-card__call-out">
+            <span className="nx-call__call-out-text">XXXXX%</span>
+          </div>
+          <div className="nx-card__text">
+            Descriptive text
+          </div>
         </div>
         <footer className="nx-card-footer"><a href="nx-link">Link</a></footer>
       </div>
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -13,97 +13,137 @@ export default function NxCardVerticalExample() {
   return (
     <div className="nx-card-container nx-card-container--column">
       <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3">Header</h3>
+        </header>
         <div className="nx-card__call-out">
-          <span className="nx-call__call-out-text">XX%</span>
+          <span className="nx-call__call-out-text">42</span>
         </div>
-        <div className="nx-card-content">
-          <header className="nx-card-header">
-            <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
-          </header>
-          <div className="nx-card__text">
-            Descriptive text
-          </div>
-        </div>
-        <footer className="nx-card-footer"><a href="nx-text-link">Link</a></footer>
-      </div>
-      <div className="nx-card nx-card--horizontal">
-        <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes}/>
-        </div>
-        <div className="nx-card-content">
-          <header className="nx-card-header">
-            <h3 className="nx-h3 nx-h3--card-header">Chiba advert hacker hotdog shoes voodoo god 3D-printed</h3>
-          </header>
-          <div className="nx-card__text">
-            Marketing systemic math- hotdog saturation point nodal point footage disposable singularity order-flow
-            systemic
-          </div>
-        </div>
-        <footer className="nx-card-footer"><a href="nx-text-link">Artisanal denim</a></footer>
-      </div>
-      <div className="nx-card nx-card--horizontal">
-        <div className="nx-card__call-out">
-          <span className="nx-call__call-out-text">XXX%</span>
-        </div>
-        <div className="nx-card-content">
-          <header className="nx-card-header">
-            <h3 className="nx-h3 nx-h3--card-header">Render-farm dolphin beef noodles</h3>
-          </header>
-          <div className="nx-card__text">
-            City advert motion apophenia film skyscraper sentient beef noodles industrial grade motion tanto BASE jump
-            girl realism
-          </div>
-        </div>
-        <footer className="nx-card-footer"><a href="nx-text-link">Film sprawl wristwatch math-engine</a></footer>
-      </div>
-      <div className="nx-card nx-card--horizontal">
-        <div className="nx-card__call-out">
-          <span className="nx-call__call-out-text">XXXX%</span>
-        </div>
-        <div className="nx-card-content">
-          <header className="nx-card-header">
-            <h3 className="nx-h3 nx-h3--card-header">Weathered plastic saturation point realism</h3>
-          </header>
-          <div className="nx-card__text">
-            Construct otaku monofilament skyscraper
-          </div>
-        </div>
-        <footer className="nx-card-footer"><a href="nx-text-link">Jabberwocky</a></footer>
-      </div>
-      <div className="nx-card nx-card--horizontal">
-        <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
-        </div>
-        <div className="nx-card-content">
-          <header className="nx-card-header">
-            <h3 className="nx-h3 nx-h3--card-header">Augmented reality grenade</h3>
-          </header>
-          <div className="nx-card__text">
-            Neural sign shrine corrupted decay fluidity paranoid nodality long-chain hydrocarbons physical shoes 8-bit
-            singularity sensory smart-boat Legba smart-sunglasses spook city
-          </div>
+        <div className="nx-card__text">
+          Text
         </div>
         <footer className="nx-card-footer">
-          <a href="nx-text-link">uplink numinous pen</a>
+          <a href="#" className="nx-text-link">Link</a>
         </footer>
       </div>
       <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3">Card header</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <NxFontAwesomeIcon icon={faShapes} />
+        </div>
+        <div className="nx-card__text">
+          Data point details
+        </div>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">Link</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3">Card header</h3>
+        </header>
+        <div className="nx-card__call-out">
+        <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
+        </div>
+        <div className="nx-card__text">
+          Large icon
+        </div>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">Link</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <span className="nx-call__call-out-text">XX%</span>
+        </div>
+        <div className="nx-card__text">
+          Descriptive text
+        </div>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">Link</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3 nx-h3--card-header">Chiba advert hacker hotdog shoes voodoo god 3D-printed</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <NxFontAwesomeIcon icon={faShapes}/>
+        </div>
+        <div className="nx-card__text">
+          Marketing systemic math- hotdog saturation
+          systemic
+        </div>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">Artisanal denim</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3 nx-h3--card-header">Render-farm dolphin beef noodles</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <span className="nx-call__call-out-text">XXX%</span>
+        </div>
+        <div className="nx-card__text">
+          City advert motion apophenia film skyscraper sentient beef
+        </div>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">Film sprawl wristwatch math-engine</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3 nx-h3--card-header">Weathered plastic saturation point realism</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <span className="nx-call__call-out-text">XXXX%</span>
+        </div>
+        <div className="nx-card__text">
+          Construct otaku monofilament skyscraper
+        </div>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">Jabberwocky</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3 nx-h3--card-header">Augmented reality grenade</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
+        </div>
+        <div className="nx-card__text">
+          Neural sign shrine corrupted decay fluidity paranoid nodality long-chain hydrocarbons physical shoes 8-bit
+          singularity sensory smart-boat Legba smart-sunglasses spook city
+        </div>
+        <footer className="nx-card-footer">
+
+            <a href="#" className="nx-text-link">uplink numino</a>
+        </footer>
+      </div>
+      <div className="nx-card nx-card--horizontal">
+        <header className="nx-card-header">
+          <h3 className="nx-h3 nx-h3--card-header">
+            Fluidity futurity order-flow city neon pre-DIY wristwatch-space bicycle digital geodesic realism sensory
+            dome otaku rain-ware lights
+          </h3>
+        </header>
         <div className="nx-card__call-out">
           <NxFontAwesomeIcon icon={faChartLine} className="nx-icon--large" />
         </div>
-        <div className="nx-card-content">
-          <header className="nx-card-header">
-            <h3 className="nx-h3 nx-h3--card-header">
-              Fluidity futurity order-flow city neon pre-DIY wristwatch-space bicycle digital geodesic realism sensory
-              dome otaku rain-ware lights
-            </h3>
-          </header>
-          <div className="nx-card__text">
-            BASE jump boy corrupted nano-skyscraper disposable sub-orbital marketing apophenia sign military-grade
-            saturation point drugs
-          </div>
+        <div className="nx-card__text">
+          BASE jump boy corrupted nano-skyscraper disposable sub-orbital marketing apophenia sign military-grade
+          saturation point drugs
         </div>
-        <footer className="nx-card-footer"><a href="nx-text-link">digital military-grade j-pop-ware</a></footer>
+        <footer className="nx-card-footer">
+          <a href="#" className="nx-text-link">digital military-grade j-pop-ware</a>
+        </footer>
       </div>
     </div>
   );
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -62,7 +62,7 @@ export default function NxCardHorizontalExample() {
       </section>
       <section className="nx-card">
         <header className="nx-card__header">
-          <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
+          <h3 className="nx-h3">Card header</h3>
         </header>
         <div className="nx-card__call-out">
           <span className="nx-card__call-out-text">XX%</span>
@@ -76,7 +76,7 @@ export default function NxCardHorizontalExample() {
       </section>
       <section className="nx-card">
         <header className="nx-card__header">
-          <h3 className="nx-h3 nx-h3--card-header">Chiba advert hacker hotdog shoes voodoo god 3D-printed</h3>
+          <h3 className="nx-h3">Chiba advert hacker hotdog shoes voodoo god 3D-printed</h3>
         </header>
         <div className="nx-card__content">
           Marketing systemic math- hotdog saturation
@@ -88,7 +88,7 @@ export default function NxCardHorizontalExample() {
       </section>
       <section className="nx-card">
         <header className="nx-card__header">
-          <h3 className="nx-h3 nx-h3--card-header">Render-farm dolphin beef noodles</h3>
+          <h3 className="nx-h3">Render-farm dolphin beef noodles</h3>
         </header>
         <div className="nx-card__call-out">
           <span className="nx-card__call-out-text">XXX%</span>
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -73,7 +73,7 @@ export default function NxCardVerticalExample() {
           <h3 className="nx-h3 nx-h3--card-header">Chiba advert hacker hotdog shoes voodoo god 3D-printed</h3>
         </header>
         <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes}/>
+          <NxFontAwesomeIcon icon={faChartLine}/>
         </div>
         <div className="nx-card__text">
           Marketing systemic math- hotdog saturation
@@ -95,53 +95,6 @@ export default function NxCardVerticalExample() {
         </div>
         <footer className="nx-card-footer">
           <a href="#" className="nx-text-link">Film sprawl wristwatch math-engine</a>
-        </footer>
-      </div>
-      <div className="nx-card">
-        <header className="nx-card-header">
-          <h3 className="nx-h3 nx-h3--card-header">Weathered plastic saturation point realism</h3>
-        </header>
-        <div className="nx-card__call-out">
-          <span className="nx-call__call-out-text">XXXX%</span>
-        </div>
-        <div className="nx-card__text">
-          Construct otaku monofilament skyscraper
-        </div>
-        <footer className="nx-card-footer">
-          <a href="#" className="nx-text-link">Jabberwocky</a>
-        </footer>
-      </div>
-      <div className="nx-card">
-        <header className="nx-card-header">
-          <h3 className="nx-h3 nx-h3--card-header">Augmented reality grenade</h3>
-        </header>
-        <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
-        </div>
-        <div className="nx-card__text">
-          Neural sign shrine corrupted decay fluidity paranoid nodality long-chain hydrocarbons physical shoes 8-bit
-          singularity sensory smart-boat Legba smart-sunglasses spook city
-        </div>
-        <footer className="nx-card-footer">
-          <a href="#" className="nx-text-link">uplink numino</a>
-        </footer>
-      </div>
-      <div className="nx-card">
-        <header className="nx-card-header">
-          <h3 className="nx-h3 nx-h3--card-header">
-            Fluidity futurity order-flow city neon pre-DIY wristwatch-space bicycle digital geodesic realism sensory
-            dome otaku rain-ware lights
-          </h3>
-        </header>
-        <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faChartLine} className="nx-icon--large" />
-        </div>
-        <div className="nx-card__text">
-          BASE jump boy corrupted nano-skyscraper disposable sub-orbital marketing apophenia sign military-grade
-          saturation point drugs
-        </div>
-        <footer className="nx-card-footer">
-          <a href="#" className="nx-text-link">digital military-grade j-pop-ware</a>
         </footer>
       </div>
     </div>
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -45,7 +45,7 @@ export default function NxCardVerticalExample() {
           <h3 className="nx-h3">Card header</h3>
         </header>
         <div className="nx-card__call-out">
-        <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
+          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
         </div>
         <div className="nx-card__text">
           Large icon
@@ -123,8 +123,7 @@ export default function NxCardVerticalExample() {
           singularity sensory smart-boat Legba smart-sunglasses spook city
         </div>
         <footer className="nx-card-footer">
-
-            <a href="#" className="nx-text-link">uplink numino</a>
+          <a href="#" className="nx-text-link">uplink numino</a>
         </footer>
       </div>
       <div className="nx-card nx-card--horizontal">
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -31,14 +31,14 @@ export default function NxCardHorizontalExample() {
           <h3 className="nx-h3">Card header</h3>
         </header>
         <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes} />
+          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--card-call-out" />
         </div>
         <div className="nx-card__content">
           Data point details
         </div>
         <footer className="nx-card__footer">
           <div className="nx-btn-bar">
-            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} className="nx-icon--large" /></NxButton>
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} /></NxButton>
           </div>
         </footer>
       </section>
@@ -47,16 +47,16 @@ export default function NxCardHorizontalExample() {
           <h3 className="nx-h3">Card header</h3>
         </header>
         <div className="nx-card__call-out">
-          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
+          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--card-call-out-xl" />
         </div>
         <div className="nx-card__content">
           Large icon
         </div>
         <footer className="nx-card__footer">
           <div className="nx-btn-bar">
-            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faTrash} className="nx-icon--large" /></NxButton>
-            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} className="nx-icon--large" /></NxButton>
-            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faTrash} className="nx-icon--large" /></NxButton>
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faTrash} /></NxButton>
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} /></NxButton>
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faTrash} /></NxButton>
           </div>
         </footer>
       </section>
--- a/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
+++ b/gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
@@ -37,7 +37,9 @@ export default function NxCardHorizontalExample() {
           Data point details
         </div>
         <footer className="nx-card__footer">
-          <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} className="nx-icon--large" /></NxButton>
+          <div className="nx-btn-bar">
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} className="nx-icon--large" /></NxButton>
+          </div>
         </footer>
       </div>
       <div className="nx-card">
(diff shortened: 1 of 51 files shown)


gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,102 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { faShapes, faChartLine } from '@fortawesome/free-solid-svg-icons';
+
+import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
+
+export default function NxCardVerticalExample() {
```
rpokorny, line 12:
```suggestion
export default function NxCardHorizontalExample() {
```

rpokorny:
I find the horizontal vs vertical naming scheme confusing.  I'm never sure whether it's referring to the layout within the cards or the layout of the cards relative to each other - with opposite implications of which is which.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return (
+    <div className="nx-card-container nx-card-container--column">
+      <div className="nx-card">
+        <header className="nx-card-header">
```
rpokorny, line 16: If this should be a `header` then should the `.nx-card` be a `section`?  Or should it all just be divs?

glenhunter: > If this should be a `header` then should the `.nx-card` be a `section`? Or should it all just be divs?

When I was writing it I hesitated over whether `<div className="nx-card-container nx-card-container--column">` should be a `<section>`. I tend to think that the cards themselves are fine as `<div>`'s

glenhunter:
> I find the horizontal vs vertical naming scheme confusing. I'm never sure whether it's referring to the layout within the cards or the layout of the cards relative to each other - with opposite implications of which is which.

I found the same thing, also with `row` and `column` for the container. I'm open to suggestions.

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card__call-out</code></td>
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">
+              Wraps the card call out (if any). The call out typically consists of a number, sparkline, icon, or
+              other small graphic. It should not be free form text.
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-call__call-out-text</code></td>
```
rpokorny, line 84: Could this be a modifier of `.nx-card__call-out` instead?

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">
+              In the case where the call out consists of text (e.g. 99%) it should be wrapped in this class (typically
+              in a <code className="nx-code">&lt;span&gt;</code>).
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card__text</code></td>
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">Free-form text. Should be concise.</td>
```
rpokorny, line 94: Should we be explicit about the "two lines or less" rule?

rpokorny:
If I add more text to an alt vertical card, what should happen?  Right now it wraps the text down to the next row, becoming kind of like a normal vertical card, except that the call out gets right-aligned:
image

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <td className="nx-cell"><code className="nx-code">.nx-card--vertical-alt</code></td>
+            <td className="nx-cell">Modifier of <code className="nx-code">.nx-card</code></td>
+            <td className="nx-cell">
+              In order to use the variant vertical layout this class should be applied to
+              {' '}<code className="nx-code">.nx-card</code> in conjunction with a
+              {' '}<code className="nx-code">&lt;div&gt;</code> wrapping the call out and description text.
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card-content</code></td>
```
rpokorny, line 111: So this cannot be used in a card that may covert to horizontal layout, correct?  Would I be correct in thinking that if `display: contents` worked in all supported browsers, we could avoid that restriction?

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">Free-form text. Should be concise.</td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card-footer</code></td>
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">Card footer. Typically provides a link to more information.</td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card--vertical-alt</code></td>
```
rpokorny, line 102: I notice that this class seems to be safe to use on horizontal cards - it has no effect on them.  Is that a deliberate fact that we want to support or is it just coincidence? I think it would be more convenient for users if it was supported.

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <td className="nx-cell"><code className="nx-code">.nx-card-container--row</code></td>
+            <td className="nx-cell">Modifier of <code className="nx-code">.nx-card-container</code></td>
+            <td className="nx-cell">
+              Container for vertically orientated cards (normal and variant). Normal vertical cards can be transformed
+              into horizontal cards by replacing this style with
+              {' '}<code className="nx-code">.nx-card-container--column</code> below.
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card-container--column</code></td>
```
rpokorny, line 54: Which layout is the default?  Or is the intent that users _must_ specify one modifier or the other?

rpokorny:
I don't see any examples of callout-only cards as shown here in zeplin: 
image

rpokorny:
@dturner32 how are callout-only cards supposed to look in the horizontal layout?

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+// Vertically oriented cards appear in a row container
+.nx-card-container--row {
+  @include container-horizontal;
+
+  .nx-card {
+    display: flex;
+    flex-direction: column;
+    -ms-flex-pack: justify;
+    justify-content: space-between;
+    margin: 0 $nx-spacing-md 0 0;
```
rpokorny, line 45: This margin could cause problems if someone wants something like rows of 3 cards in a container that is exactly wide enough for three cards plus their interior margins.  I'd consider using `column-gap` here instead of margin.

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+    text-align: center;
+    min-width: 230px;
+    max-width: 330px;
```
rpokorny, line 48: Interesting behavior with the dynamic width.  When using small words, I can't get the cards to expand beyond 286.17px wide.  It appears that they prefer to expand beyond min-width rather than wrapping text lines, up to that point.  After that they start wrapping the text and don't expand their overall width any further.  The only way I can get them wider is to include an individual word which is very long - and even then, it doesn't start to increase the width until well _after_ the word is already too wide for the container.

rpokorny:
I don't see a mockup of what should happen to the callout and text in a normal vertical card when that card's height is stretched due to an adjacent card with more content.  ie what should be happening within the circled area below. @dturner32 

image

lib/src/base-styles/_nx-card.scss
rpokorny: > Interesting behavior with the dynamic width. When using small words, I can't get the cards to expand beyond 286.17px wide. It appears that they prefer to expand beyond min-width rather than wrapping text lines, up to that point. After that they start wrapping the text and don't expand their overall width any further. The only way I can get them wider is to include an individual word which is very long - and even then, it doesn't start to increase the width until well after the word is already too wide for the container.

Coming back to this today, I see that the card in question, with the hard-coded example content (and not my custom content from messing with it) is now 301px wide :man_shrugging:

rpokorny:
Missing example of vertical card without callout:
image

And vertical cards without links:
image

And horizontal cards without links.

And horizontal cards without callouts, which aren't even in the mockups, but since the cards are supposed to be convertible from one style to the other I assume we need to support that.

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+
+  .nx-card__call-out {
+    margin-bottom: $nx-spacing-xs;
+  }
+
+  .nx-card__text {
+    margin-bottom: $nx-spacing-l;
+  }
+
+  .nx-card-footer {
```
rpokorny, line 64: Is this element necessary or could these styles just be applied directly to the `.nx-text-link`.  That's the only thing that's ever supposed to be in the footer, right?

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+    padding-top: $nx-spacing-l;
+    width: 100%;
+  }
+}
+
+// Horizontally orientated cards appear in a column
+.nx-card-container--column {
+  @include container-vertical;
+
+  flex-direction: column;
```
rpokorny, line 75: I'm wondering if this container variant should just use block layout.  That would be simplest and would fix the fact that the horizontal cards have double the intended margin between them.

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+    -ms-grid-columns: auto 1fr auto;
+    grid-template-columns: auto 1fr auto;
+    -ms-grid-rows: auto auto;
+    grid-template-areas:
+      "card-callout card-title card-footer"
+      "card-callout card-text card-footer";
+
+    margin: $nx-spacing-md 0;
+    text-align: left;
+    width: 100%;
```
rpokorny, line 89: Probably redundant, and would definitely be redundant if the container used block layout.

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+  .nx-card-header {
+    -ms-grid-row: 1;
+    -ms-grid-column: 2;
+    grid-area: card-title;
+
+    @include container-vertical;
+    border-right: 1px solid #d1d6f0;
+    border-left: 1px solid #d1d6f0;
+    margin: 0;
+    padding: 0 $nx-spacing-xxl $nx-spacing-md $nx-spacing-xxl;
```
rpokorny, line 101: Why padding and not margin?

rpokorny: > Why padding and not margin?

Oh because of the borders.  I think the borders need to be moved to the callout and the footer anyway so that they don't appear when those elements aren't present.

rpokorny:
Need example with buttons in footer: 
image

lib/src/base-styles/_nx-card.scss
rpokorny: > Is this element necessary or could these styles just be applied directly to the .nx-text-link. That's the only thing that's ever supposed to be in the footer, right?

Nevermind I see there's mockups where it has buttons instead.

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+    -ms-grid-column: 3;
+    grid-area: card-footer;
+    -ms-grid-column-align: start;
+    justify-self: start;
+
+    border: none;
+    margin: 0;
+    padding: 0 0 0 $nx-spacing-xxl;
+    // necessary to make the link wrap
+    max-width: 150px;
```
rpokorny, line 143: There's a mockup with two icon buttons in the footer, which brings it to 136px.  I'm wondering what should happen when there are more than two buttons.  I'm assuming you came up with 150px here arbitrarily; when does @dturner32 think buttons or link text should wrap?

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+      align-items: center;
+      display: flex;
+      -ms-flex-wrap: wrap;
+      flex-wrap: wrap;
+      column-gap: 24px;
+    }
+
+    .nx-card__call-out {
+      @include container-horizontal;
+      display: inline-block;
```
rpokorny, line 160: This appears to have no effect.  Even the Computed tab in the dev tools still shows `block`

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,173 @@
...
+      flex-grow: 1;
+      margin-bottom: $nx-spacing-md;
+      text-align: right;
+    }
+
+    .nx-card__text {
+      display: inline-block;
+      flex-grow: 1;
+      margin-bottom: $nx-spacing-md;
+      text-align: left;
```
rpokorny, line 170: This poses a problem when it automatically switches to normal vertical layout.  At that point I assume the text should be center aligned.  I doubt there's a way to distinguish that in CSS though.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > So this cannot be used in a card that may covert to horizontal layout, correct? Would I be correct in thinking that if `display: contents` worked in all supported browsers, we could avoid that restriction?

I didn't play with `display: contents` enough to give a very definitive answer, but I believe so yes. If it weren't for IE I would have explored this further.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > I notice that this class seems to be safe to use on horizontal cards - it has no effect on them. Is that a deliberate fact that we want to support or is it just coincidence? I think it would be more convenient for users if it was supported.

It doesn't effect horizontal cards because it only has an effect when the card is a child of `.nx-card-container--row`.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Which layout is the default? Or is the intent that users _must_ specify one modifier or the other?

I deliberately wrote it so the user must choose one layout or the other.

lib/src/base-styles/_nx-card.scss
glenhunter: > This margin could cause problems if someone wants something like rows of 3 cards in a container that is exactly wide enough for three cards plus their interior margins. I'd consider using `column-gap` here instead of margin.

I didn't do this because IE11 doesn't support `column-gap`.

glenhunter:
> I don't see a mockup of what should happen to the callout and text in a normal vertical card when that card's height is stretched due to an adjacent card with more content. ie what should be happening within the circled area below. @dturner32
> 
> image

I did discuss this with Derrek in Slack and what is shown here is the desired behaviour.

lib/src/base-styles/_nx-card.scss
glenhunter: > There's a mockup with two icon buttons in the footer, which brings it to 136px. I'm wondering what should happen when there are more than two buttons. I'm assuming you came up with 150px here arbitrarily; when does @dturner32 think buttons or link text should wrap?

Ah, I missed the examples with the icon-only buttons. The reason for this value is because without it the text never wrapped and the column would stretch far further than we wanted

lib/src/base-styles/_nx-card.scss
glenhunter: > Is this element necessary or could these styles just be applied directly to the `.nx-text-link`. That's the only thing that's ever supposed to be in the footer, right?

Well there are examples where there are buttons in the footer. I found with the grid layout that applying the grid to objects rather than containers caused all sorts of issues when it came to following the designs, and the also is obviously a problem if anybody ever adds multiple things in the footer.

glenhunter:
> Missing example of vertical card without callout:
> image
> 
> And vertical cards without links:
> image
> 
> And horizontal cards without links.
> 
> And horizontal cards without callouts, which aren't even in the mockups, but since the cards are supposed to be convertible from one style to the other I assume we need to support that.

It should be noted that just because the cards *can* be switched from one layout to another doesn't mean that they will, in fact my assumption is that it won't happen often, and in some cases the designs will preclude switching.

glenhunter:
> Missing example of vertical card without callout:
> image
> 
> And vertical cards without links:
> image
> 
> And horizontal cards without links.
> 
> And horizontal cards without callouts, which aren't even in the mockups, but since the cards are supposed to be convertible from one style to the other I assume we need to support that.

See this comment: https://github.com/sonatype/sonatype-react-shared-components/pull/170#issuecomment-709445889

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Could this be a modifier of `.nx-card__call-out` instead?

I didn't do it that way because I was trying to avoid the bare text node. `NxCard` is closer to `NxTile` in philosophy than other components in that we're developing something that is a container for other content and while we are demonstrating a variety of options we really don't know what designers will choose to use it for.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Should we be explicit about the "two lines or less" rule?

I discussed this with @dturner32 and we decided that the designers would be responsible for constraining the content not the component because as noted here https://github.com/sonatype/sonatype-react-shared-components/pull/170#discussion_r505686918 we just can't know all of the potential uses for `NxCard`

lib/src/base-styles/_nx-card.scss
glenhunter: > I'm wondering if this container variant should just use block layout. That would be simplest and would fix the fact that the horizontal cards have double the intended margin between them.

Yeah that's actually better: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/e64d35fd86922d1c3e142113691af44ff2d99555

lib/src/base-styles/_nx-card.scss
glenhunter: > Probably redundant, and would definitely be redundant if the container used block layout.

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/887f486e8e8d70ff710a91e9ba5a7cafd1da619a

lib/src/base-styles/_nx-card.scss
glenhunter: > Why padding and not margin?

It has to do with how the l/r border is put together.

glenhunter:
> Need example with buttons in footer:
> image

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/f691bf1fa5d8590b2a0336cbca00eca16e23f909

lib/src/base-styles/_nx-card.scss
glenhunter: > This appears to have no effect. Even the Computed tab in the dev tools still shows `block`

Yeah the `display` property has no effect on flex objects: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/33c3e47c8f0339b4542500c44449588376de3360

lib/src/base-styles/_nx-card.scss
glenhunter: > This poses a problem when it automatically switches to normal vertical layout. At that point I assume the text should be center aligned. I doubt there's a way to distinguish that in CSS though.

As discussed in Slack I removed the wrapping from the alternate examples. https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/f691bf1fa5d8590b2a0336cbca00eca16e23f909

gallery/src/styles/NxCard/NxCardVerticalExample.tsx
```diff
@@ -0,0 +1,108 @@
...
+          </div>
+          <div className="nx-card__text">
+            Large icon
+          </div>
+          <footer className="nx-card__footer">
+            <a href="#" className="nx-text-link">Link</a>
+          </footer>
+        </div>
+      </div>
+      <div className="nx-card-container nx-card-container--row">
```
rpokorny, line 59: I see you've got one container per row.  I was expecting a single container which would use a multi-line flexbox to arrange the cards into potentially multiple rows.  That would be significantly more helpful I think as it would allow the caller to throw any number of cards into a single container and have them lay out right.  With the container-per-row approach, the caller would have to know how many cards to fit into each row.  Which would be especially challenging give how unpredictable their widths are.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,103 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { faShapes, faChartLine, faEdit, faTrash } from '@fortawesome/free-solid-svg-icons';
+
+import { NxFontAwesomeIcon, NxButton } from '@sonatype/react-shared-components';
+
+export default function NxCardHorizontalExample() {
+  return (
+    <div className="nx-card-container nx-card-container--column">
```
rpokorny, line 14: At this point, the only thing this container effectively does is change the margin under the last card from 16px to 24px - which @dturner32 says is wrong anyway.  So I'm thinking we could just get rid of this container: if you want horizontal cards, you just use cards bare.  If you want vertical cards, you put them in the vertical card container.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
rpokorny: > When I was writing it I hesitated over whether <div className="nx-card-container nx-card-container--column"> should be a <section>. I tend to think that the cards themselves are fine as <div>'s

I'm not entirely certain, but I think the outline semantics might get confused if the cards aren't sectioning content.  I think as they are now, the header of the last card may end up applying to all content after that card up until the next header or the end of some ancestor section.  Making the cards themselves `<section>`s would avoid that ambiguity.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > At this point, the only thing this container effectively does is change the margin under the last card from 16px to 24px - which @dturner32 says is wrong anyway. So I'm thinking we could just get rid of this container: if you want horizontal cards, you just use cards bare. If you want vertical cards, you put them in the vertical card container.

It does more than that, it tells the cards within it whether they are vertically or horizontally oriented. I'm looking towards being able to toggle between the two states if desired.

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > I didn't do it that way because I was trying to avoid the bare text node. NxCard is closer to NxTile in philosophy than other components in that we're developing something that is a container for other content and while we are demonstrating a variety of options we really don't know what designers will choose to use it for.

I don't see how our bare text node concerns apply here: the issue we usually worry about with that is when a bare text node is a sibling of an element.  That's not the case here though.  Since you draw a parallel to tiles, I'll say that I feel like this is akin to requiring `.nx-tile-header__subtitle` to have another element underneath it instead of including the subtitle text directly.

gallery/src/styles/NxCard/NxCardVerticalExample.tsx
glenhunter: > I see you've got one container per row. I was expecting a single container which would use a multi-line flexbox to arrange the cards into potentially multiple rows. That would be significantly more helpful I think as it would allow the caller to throw any number of cards into a single container and have them lay out right. With the container-per-row approach, the caller would have to know how many cards to fit into each row. Which would be especially challenging given how unpredictable their widths are.

Yeah that makes sense, though it has the interesting side-effect that the cards no longer respect `container-horizontal` which means the example wraps an extra line. https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/b3baf7ff7290bce1bd0a3c1e0b7bea044822f58e

FYI I left the three minimal examples in their own rows so that they weren't stretched vertically.

lib/src/base-styles/_nx-card.scss
rpokorny: > It has to do with how the l/r border is put together.

See my second comment

lib/src/base-styles/_nx-card.scss
rpokorny: > The reason for this value is because without it the text never wrapped and the column would stretch far further than we wanted

I know, but my question is whether 150 is in fact the right width at which to wrap.

lib/src/base-styles/_nx-card.scss
rpokorny: > Yeah the display property has no effect on flex objects: 33c3e47

There was another one a few lines down.

glenhunter:
> Missing example of vertical card without callout:
> image
> 
> And vertical cards without links:
> image
> 
> And horizontal cards without links.
> 
> And horizontal cards without callouts, which aren't even in the mockups, but since the cards are supposed to be convertible from one style to the other I assume we need to support that.

Adding the no-link version required some tweaking to the way the horizontal cards work: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/b7c8761c4c8ce22a383c0e8e66cead8d0f38827e

lib/src/base-styles/_nx-card.scss
glenhunter: > > Why padding and not margin?
> 
> Oh because of the borders. I think the borders need to be moved to the callout and the footer anyway so that they don't appear when those elements aren't present.

Exactly, this was changed as a part of: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/b7c8761c4c8ce22a383c0e8e66cead8d0f38827e

rpokorny:
Vertical alignment between callout and text in alt vertical cards when text is a single line seems wrong.

Mockup: 
image

Impl: 
image

rpokorny:
Space under alt layout callout is too big:
image

Impl:
image

glenhunter:
> Vertical alignment between callout and text in alt vertical cards when text is a single line seems wrong.
> 
> Mockup:
> image
> 
> Impl:
> image

This got tweaked here: https://github.com/sonatype/sonatype-react-shared-components/commit/b7c8761c4c8ce22a383c0e8e66cead8d0f38827e

glenhunter:
> Space under alt layout callout is too big:
> image
> 
> Impl:
> image

Also was tweaked here: https://github.com/sonatype/sonatype-react-shared-components/commit/b7c8761c4c8ce22a383c0e8e66cead8d0f38827e

lib/src/base-styles/_nx-card.scss
glenhunter: > > Yeah the display property has no effect on flex objects: [33c3e47](https://github.com/sonatype/sonatype-react-shared-components/commit/33c3e47c8f0339b4542500c44449588376de3360)
> 
> There was another one a few lines down.

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/0fc435f252062d5f621c320a909a855baa3c26ff

lib/src/base-styles/_nx-card.scss
dturner32: So I just tested a three button layout in the design as a quick mock and the total width of those buttons plus the padding came out to 144px so Glen's suggestion for 150px isn't too far off. I'm ok with the 150px max

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > I didn't do it that way because I was trying to avoid the bare text node. NxCard is closer to NxTile in philosophy than other components in that we're developing something that is a container for other content and while we are demonstrating a variety of options we really don't know what designers will choose to use it for.
> 
> I don't see how our bare text node concerns apply here: the issue we usually worry about with that is when a bare text node is a sibling of an element. That's not the case here though. Since you draw a parallel to tiles, I'll say that I feel like this is akin to requiring `.nx-tile-header__subtitle` to have another element underneath it instead of including the subtitle text directly.

Right, what I'm saying is that I can see the very real possibility that something will be added here that will be a sibling of this text node. I'm anticipating that these cards will be used for far more things, far more designs, than @dturner32  could possibly lay out here so the goal is to make them as future proof as possible without going completely crazy.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > > When I was writing it I hesitated over whether  should be a . I tend to think that the cards themselves are fine as 's
> 
> I'm not entirely certain, but I think the outline semantics might get confused if the cards aren't sectioning content. I think as they are now, the header of the last card may end up applying to all content after that card up until the next header or the end of some ancestor section. Making the cards themselves `<section>`s would avoid that ambiguity.

I'm not sure entirely either but I am confident that changing the `<div>`'s to `<section>`'s doesn't hurt anything so I have done so here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/0f736b2cb618fcc1ed51598293ed5369fe894bef

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > Right, what I'm saying is that I can see the very real possibility that something will be added here that will be a sibling of this text node. I'm anticipating that these cards will be used for far more things, far more designs, than @dturner32 could possibly lay out here so the goal is to make them as future proof as possible without going completely crazy.

If someone in the future decides on some substructure that fits into this slot it would be up to that substructure to make sure it has no bare text nodes as siblings of elements.  I don't see how requiring this extra element even in cases that aren't that helps anything.

gallery/src/styles/NxCard/NxCardVerticalExample.tsx
rpokorny: > Yeah that makes sense, though it has the interesting side-effect that the cards no longer respect container-horizontal which means the example wraps an extra line. b3baf7f

I'm not sure what was happening before, but from what I can tell now that's because `.nx-card-container--row` simply doesn't include `container-horizontal`.

rpokorny:
I notice the following two mockups which I'd say are in conflict with one another in regards to how much space should be between the text and the callout:

image
image

@dturner32 can you comment on this?

rpokorny:
Still missing:
* vertical card without callout
* horizontal cards without callouts, unless @dturner32 confirms that they aren't supported

dturner32:
> Missing example of vertical card without callout:
> image
> 
> And vertical cards without links:
> image
> 
> And horizontal cards without links.
> 
> And horizontal cards without callouts, which aren't even in the mockups, but since the cards are supposed to be convertible from one style to the other I assume we need to support that.

Horizontal cards can appear without a callout and just text (I will add a mockup for that). However, horizontal cards cannot exist with only a callout and no text

rpokorny:
Mockup shows 16px of space between a row of vertical cards and a following tile.  Impl doubles up the bottom margin on the last row of a `.nx-card-container--row`:

image

dturner32:
> I notice the following two mockups which I'd say are in conflict with one another in regards to how much space should be between the text and the callout:
> 
> image
> image
> 
> @dturner32 can you comment on this?

This was intentional. Given the the line height of the text callout at its size, the 16px padding between it and the text beneath felt a bit much which is why I reduced it to 8px in this case

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,102 @@
...
+        </header>
+        <div className="nx-card__call-out">
+          <NxFontAwesomeIcon icon={faShapes} />
+        </div>
+        <div className="nx-card__content">
+          Data point details
+        </div>
+        <footer className="nx-card__footer">
+          <div className="nx-btn-bar">
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} className="nx-icon--large" /></NxButton>
```
rpokorny, line 41: `.nx-icon--large` doesn't do anything here (and in other buttons below)

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,102 @@
...
+          </div>
+        </footer>
+      </section>
+      <section className="nx-card">
+        <header className="nx-card__header">
+          <h3 className="nx-h3">Card header</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <NxFontAwesomeIcon icon={faShapes} className="nx-icon--large" />
```
rpokorny, line 50: Is it valid BEM, or a good idea, to have a modifier on a different block that is only defined in a specific context like this?  I'd think we'd instead want something tied to this block, like `.nx-card__large-call-out-icon`

rpokorny:
The spacing of the buttons in the mockup doesn't appear to match our current standards.  It's showing 25px between the button icons, when our standard is to have 24px between the buttons themselves (so the icons would actually be a bit farther than that due to the interior padding and borders of the buttons.  @dturner32

rpokorny:
> The spacing of the buttons in the mockup doesn't appear to match our current standards. It's showing 25px between the button icons, when our standard is to have 24px between the buttons themselves (so the icons would actually be a bit farther than that due to the interior padding and borders of the buttons. @dturner32

Oh now I see the buttons here are spec'd to be 8px apart.  That's different from how they are elsewhere so can you confirm this is a deliberate special case @dturner32 ?

image

gallery/src/styles/NxCard/NxCardVerticalExample.tsx
glenhunter: > > Yeah that makes sense, though it has the interesting side-effect that the cards no longer respect container-horizontal which means the example wraps an extra line. [b3baf7f](https://github.com/sonatype/sonatype-react-shared-components/commit/b3baf7ff7290bce1bd0a3c1e0b7bea044822f58e)
> 
> I'm not sure what was happening before, but from what I can tell now that's because `.nx-card-container--row` simply doesn't include `container-horizontal`.

I removed it because it wasn't doing anything useful. As far as the browser (or SCSS) is concerned in the case where we wrap the flex elements the element that corresponds to `:last-child` is the last instance of `nx-card` not the last card in any row. So in the case where we have 9 cards that have wrapped into 3 rows (as in the example), the 3rd card still has it's right margin even though it is the last card in its row, the 5th card still has it's margin even though it is the last card in its row, and the last card has its right margin removed because it is the `:last-child`. 

Screen Shot 2020-10-16 at 11 04 31 AM

`column-gap` does not solve this. Regardless I don't think that this is worth worrying about with our max-width setting as it is three cards at max width cannot fit in one row.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,102 @@
...
+          <div className="nx-btn-bar">
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faTrash} className="nx-icon--large" /></NxButton>
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faEdit} className="nx-icon--large" /></NxButton>
+            <NxButton variant="icon-only"><NxFontAwesomeIcon icon={faTrash} className="nx-icon--large" /></NxButton>
+          </div>
+        </footer>
+      </section>
+      <section className="nx-card">
+        <header className="nx-card__header">
+          <h3 className="nx-h3 nx-h3--card-header">Card header</h3>
```
rpokorny, line 65: `.nx-h3--card-header` doesn't appear to be used for anything and also isn't applied consistently across the examples.

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > Is it valid BEM, or a good idea, to have a modifier on a different block that is only defined in a specific context like this? I'd think we'd instead want something tied to this block, like `.nx-card__large-call-out-icon`

What was in my mind when I did this is whether we would want to standardize these icon sizes in some way (which still makes this wrong since the `48px` icon should also have a size related class.

dturner32:
> > The spacing of the buttons in the mockup doesn't appear to match our current standards. It's showing 25px between the button icons, when our standard is to have 24px between the buttons themselves (so the icons would actually be a bit farther than that due to the interior padding and borders of the buttons. @dturner32
> 
> Oh now I see the buttons here are spec'd to be 8px apart. That's different from how they are elsewhere so can you confirm this is a deliberate special case @dturner32 ?
> 
> image

What's shown in the mock up above is inaccurate. To your earlier point there should actually be 24px of space between the buttons as the standard. That's considering their borders so 24px between the button borders.

rpokorny:
> I found the same thing, also with row and column for the container. I'm open to suggestions.

I feel like row and column are at least a little bit clearer.  The internal layout of the cards is complex enough that my mind doesn't think of those terms as applying there.  So what I'd be in favor of is using those terms exclusively instead of the horizontal/vertical terminology.  You've got cards as they appear in a row or cards as they appear in a column.

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+    <GalleryExampleTile title="Vertical Card"
+                        id="nx-card-vertical-example"
+                        liveExample={NxCardVerticalExample}
+                        codeExamples={nxCardVerticalCode}
+                        defaultCheckeredBackground={true}>
+      Cards shown in vertical orientation. While we have shown some content heavy examples it is recomended to keep the
+      content as minimal as possible. The orientation of the cards is determined by a class applied to their container.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Horizontal Card"
```
rpokorny, line 133: I'm thinking it'd be a little nicer if the examples were reordered to put vertical and alt vertical next to each other.  I'm thinking horizontal first, then vertical, then alt vertical.

glenhunter:
> > The spacing of the buttons in the mockup doesn't appear to match our current standards. It's showing 25px between the button icons, when our standard is to have 24px between the buttons themselves (so the icons would actually be a bit farther than that due to the interior padding and borders of the buttons. @dturner32
> 
> Oh now I see the buttons here are spec'd to be 8px apart. That's different from how they are elsewhere so can you confirm this is a deliberate special case @dturner32 ?
> 
> image

They should have 24px between them which is the standard. That's what I see. 
Screen Shot 2020-10-16 at 11 16 38 AM
Screen Shot 2020-10-16 at 11 16 19 AM

dturner32:
> > > The spacing of the buttons in the mockup doesn't appear to match our current standards. It's showing 25px between the button icons, when our standard is to have 24px between the buttons themselves (so the icons would actually be a bit farther than that due to the interior padding and borders of the buttons. @dturner32
> > 
> > 
> > Oh now I see the buttons here are spec'd to be 8px apart. That's different from how they are elsewhere so can you confirm this is a deliberate special case @dturner32 ?
> > image
> 
> They should have 24px between them which is the standard. That's what I see.
> Screen Shot 2020-10-16 at 11 16 38 AM
> Screen Shot 2020-10-16 at 11 16 19 AM

that's correct

gallery/src/styles/NxCard/NxCardVerticalAlternateExample.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { faShapes, faChartLine } from '@fortawesome/free-solid-svg-icons';
+
+import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
+
+export default function NxCardVerticalAltExample() {
+  return (
+    <>
```
rpokorny, line 14: Unnecessary fragment

gallery/visualtests/nxCard.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+const { simpleTest, simpleTestLongElement } = require('./testUtils');
+
+describe('nx-card', function() {
+  beforeEach(async function() {
+    await browser.url('#/pages/nx-card');
+  });
+
+  const verticalCardSelector = '#nx-card-vertical-example .gallery-example-live',
+      horizontalCardSelector =  '#nx-card-horizontal-example .gallery-example-live',
```
rpokorny, line 15:
```suggestion
      horizontalCardSelector = '#nx-card-horizontal-example .gallery-example-live',
```

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,175 @@
...
+    font-size: 64px;
+  }
+}
+
+.nx-card__call-out-text {
+  @include bold();
+
+  font-size: 32px;
+}
+
```
rpokorny, line 34: I'm thinking `.nx-card-container--column` should include `container-vertical`.  Right now the first and last card's margins collapse out of the parent.  The parent has the same margins already, so it doesn't actually do anything, but just for consistency and cleanliness...

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,175 @@
...
+    margin: 0 0 $nx-spacing-md 0;
+    padding: 0;
+  }
+
+  .nx-card__content {
+    -ms-grid-row: 2;
+    -ms-grid-column: 2;
+    grid-area: card-text;
+
+
```
rpokorny, line 108:
```suggestion

```

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,175 @@
...
+    -ms-grid-column-align: start;
+    justify-self: start;
+
+    @include container-horizontal;
+    align-items: center;
+    border: none;
+    border-left: 1px solid #d1d6f0;
+    display: flex;
+    height: 100%;
+    margin: 0  0 0 $nx-spacing-xxl;
```
rpokorny, line 147:
```suggestion
    margin: 0 0 0 $nx-spacing-xxl;
```

lib/src/base-styles/_nx-card.scss
rpokorny: > I'm thinking .nx-card-container--column should include container-vertical. Right now the first and last card's margins collapse out of the parent. The parent has the same margins already, so it doesn't actually do anything, but just for consistency and cleanliness...

Oh it already does have it.  Well it's not working, I think you'll have to make some one-off more specific selectors.

glenhunter:
> > I found the same thing, also with row and column for the container. I'm open to suggestions.
> 
> I feel like row and column are at least a little bit clearer. The internal layout of the cards is complex enough that my mind doesn't think of those terms as applying there. So what I'd be in favor of is using those terms exclusively instead of the horizontal/vertical terminology. You've got cards as they appear in a row or cards as they appear in a column.

This has essentially what has been done, the `--vertical` and `--horizontal` modifiers are gone except in the case of the alternate variation. I'm not sure what to call that

rpokorny:
> This has essentially what has been done, the --vertical and --horizontal modifiers are gone except in the case of the alternate variation. I'm not sure what to call that

The documentation is still in terms of vertical and horizontal cards though

glenhunter:
> > This has essentially what has been done, the --vertical and --horizontal modifiers are gone except in the case of the alternate variation. I'm not sure what to call that
> 
> The documentation is still in terms of vertical and horizontal cards though

Edited here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/ad954e98bdc482d6d24adf5542cd5265c659a9cd

lib/src/base-styles/_nx-card.scss
glenhunter: > > I'm thinking .nx-card-container--column should include container-vertical. Right now the first and last card's margins collapse out of the parent. The parent has the same margins already, so it doesn't actually do anything, but just for consistency and cleanliness...
> 
> Oh it already does have it. Well it's not working, I think you'll have to make some one-off more specific selectors.

The tweak is only required on the row layout due to the bottom margin on the cards themselves: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/d2501f44b33e904af1521ebcc0eb33663e963af2

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > Right, what I'm saying is that I can see the very real possibility that something will be added here that will be a sibling of this text node. I'm anticipating that these cards will be used for far more things, far more designs, than @dturner32 could possibly lay out here so the goal is to make them as future proof as possible without going completely crazy.
> 
> If someone in the future decides on some substructure that fits into this slot it would be up to that substructure to make sure it has no bare text nodes as siblings of elements. I don't see how requiring this extra element even in cases that aren't that helps anything.

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/cd45aea566dbdab783874f540aa041f78f82fdf2

gallery/src/styles/NxCard/NxCardVerticalAlternateExample.tsx
glenhunter: > Unnecessary fragment

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/f8bff817feb04ec001547c690a2e3a010971b05d

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > I'm thinking it'd be a little nicer if the examples were reordered to put vertical and alt vertical next to each other. I'm thinking horizontal first, then vertical, then alt vertical.

Sure: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/ad954e98bdc482d6d24adf5542cd5265c659a9cd

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > `.nx-h3--card-header` doesn't appear to be used for anything and also isn't applied consistently across the examples.

copy pasta fragment from an earlier concept, removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/cc97491a57f277c77c80cd5e0e88342b611cf326

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > > Is it valid BEM, or a good idea, to have a modifier on a different block that is only defined in a specific context like this? I'd think we'd instead want something tied to this block, like `.nx-card__large-call-out-icon`
> 
> What was in my mind when I did this is whether we would want to standardize these icon sizes in some way (which still makes this wrong since the `48px` icon should also have a size related class.

In the end I clarified the classes for the icons but left them in rather than putting them inline or in the gallery. I think that there is a reasonable change that icons will be used in the call-out. https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/cd45aea566dbdab783874f540aa041f78f82fdf2

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > `.nx-icon--large` doesn't do anything here (and in other buttons below)

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/cd45aea566dbdab783874f540aa041f78f82fdf2

rpokorny:
> They should have 24px between them which is the standard. That's what I see.

> that's correct

The mockups need to be updated then @dturner32

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -20,10 +20,9 @@ const NxCardPage = () =>
   <>
     <GalleryDescriptionTile>
       <p className="nx-p">
-        Cards can be displayed in two orientations, vertical and horizontal. The vertical and horizontal cards can be
-        transformed into one another with a simple class change on their container. There is also a vertical variant
-        supplied which displays the call out and description text side-by-side when the content allows, these cannot be
-        transformed into horizontal cards.
+        Cards can be displayed in two orientations: rows and columns. Cards can be transformed between rows and columns
+        with a simple class change on their container. There is also a row variant supplied which displays the call out
+        and cosntent side-by-side, this variant cannot be transformed into column oriented cards.
```
rpokorny, line 25:
```suggestion
        and content side-by-side, this variant cannot be transformed into column oriented cards.
```

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -126,27 +116,27 @@ const NxCardPage = () =>
...
+      Cards shown in a row layout. While we have shown some content heavy examples it is recomended to keep the
+      content as minimal as possible. The layout of the cards is determined by a class applied to their container.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Alternate Vertical Card"
+                        id="nx-card-vertical-alt-example"
+                        liveExample={NxCardVerticalAltExample}
+                        codeExamples={nxCardVerticalAltCode}
+                        defaultCheckeredBackground={true}>
+      This variation of the row layout has the callout and the description text side-by-side (when content allows).
```
rpokorny, line 128: Is "(when content allows)" still true?

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export default function NxCardHorizontalExample() {
+  return (
+    <div className="nx-card-container nx-card-container--column">
+      <section className="nx-card">
+        <header className="nx-card__header">
+          <h3 className="nx-h3">Header</h3>
+        </header>
+        <div className="nx-card__call-out">
+          <span className="nx-card__call-out-text">42</span>
```
rpokorny, line 20: Outdated class

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,142 @@
...
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card__call-out-text</code></td>
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">
+              In the case where the call out consists of text (e.g. 99%) it should be wrapped in this class (typically
+              in a <code className="nx-code">&lt;span&gt;</code>).
+            </td>
+          </tr>
```
rpokorny, line 89:
```suggestion
```

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -0,0 +1,142 @@
...
+              {' '}<code className="nx-code">&lt;div&gt;</code> wraps
+              {' '}<code className="nx-code">.nx-card__call-out</code> and
+              {' '}<code className="nx-code">.nx-card__call-out-text</code> in the alternate row card layout.
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Vertical Card"
```
rpokorny, line 114: The names and ids here and below still reference the vertical/horizontal terms

gallery/src/styles/NxCard/NxCardVerticalAlternateExample.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { faShapes, faChartLine } from '@fortawesome/free-solid-svg-icons';
+
+import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
+
+export default function NxCardVerticalAltExample() {
```
rpokorny, line 12: Lets rename these to the consistent terminology

gallery/src/styles/NxCard/NxCardVerticalExample.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <footer className="nx-card__footer">
+            <a href="#" className="nx-text-link">Link</a>
+          </footer>
+        </section>
+        <section className="nx-card">
+          <header className="nx-card__header">
+            <h3 className="nx-h3">Card header</h3>
+          </header>
+          <div className="nx-card__call-out">
+            <NxFontAwesomeIcon icon={faShapes} className="nx-icon--card-call-out-xl" />
```
rpokorny, line 49: Do these `.nx-icon` modifiers need to be documented?

gallery/visualtests/nxCard.js
```diff
@@ -0,0 +1,29 @@
...
+
+  describe('nx-card vertical layout', function() {
+    it('looks right', simpleTestLongElement(verticalCardSelector));
+  });
+
+  describe('nx-card horizontal layout', function() {
+    it('looks right', simpleTestLongElement(horizontalCardSelector));
+  });
+
+  describe('nx-card vertical alternate layout', function() {
```
rpokorny, line 26: Update terminology

lib/src/base-styles/_nx-card.scss
```diff
@@ -0,0 +1,175 @@
...
+  .nx-icon--card-call-out {
+    font-size: 48px;
+  }
+
+  .nx-icon--card-call-out-xl {
+    font-size: 64px;
+  }
+}
+
+.nx-card__call-out {
```
rpokorny, line 30: Can be merged with the block above.

glenhunter: > Can be merged with the block above.

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/3369c5413e09c9c3129815857eda15c90fe845d3

gallery/visualtests/nxCard.js
glenhunter: > Update terminology

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/38ec97d38d3854439f9f6e4e7fd34936d4cd01f6

gallery/src/styles/NxCard/NxCardVerticalExample.tsx
glenhunter: > Do these `.nx-icon` modifiers need to be documented?

I waffled on that and in the end documented them here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/29682dcab30f0d9ca6ffd4b490f5764de425794e

Though I still think that we should consider making some basic icon sizing classes for common values.

gallery/src/styles/NxCard/NxCardVerticalAlternateExample.tsx
glenhunter: > Lets rename these to the consistent terminology

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/38ec97d38d3854439f9f6e4e7fd34936d4cd01f6

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > The names and ids here and below still reference the vertical/horizontal terms

https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/38ec97d38d3854439f9f6e4e7fd34936d4cd01f6

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Is "(when content allows)" still true?

Nope: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/33340d58faa9277d46178f76ac6e5ce46738f502

gallery/src/styles/NxCard/NxCardHorizontalExample.tsx
glenhunter: > Outdated class. It's in a couple of places in the examples and also still mentioned in the documentation.

Here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/d24ea69de317ef47e3418aa579407c99fce3be42 and here: https://github.com/sonatype/sonatype-react-shared-components/pull/170/commits/29682dcab30f0d9ca6ffd4b490f5764de425794e

rpokorny (approved):
+1 but @dturner32 still needs to update the mockups for https://github.com/sonatype/sonatype-react-shared-components/pull/170#issuecomment-712218543

PR #172, opened 2020-10-15
Add a donut chart component - RSC-335

mealingr:
Story: https://issues.sonatype.org/browse/CLM-17006
Build: https://jenkins.ci.sonatype.dev/job/uxui/job/sonatype-react-shared-components/job/RSC-335_donut_chart/lastBuild

I know there was some mention of improvements/fancier things in 
https://sonatype.slack.com/archives/CJUF9G7T2/p1601917616089900
but this should hopefully get us the status quo of what we have now/something to build on


rpokorny:
> jenkins is repporting hudson.AbortException: No test report files were found. Configuration error?, not sure how to fix that yet or if it's a problem with this code

That's fallout from an earlier error, you need to scroll up to see what went wrong.  But most likely, it's the same thing I described to someone else recently on slack:

> But the first thing that comes to mind that you likely don't know is that builds of our master branch automatically publish releases.  In support of that, we have logic in the dev branch builds that enforce that they must have a version higher than master.  So you want to bump the version on your branch.  use the procedure in the readme.  In your case it should be a minor version

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.1.2",
+  "version": "2.1.3",
```
rpokorny, line 3: That's a patch version.  A minor version bump would be 2.2.0

mealingr: > That's a patch version. A minor version bump would be 2.2.0

Oops I'll fix

lib/src/components/NxDonutChart/NxDonutChart.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { FunctionComponent, ReactElement } from 'react';
+import { Props, propTypes } from './types';
+export { Props } from './types';
+
+const NxDonutChart: FunctionComponent<Props> =
```
rpokorny, line 11: Given that we expect to have more full-fledged donut charts in the future (ones with more than two slices), I wonder if this should have a slightly more specific name, like `NxCompletionDonutChart` or `NxProgressDonutChart`

lib/src/components/NxDonutChart/NxDonutChart.tsx
```diff
@@ -0,0 +1,44 @@
...
+  function NxDonutChart(props): ReactElement<Props> {
```
rpokorny, line 12: Use `forwardRef` so that tooltips could be attached.  See `NxButton` for an example.

lib/src/components/NxDonutChart/types.ts
```diff
@@ -0,0 +1,23 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import * as PropTypes from 'prop-types';
+
+export interface Props {
+  id?: string | null;
+  percent: number;
+  outerRadius: number;
+  innerRadius: number;
+  fillColors: string[];
```
rpokorny, line 14: I question whether we want all this stuff to be configurable.  The overall size of the donut and the colors can be done in CSS and the inner proportions of the donut could just as well be an implementation detail, I would think.

gallery/package.json
mealingr: > That's a patch version. A minor version bump would be 2.2.0

Should be fixed by https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/c7c3784a194b3348b844eaecb20a5224b090a2c2

lib/src/components/NxDonutChart/types.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import * as PropTypes from 'prop-types';
+
+export interface Props {
```
rpokorny, line 9: Should probably allow all svg attributes rather than just id in particular.  See the `types.ts` files of `NxButton` and `NxTableBody` for various ways to include native attrs in your allowed props.

lib/src/components/NxDonutChart/NxDonutChart.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const circumference = 2 * Math.PI * r;
+    const strokeDasharray = [decimal * circumference, (1.0 - decimal) * circumference];
+    const strokeDashoffset = (0.25 * circumference) - (1.0 - decimal) * circumference;
+    return (
+      <svg width={diameter} height={diameter}>
+        <circle cx={props.outerRadius}
+                cy={props.outerRadius}
+                strokeWidth={strokeWidth}
+                r={r}
+                fill="transparent"
```
rpokorny, line 26: Can be in css.  Also "none" would probably be the more appropriate value.

lib/src/components/NxDonutChart/NxDonutChart.tsx
```diff
@@ -0,0 +1,44 @@
...
+  function NxDonutChart(props): ReactElement<Props> {
+    const decimal = props.percent / 100;
+    const diameter = props.outerRadius * 2;
+    const strokeWidth = props.outerRadius - props.innerRadius;
+    const r = props.innerRadius + strokeWidth / 2;
+    const circumference = 2 * Math.PI * r;
+    const strokeDasharray = [decimal * circumference, (1.0 - decimal) * circumference];
+    const strokeDashoffset = (0.25 * circumference) - (1.0 - decimal) * circumference;
+    return (
+      <svg width={diameter} height={diameter}>
```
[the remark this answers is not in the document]
rpokorny: > Rather than setting width and height here, you can leave that to CSS and use a viewBox to create an arbitrary (and more convenient) "user coordinate space". For instance, `viewBox="-1 -1 2 2" would give you a coordinate space with (0,0) in the center and the edges 1 unit away from that center. This would likely simplify your math above, and would allow the on-screen size of the chart to be controlled in css as it ought to be.

Additionally, if you center around (0,0), it makes it relatively easy to use hard-coded transforms to get the start/finish of the arc where you want them rather than having to calculate the dashoffset.

lib/src/components/NxDonutChart/NxDonutChart.tsx
```diff
@@ -0,0 +1,44 @@
...
+export { Props } from './types';
+
+const NxDonutChart: FunctionComponent<Props> =
+  function NxDonutChart(props): ReactElement<Props> {
+    const decimal = props.percent / 100;
+    const diameter = props.outerRadius * 2;
+    const strokeWidth = props.outerRadius - props.innerRadius;
+    const r = props.innerRadius + strokeWidth / 2;
+    const circumference = 2 * Math.PI * r;
+    const strokeDasharray = [decimal * circumference, (1.0 - decimal) * circumference];
```
rpokorny, line 18: I've discovered that there's a [`pathLength`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pathLength) attribute that allows us to define custom sizing semantics for things like the dasharray.  That is, if you set `pathLength="100"`, you can then just set the dasharray using the actual percentage followed by 100 - dashoffset notwithstanding that is, but see one of the other threads on my idea to avoid that confusing attribute.

lib/src/components/NxDonutChart/NxDonutChart.tsx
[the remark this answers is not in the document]
rpokorny: > Additionally, if you center around (0,0), it makes it relatively easy to use hard-coded transforms to get the start/finish of the arc where you want them rather than having to calculate the dashoffset.

These are the transforms I'm thinking of: `transform: rotate(-90deg) scaleY(-1);`

lib/src/components/NxDonutChart/NxDonutChart.tsx
mealingr: > Given that we expect to have more full-fledged donut charts in the future (ones with more than two slices), I wonder if this should have a slightly more specific name, like `NxCompletionDonutChart` or `NxProgressDonutChart`

I thought `NxBinaryDonutChart` might also be alright if that works for you https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/47355af1906408054b5f20f7a7b2cea5319f34f5

lib/src/components/NxDonutChart/NxDonutChart.tsx
mealingr: > Use `forwardRef` so that tooltips could be attached. See `NxButton` for an example.

Changed to use `forwardRef` in https://github.com/sonatype/sonatype-react-shared-components/commit/47355af1906408054b5f20f7a7b2cea5319f34f5

mealingr: > You still need to actually attach the ref to the `<svg>`

Should be done with https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/5a3e4fd020df42fb5582f6d3dd79160c5e40120b

lib/src/components/NxDonutChart/NxDonutChart.tsx
mealingr: > I've discovered that there's a [`pathLength`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pathLength) attribute that allows us to define custom sizing semantics for things like the dasharray. That is, if you set `pathLength="100"`, you can then just set the dasharray using the actual percentage followed by 100 - dashoffset notwithstanding that is, but see one of the other threads on my idea to avoid that confusing attribute.

Took me a few attempts to iron things out but this worked nicely in the end https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/4cafff19731edc0c160cde6940e2cc1d48332d60

lib/src/components/NxDonutChart/NxDonutChart.tsx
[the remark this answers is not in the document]
mealingr: > use a `viewBox`

> These are the transforms I'm thinking of: `transform: rotate(-90deg) scaleY(-1);`

Used a `viewBox` and transforms in https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/5a3e4fd020df42fb5582f6d3dd79160c5e40120b seem to work nicely

lib/src/components/NxDonutChart/types.ts
mealingr: > I question whether we want all this stuff to be configurable. The overall size of the donut and the colors can be done in CSS and the inner proportions of the donut could just as well be an implementation detail, I would think.

Moved what I can to css in https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/5a3e4fd020df42fb5582f6d3dd79160c5e40120b and https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/2afc66af8da46c8fedc9a79e40683da1fa841a9d

lib/src/components/NxDonutChart/NxDonutChart.tsx
mealingr: > Can be in css. Also "none" would probably be the more appropriate value.

Moved in https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/5a3e4fd020df42fb5582f6d3dd79160c5e40120b

rpokorny:
In Chrome (but not Firefox!) I'm getting this weird effect on the large-hole example where the edge of the colors isn't radial, its at an angle:
image

rpokorny:
> In Chrome (but not Firefox!) I'm getting this weird effect on the large-hole example where the edge of the colors isn't radial, its at an angle:

Even stranger, it fixes itself if I zoom in even a little bit (and gets worse if I zoom out).  We may have to chalk that one up to a browser bug

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.scss
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-binary-donut-chart-examples {
+  svg {
+    width: 200px;
+    height: 200px;
```
rpokorny, line 10: I'd say it should have some default size, probably matching what you want it to be in the app report.  Applications would be free to change that using CSS, but there'd be no need to demonstrate it.

gallery/src/pageConfig.ts
```diff
@@ -103,7 +104,8 @@ const pageConfig: PageConfig = {
     NxStatefulTreeViewMultiSelect: NxStatefulTreeViewMultiSelect,
     NxTreeViewRadioSelect: NxTreeViewRadioSelectPage,
     NxStatefulTreeViewRadioSelect: NxStatefulTreeViewRadioSelectPage,
-    NxVulnerabilityDetails: NxVulnerabilityDetailsPage
+    NxVulnerabilityDetails: NxVulnerabilityDetailsPage,
+    NxBinaryDonutChart: NxBinaryDonutChartPage
```
rpokorny, line 108: This list should be in alphabetical order

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+                        className="nx-binary-donut-chart-examples"
+                        codeExamples={nxBinaryDonutChartLargeHoleExample}
+                        liveExample={NxBinaryDonutChartLargeHoleExample}>
+        An example of a <code className="nx-code">NxBinaryDonutChart</code> with a large hole.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Skewed Example"
+                        codeExamples={nxBinaryDonutChartSkewedExample}
+                        liveExample={NxBinaryDonutChartSkewedExample}>
+        An example of a <code className="nx-code">NxBinaryDonutChart</code> which is skewed.
```
rpokorny, line 54: While this capability to do this follows naturally from the ability to pass arbitrary svg attributes, it doesn't necessarily seem like something anyone would ever be likely to really want to do, so I'm not sure it needs an example.  At the least, if it is kept then maybe it should be called out that the point of the example is to demonstrate that arbitrary attributes can be passed.

Even then though, I worry that this example might mislead people into thinking that size adjustments to this component should generally be done using props, when that isn't the case and we'd want them to use svg for that like they would on any other element.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.scss
```diff
@@ -0,0 +1,19 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-colors';
+
+.nx-binary-donut-chart-circle {
+  fill: none;
+  transform: rotate(-90deg);
+
+  &:nth-of-type(1) {
```
rpokorny, line 13: nth child selectors are to be avoided where possible, as they are fragile.  What I'd rather see here is a class for the svg itself (`nx-binary-donut`) which could probably house the fill and transform, and then meaningful classes for the two circles (`nx-binary-donut__background` and `nx-binary-donut__arc` maybe) which set their stroke color.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,40 @@
...
+const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
+    function NxBinaryDonutChart(props, ref) {
+      const innerRadiusPercent = props.innerRadiusPercent === 0 ? props.innerRadiusPercent :
+        (props.innerRadiusPercent || 50);
+
+      const strokeWidth = 100 - innerRadiusPercent;
+      const r = innerRadiusPercent + (strokeWidth / 2);
+      const strokeDasharray = [(100 - props.percent), props.percent];
+      return (
+        <svg ref={ref} viewBox="-100 -100 200 200" {...props}>
```
rpokorny, line 22: Passing `...props` will cause the custom props to get passed to the native element as well, which is why you're getting 
```Warning: React does not recognize the `innerRadiusPercent` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `innerradiuspercent` instead. If you accidentally passed it from a parent component, remove it from the DOM element
```
in the console.  What I'd do is start the function with a line like
```
const { innerRadiusPercent, percent, ...svgAttrs } = props;
```
and then you don't have to use `props.` when looking up the other two props either.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <circle className="nx-binary-donut-chart-circle"
+                  strokeWidth={strokeWidth}
+                  r={r}>
+          </circle>
+          <circle className="nx-binary-donut-chart-circle"
```
rpokorny, line 27: So this actually represents the amount of the the donut that's not filled in, right?

rpokorny:
Oh boy, so I adjusted the minimal example to 50% and the angle issues get weirder/worse:

image

I'll investigate but one possibility would be to use a `<path>` instead of the dasharray.  That'll required some trigonometry though

rpokorny:
They're also really messed up in IE:
image

rpokorny:
> I'll investigate but one possibility would be to use a <path> instead of the dasharray. That'll required some trigonometry though

Quick and dirty POC here: https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-335_donut_chart...RSC-335_donut_chart_path

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { forwardRef } from 'react';
+import { Props, propTypes } from './types';
+export { Props } from './types';
+
+import './NxBinaryDonutChart.scss';
+
+const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
+    function NxBinaryDonutChart(props, ref) {
```
rpokorny, line 14: We may want to validate that the percent is in the range [0, 100] and either clamp it to that range or throw an exception otherwise

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,40 @@
...
+      const innerRadiusPercent = props.innerRadiusPercent === 0 ? props.innerRadiusPercent :
+        (props.innerRadiusPercent || 50);
```
rpokorny, line 16:
```suggestion
      const innerRadiusPercent = props.innerRadiusPercent == null ? 50 : props.innerRadiusPercent;
```

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.scss
mealingr: > I'd say it should have some default size, probably matching what you want it to be in the app report. Applications would be free to change that using CSS, but there'd be no need to demonstrate it.

Set it to 30px by 30px as default (same as the app report size I think) https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

gallery/src/pageConfig.ts
mealingr: > This list should be in alphabetical order

Reordered to alphabetical https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mealingr: > it doesn't necessarily seem like something anyone would ever be likely to really want to do, so I'm not sure it needs an example

Removed this example https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.scss
mealingr: > `nx-binary-donut__background` and `nx-binary-donut__arc` maybe

Changed to use these classes instead of nth child selectors https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mealingr: > start the function with a line like

Thanks done in https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mealingr: > So this actually represents the amount of the the donut that's not filled in, right?

Yeah your arc class suggestion makes it clearer now I think https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mealingr: > We may want to validate that the percent is in the range [0, 100] and either clamp it to that range or throw an exception otherwise

Tried to add some validation https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mealingr: Changed in https://github.com/sonatype/sonatype-react-shared-components/pull/172/commits/0b2fad02e32cf9927810950d59c52abe24ac9a2f thanks

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.scss
rpokorny: > Set it to 30px by 30px as default (same as the app report size I think)

I recommend setting only the height.  It's got an inherent aspect ratio so the width will come out equal to the height automatically, and this way it's one less property for a consuming app to have to override if they want to customize it

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxBinaryDonutChart } from '@sonatype/react-shared-components';
+
+export default function NxBinaryDonutChartMinimalExample() {
+  return <NxBinaryDonutChart percent={90} />;
```
rpokorny, line 11: You'll notice that in my branch I added more examples at different percentages.  Keep in mind that the gallery actually has (at least) two purposes:  not just to demonstrate to consumers how to use the components, but also to prove to ourselves that the styles (and logic) work correctly.  Therefore I do think you should add donuts at a variety of percentages to at least some of the examples.  And especially covering the edge cases.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+const nxBinaryDonutChartMinimalExampleCode = require('!!raw-loader!./NxBinaryDonutChartMinimalExample').default;
+const nxBinaryDonutChartNoHoleExample = require('!!raw-loader!./NxBinaryDonutChartNoHoleExample').default;
+const nxBinaryDonutChartLargeHoleExample = require('!!raw-loader!./NxBinaryDonutChartLargeHoleExample').default;
+
+const NxBinaryDonutChartPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        <code className="nx-code">NxBinaryDonutChart</code> represents a binary donut chart.
+      </p>
```
rpokorny, line 23: Should have a table of props documentation like we do for other components.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Minimal Example"
+                        className="nx-binary-donut-chart-examples"
```
rpokorny, line 27: We don't typically have classNames here but we do often have ids.  Those ids pretty much only important for visual tests.  Which, btw, you are missing

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.scss
```diff
@@ -0,0 +1,24 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-colors';
+
+.nx-binary-donut-chart {
+  width: 30px;
+  height: 30px;
+
+  circle, path {
```
rpokorny, line 13: If I'm not mistaken you can set `fill: none` right on the svg (e.g. on `.nx-binary-donut-chart`) and it will inherit down.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+      // Add 0.5π because we want the angle from the top of the circle, not the right
+      const arcEndAngle = (Math.PI / 2) + (2 * Math.PI * (percent / 100));
+      const arcEndX = r * Math.cos(arcEndAngle);
+      // Multiply by -1 because the y-axis origin (0,0) is at the top-left rather than bottom-left
+      const arcEndY = -1 * r * Math.sin(arcEndAngle);
+      const largeArc = percent > 50 ? 1 : 0;
+
+      return (
+        <svg ref={ref} viewBox="-100 -100 200 200" className="nx-binary-donut-chart" {...svgAttrs}>
```
rpokorny, line 30: You'll want to manually merge the className here with any className that comes in on the props.  You can use the `classnames` utility for that, as demonstrated in many of the other components.  Otherwise a `className` specified by the caller will overwrite `nx-binary-donut-chart`.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,49 @@
...
+      const innerRadiusPercentNotNull = innerRadiusPercent == null ? 50 : innerRadiusPercent < 0 ? 0 :
+        innerRadiusPercent > 100 ? 100 : innerRadiusPercent;
+
+      const strokeWidth = 100 - innerRadiusPercentNotNull;
+      const r = innerRadiusPercentNotNull + (strokeWidth / 2);
+
+      // Add 0.5π because we want the angle from the top of the circle, not the right
+      const arcEndAngle = (Math.PI / 2) + (2 * Math.PI * (percent / 100));
+      const arcEndX = r * Math.cos(arcEndAngle);
+      // Multiply by -1 because the y-axis origin (0,0) is at the top-left rather than bottom-left
```
rpokorny, line 25: This comment isn't quite right, the origin is in the center using this viewbox.  I'd say y increases going downwards on the screen, not upwards

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const paths = getShallowComponent().find('path');
+
+    expect(circles.length).toBe(1);
+    expect(circles.at(0)).toHaveClassName('.nx-binary-donut-chart__background');
+    expect(circles.at(0)).toHaveProp('strokeWidth', 50);
+    expect(circles.at(0)).toHaveProp('r', 75);
+    expect(paths.length).toBe(0);
+  });
+
+  it('renders only a background circle with the expected properties when percent is 0', function() {
```
rpokorny, line 31: Isn't this the same as the previous test?

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const paths = getShallowComponent().find('path');
+
+    expect(circles.length).toBe(1);
+    expect(circles.at(0)).toHaveClassName('.nx-binary-donut-chart__background');
+    expect(circles.at(0)).toHaveProp('strokeWidth', 50);
+    expect(circles.at(0)).toHaveProp('r', 75);
+    expect(paths.length).toBe(1);
+    expect(paths.at(0)).toHaveClassName('.nx-binary-donut-chart__arc');
+    expect(paths.at(0)).toHaveProp('d', 'M 0 -75 A 75 75 0 0 0 ' + (75 * Math.cos(1.5 * Math.PI))
+        + ' ' + (-75 * Math.sin(1.5 * Math.PI)));
```
rpokorny, line 67: These trig calcs should come out to 0 and 1 respectively

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    expect(circles.length).toBe(1);
+    expect(circles.at(0)).toHaveClassName('.nx-binary-donut-chart__arc');
+    expect(circles.at(0)).toHaveProp('strokeWidth', 50);
+    expect(circles.at(0)).toHaveProp('r', 75);
+    expect(paths.length).toBe(0);
+  });
+
+  it('renders a background circle and arc path with expected properties when 0 < percent < 100', function() {
+    const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, { percent: 50 });
```
rpokorny, line 56: may want to try a few different percentages to make sure all the positives and negatives come out right.  And the large arc flag.

lib/src/components/NxBinaryDonutChart/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import { SVGAttributes } from 'react';
+import * as PropTypes from 'prop-types';
+import { requiredPercentNumber, optionalPercentNumber } from '../../util/customPropTypes';
+
+export interface Props extends SVGAttributes<SVGSVGElement>{
+  percent: number;
+  innerRadiusPercent?: number | null | undefined;
```
rpokorny, line 13: `undefined` is implicit in `?`
```suggestion
  innerRadiusPercent?: number | null;
```

lib/src/util/__tests__/customPropTypes.test.ts
```diff
@@ -43,4 +43,79 @@ describe('Custom PropTypes', function() {
...
+    it('should return an error if a value greater than 100 is supplied', function() {
+      const props = { requiredPercent: 101 },
+          retVal = requiredPercentNumber(props, propName, componentName, '', ''),
+          err = `${componentName}: prop "${propName}" must be a number between 0 and 100 inclusive; received ` +
+              props[propName];
+
+      expect(retVal).toBeInstanceOf(TypeError);
+      expect(retVal).toEqual(new TypeError(err));
+    });
+  });
```
rpokorny, line 101: Should test when null and undefined are supplied to this one too

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.scss
mr-zepol: > > Set it to 30px by 30px as default (same as the app report size I think)
> 
> I recommend setting only the height. It's got an inherent aspect ratio so the width will come out equal to the height automatically, and this way it's one less property for a consuming app to have to override if they want to customize it

Done e8460440bc68f3dd35da44cd427a523a3659fa99

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
mr-zepol: > You'll notice that in my branch I added more examples at different percentages. Keep in mind that the gallery actually has (at least) two purposes: not just to demonstrate to consumers how to use the components, but also to prove to ourselves that the styles (and logic) work correctly. Therefore I do think you should add donuts at a variety of percentages to at least some of the examples. And especially covering the edge cases.

I added three more examples ( 25, 50, 100), I can add more if needed e8460440bc68f3dd35da44cd427a523a3659fa99

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.scss
mr-zepol: > If I'm not mistaken you can set `fill: none` right on the svg (e.g. on `.nx-binary-donut-chart`) and it will inherit down.

Done e8460440bc68f3dd35da44cd427a523a3659fa99

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mr-zepol: > You'll want to manually merge the className here with any className that comes in on the props. You can use the `classnames` utility for that, as demonstrated in many of the other components. Otherwise a `className` specified by the caller will overwrite `nx-binary-donut-chart`.

Done d08eba2ce5e2da5bcf001ffde1cc40e342e7e2c7

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
mr-zepol: > These trig calcs should come out to 0 and 1 respectively

I am checking this and they are not, also, AFAIK the values should be the same for any percentage >0 and <100 right? I was testing with 72 and it's different, so maybe there's something going on

lib/src/util/__tests__/customPropTypes.test.ts
mr-zepol: > Should test when null and undefined are supplied to this one too

Done 0233f50e78105bae19a8d84cce928a4d31c13944

lib/src/components/NxBinaryDonutChart/types.ts
mr-zepol: Done 0233f50e78105bae19a8d84cce928a4d31c13944

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mr-zepol: > This comment isn't quite right, the origin is in the center using this viewbox. I'd say y increases going downwards on the screen, not upwards

Changed 0233f50e78105bae19a8d84cce928a4d31c13944

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
mr-zepol: > Isn't this the same as the previous test?

I modified the test for another percentage ec3dc386383acec3217b81fca65f5651fd594ce8

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mr-zepol: > We don't typically have classNames here but we do often have ids. Those ids pretty much only important for visual tests. Which, btw, you are missing

I added one visual test a63e101f8437f7be35a7e2cea5eb4c594c505506 , since basically there are no actions on that component I didn't think it was necessary to do the same for the other percentages, but I can add more if you think it's a good idea

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mr-zepol: Added 2471b90185acb2beb99a29a9678dd6f7a1e99006

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,103 @@
...
+const nxBinaryDonutChartFullExample = require('!!raw-loader!./NxBinaryDonutChartFullExample').default;
+const nxBinaryDonutChartHalfExample = require('!!raw-loader!./NxBinaryDonutChartHalfExample').default;
+const nxBinaryDonutChartMinimalExampleCode = require('!!raw-loader!./NxBinaryDonutChartMinimalExample').default;
+const nxBinaryDonutChartNoHoleExample = require('!!raw-loader!./NxBinaryDonutChartNoHoleExample').default;
+const nxBinaryDonutChartLargeHoleExample = require('!!raw-loader!./NxBinaryDonutChartLargeHoleExample').default;
+const nxBinaryDonutChartQuarterExample = require('!!raw-loader!./NxBinaryDonutChartQuarterExample').default;
+
+const NxBinaryDonutChartPage = () =>
+  <>
+      <GalleryDescriptionTile>
```
rpokorny, line 26: It seems your editor is set to quad-indent jsx.  There are eslint errors all through here.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
rpokorny: > I added three more examples ( 25, 50, 100), I can add more if needed e846044

Two things:
1. I don't think each one needs to be a separate example.  You can have a single example for each "style" (e.g. each `innerRadius`) and within that example have a series of donut charts with varying percentages.  That would keep the verbosity of the page down.
2. I think we definitely also need one for 0%

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
rpokorny: > I added one visual test a63e101 , since basically there are no actions on that component I didn't think it was necessary to do the same for the other percentages, but I can add more if you think it's a good idea

You used the same value for each id.  All ids on an HTML page at any given time must be unique.

Additionally.  I think visual tests of varying percentages are necessary for the same reason that examples of varying percentages are necessary: to make sure they look right :)  Similarly, I do also think the example with a 0 inner radius should have a visual test as that's kind of a special case that the browser would be more likely to screw up rendering.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
rpokorny: > Done d08eba2

That is not at all what I meant.  Observe the usage of `classnames` in `NxCloseButton` as an example.

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
rpokorny: > I modified the test for another percentage ec3dc38

"it renders only a background circle with the expected properties when percent is 50" is not desired to be a true statement, nor is it what that test actually tests.

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
rpokorny: > I am checking this and they are not (-1.3777276490407723e-14 and 75)

Sorry, I meant they should come out to 0 and `r`, which they basically do.  Unfortunately, what with π being an irrational number, there are going to be rounding errors which is where that decimal is coming from.

> also, AFAIK the values should be the same for any percentage >0 and <100 right?

No.  These values define where the arc stops.  That they are different for different percentage values is the core functionality of the component.

lib/src/util/__tests__/customPropTypes.test.ts
rpokorny: > Done 0233f50

Do those tests pass?  In my reading of the component's source code, it doesn't seem to be written to error out in these cases but rather to clamp to the [0, 100] range.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,103 @@
...
+                  <th className="nx-cell nx-cell--header">Details</th>
+              </tr>
+              </thead>
+              <tbody>
+              <tr className="nx-table-row">
+                  <td className="nx-cell">percent</td>
+                  <td className="nx-cell">number</td>
+                  <td className="nx-cell">Yes</td>
+                  <td className="nx-cell">
+                      Percentage of the donut
```
rpokorny, line 45:
```suggestion
                      Percentage which this donut represents.  E.g. when 0 the donut is empty, and as it increases towards 100 the amount of the donut which is filled in increases.
```

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,103 @@
...
+                  </td>
+              </tr>
+              <tr className="nx-table-row">
+                  <td className="nx-cell">innerRadiusPercent</td>
+                  <td className="nx-cell">number</td>
+                  <td className="nx-cell">No</td>
+                  <td className="nx-cell">
+                      Represent the radius of the value displayed in the donut, the default value is 50
```
rpokorny, line 53:
```suggestion
                      The size of the hole in the donut, as a percentage of the donut's overall size.  The default value is 50.
```

gallery/visualtests/nxBinaryDonutChart.js
```diff
@@ -0,0 +1,17 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+const { simpleTest } = require('./testUtils');
+
+describe('NxBinaryDonutChart', function() {
+  beforeEach(async function() {
+    await browser.url('#/pages/NxBinaryDonutChart');
+  });
+
+  const simpleDonutSelector = 'nx-binary-donut-chart-examples';
```
rpokorny, line 14: I doubt this is working.
```suggestion
  const simpleDonutSelector = '#nx-binary-donut-chart-examples';
```

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,53 @@
...
+import classnames from "classnames";
+
+const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
+    function NxBinaryDonutChart(props, ref) {
+      const { innerRadiusPercent, percent, ...svgAttrs } = props;
+
+      const donutClasses = classnames('nx-binary-donut-chart', { innerRadiusPercent });
+
+      const innerRadiusPercentNotNull = innerRadiusPercent == null ? 50 : innerRadiusPercent < 0 ? 0 :
+        innerRadiusPercent > 100 ? 100 : innerRadiusPercent;
```
rpokorny, line 21: I prefer to format nested ternaries as follows, so that it becomes an "if-then" on each line followed by a default value:

```suggestion
      const innerRadiusPercentNotNull = innerRadiusPercent == null ? 50 :
        innerRadiusPercent < 0 ? 0 :
        innerRadiusPercent > 100 ? 100 :
        innerRadiusPercent;
```

lib/src/util/__tests__/customPropTypes.test.ts
rpokorny: > Do those tests pass? In my reading of the component's source code, it doesn't seem to be written to error out in these cases but rather to clamp to the [0, 100] range.

Sorry, I forgot what these tests were for.  I was thinking of the component not the propTypes.  Nevermind.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
mr-zepol: > > I added three more examples ( 25, 50, 100), I can add more if needed [e846044](https://github.com/sonatype/sonatype-react-shared-components/commit/e8460440bc68f3dd35da44cd427a523a3659fa99)
> 
> Two things:
> 
> 1. I don't think each one needs to be a separate example.  You can have a single example for each "style" (e.g. each `innerRadius`) and within that example have a series of donut charts with varying percentages.  That would keep the verbosity of the page down.
> 2. I think we definitely also need one for 0%

Done b9f8078b190bac0c253d7a7c363ad492ea6a5587

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mr-zepol: > > I added one visual test [a63e101](https://github.com/sonatype/sonatype-react-shared-components/commit/a63e101f8437f7be35a7e2cea5eb4c594c505506) , since basically there are no actions on that component I didn't think it was necessary to do the same for the other percentages, but I can add more if you think it's a good idea
> 
> You used the same value for each id. All ids on an HTML page at any given time must be unique.
> 
> Additionally. I think visual tests of varying percentages are necessary for the same reason that examples of varying percentages are necessary: to make sure they look right :) Similarly, I do also think the example with a 0 inner radius should have a visual test as that's kind of a special case that the browser would be more likely to screw up rendering.

Added b9f8078b190bac0c253d7a7c363ad492ea6a5587

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mr-zepol: > It seems your editor is set to quad-indent jsx. There are eslint errors all through here.

Fixed b9f8078b190bac0c253d7a7c363ad492ea6a5587

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mr-zepol: > 

Changed b9f8078b190bac0c253d7a7c363ad492ea6a5587

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
mr-zepol: > 

Changed b9f8078b190bac0c253d7a7c363ad492ea6a5587

gallery/visualtests/nxBinaryDonutChart.js
mr-zepol: > I doubt this is working.

I added a visual test for each example we have for the gallery b9f8078b190bac0c253d7a7c363ad492ea6a5587

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
mr-zepol: > > Done [d08eba2](https://github.com/sonatype/sonatype-react-shared-components/commit/d08eba2ce5e2da5bcf001ffde1cc40e342e7e2c7)
> 
> That is not at all what I meant. Observe the usage of `classnames` in `NxCloseButton` as an example.

Changed 655a6b02744a48b1a1446c23e842355303057d68

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
mr-zepol: > > I modified the test for another percentage [ec3dc38](https://github.com/sonatype/sonatype-react-shared-components/commit/ec3dc386383acec3217b81fca65f5651fd594ce8)
> 
> "it renders only a background circle with the expected properties when percent is 50" is not desired to be a true statement, nor is it what that test actually tests.

I removed this test since I realized 50 was already being tested, I added two more with different values

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
mr-zepol: > may want to try a few different percentages to make sure all the positives and negatives come out right. And the large arc flag.

I added two more tests with different values

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartEmptyExample.tsx
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxBinaryDonutChart } from '@sonatype/react-shared-components';
+
+export default function NxBinaryDonutChartEmptyExample() {
+  return <NxBinaryDonutChart percent={0} />;
```
rpokorny, line 11: This could just be merged into the "minimal" example

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartFullExample.tsx
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxBinaryDonutChart } from '@sonatype/react-shared-components';
+
+export default function NxBinaryDonutChartFullExample() {
+  return <NxBinaryDonutChart percent={100} />;
```
rpokorny, line 11: Redundant with the "mimimal" example

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+          </tr>
+        </tbody>
+      </table>
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Minimal Examples"
+                        id="nx-binary-donut-chart-minimal-examples"
+                        codeExamples={nxBinaryDonutChartMinimalExampleCode}
+                        liveExample={NxBinaryDonutChartMinimalExample}>
+        Minimal examples of a <code className="nx-code">NxBinaryDonutChart</code> with different values.
```
rpokorny, line 63:
```suggestion
        Minimal examples of <code className="nx-code">NxBinaryDonutChart</code>s with different values.
```

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React, { forwardRef } from 'react';
+import { Props, propTypes } from './types';
+export { Props } from './types';
+
+import './NxBinaryDonutChart.scss';
+import classnames from 'classnames';
+
+const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
+    function NxBinaryDonutChart({className, ...props}, ref) {
+      const { innerRadiusPercent, percent, ...svgAttrs } = props;
```
rpokorny, line 16:
```suggestion
    function NxBinaryDonutChart(props, ref) {
      const { innerRadiusPercent, percent, className, ...svgAttrs } = props;
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+        + ' ' + (-75 * Math.sin(1.5 * Math.PI)));
+    expect(paths.at(0)).toHaveProp('strokeWidth', 50);
+  });
+
+  it('renders only an arc circle with the expected properties when percent is 72', function() {
+    const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, { percent: 72 });
+    const circles = getShallowComponent().find('circle');
+    const paths = getShallowComponent().find('path');
+
+    const arcEnd = Math.PI / 2 + 1.44 * Math.PI;
```
rpokorny, line 76:
```suggestion
    // (2* percent + 0.5)π
    const arcEnd = 1.94 * Math.PI;
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+        + ' ' + (-75 * Math.sin(arcEnd)));
+    expect(paths.at(0)).toHaveProp('strokeWidth', 50);
+  });
+
+  it('renders only an arc circle with the expected properties when percent is 11', function() {
+    const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, { percent: 11 });
+    const circles = getShallowComponent().find('circle');
+    const paths = getShallowComponent().find('path');
+
+    const arcEnd = Math.PI / 2 + 0.22 * Math.PI;
```
rpokorny, line 94:
```suggestion
    // (2*percent + 0.5)π
    const arcEnd = 0.72 * Math.PI;
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
+import NxBinaryDonutChart from '../NxBinaryDonutChart';
+
+describe('NxBinaryDonutChart', function() {
+  it('renders an svg with the expected properties', function() {
+    const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, { percent: 90 });
```
rpokorny, line 12: This isn't really the right/most convenient way to use `getShallowComponent`.  See `NxCloseButton.test.tsx` for a simple set of examples.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartEmptyExample.tsx
mr-zepol: > This could just be merged into the "minimal" example

Done 76104c42b68ca9b5b18665b3149a349594ed02e3

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartFullExample.tsx
mr-zepol: > Redundant with the "mimimal" example

Removed 76104c42b68ca9b5b18665b3149a349594ed02e3

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    expect(circles.at(0)).toHaveProp('strokeWidth', 50);
+    expect(circles.at(0)).toHaveProp('r', 75);
+    expect(paths.length).toBe(1);
+    expect(paths.at(0)).toHaveClassName('.nx-binary-donut-chart__arc');
+    expect(paths.at(0)).toHaveProp('d', 'M 0 -75 A 75 75 0 0 0 ' + (75 * Math.cos(1.5 * Math.PI))
+        + ' ' + (-75 * Math.sin(1.5 * Math.PI)));
+    expect(paths.at(0)).toHaveProp('strokeWidth', 50);
+  });
+
+  it('renders only an arc circle with the expected properties when percent is 72', function() {
```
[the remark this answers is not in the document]
mr-zepol: > This test name isn't quite right - it doesn't only render the arc circle, it renders the background circle too. Same with the test below

Changed the description 76104c42b68ca9b5b18665b3149a349594ed02e3

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
mr-zepol: > This isn't really the right/most convenient way to use `getShallowComponent`. See `NxCloseButton.test.tsx` for a simple set of examples.

Done 76104c42b68ca9b5b18665b3149a349594ed02e3

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    expect(circles.length).toBe(1);
+    expect(circles.at(0)).toHaveClassName('.nx-binary-donut-chart__background');
+    expect(circles.at(0)).toHaveProp('strokeWidth', 50);
+    expect(circles.at(0)).toHaveProp('r', 75);
+    expect(paths.length).toBe(0);
+  });
+
+  it('renders a background circle and arc path with expected properties when 0 < percent < 100', function() {
+    const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, { percent: 50 });
```
rpokorny, line 53: Don't need this line anymore
```suggestion
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    expect(paths.length).toBe(1);
+    expect(paths.at(0)).toHaveClassName('.nx-binary-donut-chart__arc');
+    expect(paths.at(0)).toHaveProp('d', 'M 0 -75 A 75 75 0 0 0 ' + (75 * Math.cos(1.5 * Math.PI))
+        + ' ' + (-75 * Math.sin(1.5 * Math.PI)));
+    expect(paths.at(0)).toHaveProp('strokeWidth', 50);
+  });
+
+  it('renders an arc circle and its background with the expected properties when percent is 72', function() {
+    const circles = getShallowComponent({ percent: 72}).find('circle');
+    const paths = getShallowComponent({ percent: 72}).find('path');
```
rpokorny, line 70: Unbalanced spacing
```suggestion
    const circles = getShallowComponent({ percent: 72 }).find('circle');
    const paths = getShallowComponent({ percent: 72 }).find('path');
```

rpokorny (approved):
+1 aside from last few minor comments.  Btw, I see you rebased; we prefer merging over rebasing in this repo.

PR #173, opened 2020-10-16
Tile footer spacing bug RSC-327

glenhunter:
https://issues.sonatype.org/browse/RSC-327

Simple fix to remove the bottom padding of `nx-footer` when it appears with `nx-tile`

### DIFF
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -22,6 +22,10 @@ $nx-tile-background-color: $nx-white;
   border-radius: $nx-border-radius;
   margin: $nx-spacing-xs 0 $nx-spacing-md 0;
   padding: $nx-spacing-l;
+
+  .nx-footer {
+    padding-bottom: 0;
+  }
 }
 
 // Special tile for containing grids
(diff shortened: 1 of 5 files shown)


lib/src/base-styles/_nx-tile.scss
```diff
@@ -22,6 +22,10 @@ $nx-tile-background-color: $nx-white;
   border-radius: $nx-border-radius;
   margin: $nx-spacing-xs 0 $nx-spacing-md 0;
   padding: $nx-spacing-l;
+
+  .nx-footer {
+    padding-bottom: 0;
```
rpokorny, line 27: Should the footer's bottom padding have actually been margin?  There's no border on that end which suggests, by our standard padding and margin rules, that it should have been.  In that case the `container-vertical` on `.nx-tile` would have taken care of this.

rpokorny: > Should the footer's bottom padding have actually been margin? There's no border on that end which suggests, by our standard padding and margin rules, that it should have been. In that case the container-vertical on .nx-tile would have taken care of this.

On second thought, why does the footer ever have bottom spacing?  It's meant to always be the last thing in a container which should remove any bottom spacing it has anyway.

glenhunter: > > Should the footer's bottom padding have actually been margin? There's no border on that end which suggests, by our standard padding and margin rules, that it should have been. In that case the container-vertical on .nx-tile would have taken care of this.
> 
> On second thought, why does the footer ever have bottom spacing? It's meant to always be the last thing in a container which should remove any bottom spacing it has anyway.

I assumed that since you created the bug and did the `nx-footer` consideration you knew a reason why only this particular application of `nx-footer` was wrong. I agree with your points above.

PR #176, opened 2020-10-20
rsc colours blues RSC-262 

glenhunter:
https://issues.sonatype.org/browse/RSC-262

I created specific variables for hover, selected, and active states which are consistent across a few components. There were a couple of places where some border colours were duplicated but I didn't feel that they were consistent enough to create a unique variable.

### DIFF
--- a/lib/src/base-styles/_nx-alert.scss
+++ b/lib/src/base-styles/_nx-alert.scss
@@ -66,7 +66,7 @@
 }
 
 .nx-alert--info {
-  background-color: #f1f7ff;
+  background-color: $nx-blue-50;
   border-color: $nx-blue-600;
 
   > .nx-icon {
--- a/lib/src/base-styles/_nx-alert.scss
+++ b/lib/src/base-styles/_nx-alert.scss
@@ -67,10 +67,10 @@
 
 .nx-alert--info {
   background-color: #f1f7ff;
-  border-color: #045dd1;
+  border-color: $nx-blue-600;
 
   > .nx-icon {
-    color: #045dd1;
+    color: $nx-blue-600;
   }
 }
 
--- a/lib/src/scss-shared/_nx-colors.scss
+++ b/lib/src/scss-shared/_nx-colors.scss
@@ -117,4 +117,3 @@ $nx-threat-color-low-light: #92cbff;
 $nx-threat-color-moderate-light: #fdecaf;
 $nx-threat-color-severe-light: #ffd6a2;
 $nx-threat-color-critical-light: #ff9493;
-
--- a/lib/src/scss-shared/_nx-colors.scss
+++ b/lib/src/scss-shared/_nx-colors.scss
@@ -81,7 +81,7 @@ $nx-text-color-disabled: #949494;
 $nx-text-color-error: $nx-fail-color;
 
 $nx-link-color: $nx-blue-600;
-$nx-link-color-active: #2783fb;
+$nx-link-color-active: $nx-blue-400;
 $nx-link-color-hover: $nx-blue-800;
 $nx-link-color-disabled: #949494;
 
--- a/lib/src/scss-shared/_nx-colors.scss
+++ b/lib/src/scss-shared/_nx-colors.scss
@@ -105,7 +105,7 @@ $nx-border-row-dark: 1px solid $nx-grey-5;
 
 // Threat colors
 $nx-threat-color-unspecified: #d1d6f0;
-$nx-threat-color-none: #c5deff;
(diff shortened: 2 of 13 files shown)


dturner32:
> https://issues.sonatype.org/browse/RSC-262
> 
> There are some commonalities that go across different components, for example in lists and tables the `active`, `selected`, and `hover` state background colours are the same. We could create default colour variables for these states: `$nx-active-background` and `$nx-selected-background`. When we get to the grey colours most of the disabled states are the same across _all_ components so we could have standard variables for disabled borders, backgrounds, and text colours.
> 
> What do you think?

that makes sense to me

lib/src/base-styles/_nx-alert.scss
```diff
@@ -66,11 +66,11 @@
 }
 
 .nx-alert--info {
-  background-color: #f1f7ff;
-  border-color: #045dd1;
+  background-color: $nx-blue-50;
+  border-color: $nx-blue-600;
```
rpokorny, line 70: I see that these numbers are copied from zeplin, but what do they mean/measure?

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -114,3 +117,4 @@ $nx-threat-color-low-light: #92cbff;
 $nx-threat-color-moderate-light: #fdecaf;
 $nx-threat-color-severe-light: #ffd6a2;
 $nx-threat-color-critical-light: #ff9493;
+
```
rpokorny, line 120: Extraneous blank line

rpokorny:
So we aren't doing the semantic variable names yet?

lib/src/base-styles/_nx-alert.scss
glenhunter: > I see that these numbers are copied from zeplin, but what do they mean/measure?

I was under the impression that those numbers were essentially arbitrary, but the scale would be consistent between colours so we'd know that towards 1000 is darker and towards 0 is lighter. @dturner32 ?

rpokorny: > but the scale would be consistent between colours so we'd know that towards 1000 is darker and towards 0 is lighter.

If that were the case I'd think we'd use adjacent numbers.  Or is the idea to build in gaps in case we want in-between colors in the future?

dturner32: those numbers are a value scale from lightest to darkest, 50 being the lightest and 1000 being the darkest.

glenhunter: > > but the scale would be consistent between colours so we'd know that towards 1000 is darker and towards 0 is lighter.
> 
> If that were the case I'd think we'd use adjacent numbers. Or is the idea to build in gaps in case we want in-between colors in the future?

Exactly

glenhunter:
> So we aren't doing the semantic variable names yet?

That was the purpose of the draft PR to solicit some opinions on the semantic variables.

lib/src/base-styles/_nx-alert.scss
```diff
@@ -66,11 +66,11 @@
...
 
   > .nx-icon {
-    color: #045dd1;
+    color: $nx-blue-600;
```
rpokorny, line 73: So is the idea that we'll only use the semantic variables for cases where a lot of reuse is expected?

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -77,14 +80,15 @@ $nx-text-color: #474747;
...
+$nx-link-color-hover: $nx-blue-800;
 $nx-link-color-disabled: #949494;
 
-$nx-background-color-interactive: #2476c3;
-$nx-background-color-interactive-active: #0046b1;
-$nx-background-color-interactive-disabled: #97cbee;
+// Basic interactive states
+$nx-selected-background: $nx-blue-200;
+$nx-active-background: $nx-blue-200;
+$nx-hover-background: #f1f3fa;
```
rpokorny, line 91: Is this going to be in the teal palette?

glenhunter: > Is this going to be in the teal palette?

Yes it is, I probably should have removed this, but I'm going to do the grey and then the teal. I'll remove it if you would prefer.

lib/src/base-styles/_nx-alert.scss
glenhunter: > So is the idea that we'll only make semantic variables for cases where a lot of reuse is expected?

Yes.

rpokorny (approved):

PR #179, opened 2020-10-22
NxFilterInput nx-text-input class - RSC-365

rpokorny:
https://issues.sonatype.org/browse/RSC-365

### DIFF
--- a/lib/src/components/NxFilterInput/NxFilterInput.tsx
+++ b/lib/src/components/NxFilterInput/NxFilterInput.tsx
@@ -17,7 +17,7 @@ export { Props } from './types';
 const NxFilterInput = forwardRef<HTMLDivElement, Props>(
     function NxFilterInput(props, ref) {
       const { value, placeholder, onChange, className, inputId, disabled, list, ...otherProps } = props,
-          classes = classnames('nx-filter-input', className, {
+          classes = classnames('nx-filter-input', 'nx-text-input', className, {
             'nx-filter-input--disabled': disabled
           });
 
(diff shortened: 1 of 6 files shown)


lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -17,7 +17,7 @@ export { Props } from './types';
 const NxFilterInput = forwardRef<HTMLDivElement, Props>(
     function NxFilterInput(props, ref) {
       const { value, placeholder, onChange, className, inputId, disabled, list, ...otherProps } = props,
-          classes = classnames('nx-filter-input', className, {
+          classes = classnames('nx-filter-input', 'nx-text-input', className, {
```
glenhunter, line 20: This adds `width: 300px` which contradicts the Zeplin files which show the filters has having a width of `250px`.

glenhunter:
`.nx-filter-input` has `display: inline-block` which is also set in `nx-text-input` so it should be removed from `nx-filter-input`.

rpokorny:
> .nx-filter-input has display: inline-block which is also set in nx-text-input so it should be removed from nx-filter-input.

1025ce33

lib/src/components/NxFilterInput/NxFilterInput.tsx
rpokorny: > This adds width: 300px which contradicts the Zeplin files which show the filters has having a width of 250px.

@dturner32 says the 250 only applies in the table filter header.  Implemented in 7cdb2ee4

glenhunter (approved):

PR #182, opened 2020-10-23
Update NxSubmitMask design RSC-147

glenhunter:
https://issues.sonatype.org/browse/RSC-147

Screen Shot 2020-10-23 at 2 42 43 PM

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.4.1",
+  "version": "2.5.0",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.6.0",
+  "version": "2.5.2",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.5.2",
+  "version": "2.5.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.5.0",
+  "version": "2.6.0",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -44,6 +44,6 @@
 
 .nx-submit-mask--success {
   .nx-submit-mask__message {
-    color: #099843;
+    color: #1a832b;
   }
 }
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -11,7 +11,7 @@
 .nx-submit-mask {
   align-items: center;
 
-  background-color: rgba(#555, 0.75);
+  background-color: rgba(85, 85, 85, 0.75);
 
   display: flex;
   justify-content: center;
@@ -35,6 +35,7 @@
 
   background-color: #fff;
   border-radius: $nx-border-radius;
+  box-sizing: border-box;
   margin: 0;
   padding: $nx-spacing-l;
   min-width: 124px;
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -11,7 +11,7 @@
 .nx-submit-mask {
   align-items: center;
 
-  background-color: rgba(85, 85, 85, 0.75);
+  background-color: transparentize(#555555, 0.25);
 
   display: flex;
   justify-content: center;
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -6,11 +6,12 @@
  */
 @import '../../scss-shared/_nx-variables';
 @import '../../scss-shared/_nx-container-helpers';
+@import '../../scss-shared/_nx-text-helpers';
 
 .nx-submit-mask {
   align-items: center;
 
-  background-color: rgba($nx-off-white, 0.9);
+  background-color: rgba(#555, 0.75);
 
   display: flex;
   justify-content: center;
@@ -30,17 +31,18 @@
 
 .nx-submit-mask__message {
   @include container-vertical;
+  @include semi-bold();
 
-  background-color: $nx-off-white;
-  border: 2px solid $nx-slate;
-  box-shadow: -4px 4px 0px rgba($nx-slate, 0.65);
+  background-color: #fff;
+  border-radius: $nx-border-radius;
   margin: 0;
-  padding: 22px 12px;
+  padding: $nx-spacing-l;
+  min-width: 124px;
+  max-width: 250px;
 }
 
 .nx-submit-mask--success {
   .nx-submit-mask__message {
-    border-color: $nx-dark-green;
-    color: $nx-dark-green;
+    color: #099843;
   }
 }
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -27,6 +27,11 @@
 
 .nx-submit-mask--fullscreen {
   position: fixed;
+
+  .nx-submit-mask__message {
+    position: absolute;
+    top: 200px;
+  }
 }
 
 .nx-submit-mask__message {
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -44,6 +44,7 @@
   box-sizing: border-box;
   margin: 0;
   padding: $nx-spacing-l;
+  text-align: center;
   min-width: 124px;
   max-width: 250px;
 }
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -30,6 +30,7 @@
 }
 
 .nx-submit-mask__message {
+  @include container-horizontal;
   @include container-vertical;
   @include semi-bold();
 
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.scss
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.scss
@@ -29,8 +29,8 @@
   position: fixed;
 
   .nx-submit-mask__message {
-    position: absolute;
-    top: 200px;
+    align-self: start;
+    margin-top: 200px;
   }
 }
 
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
@@ -33,7 +33,7 @@ const NxSubmitMask: FunctionComponent<Props> =
           { success ?
             <>
               <NxFontAwesomeIcon icon={faCheckCircle} className="nx-loading-spinner__icon" />
-              <span>{showMessage}</span>
+              <span className="nx-submit-mask__message-text">{showMessage}</span>
             </> :
             <NxLoadingSpinner>{showMessage || 'Submitting...'}</NxLoadingSpinner>
           }
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
@@ -25,8 +25,7 @@ const NxSubmitMask: FunctionComponent<Props> =
           'nx-submit-mask--success': !!success
         }),
 
-        // the default Loading... message is provided by NxLoadingSpinner
-        spinnerMessage = success ? (successMessage || 'Success!') : message;
+        showMessage = success ? (successMessage || 'Success!') : message;
 
     return (
       <div className={classes}>
@@ -34,9 +33,9 @@ const NxSubmitMask: FunctionComponent<Props> =
           { success ?
             <>
               <NxFontAwesomeIcon icon={faCheckCircle} className="nx-loading-spinner__icon" />
-              <span>{spinnerMessage || 'Loading…'}</span>
+              <span>{showMessage}</span>
             </> :
-            <NxLoadingSpinner>{spinnerMessage}</NxLoadingSpinner>
+            <NxLoadingSpinner>{showMessage || 'Submitting...'}</NxLoadingSpinner>
           }
         </div>
       </div>
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
@@ -25,7 +25,8 @@ const NxSubmitMask: FunctionComponent<Props> =
           'nx-submit-mask--success': !!success
         }),
 
-        showMessage = success ? (successMessage || 'Success!') : message;
+        showSuccessMessage = (successMessage || 'Success!'),
+        showSubmittingMessage = (message || 'Submitting…');
 
     return (
       <div className={classes}>
@@ -33,9 +34,9 @@ const NxSubmitMask: FunctionComponent<Props> =
           { success ?
             <>
               <NxFontAwesomeIcon icon={faCheckCircle} className="nx-loading-spinner__icon" />
-              <span className="nx-submit-mask__message-text">{showMessage}</span>
+              <span className="nx-submit-mask__message-text">{showSuccessMessage}</span>
             </> :
-            <NxLoadingSpinner>{showMessage || 'Submitting…'}</NxLoadingSpinner>
+            <NxLoadingSpinner>{showSubmittingMessage}</NxLoadingSpinner>
           }
         </div>
       </div>
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
@@ -23,10 +23,7 @@ const NxSubmitMask: FunctionComponent<Props> =
     const classes = classnames('nx-submit-mask', {
           'nx-submit-mask--fullscreen': !!fullscreen,
           'nx-submit-mask--success': !!success
-        }),
-
-        showSuccessMessage = (successMessage || 'Success!'),
-        showSubmittingMessage = (message || 'Submitting…');
+        })
 
     return (
       <div className={classes}>
@@ -34,9 +31,9 @@ const NxSubmitMask: FunctionComponent<Props> =
           { success ?
             <>
               <NxFontAwesomeIcon icon={faCheckCircle} className="nx-loading-spinner__icon" />
-              <span className="nx-submit-mask__message-text">{showSuccessMessage}</span>
+              <span className="nx-submit-mask__message-text">{successMessage || 'Success!'}</span>
             </> :
-            <NxLoadingSpinner>{showSubmittingMessage}</NxLoadingSpinner>
+            <NxLoadingSpinner>{message || 'Submitting…'}</NxLoadingSpinner>
           }
         </div>
       </div>
--- a/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
+++ b/lib/src/components/NxSubmitMask/NxSubmitMask.tsx
@@ -6,7 +6,9 @@
  */
 import React, { FunctionComponent } from 'react';
 import classnames from 'classnames';
+import { faCheckCircle } from '@fortawesome/free-solid-svg-icons';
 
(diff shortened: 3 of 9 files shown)


rpokorny:
I've been lately second-guessing my original decision to include that goofy mid-line ellipsis in the example.  The point was just to demonstrate that you can change the punctuation, but everyone's first instinct when they see it is that it's a bug.  This would be a good time to change it to something else.

glenhunter:
> I've been lately second-guessing my original decision to include that goofy mid-line ellipsis in the example. The point was just to demonstrate that you can change the punctuation, but everyone's first instinct when they see it is that it's a bug. This would be a good time to change it to something else.

The goofy ellipsis is no more: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/4872c1af58de09e8cbbe621101b432ed7f625531

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.5.0",
+  "version": "2.6.0",
```
rpokorny, line 3: Shouldn't this be a patch version?

lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -6,11 +6,12 @@
  */
 @import '../../scss-shared/_nx-variables';
 @import '../../scss-shared/_nx-container-helpers';
+@import '../../scss-shared/_nx-text-helpers';
 
 .nx-submit-mask {
   align-items: center;
 
-  background-color: rgba($nx-off-white, 0.9);
+  background-color: rgba(#555, 0.75);
```
rpokorny, line 14: I don't think this is technically valid css syntax, but scss seems to be converting it so I guess it's fine

lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -30,17 +31,18 @@
...
 
-  background-color: $nx-off-white;
-  border: 2px solid $nx-slate;
-  box-shadow: -4px 4px 0px rgba($nx-slate, 0.65);
+  background-color: #fff;
+  border-radius: $nx-border-radius;
   margin: 0;
-  padding: 22px 12px;
+  padding: $nx-spacing-l;
+  min-width: 124px;
```
rpokorny, line 40: This was intended as a border-box width but it's currently the content-box width

rpokorny:
Should the content be centered when it is less than the minimum width?  It isn't currently.

lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -30,17 +31,18 @@
...
+  max-width: 250px;
 }
 
 .nx-submit-mask--success {
   .nx-submit-mask__message {
-    border-color: $nx-dark-green;
-    color: $nx-dark-green;
+    color: #099843;
```
rpokorny, line 46: I'm seeing #1a832b in zeplin

rpokorny:
Success icon has left margin; it shouldn't
image

rpokorny:
The Placement/Layout section of the mockups indicates that the message box should be 200px from the top of the viewport.  Fullscreen submit message boxes are currently centered in the viewport.  I'm also wondering if @dturner32 is aware that the masks are not fullscreen by default, and whether he intends that we continue to support that or that we drop it.

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
```diff
@@ -28,9 +30,15 @@ const NxSubmitMask: FunctionComponent<Props> =
...
     return (
       <div className={classes}>
-        <h3 className="nx-h3 nx-submit-mask__message">
-          <NxLoadingSpinner>{spinnerMessage}</NxLoadingSpinner>
-        </h3>
+        <div className="nx-submit-mask__message">
+          { success ?
+            <>
+              <NxFontAwesomeIcon icon={faCheckCircle} className="nx-loading-spinner__icon" />
+              <span>{spinnerMessage || 'Loading…'}</span>
```
rpokorny, line 37: This `|| 'Loading...'` is wrong.  This is the success message here.  At the same time, `spinnerMessage` will always have a truthy value here based on the way it's created above.

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
```diff
@@ -28,9 +30,15 @@ const NxSubmitMask: FunctionComponent<Props> =
...
+            </> :
+            <NxLoadingSpinner>{spinnerMessage}</NxLoadingSpinner>
```
rpokorny, line 39: I think the variables are a little bit tangled up.  What you have works but it uses variables names that aren't quite accurate to what the variables do.  What I'd do instead is pass the `message` prop directly to the `<NxLoadingSpinner>`, and then instead of having a variable called `spinnerMessage`, use the `successMessage` prop in the JSX construction for the success case, with the `|| 'Success!'` part in there for the default.

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
```diff
@@ -42,17 +41,18 @@ describe('NxSubmitMask', function() {
 
   it('passes the successMessage prop as the child of NxLoadingSpinner when the success prop is true', function() {
     const spinnerChildren = shallow(<NxSubmitMask message="foo" successMessage="bar" success />)
-        .find(NxLoadingSpinner).children();
+        .find('div.nx-submit-mask__message span').children();
```
rpokorny, line 44: Selecting unclassed elements like this isn't terribly maintainable.  I'd rather there be a class added to the span.

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
```diff
@@ -11,14 +11,13 @@ import NxLoadingSpinner from '../../NxLoadingSpinner/NxLoadingSpinner';
...
+  it('renders a .nx-submit-mask div containing a .nx-submit-mask__message div containing a NxLoadingSpinner',
       function() {
         const component = shallow(<NxSubmitMask/>);
 
         expect(component).toMatchSelector('div.nx-submit-mask');
-        expect(component.find('h3')).toHaveClassName('nx-h3');
-        expect(component.find('h3')).toHaveClassName('nx-submit-mask__message');
-        expect(component.find('h3').find(NxLoadingSpinner)).toExist();
+        expect(component.find('div.nx-submit-mask div')).toHaveClassName('nx-submit-mask__message');
+        expect(component.find('div.nx-submit-mask__message').find(NxLoadingSpinner)).toExist();
```
rpokorny, line 20: Btw you don't need to add the `div` part at the beginning, you can start with just the classname `.nx-submit-mask__message`.  That's what we'd typically do

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
```diff
@@ -42,17 +41,18 @@ describe('NxSubmitMask', function() {
 
   it('passes the successMessage prop as the child of NxLoadingSpinner when the success prop is true', function() {
```
rpokorny, line 42: This test name is no longer accurate

glenhunter: > This test name is no longer accurate

https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/1b1a50bff29509928cd95a565554b8b8af968045

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
glenhunter: > Btw you don't need to add the `div` part at the beginning, you can start with just the classname `.nx-submit-mask__message`. That's what we'd typically do

It didn't work for me before, but now it does so I must have screwed something else up: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/7eaad8766f8001a37f8fff6f890d10ea7a8be8af

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
glenhunter: > Selecting unclassed elements like this isn't terribly maintainable. I'd rather there be a class added to the span.

Done: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/7eaad8766f8001a37f8fff6f890d10ea7a8be8af

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
glenhunter: > I think the variables are a little bit tangled up. What you have works but it uses variables names that aren't quite accurate to what the variables do. What I'd do instead is pass the `message` prop directly to the `<NxLoadingSpinner>`, and then instead of having a variable called `spinnerMessage`, use the `successMessage` prop in the JSX construction for the success case, with the `|| 'Success!'` part in there for the default.

I did a bunch of work on it here, I think it's much better now: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/7eaad8766f8001a37f8fff6f890d10ea7a8be8af

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
glenhunter: > This `|| 'Loading...'` is wrong. This is the success message here. At the same time, `spinnerMessage` will always have a truthy value here based on the way it's created above.

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/7eaad8766f8001a37f8fff6f890d10ea7a8be8af

glenhunter:
> The Placement/Layout section of the mockups indicates that the message box should be 200px from the top of the viewport. Fullscreen submit message boxes are currently centered in the viewport. I'm also wondering if @dturner32 is aware that the masks are not fullscreen by default, and whether he intends that we continue to support that or that we drop it.

I adjusted it so that the message is now `200px` from the top in fullscreen mode.

glenhunter:
> Success icon has left margin; it shouldn't
> image

I adjusted this, and also fixed the icon right spacing here and in NxLoadingSpinner since I assume it should be the same, right @dturner32 ? https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/ffc08f8c85b6cebc446c70318c0baaa27faf0011

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > I'm seeing #1a832b in zeplin

I remember grabbing the green from Zeplin, but it's clearly wrong so it's fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/a501adad683c58369ef36082dd474034170848da

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > I don't think this is technically valid css syntax, but scss seems to be converting it so I guess it's fine

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/12f0e8dbc13f4e55843306be3219208c4a9eaba3

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > This was intended as a border-box width but it's currently the content-box width

Fixed here by changing it to `border-box`: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/12f0e8dbc13f4e55843306be3219208c4a9eaba3

gallery/package.json
glenhunter: > Shouldn't this be a patch version?

Yeah it should: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/aac0a6d958e152b02a2c6080a060e75e9cb52bd2

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > > I don't think this is technically valid css syntax, but scss seems to be converting it so I guess it's fine
> 
> Fixed here: [12f0e8d](https://github.com/sonatype/sonatype-react-shared-components/commit/12f0e8dbc13f4e55843306be3219208c4a9eaba3)

This is a better fix, more consistent with how we do it elsewhere: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/eebce7f8a2e4002495a1b1d2ece4669d033a5c15

gallery/package.json
rpokorny: > Yeah it should: aac0a6d

Not .1?

lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -26,21 +27,29 @@
 
 .nx-submit-mask--fullscreen {
   position: fixed;
+
+  .nx-submit-mask__message {
+    position: absolute;
```
rpokorny, line 32: It seems slightly weird that the positioning of this element is done absolutely in this case, but using flexbox in the normal case.  Could use `align-self` and `margin-top` to keep it as flexbox in both cases.

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
```diff
@@ -23,14 +25,19 @@ const NxSubmitMask: FunctionComponent<Props> =
...
-        <h3 className="nx-h3 nx-submit-mask__message">
-          <NxLoadingSpinner>{spinnerMessage}</NxLoadingSpinner>
-        </h3>
+        <div className="nx-submit-mask__message">
+          { success ?
+            <>
+              <NxFontAwesomeIcon icon={faCheckCircle} className="nx-loading-spinner__icon" />
+              <span className="nx-submit-mask__message-text">{showMessage}</span>
+            </> :
+            <NxLoadingSpinner>{showMessage || 'Submitting...'}</NxLoadingSpinner>
```
rpokorny, line 38:
```suggestion
            <NxLoadingSpinner>{showMessage || 'Submitting…'}</NxLoadingSpinner>
```

rpokorny, line 38: So we are changing the default text here?

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
```diff
@@ -23,14 +25,19 @@ const NxSubmitMask: FunctionComponent<Props> =
           'nx-submit-mask--success': !!success
         }),
 
-        // the default Loading... message is provided by NxLoadingSpinner
-        spinnerMessage = success ? (successMessage || 'Success!') : message;
+        showMessage = success ? (successMessage || 'Success!') : message;
```
rpokorny, line 28: This is still confusing.  There's no reason to conflate the `message` and `successMessage` into the same variable.

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
```diff
@@ -11,14 +11,13 @@ import NxLoadingSpinner from '../../NxLoadingSpinner/NxLoadingSpinner';
...
+  it('renders a .nx-submit-mask div containing a .nx-submit-mask__message div containing a NxLoadingSpinner',
       function() {
         const component = shallow(<NxSubmitMask/>);
 
         expect(component).toMatchSelector('div.nx-submit-mask');
-        expect(component.find('h3')).toHaveClassName('nx-h3');
-        expect(component.find('h3')).toHaveClassName('nx-submit-mask__message');
-        expect(component.find('h3').find(NxLoadingSpinner)).toExist();
+        expect(component.find('.nx-submit-mask div')).toHaveClassName('nx-submit-mask__message');
+        expect(component.find('.nx-submit-mask__message').find(NxLoadingSpinner)).toExist();
```
rpokorny, line 20:
```suggestion
        expect(component.find('.nx-submit-mask  .nx-submit-mask__message').find(NxLoadingSpinner)).toExist();
```

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
```diff
@@ -35,24 +34,23 @@ describe('NxSubmitMask', function() {
...
 
-  it('passes the successMessage prop as the child of NxLoadingSpinner when the success prop is true', function() {
+  it('passes the successMessage prop as the success message text when the success prop is true', function() {
     const spinnerChildren = shallow(<NxSubmitMask message="foo" successMessage="bar" success />)
-        .find(NxLoadingSpinner).children();
+        .find('.nx-submit-mask__message .nx-submit-mask__message-text').children();
 
     expect(spinnerChildren).toHaveText('bar');
     expect(spinnerChildren).not.toHaveText('foo');
   });
```
rpokorny, line 47: The new default text for the `NxLoadingSpinner` ("Submitting") needs a test

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
```diff
@@ -35,24 +34,23 @@ describe('NxSubmitMask', function() {
...
 
   it('passes "Success!" as the child of NxLoadingSpinner when the success is true and successMessage is not specified',
       function() {
-        expect(shallow(<NxSubmitMask success />).find(NxLoadingSpinner).children()).toHaveText('Success!');
-        expect(shallow(<NxSubmitMask successMessage={null} success />).find(NxLoadingSpinner).children())
-            .toHaveText('Success!');
+        expect(shallow(<NxSubmitMask success />).find('.nx-submit-mask__message .nx-submit-mask__message-text')
+            .children()).toHaveText('Success!');
+        expect(shallow(<NxSubmitMask successMessage={null} success />)
+            .find('.nx-submit-mask__message .nx-submit-mask__message-text').children()).toHaveText('Success!');
```
rpokorny, line 53:
```suggestion
        expect(shallow(<NxSubmitMask success />).find('.nx-submit-mask__message-text')).toHaveText('Success!');
        expect(shallow(<NxSubmitMask successMessage={null} success />)
            .find('.nx-submit-mask__message-text')).toHaveText('Success!');
```

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
glenhunter: > So we are changing the default text here?

Yes, I think it's odd for the default task of the Submit Mask to be "Loading...", it isn't loading, it's submitting.

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > It seems slightly weird that the positioning of this element is done absolutely in this case, but using flexbox in the normal case. Could use `align-self` and `margin-top` to keep it as flexbox in both cases.

Flexbox aligning: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/457c922130aba6edad12f7121732a4530f493164

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
glenhunter: > This is still confusing. There's no reason to conflate the `message` and `successMessage` into the same variable.

How is this? https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/d88bd9fe8a996785f7bf86a97feed31447f2250d

lib/src/components/NxSubmitMask/__tests__/NxSubmitMask.test.tsx
glenhunter: > The new default text for the `NxLoadingSpinner` ("Submitting") needs a test

https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/b118888cce5bb67e6a8ce8fe9b33ed3f72b9c55f

gallery/package.json
glenhunter: > > Yeah it should: [aac0a6d](https://github.com/sonatype/sonatype-react-shared-components/commit/aac0a6d958e152b02a2c6080a060e75e9cb52bd2)
> 
> Not .1?

You suggested patch instead of minor version so that means I back up from 2.6.0 to 2.5.x, then increment up 0.0.1 from there

rpokorny: > You suggested patch instead of minor version so that means I back up from 2.6.0 to 2.5.x, then increment up 0.0.1 from there

But you went up by 0.0.2:
image

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
rpokorny: > How is this? d88bd9f

Better but the variables names you've chosen seem suggestive of being booleans, when they are actually strings.  You don't really need the variables at all; you can move the `foo || 'bar'` expressions directly into the JSX.

gallery/package.json
glenhunter: > > You suggested patch instead of minor version so that means I back up from 2.6.0 to 2.5.x, then increment up 0.0.1 from there
> 
> But you went up by 0.0.2:

Hmmm, odd, merged from Master and bumped again: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/3e9ac67844b18cbbbb5b4912d04cec8e5220863f

lib/src/components/NxSubmitMask/NxSubmitMask.tsx
glenhunter: > > How is this? [d88bd9f](https://github.com/sonatype/sonatype-react-shared-components/commit/d88bd9fe8a996785f7bf86a97feed31447f2250d)
> 
> Better but the variables names you've chosen seem suggestive of being booleans, when they are actually strings. You don't really need the variables at all; you can move the `foo || 'bar'` expressions directly into the JSX.

Oh cool! I can: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/a1c4908b760082684f29f30f6f839b54d5f3f397

glenhunter:
> Should the content be centered when it is less than the minimum width? It isn't currently.

Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/182/commits/5c7e6fd28e6027c5593e578973c6951fc89a9438

rpokorny (approved):

PR #184, opened 2020-10-26
RSC-369: Pass HtmlInputChange Event

chriswininger:
This adds a second argument to the NxTextInput onChange callback which
receives the originating change event.

This will maintain backwards compatibility for the majority use case
where only the updated value is desired but still allow for other use
cases where additional information from the event is needed

### DIFF
--- a/lib/src/components/NxTextInput/NxTextInput.tsx
+++ b/lib/src/components/NxTextInput/NxTextInput.tsx
@@ -12,11 +12,9 @@ import { faExclamationCircle, faCheck } from '@fortawesome/free-solid-svg-icons'
 import './NxTextInput.scss';
 
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
-import { Props, propTypes } from './types';
+import { Props, propTypes, TextInputElement } from './types';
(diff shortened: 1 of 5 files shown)


lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -96,7 +94,7 @@ const NxTextInput = forwardRef<TextInputElement, Props>(
 
       function inputOnChange(e: FormEvent<TextInputElement>) {
         if (onChange) {
-          onChange(e.currentTarget.value);
+          onChange(e.currentTarget.value, e);
```
Blacktiger, line 97: Personally, I think these should be flipped so that we can use the `onChange` just like the normal event from React. That would make this a breaking change though, so I guess this is fine.

Blacktiger (approved):

rpokorny (approved):
Thanks for the contribution!

lib/src/components/NxTextInput/NxTextInput.tsx
chriswininger: I can definitely see that argument, but seems like keeping backwards compatibility trumps that, plus I buy Ross's idea that in many cases it would be better to use the abstracted value

PR #185, opened 2020-10-26
Update RSC min max res RSC-368

glenhunter:
https://issues.sonatype.org/browse/RSC-368

Update RSC with the new page width values from design: 1366px and 1600px.

Also includes ensuring that we have horizontal scrolling below `1366px`


rpokorny:
> Also includes ensuring that we have horizontal scrolling below 1366px

That's not working for me. 
localhost_4043_(1024x768) (1)

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.5.1",
+  "version": "2.5.2",
```
rpokorny, line 3: I feel like this creates changes that are visible enough that it warrants a minor version

lib/src/base-styles/_nx-responsive.scss
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+@media (max-width: 1365px) {
+  .nx-body {
+    overflow: auto;
```
rpokorny, line 10: Does this really need a media query?  Wouldn't it work to just always have this style in conjunction with the `min-width` and `max-width`?

rpokorny:
Did you confirm with design that we still want the 24/76 split between the sidebar and main content? Also that this results in tiles and things being up to 1552px wide (when there isn't a sidebar).

glenhunter:
> Did you confirm with design that we still want the 24/76 split between the sidebar and main content? Also that this results in tiles and things being up to 1552px wide (when there isn't a sidebar).

I did not. I expected @dturner32 to understand all of that. If the 24/76 split is changing I would also expect that to be a separate ticket.

glenhunter:
> > Also includes ensuring that we have horizontal scrolling below 1366px
> 
> That's not working for me.

One thing that I noticed and cannot figure out how to change is the location of the horizontal scrollbar - it is at the bottom of the page so you must scroll down to see it. With mice that allow horizontal scrolling the page can be scrolled horizontally from any point.

gallery/package.json
glenhunter: > I feel like this creates changes that are visible enough that it warrants a minor version

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/185/commits/2446d608fbf8db4370111dc18fffe7b11cef6cb3

lib/src/base-styles/_nx-responsive.scss
glenhunter: > Does this really need a media query? Wouldn't it work to just always have this style in conjunction with the `min-width` and `max-width`?

Not only do we not need the media query, we don't even need the `overflow: hidden` value that it overrides. https://github.com/sonatype/sonatype-react-shared-components/pull/185/commits/8d95298f9b3e84036d29fe26a1dbdc521051aebc

rpokorny:
> One thing that I noticed and cannot figure out how to change is the location of the horizontal scrollbar when page scrolling is enabled (as it is by default in the gallery) - it is at the bottom of the page so you must scroll down to see it. With mice that enable horizontal scrolling the page can be scrolled horizontally from any point.

I think I've found several ways to fix that.  The cleanest is probably to use `overflow: visible` on the `<html>` and `<body>`.  Alternatively, applying `width: fit-content` to the body (not supported in IE), or `position: absolute` to the `<body>`.

glenhunter:
> > One thing that I noticed and cannot figure out how to change is the location of the horizontal scrollbar when page scrolling is enabled (as it is by default in the gallery) - it is at the bottom of the page so you must scroll down to see it. With mice that enable horizontal scrolling the page can be scrolled horizontally from any point.
> 
> I think I've found several ways to fix that. The cleanest is probably to use `overflow: visible` on the `<html>` and `<body>`. Alternatively, applying `width: fit-content` to the body (not supported in IE), or `position: absolute` to the `<body>`.

I am annoyed that I missed `overflow: visible`. It only needs to be on `nx-body`. https://github.com/sonatype/sonatype-react-shared-components/pull/185/commits/6c923b6595a0a44576b5549afb1af8314d6dd6d6

Interestingly `position: absolute;` essentially turned the page scrolling layout into a non-page scrolling layout.

rpokorny (approved):

PR #187, opened 2020-10-28
Add default bottom margin to text inputs RSC-357

glenhunter:
https://issues.sonatype.org/browse/RSC-357

I used the same bottom margin (32px) as for the form groups because we know it accommodates the validation message plus a bit of extra space.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.6.1",
+  "version": "2.6.2",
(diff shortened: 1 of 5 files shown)


rpokorny:
Needs to be cancelled out when in a form group.  Margin is doubled up in the form row example.

glenhunter:
> Needs to be cancelled out when in a form group. Margin is doubled up in the form row example.

https://github.com/sonatype/sonatype-react-shared-components/pull/187/commits/888abaf951cd5956671ebde0eb072dd60dccb263

rpokorny (approved):

PR #190, opened 2020-11-02
Pagination Controls - RSC-181

rpokorny:
Introduces two components: `NxPagination` for when you know how many pages there are, and `NxIndeterminatePagination` for when you don't (i.e. just back and forward arrows).
image
image


glenhunter:
I would expect that when I'm on the last page (41 in this example) that I would see 40, 39, 38, etc, as it appears when I'm on page 40.

Screen Shot 2020-11-02 at 1 42 15 PM
Screen Shot 2020-11-02 at 1 42 11 PM

gallery/src/components/NxPagination/NxPaginationPage.tsx
```diff
@@ -0,0 +1,142 @@
...
+            A button for the current page will always be visible and is distinguished via a dark blue background.
+            Clicking this button has no effect - the component's <code className="nx-code">onChange</code> handler
+            does not get triggered.
+          </span>
+        </li>
+        <li className="nx-list__item">
+          <span className="nx-list__text">Next/previous page buttons</span>
+          <span className="nx-list__subtext">
+            Whenever there are pages either above or below the current page, arrow buttons are rendered as
+            applicable to navigate the pages sequentially.
```
glenhunter, line 43:
```suggestion
            Whenever there are pages either above or below the current page, arrow buttons are rendered and can be used to navigate the pages sequentially.
```

gallery/src/components/NxPagination/NxPaginationPage.tsx
```diff
@@ -0,0 +1,142 @@
...
+          </span>
+        </li>
+        <li className="nx-list__item">
+          <span className="nx-list__text">Nearby page buttons</span>
+          <span className="nx-list__subtext">
+            Buttons for up to 4 additional pages that precede or follow the current page will be present as
+            applicable. The lowest "nearby page" to get a rendered button always has a page index that is one
```
glenhunter, line 50:
```suggestion
            Buttons for up to 4 additional pages that precede or follow the current page will be present. The lowest "nearby page" to get a rendered button always has a page index that is one
```

lib/src/base-styles/_nx-pagination.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  &.selected {
+    background-color: #232c50;
+    border-color: #232c50;
+    color: #fff;
+  }
+
+  &:focus {
+    border: solid 1px #02a6d9;
+    box-shadow: 0 0 3px 1px rgba(2, 166, 217, 0.3);
```
glenhunter, line 47: I thought we were trying to standardize on using `transparentize` with the hex value for these?

glenhunter: Also this is not the default `:focus` state for all buttons so I'm not sure if it would be better to remove it here and implement it to `.nx-btn` as a separate ticket, or if we should implement it in `.nx-btn` here and now. I did create a ticket for the `.nx-btn` `:focus` states here: https://issues.sonatype.org/browse/RSC-384

lib/src/base-styles/_nx-pagination.scss
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+.nx-btn-bar--pagination {
+  text-align: center;
+
+  .nx-btn {
+    height: 40px;
+    min-width: 40px;
+    margin: 0 $nx-spacing-xs 0 0;
+    padding: 0 $nx-spacing-md;
```
glenhunter, line 15: The buttons are ending up `42px` wide rather than `40px` which I wasn't seeing before but now I am, I'm thinking that we should subtract the borders from the padding as we have done elsewhere.

rpokorny:
> I would expect that when I'm on the last page (41 in this example) that I would see 40, 39, 38, etc, as it appears when I'm on page 40.

So you're saying that you want a special case for when the current page is one greater than a multiple of 5, that it doesn't get it's own group but gets added to the previous group?  I could probably make that happen but it would make things a little more complicated.  @dturner32 what do you think?

gallery/src/components/NxPagination/NxPaginationPage.tsx
rpokorny: >  + Whenever there are pages either above or below the current page, arrow buttons are rendered and can be used to navigate the pages sequentially.

I don't like this suggestion.  I included the phrase "as applicable" deliberately, to imply that you won't get the left arrow when you're on the first page and won't get the right arrow when you're on the last page.

gallery/src/components/NxPagination/NxPaginationPage.tsx
rpokorny: > \+ Buttons for up to 4 additional pages that precede or follow the current page will be present. The lowest "nearby page" to get a rendered button always has a page index that is one

Again, "as applicable" was important here.  If you're close to the end of list of pages you might not get a full 4 nearby pages.

lib/src/base-styles/_nx-pagination.scss
rpokorny: > I thought we were trying to standardize on using transparentize with the hex value for these?

Fixed in 150d6419.  There are now at least three places this exact box-shadow setting is used.  We should probably figure out in what sense it is standard and make a variable for it (maybe even something like `$nx-focus-box-shadow` if it is in fact meant to be used for focus everywhere.

rpokorny: > Also this is now the default :focus state for all buttons so I'm not sure if it would be better to remove it here and implement it to .nx-btn as a separate ticket, or if we should implement it in .nx-btn now. I did create a ticket for the .nx-btn :focus states here: https://issues.sonatype.org/browse/RSC-384

Separate ticket I think. And in that ticket the bespoke impl here would get cleaned up.

When did this become the intended focus state for all buttons?  It's feeling like one of those snuck-in-after-the-fact changes to me which indicates there's still a breakdown in our process.

lib/src/base-styles/_nx-pagination.scss
rpokorny: > The buttons are ending up 42px wide rather than 40px which I wasn't seeing before but now I am, I'm thinking that we should subtract the borders from the padding as we have done elsewhere.

See https://sonatype.slack.com/archives/CJUF9G7T2/p1604064001371900 .  The 40px is a minimum which doesn't really have any effect anywhere now that I've made the arrows slightly larger than they were.

glenhunter:
> > I would expect that when I'm on the last page (41 in this example) that I would see 40, 39, 38, etc, as it appears when I'm on page 40.
> 
> So you're saying that you want a special case for when the current page is one greater than a multiple of 5, that it doesn't get it's own group but gets added to the previous group? I could probably make that happen but it would make things a little more complicated. @dturner32 what do you think?

I don't think that it's a special case, why wouldn't we want the last page to also have the preceding 4 pages listed? It seems much weirder to me to click on 41 and have buttons 36 - 40 disappear.

rpokorny:
> I don't think that it's a special case, why wouldn't we want the last page to also have the preceding 4 pages listed? It seems much weirder to me to click on 41 and have buttons 36 - 40 disappear.

See this message and the two that follow: https://sonatype.slack.com/archives/CJUF9G7T2/p1604000009366500?thread_ts=1603998730.366000&cid=CJUF9G7T2

The decision was that the page groups should always start at one greater than a multiple of 5, and end on a multiple of 5 or the last page, whichever comes first.

dturner32:
> > I would expect that when I'm on the last page (41 in this example) that I would see 40, 39, 38, etc, as it appears when I'm on page 40.
> 
> So you're saying that you want a special case for when the current page is one greater than a multiple of 5, that it doesn't get it's own group but gets added to the previous group? I could probably make that happen but it would make things a little more complicated. @dturner32 what do you think?

looking at the example Glen posted, that makes sense to me to have the final page just be part of that last set of numbers so that this becomes a case where you see 6 numbers in the final set instead of 5.

dturner32:
> > I don't think that it's a special case, why wouldn't we want the last page to also have the preceding 4 pages listed? It seems much weirder to me to click on 41 and have buttons 36 - 40 disappear.
> 
> See this message and the two that follow: https://sonatype.slack.com/archives/CJUF9G7T2/p1604000009366500?thread_ts=1603998730.366000&cid=CJUF9G7T2
> 
> The decision was that the page groups should always start at one greater than a multiple of 5, and end on a multiple of 5 or the last page, whichever comes first.

I still stand with the decision to show whatever the last remaining pages are and they don't necessarily have to be a set of five so if the last set is just 41, 42, 43 i'm fine with that. However, if the last set is just one page number then I would rather have it show up in the previous set than to be on a page by itself. So if you have 41-45 and 46 is the total number of pages, I'd prefer it be added to the previous set so it shows 41-46

rpokorny:
> I still stand with the decision to show whatever the last remaining pages are and they don't necessarily have to be a set of five so if the last set is just 41, 42, 43 i'm fine with that. However, if the last set is just one page number then I would rather have it show up in the previous set than to be on a page by itself. So if you have 41-45 and 46 is the total number of pages, I'd prefer it be added to the previous set so it shows 41-46

Special case applied in 44917f3 and 28a174f

lib/src/components/NxPagination/NxPagination.tsx
```diff
@@ -0,0 +1,138 @@
...
+  else {
+    if (pageCount === 0) {
+      throw new TypeError(`currentPage must not be defined when pageCount is 0, was ${currentPage}`);
+    }
+
+    if (currentPage < 0 || currentPage >= pageCount || !Number.isInteger(currentPage)) {
+      throw new TypeError(`currentPage is out of bounds or not an integer: ${currentPage}`);
+    }
+  }
+}
```
Blacktiger, line 38: The following incorporates several suggestions:

* Checks that the values are numeric before checking the range
* Separates checking the pageCount and currentPage into a single condition
* Always prefer `===` to `==` to avoid type coercion errors
* Avoid template strings when not necessary
* Reduce the number of nested conditions by taking advantage of the early return due to throw statements

```suggestion
  if (!Number.isInteger(pageCount) || pageCount <= 0) {
    throw new TypeError(`pageCount must be > 0 if it is defined, was ${pageCount}`);
  }

  if (pageCount && (currentPage === null || currentPage === undefined)) {
    throw new TypeError('currentPage must be defined when pageCount is > 0');
  }

  if (!Number.isInteger(currentPage) || currentPage < 0) {
    throw new TypeError(`currentPage must be a positive integer when the pageCount was defined, was ${currentPage}`);
  }

  if (currentPage >= pageCount) {
    throw new TypeError(`currentPage must be less than the pageCount, currentPage was ${currentPage} and pageCount was ${pageCount}`);
  }
```

Blacktiger (approved):

lib/src/base-styles/_nx-pagination.scss
glenhunter: > > Also this is now the default :focus state for all buttons so I'm not sure if it would be better to remove it here and implement it to .nx-btn as a separate ticket, or if we should implement it in .nx-btn now. I did create a ticket for the .nx-btn :focus states here: https://issues.sonatype.org/browse/RSC-384
> 
> Separate ticket I think. And in that ticket the bespoke impl here would get cleaned up.
> 
> When did this become the intended focus state for all buttons? It's feeling like one of those snuck-in-after-the-fact changes to me which indicates there's still a breakdown in our process.

I think that @dturner32 was asked to make it the default a little while ago when the `:focus` states settled down into a consistent state.

glenhunter:
I noticed that the selected button still responds to mouse clicks (the background changes, cursor changes), I think we should probably override those styles

lib/src/base-styles/_nx-pagination.scss
dturner32: > > > Also this is now the default :focus state for all buttons so I'm not sure if it would be better to remove it here and implement it to .nx-btn as a separate ticket, or if we should implement it in .nx-btn now. I did create a ticket for the .nx-btn :focus states here: https://issues.sonatype.org/browse/RSC-384
> > 
> > 
> > Separate ticket I think. And in that ticket the bespoke impl here would get cleaned up.
> > When did this become the intended focus state for all buttons? It's feeling like one of those snuck-in-after-the-fact changes to me which indicates there's still a breakdown in our process.
> 
> I think that @dturner32 was asked to make it the default a little while ago when the `:focus` states settled down into a consistent state.

yes, I added the shadow for the focus state a while back. Glen and I had a conversation around this because the focus border on its own wasn't prominent enough in some cases.

lib/src/components/NxPagination/NxPagination.tsx
rpokorny: > The following incorporates several suggestions:

The suggested code doesn't compile.  I also don't necessarily agree with some of the changes:

> Always prefer === to == to avoid type coercion errors

`== null` is explicitly the only time we consider it ok/preferable to use double-equals.  It is the most succinct way to check for both null and undefined.  You'll notice we specifically have our eslint rules set up to allow it.

rpokorny:
> I noticed that the selected button still responds to mouse clicks (the background changes, cursor changes), I think we should probably override those styles

I had assumed it was supposed to, and set it up that way on purpose.  @dturner32 should it?

dturner32:
> > I noticed that the selected button still responds to mouse clicks (the background changes, cursor changes), I think we should probably override those styles
> 
> I had assumed it was supposed to, and set it up that way on purpose. @dturner32 should it?

in selected state the button shouldn't respond to any mouse click/cursor change effects.

lib/src/components/NxPagination/NxPagination.tsx
Blacktiger: Using `==` for null is fair I suppose - it just seems dangerous to me as it would be easy to miss that when changing the code but as long as we have an eslint rule blocking `==` comparisons against anything other than null I don't mind.

I just quickly wrote that code by hand so I wouldn't be surprised if I missed something syntatically. Maybe I shouldn't have used the suggestion feature but I wanted to make it clear what code would be generally replaced.

rpokorny:
> in selected state the button shouldn't respond to any mouse click/cursor change effects.

94372db3

lib/src/base-styles/_nx-pagination.scss
rpokorny: > Separate ticket I think. And in that ticket the bespoke impl here would get cleaned up.

That other ticket has been merged to master and this PR now relies on the default button focus styles.  Note that an additional issue has been found with button focus styles which I have put in as another ticket: https://issues.sonatype.org/browse/RSC-391

glenhunter:
> > in selected state the button shouldn't respond to any mouse click/cursor change effects.
> 
> [94372db](https://github.com/sonatype/sonatype-react-shared-components/commit/94372db32d860047bb1ea0d64d61248240857585)

The cursor is still the pointer on `:hover` and it still appears in the tab order.

glenhunter:
> > > in selected state the button shouldn't respond to any mouse click/cursor change effects.
> > 
> > 
> > [94372db](https://github.com/sonatype/sonatype-react-shared-components/commit/94372db32d860047bb1ea0d64d61248240857585)
> 
> The cursor is still the pointer on `:hover` and it still appears in the tab order. I would expect that as you tabbed through the pagination buttons that it would skip the `selected` button.

lib/src/base-styles/_nx-pagination.scss
rpokorny: > Note that an additional issue has been found with button focus styles which I have put in as another ticket: https://issues.sonatype.org/browse/RSC-391

RSC-291 is now Done and merged into this branch as well.

rpokorny:
> The cursor is still the pointer on :hover and it still appears in the tab order. I would expect that as you tabbed through the pagination buttons that it would skip the selected button.

c7ac2644

glenhunter (approved):

PR #193, opened 2020-11-04
NxDropdown right-floated buttons - RSC-348

rpokorny:
https://issues.sonatype.org/browse/RSC-348

This turned out to be more than just adding an example, but rather creating new styles to make this possible.  As much as I hate floats, this seems like it was actually a correct time to use them due to the fact that we really wanted the new button to be on top of the existing menu row button.


gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
```diff
@@ -14,10 +14,7 @@ function NxDropdownNavigationExample() {
       onClick = () => { alert('click'); };
 
   return (
-    <NxDropdown label="Navigation"
-                className="nx-dropdown--navigation"
```
rpokorny, line 18: This class wasn't doing anything anymore (the CSS props it added were already there) so I removed it.

PR #200, opened 2020-11-06
Consolidate NxTabs code - RSC-293

rpokorny:
https://issues.sonatype.org/browse/RSC-293

We found the separate directories and scss files pretty inconvenient to work with, so this ticket is to merge them.  I left the components as separate files (in the same directory) since the main `NxTabs` file already has a lot going on.


lib/src/components/NxTabs/TabContext.ts
```diff
@@ -1,13 +1,17 @@
-/**
+/*
```
rpokorny, line 1: Lol at git thinking these two files are related

Blacktiger (approved):

glenhunter (approved):

PR #203, opened 2020-11-10
nx-form-select RSC-373

glenhunter:
https://issues.sonatype.org/browse/RSC-373


lib/src/base-styles/_nx-form-select.scss
```diff
@@ -0,0 +1,43 @@
...
+    <polygon points='0,0 100,0 50,50'/></svg>") no-repeat;
+  background-size: 12px;
+  background-position: calc(100% - 16px) 16px;
+  background-repeat: no-repeat;
+  border: 1px solid #2c2c2c;
+  border-radius: $nx-border-radius;
+  color: $nx-text-color;
+  display: inline-block;
+  font-size: $nx-font-size;
+  line-height: 1.5;
```
rpokorny, line 25:
```suggestion
  line-height: $nx-line-height;
```

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -90,6 +90,18 @@ export default function NxFormLayoutExample() {
           Blue
         </NxRadio>
       </fieldset>
+      <div className="nx-form-group">
+        <label className="nx-label">
+          <span className="nx-label__text">Select</span>
+          <select className="nx-form-select">
+            <option className="nx-form-select__option">Option 1</option>
```
rpokorny, line 97: What's the purpose of this class?

gallery/visualtests/nxFormSelect.js
```diff
@@ -0,0 +1,34 @@
...
+    it('has a light border by default', simpleTest(selector));
+
+    it('has a dark border when hovered',
+        hoverTest(selector));
+
+        it('has a blue border when focused',
+        focusTest(selector));
+
+    it('has a blue border when hovered and focused',
+        focusAndHoverTest(selector));
```
rpokorny, line 28: Formatting
```suggestion
    it('has a light border by default', simpleTest(selector));
    it('has a dark border when hovered', hoverTest(selector));
    it('has a blue border when focused', focusTest(selector));
    it('has a blue border when hovered and focused', focusAndHoverTest(selector));
```

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -90,6 +90,18 @@ export default function NxFormLayoutExample() {
           Blue
         </NxRadio>
       </fieldset>
+      <div className="nx-form-group">
```
rpokorny, line 93: We probably need an example in an nx-form-row as well (assuming that's intended to be supported)

lib/src/base-styles/_nx-form-select.scss
```diff
@@ -0,0 +1,47 @@
...
+  background: #fff url("data:image/svg+xml;utf8,
+    <svg xmlns='http://www.w3.org/2000/svg' width='100' height='100' fill='%232c2c2c'>
+    <polygon points='0,0 100,0 50,50'/></svg>") no-repeat;
+  background-size: 12px;
+  background-position: calc(100% - 16px) 16px;
+  background-repeat: no-repeat;
+  border: 1px solid #d1d6f0;
+  border-radius: $nx-border-radius;
+  color: $nx-text-color;
+  display: inline-block;
```
rpokorny, line 23: I think this is redundant

gallery/src/styles/NxFormSelect/NxFormSelectExample.tsx
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+const NxFormSelectExample = () =>
```
rpokorny, line 9: Couldn't these examples be plain HTML?

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -90,6 +90,18 @@ export default function NxFormLayoutExample() {
...
+        <label className="nx-label">
+          <span className="nx-label__text">Select</span>
+          <select className="nx-form-select">
```
rpokorny, line 96: This example should probably use the `select` as a controlled component (e.g. track its value with `useState` like we do for the text inputs).

gallery/src/styles/NxFormSelect/NxFormSelectExample.tsx
glenhunter: > Couldn't these examples be plain HTML?

They could, I've just gotten used to making the examples in React, it's good practice for me. I can switch them to HTML if you think that it matters.

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > We probably need an example in an nx-form-row as well (assuming that's intended to be supported)

https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/b1b146f069e70a322c4dc4df843410c9f09edae5

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > What's the purpose of this class?

I was trying something and forgo to remove it: https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/407d9a451b1dd4719f51e388ffe90c8e88bc7c55

lib/src/base-styles/_nx-form-select.scss
glenhunter: > I think this is redundant

https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/5f7c72d74437bcf124f3d5cee6797ab540644de8

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > This example should probably use the `select` as a controlled component (e.g. track its value with `useState` like we do for the text inputs).

This is probably wrong (or incomplete) but how's this? https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/2ce5822229084a320e3f375e6bb9a074b18b4279

rpokorny: > This is probably wrong (or incomplete) but how's this? 2ce5822

Yeah that's not right.  What you want is a single `useState` that tracks the current value of the `select`.  The value of that state would be set on the `select` using its `value` prop, and the `select`'s `onChange` callback would be used to update the state.

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -90,6 +96,18 @@ export default function NxFormLayoutExample() {
           Blue
         </NxRadio>
       </fieldset>
+      <div className="nx-form-group">
+        <label className="nx-label">
+          <span className="nx-label__text">Select</span>
+          <select className="nx-form-select" value={selectVal} onChange={onSelectChange}>
+            <option>Option 1</option>
```
rpokorny, line 103: The typical way to use `<option>` elements is to give each one a `value` that is more of the machine-readable value whereas the child text is the human readable form.

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -90,6 +96,18 @@ export default function NxFormLayoutExample() {
           Blue
         </NxRadio>
       </fieldset>
+      <div className="nx-form-group">
+        <label className="nx-label">
+          <span className="nx-label__text">Select</span>
+          <select className="nx-form-select" value={selectVal} onChange={onSelectChange}>
```
rpokorny, line 102: Notice that when the page first loads, the `selectVal` variable holds `''` but the select is set to "Option 1".  So at that point is time the state variable and the actual UI are out of sync which is undesirable.  To avoid this, set the default value of the state variable (that is, the value passed into the `useState` call), to a value that is actually one of the valid choices from the `<select>`

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -47,6 +47,26 @@ export default function NxFormLayoutExample() {
...
+        <div className="nx-form-group">
+          <label className="nx-label nx-label--optional">
+            <span className="nx-label__text">Label</span>
+            <NxStatefulTextInput/>
+          </label>
+        </div>
+        <div className="nx-form-group">
+          <label className="nx-label">
+            <span className="nx-label__text">Label</span>
+            <select className="nx-form-select">
```
rpokorny, line 60: This one needs state management too

gallery/visualtests/nxFormSelect.js
```diff
@@ -0,0 +1,28 @@
...
+  beforeEach(async function() {
+    await browser.url('#/pages/nx-form-select');
+  });
+
+  describe('Simple NxFormSelect', function() {
+    it('has a light border by default', simpleTest(selector));
+    it('has a dark border when hovered', hoverTest(selector));
+    it('has a blue border when focused', focusTest(selector));
+    it('has a blue border when hovered and focused', focusAndHoverTest(selector));
+  });
```
rpokorny, line 23: Should we have a test when the menu is open?

glenhunter: > Should we have a test when the menu is open?

I thought about that and didn't think that it was worth it since we don't style the open state in any way so we'd have a visual test for something we don't style that is 100% controlled by the browser

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > The typical way to use `<option>` elements is to give each one a `value` that is more of the machine-readable value whereas the child text is the human readable form.

The task was just to style the `<select>` and I would hope that any developer can figure that out themselves without us having to create an example

glenhunter: > The typical way to use `<option>` elements is to give each one a `value` that is more of the machine-readable value whereas the child text is the human readable form.

https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/a329a17b7cda35bbf29e06fca284f1b3ae6255ea

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > Notice that when the page first loads, the `selectVal` variable holds `''` but the select is set to "Option 1". So at that point is time the state variable and the actual UI are out of sync which is undesirable. To avoid this, set the default value of the state variable (that is, the value passed into the `useState` call), to a value that is actually one of the valid choices from the `<select>`

What I should have done is have a "Pick one" option specified first: https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/a329a17b7cda35bbf29e06fca284f1b3ae6255ea

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
glenhunter: > This one needs state management too

https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/a329a17b7cda35bbf29e06fca284f1b3ae6255ea

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > > This is probably wrong (or incomplete) but how's this? [2ce5822](https://github.com/sonatype/sonatype-react-shared-components/commit/2ce5822229084a320e3f375e6bb9a074b18b4279)
> 
> Yeah that's not right. What you want is a single `useState` that tracks the current value of the `select`. The value of that state would be set on the `select` using its `value` prop, and the `select`'s `onChange` callback would be used to update the state.

https://github.com/sonatype/sonatype-react-shared-components/pull/203/commits/73a62067e8611c9cfb541ead62e8f413ca8e829d

rpokorny (approved):

PR #204, opened 2020-11-11
NxAccordion - RSC-229

rpokorny:
https://issues.sonatype.org/browse/RSC-229


gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -0,0 +1,167 @@
...
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell><code className="nx-code">nx-accordion__header-title</code></NxTableCell>
+              <NxTableCell>First child of <code className="nx-code">NxAccordion.Header</code></NxTableCell>
+              <NxTableCell>
+                It is expected that the first child of <code className="nx-code">NxAccordion.Header</code>
+                will always be a <code className="nx-code">.nx-accordion__header-title</code> containing
+                the text content of the always visible section of the accordion. This element should be
+                and <code className="nx-code">h*</code>  heading. Note however
```
glenhunter, line 104:
```suggestion
                an <code className="nx-code">h*</code>  heading. Note however
```

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <NxTableHead>
+            <NxTableRow>
+              <NxTableCell>Name</NxTableCell>
+              <NxTableCell>Location</NxTableCell>
+              <NxTableCell>Description</NxTableCell>
+            </NxTableRow>
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell><code className="nx-code">nx-accordion__header-title</code></NxTableCell>
```
glenhunter, line 98: I don't understand why we're bothering with this class when we are effectively setting it to the same values as an `h3`. I also don't understand why we'd go to the trouble of stating that it can be any `h#` value when the class overrides the font-size which is the main visual differentiator between the headings, and as mentioned below the semantics are largely irrelevant.

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -0,0 +1,167 @@
...
+              <NxTableCell>Last child of <code className="nx-code">NxAccordion.Header</code></NxTableCell>
+              <NxTableCell>
+                <code className="nx-code">NxAccordion.Header</code> supports the inclusion of buttons on
+                its right-hand side. This is accomplished by adding
+                a <code className="nx-code">.nx-btn-bar</code> after
+                the <code className="nx-code">.nx-accordion__header-title</code>.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell><code className="nx-code">nx-accordion__subheader</code></NxTableCell>
```
glenhunter, line 125: Similar to the header above I don't see any need for this class as it essentially duplicates the `h3` styles. Why not just recommend an `h3`?

glenhunter:
@dturner32 the Zeplin file states that there are no interaction states for the accordion but the implementation demonstrates `:focus` states (which are necessary/correct IMO), and also a cursor change on `:hover` (also necessary/correct IMO). I think the Zelin file should be updated.

dturner32:
> @dturner32 the Zeplin file states that there are no interaction states for the accordion but the implementation demonstrates `:focus` states (which are necessary/correct IMO), and also a cursor change on `:hover` (also necessary/correct IMO). I think the Zelin file should be updated.

right, I know we discussed that yesterday in WW so I will update the zeplin mocks accordingly

gallery/src/components/NxAccordion/NxAccordionPage.tsx
rpokorny: > Similar to the header above I don't see any need for this class as it essentially duplicates the h3 styles. Why not just recommend an h3?

Should the subheader be an `h*`? in places like the nx-tile-header, it isn't.  Semantically I believe having it as another `h3` (after the `h3` in the `summary`) would start a new outline section that's a sibling of the section in the summary. This is not what we want.  Therefore I believe it needs to be a div.  Now, it could be a div with the `nx-h3` class, but I wasn't sure how you'd feel about that and further it would make it harder for us to separate the style of this element from the general h3 style in the future, if that's what we choose to do.

gallery/src/components/NxAccordion/NxAccordionPage.tsx
rpokorny: > I don't understand why we're bothering with this class when we are effectively setting it to the same values as an h3.

TBH I hadn't really noticed that it was the same as `nx-h3`.  I'd be alright with getting rid of this custom class and just using `.nx-h3` here.  As far as the note about the sectioning root goes, I was thinking there are multiple places/ways accordions could be used which might fall at different levels in the outline structure.  For instance an accordion which is _not_ wrapped in a tile would likely be at the `h2` level.  That note is essentially to tell the user not to worry about that, because the sectioning root makes it irrelevant.

rpokorny: > > I don't understand why we're bothering with this class when we are effectively setting it to the same values as an h3.
>
> TBH I hadn't really noticed that it was the same as nx-h3. I'd be alright with getting rid of this custom class and just using .nx-h3 here. As far as the note about the sectioning root goes, I was thinking there are multiple places/ways accordions could be used which might fall at different levels in the outline structure. For instance an accordion which is not wrapped in a tile would likely be at the h2 level. That note is essentially to tell the user not to worry about that, because the sectioning root makes it irrelevant.

1f32fd51

gallery/src/components/NxAccordion/NxAccordionPage.tsx
glenhunter: > > Similar to the header above I don't see any need for this class as it essentially duplicates the h3 styles. Why not just recommend an h3?
> 
> Should the subheader be an `h*`? in places like the nx-tile-header, it isn't. Semantically I believe having it as another `h3` (after the `h3` in the `summary`) would start a new outline section that's a sibling of the section in the summary. This is not what we want. Therefore I believe it needs to be a div. Now, it could be a div with the `nx-h3` class, but I wasn't sure how you'd feel about that and further it would make it harder for us to separate the style of this element from the general h3 style in the future, if that's what we choose to do.

I understand what you're saying, but I don't see these as sub-headers. They aren't directly below the header and always visible like a tile sub-header would be. To me they are headings within the accordion content. It feels weird to use a `<div>` for them, what if we used `<h4>`'s but styled them using `.nx-h3` to match the design?

rpokorny: > I understand what you're saying, but I don't see these as sub-headers. They aren't directly below the header and always visible like a tile sub-header would be. To me they are headings within the accordion content. It feels weird to use a <div> for them, what if we used <h4>'s but styled them using .nx-h3 to match the design?

The only semantics hint I had to go on was that in the zeplin mockup, this element's text content is "Sub Header".  I think we need to ask @dturner32 for more clarification on the role that this element is intended to fulfill and if he has any real-world examples from upcoming product designs.

dturner32: for clarification, I have the same idea in mind that Glen has. The naming of the element may be confusing things here. I guess technically it shouldn't be called a subheader, however it does reflect the h3 styling. They are headings as part of the accordion content area. Maybe 'content header' or just 'header' would make more sense.

rpokorny: > for clarification, I have the same idea in mind that Glen has. The naming of the element may be confusing things here. I guess technically it shouldn't be called a subheader, however it does reflect the h3 styling. They are headings as part of the accordion content area. Maybe 'content header' or just 'header' would make more sense.

Ok, so that sounds like they should be `h*` elements, but at a lower level than the `h*` in the top of the accordion, as they represent subsections.  So I think, semantically speaking (not stylistically) we need to either change the accordion top header to an `h2` or `h1`, or alternatively change these subheaders to `h4`.  I'd prefer the former

glenhunter: > > for clarification, I have the same idea in mind that Glen has. The naming of the element may be confusing things here. I guess technically it shouldn't be called a subheader, however it does reflect the h3 styling. They are headings as part of the accordion content area. Maybe 'content header' or just 'header' would make more sense.
> 
> Ok, so that sounds like they should be `h*` elements, but at a lower level than the `h*` in the top of the accordion, as they represent subsections. So I think, semantically speaking (not stylistically) we need to either change the accordion top header to an `h2` or `h1`, or alternatively change these subheaders to `h4`. I'd prefer the former

I vote for `h2`'s with the subsections as `h3`'s. I think `h1` should be reserved for the page title.

rpokorny: > I vote for h2's with the subsections as h3's. I think h1 should be reserved for the page title.

What are your thoughts on the CSS class for that `h2`?  Is is alright/better to give it the `nx-h3` class, or should I go back to it having a custom class that happens to match our usual `h3` styling?

rpokorny: > > I vote for h2's with the subsections as h3's. I think h1 should be reserved for the page title.
> 
> What are your thoughts on the CSS class for that h2? Is is alright/better to give it the nx-h3 class, or should I go back to it having a custom class that happens to match our usual h3 styling?

Subheaders changed to h3's in d4e4d152.  Awaiting your reply to the above before changing the accordion header.

glenhunter: > > I vote for h2's with the subsections as h3's. I think h1 should be reserved for the page title.
> 
> What are your thoughts on the CSS class for that `h2`? Is is alright/better to give it the `nx-h3` class, or should I go back to it having a custom class that happens to match our usual `h3` styling?

IMO part of the value of having `nx-h#` classes is that it gives us the ability to say "this is semantically an `h#`, but stylistically we want it to have the same attributes as an `h#`." So I'm fine with styling the `<h2>` with `.nx-h3`.

rpokorny: I thought about it a bit and decided to go back to the custom class as I had originally (but with `<h2>` this time).  I feel like if we went with the `.nx-h3` class it would be too easy for consumers to miss that the class and element don't match and to implement it incorrectly. Additionally, if in the future we wanted this area to be styled differently than a normal `h3` the custom class allows that to be accomplished more easily/cleanly.

> IMO part of the value of having nx-h# classes is that it gives us the ability to say "this is semantically an h#, but stylistically we want it to have the same attributes as an h#."

I haven't really been thinking of that as a use of those classes.  As you know, I'm not that comfortable with classes whose _meaning_ is to designate styling.  To me, the meaning of the `nx-h3` class is the same as the meaning of `<h3>` itself, just within the RSC context.  And if we didn't have to support apps that are partially RSC and partially not, I'd be in favor of dropping the `nx-h3` etc classes and instead using bare element selectors to specify those styles.

glenhunter (approved):

PR #206, opened 2020-11-13
NxTreeView Style Update - RSC-146

rpokorny:
https://issues.sonatype.org/browse/RSC-146


lib/src/components/NxTreeView/NxTreeView.tsx
```diff
@@ -4,55 +4,81 @@
...
-            <div className="nx-tree-view__twisty">
-              <span className="nx-tree-view__twisty-icon"/>
-            </div>
+          <button className="nx-tree-view__trigger"
+                  onClick={onToggleCollapse || undefined}
+                  aria-controls={treeViewId}
+                  aria-expanded={isExpanded}
+                  aria-disabled={disabled || isEmpty || undefined}>
+            <NxFontAwesomeIcon className="nx-tree-view__twisty" icon={faCaretRight} />
             <div className="nx-tree-view__text">
```
glenhunter, line 40: IMO this should be a `<span>`

lib/src/components/NxTreeView/NxTreeView.tsx
```diff
@@ -4,55 +4,81 @@
...
         ),
         triggerTooltipProps = typeof triggerTooltip === 'string' ? { title: triggerTooltip } : triggerTooltip;
 
     return (
-      <div className={treeViewClasses}
-           id={id || undefined}>
+      <div className={treeViewClasses} id={treeViewId} role="tree">
         { triggerTooltipProps ? <NxTooltip { ...triggerTooltipProps } >{trigger}</NxTooltip> : trigger }
-        <div className="nx-tree-view__children">
+        <div className="nx-tree-view__children" role="group">
```
glenhunter, line 50: In the W3 example `role="tree item"` appeared at this level. I experimented with this locally and found that it made no difference at all as far as Voice Over was concerned, and frankly I find that how you have it makes more sense semantically how you have it.

glenhunter:
I checked this in Voice Over and found that it is still reading the tree view as a table, I have no idea why. I tried fiddling with a bunch of things and it didn't change.

Navigating the tree views is still an awkward mix of VO commands and tabbing but I don't think that there is anything that we can do about that. 

VO does announce the tree view as a expanded and collapsed once you activate the trigger (but not before).

The tooltip is read almost as if it is a part of the label with no indication that it is a tooltip.

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -8,88 +8,92 @@
...
+  background: none;
+  border: none;
+  box-sizing: border-box;
+  color: #2c2c2c;
   cursor: pointer;
   display: flex;
-  height: 30px;
-  line-height: 32px;
+  font: inherit;
+  font-weight: 600;
```
glenhunter, line 68: Why not use `@include: semi-bold();`?

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -8,88 +8,92 @@
...
-  display: block;
-  flex-shrink: 0;
-  height: 17px;
-  margin-right: 5px;
-  width: 16px;
+  &:focus {
+    outline: none;
+
+    .nx-tree-view__twisty {
+      filter: drop-shadow(0 0 3px transparentize(#02a6d9, 0.7));
```
glenhunter, line 81: Just out of curiosity why do this rather than use our standard :focus box-shadow variable?

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -8,88 +8,92 @@
...
 }
 
 .nx-tree-view__trigger {
   @include container-horizontal;
   align-items: center;
-  border: 1px solid transparent;
+  background: none;
+  border: none;
+  box-sizing: border-box;
+  color: #2c2c2c;
```
glenhunter, line 64:
```suggestion
  color: $nx-text-color-dark;
```

lib/src/components/NxTreeView/NxTreeView.tsx
rpokorny: > IMO this should be a <span>

7554222f

lib/src/components/NxTreeView/NxTreeView.tsx
rpokorny: > In the W3 example role="tree item" appeared at this level. I experimented with this locally and found that it made no difference at all as far as Voice Over was concerned, and frankly I find that how you have it makes more sense semantically how you have it.

Keep in mind that the concept they have of what a tree is is a little bit different than our concept.  I felt like putting this role here made sense given how we have it set up overall.

lib/src/components/NxTreeView/NxTreeView.scss
rpokorny: > Why not use @include: semi-bold();?

Ah, right.  Generally I avoid that as just setting the `font-weight` is just as easy, but in this case it makes sense since we have to reset the font face as well.  622b3422

glenhunter (approved):

lib/src/components/NxTreeView/NxTreeView.scss
rpokorny: > Ah, right. Generally I avoid that as just setting the font-weight is just as easy, but in this case it makes sense since we have to reset the font face as well. 622b342

I've had to revert this.  The thing is, this element is a `<button>` and has all sorts of defaults set, including `font-size`.  The mixin doesn't affect `font-size`.  To me it seems easier to reset all of the font properties with `font: inherit` and then customize the weight. be9f758c

PR #208, opened 2020-11-17
IWeek: NxPolicyThreatSlider Accessibility

rpokorny:
This PR adds an aria-valuetext attribute to each slider so that they are read as both their numeric value and threat category.  Note that in ChromeVox this prevents the reading of the min and max values which was happening previously. My understanding is that that was not happening in VoiceOver anyway.

I would like it if someone could test in VO


jflinchbaugh (approved):
+1

Blacktiger (approved):

PR #209, opened 2020-11-17
iWeek nxCloseBtn accessibility

glenhunter:
Added an `aria-label="Dismiss"` attribute. The only place we specifically use this is in NxAlert (right now). I found that if `NxAlert` was given the attribute `role="alert"` VoiceOver read the text of the alert first even when the dismiss button was focused, and then identified the button in the context of the dismiss button of the alert (which seems very clever).

### DIFF
--- a/lib/src/components/NxCloseButton/NxCloseButton.tsx
+++ b/lib/src/components/NxCloseButton/NxCloseButton.tsx
@@ -16,7 +16,12 @@ const NxCloseButton = forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HTMLBut
       const btnClasses = classnames('nx-btn--close', className);
 
       return (
-        <NxButton ref={ref} type="button" className={btnClasses} variant="icon-only" { ...otherProps }>
+        <NxButton ref={ref}
+                  type="button"
+                  className={btnClasses}
+                  variant="icon-only"
+                  { ...otherProps }
+                  aria-label="Dismiss">
           <Close/>
         </NxButton>
       );
--- a/lib/src/components/NxCloseButton/NxCloseButton.tsx
+++ b/lib/src/components/NxCloseButton/NxCloseButton.tsx
@@ -20,8 +20,8 @@ const NxCloseButton = forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HTMLBut
                   type="button"
                   className={btnClasses}
                   variant="icon-only"
-                  { ...otherProps }
-                  aria-label="Dismiss">
+                  aria-label="Dismiss"
+                  { ...otherProps }>
           <Close/>
         </NxButton>
       );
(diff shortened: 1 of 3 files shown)


lib/src/components/NxCloseButton/NxCloseButton.tsx
```diff
@@ -16,7 +16,12 @@ const NxCloseButton = forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HTMLBut
       const btnClasses = classnames('nx-btn--close', className);
 
       return (
-        <NxButton ref={ref} type="button" className={btnClasses} variant="icon-only" { ...otherProps }>
+        <NxButton ref={ref}
+                  type="button"
+                  className={btnClasses}
+                  variant="icon-only"
+                  { ...otherProps }
+                  aria-label="Dismiss">
```
rpokorny, line 24: By putting this last, you prevent `otherProps` from overriding it.  Is that desired?

rpokorny, line 24: Why "Dismiss" and not "Close"?

glenhunter: > By putting this last, you prevent `otherProps` from overriding it. Is that desired?

I didn't know that the order mattered, changed here: https://github.com/sonatype/sonatype-react-shared-components/pull/209/commits/253dab64fe05eb1007dba216df219445e66bacb8

glenhunter: > Why "Dismiss" and not "Close"?

I wavered back and further between the two. To me you close a window, you dismiss an alert. I could be easily convinced otherwise.

rpokorny: > I wavered back and forth between the two. To me you close a window, you dismiss an alert. I could be easily convinced otherwise.

I feel like, given that we made this a separate component, it isn't intended to assume that it is only used within NxAlert.  Additionally we named it NxCloseButton, not NxDismissButton, so I feel like to the same extent that there's a difference in meaning between those two terms, the component name and default label should probably match.

NxAlert can always override it with "Dismiss" if it matters.

jflinchbaugh (approved):
+1

lib/src/components/NxCloseButton/NxCloseButton.tsx
glenhunter: > > I wavered back and forth between the two. To me you close a window, you dismiss an alert. I could be easily convinced otherwise.
> 
> I feel like, given that we made this a separate component, it isn't intended to assume that it is only used within NxAlert. Additionally we named it NxCloseButton, not NxDismissButton, so I feel like to the same extent that there's a difference in meaning between those two terms, the component name and default label should probably match.
> 
> NxAlert can always override it with "Dismiss" if it matters.

That's a good point about the text matching the component name: https://github.com/sonatype/sonatype-react-shared-components/pull/209/commits/33d171f9818d59eb7f09f13017df4276eecd099e

rpokorny (approved):

santdelv (approved):

PR #210, opened 2020-11-17
Iweek nxthreatindicator accessibility

santdelv:
Ticket: https://trello.com/c/LNskx4dF

### DIFF
--- a/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
+++ b/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
@@ -24,7 +24,12 @@ const NxThreatIndicator: FunctionComponent<Props> =
         className = `nx-threat-indicator nx-threat-indicator--${category}`,
         label = `threat level ${category}`;
 
-    return <NxFontAwesomeIcon aria-hidden="false" aria-label={label} className={className} icon={faCircle}/>;
+    // @ts-ignore
+    return <NxFontAwesomeIcon aria-hidden="false"
+                              aria-label={label}
+                              role="presentation"
+                              className={className}
+                              icon={faCircle}/>;
   };
 
 NxThreatIndicator.propTypes = propTypes;
--- a/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
+++ b/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
@@ -22,9 +22,9 @@ const NxThreatIndicator: FunctionComponent<Props> =
           'unspecified',
 
         className = `nx-threat-indicator nx-threat-indicator--${category}`,
-        title = `threat level ${category}`;
+        label = `threat level ${category}`;
 
-    return <NxFontAwesomeIcon title={title} className={className} icon={faCircle}/>;
+    return <NxFontAwesomeIcon aria-hidden="false" aria-label={label} className={className} icon={faCircle}/>;
   };
 
 NxThreatIndicator.propTypes = propTypes;
--- a/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
+++ b/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
@@ -22,7 +22,7 @@ const NxThreatIndicator: FunctionComponent<Props> =
           'unspecified',
 
         className = `nx-threat-indicator nx-threat-indicator--${category}`,
-        title = `${category} threat level indicator`;
+        title = `threat level ${category}`;
 
     return <NxFontAwesomeIcon title={title} className={className} icon={faCircle}/>;
   };
--- a/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
+++ b/lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
@@ -24,7 +24,7 @@ const NxThreatIndicator: FunctionComponent<Props> =
         className = `nx-threat-indicator nx-threat-indicator--${category}`,
         label = `threat level ${category}`;
 
-    return <NxFontAwesomeIcon aria-hidden="false"
+    return <NxFontAwesomeIcon aria-hidden={false}
                               aria-label={label}
                               className={className}
                               icon={faCircle}/>;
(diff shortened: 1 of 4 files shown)


lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
```diff
@@ -21,9 +21,10 @@ const NxThreatIndicator: FunctionComponent<Props> =
           policyThreatLevel != null ? categoryByPolicyThreatLevel[policyThreatLevel] :
           'unspecified',
 
-        className = `nx-threat-indicator nx-threat-indicator--${category}`;
+        className = `nx-threat-indicator nx-threat-indicator--${category}`,
+        title = `${category} threat level indicator`;
```
rpokorny, line 25: We want the text to flow decently well with what comes after.  How about `threat level ${category}`?

lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
```diff
@@ -21,9 +21,10 @@ const NxThreatIndicator: FunctionComponent<Props> =
...
 
-    return <NxFontAwesomeIcon className={className} icon={faCircle} />;
+    return <NxFontAwesomeIcon title={title} className={className} icon={faCircle}/>;
```
rpokorny, line 27: Using `title` activates browser-native tooltips.  We want to avoid this because we have our own tooltips which look different.  If we actually want this text to be visible in a tooltip, we should use `NxTooltip`.  If we don't, we should use `aria-label`.

lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
santdelv: Sounds good f7dbeeeaf4eac36f29d4354de5fa8a99df862565

lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
santdelv: 85f72f39f457d112a10b8152d8b6f90b848fcfbd

rpokorny: > Using title activates browser-native tooltips. We want to avoid this because we have our own tooltips which look different. If we actually want this text to be visible in a tooltip, we should use NxTooltip. If we don't, we should use aria-label.

After more investigation I want to elaborate a little bit.  It turns out that `FontAwesomeIcon` (which back `NxFontAwesomeIcon`) doesn't apply the `title` attribute natively, but rather turns it into an svg `<title>` element which it then sets up as the describedby on the svg.  This still has the effect of causing the browser to render a native tooltip, which we don't really want.  What else is interesting though is that the `<svg>` has a role of "img".  If you don't specify a title on it, FontAwesome sets it up with `aria-hidden` causing it not to be read at all.  If you add the `title` however, `aria-hidden` is not applied which causes it to be read out (in ChromeVox at least) as "severe threat level indicator. image."  The word "image" being in there doesn't seem ideal to me so I wonder if we can/should override that role.  I also wonder if VoiceOver does that too or just ChromeVox.

santdelv: VO also reads the aria label (or title) and then "image".
About the role, I'm not sure what should we use then, does role="icon" sound good?

rpokorny: > does role="icon" sound good?

As far as I can tell that isn't a valid role.  Here's the list I've been using: https://www.w3.org/WAI/PF/aria/roles#role_definitions

santdelv: I don't see a role more appropriate than image. We could put role="", what do you think @rpokorny ?

rpokorny: > I don't see a role more appropriate than image. We could put role="", what do you think @rpokorny ?

`role="none"` or `role="presentation"` seem fine to me, as long as they don't disable the label/title reading.  I'm not sure leaving it blank is valid.

santdelv: I used presentation, none did not sound right to me
cab3cd74528ef3b4266b9328b5a4ad6ef4a25d95

lib/src/components/NxThreatIndicator/NxThreatIndicator.tsx
```diff
@@ -21,9 +21,15 @@ const NxThreatIndicator: FunctionComponent<Props> =
           policyThreatLevel != null ? categoryByPolicyThreatLevel[policyThreatLevel] :
           'unspecified',
 
-        className = `nx-threat-indicator nx-threat-indicator--${category}`;
+        className = `nx-threat-indicator nx-threat-indicator--${category}`,
+        label = `threat level ${category}`;
 
-    return <NxFontAwesomeIcon className={className} icon={faCircle} />;
+    // @ts-ignore
```
rpokorny, line 27: What's this?

santdelv: Is to suppress a warning that my IDE was throwing when adding role="presentation".

```
TS2322: Type '{ aria-hidden: string; aria-label: string; role: string; className: string; icon: IconDefinition; }' is not assignable to type 'IntrinsicAttributes & Props & { children?: ReactNode; }'.
  Property 'role' does not exist on type 'IntrinsicAttributes & Props & { children?: ReactNode; }'.
```

rpokorny: > Is to suppress a warning that my IDE was throwing when adding role="presentation".

Ah, so it seems that's not really supported.  There's no guarantee that passing the `role` prop to the underlying third-party `FontAwesomeIcon` component will cause that component to set it on the `<svg>`. It works now but might not on a future upgrade.  Further, in testing again locally, it's no longer reading the label at all for me.  Maybe leaving the role alone (as img) was better.

santdelv: be5eb73cabe4fbff96707c1d047f2fd3fb4498a6
It was still being read correctly for me with VO, but yes, seems like a good idea to leave image to avoid future compatibility problems.

rpokorny (approved):

Blacktiger (approved):

PR #211, opened 2020-11-17
IWeek NxAlert a11y

rpokorny:
https://trello.com/c/1P0pfFRg/22-nxalert

Allows NxAlert users to specify the role and a label on the icon.  Provides default values for that role and label for the various NxAlert specializations.

Note: in ChromeVox, these changes made almost no difference.  I'd want someone with VoiceOver to check them as well.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.9.3",
+  "version": "2.9.4",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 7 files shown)


Blacktiger (approved):

santdelv:
I can see the aria-label in the code but VO doesn't actually read the icon, I think you should set aria-hidden to false in the NxFontAwesomeIcon.

rpokorny:
> I can see the aria-label in the code but VO doesn't actually read the icon, I think you should set aria-hidden to false in the NxFontAwesomeIcon.

Try it with my latest changes please.

santdelv:
> > I can see the aria-label in the code but VO doesn't actually read the icon, I think you should set aria-hidden to false in the NxFontAwesomeIcon.
> 
> Try it with my latest changes please.

It is being read correctly by VO now.

santdelv (approved):

PR #212, opened 2020-11-17
NxDropdown accessibility improvements

glenhunter:
Added attributes for `aria-haspopup`, `aria-pressed`, and `aria-expanded`

Fixed examples with anchor tags so that they have dummy `href`'s and will be recognized by screen readers.

### DIFF
--- a/gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
+++ b/gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
@@ -25,16 +25,16 @@ function NxDropdownNavigationExample() {
         Text link 3
       </a>
       <button onClick={onClick} className="nx-dropdown-button">
-        Nav Link4 - this link should trigger truncation
+        Button Link 4 - this link should trigger truncation
       </button>
       <button onClick={onClick} className="nx-dropdown-button">
-        Nav Link5
+        Button Link 5
       </button>
       <button onClick={onClick} className="nx-dropdown-button">
-        Nav Link6
+        Button Link 6
       </button>
       <button className="disabled nx-dropdown-button">
-        Nav Link7 Disabled
+        Button Link 7 Disabled
       </button>
     </NxDropdown>
   );
--- a/gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
+++ b/gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
@@ -15,14 +15,14 @@ function NxDropdownNavigationExample() {
 
   return (
     <NxDropdown label="Navigation" isOpen={isOpen} onToggleCollapse={onToggleCollapse}>
-      <a onClick={onClick} className="nx-dropdown-button">
-        Nav Link1
+      <a onClick={onClick} href="#/pages/NxDropdown" className="nx-dropdown-button">
+        Text link 1
       </a>
-      <a onClick={onClick} className="nx-dropdown-button">
-        Nav Link2
+      <a onClick={onClick} href="#/pages/NxDropdown" className="nx-dropdown-button">
+        Text link 2
       </a>
-      <a onClick={onClick} className="nx-dropdown-button">
-        Nav Link3
+      <a onClick={onClick} href="#/pages/NxDropdown" className="nx-dropdown-button">
+        Text link 3
       </a>
       <button onClick={onClick} className="nx-dropdown-button">
         Nav Link4 - this link should trigger truncation
--- a/lib/src/components/NxDropdown/NxDropdown.tsx
+++ b/lib/src/components/NxDropdown/NxDropdown.tsx
@@ -40,7 +40,7 @@ const NxDropdown: FunctionComponent<Props> = function NxDropdown(props) {
               className={buttonClasses}
               onClick={!disabled && onToggleCollapse || undefined}
               aria-haspopup="true"
-              aria-expanded={isOpen ? true : false}>
+              aria-expanded={isOpen}>
       <span className="nx-dropdown__toggle-label">{ label }</span>
       <NxFontAwesomeIcon icon={isOpen ? faCaretUp : faCaretDown}/>
     </NxButton>
--- a/lib/src/components/NxDropdown/NxDropdown.tsx
+++ b/lib/src/components/NxDropdown/NxDropdown.tsx
@@ -40,7 +40,6 @@ const NxDropdown: FunctionComponent<Props> = function NxDropdown(props) {
               className={buttonClasses}
               onClick={!disabled && onToggleCollapse || undefined}
               aria-haspopup="true"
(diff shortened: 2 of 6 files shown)


lib/src/components/NxDropdown/NxDropdown.tsx
```diff
@@ -38,7 +38,10 @@ const NxDropdown: FunctionComponent<Props> = function NxDropdown(props) {
     <NxButton type="button"
               variant={variant || 'tertiary'}
               className={buttonClasses}
-              onClick={!disabled && onToggleCollapse || undefined}>
+              onClick={!disabled && onToggleCollapse || undefined}
+              aria-haspopup="true"
+              aria-pressed={isOpen ? true : false}
+              aria-expanded={isOpen ? true : false}>
```
rpokorny, line 44: These seem redundant with each other to me.  ChromeVox reads the gallery example as "Navigation. Popup button expanded pressed" or "Navigation. Popup button collapsed not pressed".

rpokorny:
Some relevant docs that I found:

https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html
https://www.w3.org/WAI/PF/aria/roles#menu
https://www.w3.org/WAI/PF/aria/usage#managingfocus

So it seems we should be using the `menu` and `menuitem` roles. More concerningly, these docs seem to suggest that we are required to make it so that the menu overall is focusable, rather than the individual menuitems, and then to make it so the menuitems are navigable by arrow keys.  Which seems like it would be counterintuitive to use and annoying to implement to me.

lib/src/components/NxDropdown/NxDropdown.tsx
glenhunter: > These seem redundant with each other to me. ChromeVox reads the gallery example as "Navigation. Popup button expanded pressed" or "Navigation. Popup button collapsed not pressed".

I went back and re-read the MDN entry for pressed and I think you're right, it's for something that acts as a toggle control. Removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/212/commits/96f2d4d67404f48081ee15b76ddc3e3ef0085913

lib/src/components/NxDropdown/NxDropdown.tsx
```diff
@@ -38,7 +38,9 @@ const NxDropdown: FunctionComponent<Props> = function NxDropdown(props) {
     <NxButton type="button"
               variant={variant || 'tertiary'}
               className={buttonClasses}
-              onClick={!disabled && onToggleCollapse || undefined}>
+              onClick={!disabled && onToggleCollapse || undefined}
+              aria-haspopup="true"
+              aria-expanded={isOpen ? true : false}>
```
rpokorny, line 43:
```suggestion
              aria-expanded={isOpen}>
```

glenhunter:
> Some relevant docs that I found:
> 
> https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html
> https://www.w3.org/WAI/PF/aria/roles#menu
> https://www.w3.org/WAI/PF/aria/usage#managingfocus
> 
> So it seems we should be using the `menu` and `menuitem` roles. More concerningly, these docs seem to suggest that we are required to make it so that the menu overall is focusable, rather than the individual menuitems, and then to make it so the menuitems are navigable by arrow keys. Which seems like it would be counterintuitive to use and annoying to implement to me.

I was going off this: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role 

Interestingly MDN doesn't even list `menu` or `menu item` as a role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

rpokorny:
> Interestingly MDN doesn't even list menu or menu item as a role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

From that page:

> This page lists reference pages covering all the WAI-ARIA roles discussed on MDN. For a full list of roles, see Using ARIA: Roles, States, and Properties

rpokorny:
> Found this article which clearly suggests that we should not be using menu and menu item: https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/

:thinking: hmm my takeaway from that article is that a dropdown of links is (often) not a menu but that a dropdown of buttons is a menu... we use `NxDropdown` for both and even support and demonstrate mixing the two.  So the issue is only more confused for me now.  One important thing to note is that `NxDropdown` is _not_ one of those hover-to-open nav menu things, which seem to be what is most incompatible with "true menu" behavior.

On a side note, I think it's a little funny that the author of that article went through all of that hair-splitting, only to use an example for the "true menu" that should probably just be a select box.

glenhunter:
> > Found this article which clearly suggests that we should not be using menu and menu item: https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/
> 
> 🤔 hmm my takeaway from that article is that a dropdown of links is (often) not a menu but that a dropdown of buttons is a menu... we use `NxDropdown` for both and even support and demonstrate mixing the two. So the issue is only more confused for me now. One important thing to note is that `NxDropdown` is _not_ one of those hover-to-open nav menu things, which seem to be what is most incompatible with "true menu" behavior.
> 
> On a side note, I think it's a little funny that the author of that article went through all of that hair-splitting, only to use an example for the "true menu" that should probably just be a select box.

My feeling is that what's here is good enough semantically, and functionally (in VoiceOver anyway) works very well.

gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
```diff
@@ -15,14 +15,14 @@ function NxDropdownNavigationExample() {
...
       </a>
-      <a onClick={onClick} className="nx-dropdown-button">
-        Nav Link2
+      <a onClick={onClick} href="#/pages/NxDropdown" className="nx-dropdown-button">
+        Text link 2
       </a>
-      <a onClick={onClick} className="nx-dropdown-button">
-        Nav Link3
+      <a onClick={onClick} href="#/pages/NxDropdown" className="nx-dropdown-button">
+        Text link 3
```
rpokorny, line 25: If you're going to change these to make them less confusing, you should probably also change the buttons ones below.  Having them referred to have "Nav Links" doesn't really make sense.

rpokorny (approved):

gallery/src/components/NxDropdown/NxDropdownNavigationExample.tsx
glenhunter: > If you're going to change these to make them less confusing, you should probably also change the buttons ones below. Having them referred to have "Nav Links" doesn't really make sense.

Made them consistent across all examples: https://github.com/sonatype/sonatype-react-shared-components/pull/212/commits/426c00100d3915785033dcd7766b8a6191beea43

santdelv (approved):

PR #214, opened 2020-11-18
Improve a11y experience for table sorting

Blacktiger:
https://trello.com/c/jIuN4TOA/6-nxtable-sorting


lib/src/components/NxTable/NxTableCell.tsx
```diff
@@ -39,30 +41,42 @@ const NxTableCell = function NxTableCell(props: NxTableCellProps) {
...
   }
 
-  const Tag = isHeader ? 'th' : 'td';
 
-  return (
-    <Tag className={classes} {...attrs}>
+  const Tag = isHeader ? 'th' : 'td';
+  const tabIndex = isSortable? 0 : undefined;
+  const cell = (
+    <Tag className={classes} tabIndex={tabIndex} aria-sort={ariaSort} {...attrs}>
```
Blacktiger, line 79: `aria-sort` is not currently supported by VO or Chrome-Vox from what I saw, but other screen readers do support the attribute so I'm assuming VO will eventually gain support.

lib/src/components/NxTable/NxTableCell.tsx
```diff
@@ -39,30 +41,42 @@ const NxTableCell = function NxTableCell(props: NxTableCellProps) {
...
-  }
-  else {
-    maskedSort = <NxFontAwesomeIcon icon={faSort} />;
+  let ariaSort: 'ascending' | 'descending' | 'none' | undefined;
+  let ariaLabel;
+  if (isSortable) {
+    const text = textContent(children);
+    if (sortDir === 'asc') {
+      ariaSort = 'ascending';
+      ariaLabel = `${text} ${ariaSort}`;
```
Blacktiger, line 50: I found a website that suggested the aria-label should contain text telling the user what will happen when they click the cell, but in my testing that didn't work well. Instead I'm just labelling how the column is sorted. (It turns out when you click the cell it announces the _next_ label which isn't what I expected initially.)

rpokorny:
Side note: the table header doesn't seem to have good focus styles

lib/src/components/NxTooltip/NxTooltip.tsx
```diff
@@ -36,7 +37,8 @@ function fixOptional(props: Props): TooltipProps {
     onClose: props.onClose || undefined,
     open: props.open === null ? undefined : props.open,
     placement: convertPlacement(props.placement),
-    title: props.title || ''
+    title: props.title || '',
+    'aria-label': textContent(props.title)
```
rpokorny, line 41: Btw this makes it read out twice in ChromeVox

rpokorny, line 41: This `fixOptional` function isn't really the right place to do this

lib/src/util/childUtil.ts
```diff
@@ -30,3 +30,30 @@ export function only(children: ReactNode, type: React.ComponentType): React.Reac
   });
   return matchingChildren.length ? matchingChildren[0] : null;
 }
+
+/**
+ * @param children to search
+ * @returns the concatenated string contents of thie children
+ */
+export function textContent(children: ReactNode): string {
```
rpokorny, line 38: I was thinking about this function more and I realized that even doing something like this is limited: it wouldn't work right if actual components were part of the `children`.  What I wish we could find is something like `ReactDOMServer.renderToString()` that only rendered the text nodes and not the element markup.  Though come to think of it even that would have problems if any of the child elements were supposed to be invisible. I wonder if maybe it would be best to just require the caller of `NxTable` to pass in the label strings, or at least to pass them in any time the `children` aren't themselves a single string.

Blacktiger: Why wouldn't this work right with components instead of children? It will just recurse down and combine all text found into a single string.

rpokorny: > Why wouldn't this work right with components instead of children? It will just recurse down and combine all text found into a single string.

I'm referring to essentially the difference between shallow rendering and mounting.  Text that is defined within the implementation of child components will be missed unless you actually render the components.

lib/src/components/NxTooltip/NxTooltip.tsx
Blacktiger: I wasn't able to duplicate with ChromeVox. That reader for me only reads the aria-label if I tab to the `<th>`, if I use the cv+arrow keys to navigate around it just stops on the name element and reads the element contents instead of the label when the NxTooltip is applied to the `<th>`.

In VO+Safari the reader reads out "name unsorted name" instead. I tried several different options but I can't seem to get any better behavior here.

lib/src/util/childUtil.ts
Blacktiger: I see what you are getting at, though I suspect that's going to fairly rare for a sortable header. Table headers tend to be pretty static in general.

As an alternative we could define a special component for the sortable headers (`NxSortableHeader` maybe). That would have the advantage of allowing us to define separate requirements such as "simple strings only for children" while still allowing the `NxTableCell` to allow for more complex headers.

lib/src/components/NxTooltip/NxTooltip.tsx
Blacktiger: I wasn't able to duplicate this with ChromeVox locally.

lib/src/util/childUtil.ts
rpokorny: > As an alternative we could define a special component for the sortable headers (NxSortableHeader maybe). That would have the advantage of allowing us to define separate requirements such as "simple strings only for children" while still allowing the NxTableCell to allow for more complex headers.

That would make it so that, for instance, a sortable header can't have an icon next to the text.  I'd rather stick with what you have now, documenting the caveat, or force users to provide the accessible label separately.

lib/src/components/NxTooltip/NxTooltip.tsx
rpokorny: > I wasn't able to duplicate with ChromeVox. That reader for me only reads the aria-label if I tab to the <th>, if I use the cv+arrow keys to navigate around it just stops on the name element and reads the element contents instead of the label when the NxTooltip is applied to the <th>.

That's what I'm getting now too. Not sure what was different before.  However, this particular comment thread wasn't about behavior, it was about code organization and how this line does not fit the purpose of the function that it is in.

Blacktiger: 5d513469f1c3e8bdbc9e91fa472b5fac4fbcab96

rpokorny (approved):

glenhunter (approved):

PR #215, opened 2020-11-18
nx-tile accessibility improvements

glenhunter:
Make sure that all `<div class="nx-tile">` are changed to use `<section>`
Adjust headers inside of `nx-tile` to use `<header>`
Use `aria-labelledby` tied to an ID on the tile tile h#

Despite all of the documentation stating that adding `aria-labelledby` to the `<section>` would cause the `<section>` to be seen as a landmark it did not appear in the voice over "rotor" in the landmark section. Nonetheless I think that the `<section>` and `<header>` changes are worthwhile and maybe somebody can see if they get different results from the  aria changes


santdelv (approved):

gallery/src/styles/NxTile/NxTileWithHorizontalRuleExample.html
```diff
@@ -1,26 +0,0 @@
-<!--
```
jflinchbaugh, line 1: why remove this file? unused?

jflinchbaugh (approved):
+1

jflinchbaugh:
do you wait and do the version bump after approvals?

gallery/src/styles/NxTile/NxSimpleTileExample.html
```diff
@@ -6,10 +6,10 @@
     distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
 
 -->
-<div class="nx-tile">
-  <div class="nx-tile-header">
+<section class="nx-tile" aria-label="Example of nx-tile">
```
rpokorny, line 9: Why doesn't the label match the header?

gallery/src/styles/NxTile/NxTileWithHorizontalRuleExample.html
glenhunter: > why remove this file? unused?

Yeah, we removed that example last week and didn't remove the file, so it's just a bit of housekeeping

glenhunter:
> Make sure that all `<div class="nx-tile">` are changed to use `<section>`
> Adjust headers inside of `nx-tile` to use `<header>`
> Use `aria-labelledby` tied to an ID on the tile tile h#
> 
> Despite all of the documentation stating that adding `aria-labelledby` to the `<section>` would cause the `<section>` to be seen as a landmark it did not appear in the voice over "rotor" in the landmark section.



> do you wait and do the version bump after approvals?

Yes, after approvals we make sure that we have a passing build in Jenkins (which checks the versioning), before we merge.

gallery/src/styles/NxTile/NxSimpleTileExample.html
glenhunter: > Why doesn't the label match the header?

I was trying to make the labels be more logical when read out loud, I should have had the word "simple" in there though. https://github.com/sonatype/sonatype-react-shared-components/pull/215/commits/92cc96f11ece11c47c9d259e9b2e4ccee8bca33f

rpokorny (approved):

PR #216, opened 2020-11-18
Add scope="col" to example pages

Blacktiger:
https://trello.com/c/JZsZSAci/43-nxtable-nx-table-column-headers


Blacktiger:
Nevermind, this isn't needed.

PR #218, opened 2020-11-18
iweek demo nx-card as landmark

jflinchbaugh:
add aria-label to each example card

trello: https://trello.com/c/z7TIZT9Y/30-nx-card
build: https://jenkins.ci.sonatype.dev/job/uxui/job/sonatype-react-shared-components/job/iweek-nx-card-landmark/

Screen Shot 2020-11-18 at 3 32 10 PM


gallery/src/styles/NxCard/NxCardColumnLayoutExample.tsx
```diff
@@ -12,7 +12,7 @@ import { NxFontAwesomeIcon, NxButton } from '@sonatype/react-shared-components';
 export default function NxCardColumnLayoutExample() {
   return (
     <div className="nx-card-container nx-card-container--column">
-      <section className="nx-card">
+      <section className="nx-card" aria-label="Card 1 in Column Layout">
```
Blacktiger, line 15: I've noticed most of the aria-label examples I've seen seem to use only lower-case. I'm guessing it's just because the screen readers don't pay attention to the capitalization anyway and laziness.

Blacktiger (approved):

gallery/src/styles/NxCard/NxCardColumnLayoutExample.tsx
```diff
@@ -12,7 +12,7 @@ import { NxFontAwesomeIcon, NxButton } from '@sonatype/react-shared-components';
 export default function NxCardColumnLayoutExample() {
   return (
     <div className="nx-card-container nx-card-container--column">
-      <section className="nx-card">
+      <section className="nx-card" aria-label="card 1 in column layout">
```
rpokorny, line 15: Similarly to a comment I left on the tile PR, I'm wondering why the labels wouldn't be the same as the headers?

jflinchbaugh: If I do that, should I update the headers to be more _meaningful_ as examples? Consider that if I match them up, then they'll read twice: once upon selecting the card, and again as the header.

glenhunter: > If I do that, should I update the headers to be more _meaningful_ as examples? Consider that if I match them up, then they'll read twice: once upon selecting the card, and again as the header. Many of them might just read: "header region", then "header" again upon proceeding.
> 
> I chose the descriptions I did to sort of capture the reason these cards were being presented here (context?). I would have liked to offer more meaning, but nature of the visual demo didn't offer much meaning. I could see it being more meaningful to call a card "Weather Forecast" or "Evaluation Results".
> 
> The aria-labelledby attr seemed to offer the idea of building a label from a hierarchy of elements for context.

I did something similar with the tiles, I tried to write the labels to capture the heading and also provide some context that I thought I would want if I were only hearing the description. I think that because the cards are so much more space constrained it would be even more useful to have a slightly "fuller" label.

rpokorny (approved):
Weirdly ChromeVox only reads these labels if I click on the cards, not if I keyboard-navigate to them.  I've had the same problem with other elements; I'm sure it's just a ChromeVox problem.

glenhunter (approved):

santdelv (approved):

PR #219, opened 2020-11-18
IWeek: NxTextInput and labels ARIA

rpokorny:
https://trello.com/c/ZC6glZAu/9-nxtextinput

This PR adds validation ARIA validation handling via the `aria-invalid` and `aria-errormessage` attrs.  It also demonstrates how to use `aria-required`, `aria-describedby` and `for` with labels to try to get the intended effect in regards to labels and sublabels (that is, I assume we want the sublabel to be the description, not part of the accessible name).  

For now, I am mainly interested in @glenhunter 's feedback as this introduces a new way to structure the form groups, which luckily worked without any styling updates.  Additionally I'd like feedback on how the sublabels are called out in VoiceOver.  In ChromeVox they aren't called out in any special way, despite Chrome's accessibility tab recognizing them correctly.  In fact, when navigating using ChromeVox shortcuts the sublabel now gets announced before the label, due to the fact that it waits to announce the label until it you navigate to the text input itself.  Which isn't ideal.


gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -36,25 +36,27 @@ export default function NxFormLayoutExample() {
     <form className="nx-form" onSubmit={onSubmit}>
       <div className="nx-form-group">
         <label className="nx-label">
-          <span className="nx-label__text">Label</span>
-          <NxStatefulTextInput validator={validator}/>
+          <span className="nx-label__text">A Field to Fill in</span>
+          <NxStatefulTextInput aria-required={true} validator={validator}/>
```
glenhunter, line 40: Shouldn't this be outside the `<label>` to be consistent?

gallery/src/styles/NxTile/NxTileFormExample.tsx
```diff
@@ -28,7 +28,7 @@ export default function NxTileFormExample() {
           <div className="nx-form-group">
             <label className="nx-label">
               <span className="nx-label__text">Label</span>
-              <NxStatefulTextInput validator={validator}/>
+              <NxStatefulTextInput aria-required={true} validator={validator}/>
```
glenhunter, line 31: As above, if we're going to change the layout I would prefer that we are consistent about whether the `input` is inside or outside the `<label>`

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
rpokorny: > Shouldn't this be outside the <label> to be consistent?

No, the idea was to deliberately support both, since the old way is already used by consumers of RSC.

glenhunter: > > Shouldn't this be outside the `<label>` to be consistent?
> 
> No, the idea was to deliberately support both, since the old way is already used by consumers of RSC.

Don't forget to update the docs to make it clear what the preferred layout is.

glenhunter (approved):

lib/src/setupTests.js
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+const nodeCrypto = require('crypto');
+
+// typescript doesn't like this which is why it's in a JS file
+window.crypto = {
+  getRandomValues: function(buffer) { return nodeCrypto.randomFillSync(buffer); }
```
Blacktiger, line 11: why not just put this in setupTests.ts or move everything frome setupTests.ts to setupTests.js?

rpokorny: > why not just put this in setupTests.ts

See the code comment.  Typescript claims that `window.crypto` is readonly (which it is in a real browser) and doesn't let me override it.

> or move everything frome setupTests.ts to setupTests.js?

At least some of the lines in `setupTests.ts` are necessary to get typescript to understand that types for the test functions (for instance that the jest-enzyme matchers exist).

Blacktiger: Could we just ts-ignore the window.crypto error instead?

Blacktiger (approved):

PR #220, opened 2020-11-18
NxPagination accessibility

Blacktiger:
https://trello.com/c/uceaTz7m/2-nxpagination


lib/src/components/NxPagination/__tests__/NxPagination.test.tsx
```diff
@@ -85,105 +85,105 @@ describe('NxPagination', function() {
...
       expect(getShallowComponent({ pageCount: 2, currentPage: 0 })).toMatchElement(
-        <div>
-          <NxButton>1</NxButton>
-          <NxButton>2</NxButton>
-          <NxButton><NxFontAwesomeIcon icon={{} as IconProp} /></NxButton>
-        </div>
+        <nav>
+          <NxButton aria-label="first page">1</NxButton>
+          <NxButton aria-label="page 2">2</NxButton>
+          <NxButton aria-label="next page"><NxFontAwesomeIcon icon={{} as IconProp} /></NxButton>
```
rpokorny, line 91: `toMatchElement` does not check props.  You'll need to assert these separately.

If you're wondering about the `icon` prop on `NxFontAwesomeIcon`, it's just to satisfy the type checker.

rpokorny:
I find the usability to still be pretty lacking.  When you first navigate to the pagination, you start on the previous page button - but it doesn't tell you either what page you're currently on nor what page the previous page that you'd be going to is.

Tabbing through the component, it calls out each page that you can click on, and just skips the current page.  Unless you know to listen for that skip, there's no way at all to get the component to tell you which page is the current one.

I also wonder how the user should know what the pagination bar is paging.  Of course, in these examples it's standalone, but should we document that the pagination should use `aria-controls` to refer to the table or whatever that's being paged? And would doing so actually make a difference to screen readers?

I'm also conflicted on it being a `<nav>` as it won't, in typical usage, navigate to a different URL or anything, but would just update the data in a particular table or something.

Blacktiger:
> I find the usability to still be pretty lacking. When you first navigate to the pagination, you start on the previous page button - but it doesn't tell you either what page you're currently on nor what page the previous page that you'd be going to is.
> 
> Tabbing through the component, it calls out each page that you can click on, and just skips the current page. Unless you know to listen for that skip, there's no way at all to get the component to tell you which page is the current one.

Interesting, when testing via VO+Safari it stopped on the current element and called it out for being the current element.

> I also wonder how the user should know what the pagination bar is paging. Of course, in these examples it's standalone, but should we document that the pagination should use `aria-controls` to refer to the table or whatever that's being paged? And would doing so actually make a difference to screen readers?

Looks to me like we should be using `aria-owns` to make a parent-child relationship between the pagination controls and whatever it pages through.

> I'm also conflicted on it being a `<nav>` as it won't, in typical usage, navigate to a different URL or anything, but would just update the data in a particular table or something.

I read the docs for the navigation role and thought that it should apply here. (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role) This doc also indicates that you should use `nav` for pagination: https://www.a11ymatters.com/pattern/pagination/

rpokorny:
> Looks to me like we should be using aria-owns to make a parent-child relationship between the pagination controls and whatever it pages through.

Why `owns` and not `controls`?

Blacktiger:
> > Looks to me like we should be using aria-owns to make a parent-child relationship between the pagination controls and whatever it pages through.
> 
> Why `owns` and not `controls`?

Doing some more reading on this I think I was incorrect.

lib/src/components/NxPagination/__tests__/NxPagination.test.tsx
Blacktiger: 3cc14e7475ed7e7a139a659b51b5d4bb65cd5635

rpokorny:
Navigating left and right through the buttons is really wonky for me (seems to skip over the central group of buttons sometimes and go to either end) but that might just be chromevox.  It also doesn't read the nav's label ifI keyboard navigate to it.  It _does_ read it if I click in just the right place (essentially, in the padding of the surrounding `gallery-example-live`), so that's pretty weird.

lib/src/components/NxPagination/NxPagination.tsx
```diff
@@ -115,27 +135,31 @@ export default function NxPagination({ className, pageCount, currentPage, onChan
 
         { numPagesAboveRange > 1 &&
           // Rigth '...' - forward one page group
-          <NxButton className={getBtnClasses()} onClick={handleBtnClick(currentPageRangeEnd)}>
+          <NxButton aria-label={`show next ${PAGE_RANGE_SIZE} pages`}
```
rpokorny, line 138: The ellipsis buttons don't just show the next set of buttons, they navigate to a different page.  I think they should have text similar to what the other buttons has which simply says which page they go to.

rpokorny:
When I click a button which disappears as a result of the click (for instance clicking ... until I get to the end of the list), both the focus and the ChromVox nav appear to get "lost".  Trying to go to the next element results it is wrapping back to the top of the entire page.

lib/src/components/NxPagination/NxPagination.tsx
Blacktiger: 3f1d6844

rpokorny:
> When I click a button which disappears as a result of the click (for instance clicking ... until I get to the end of the list), both the focus and the ChromeVox nav appear to get "lost". Trying to go to the next element results it is wrapping back to the top of the entire page.

@Blacktiger has established that in VO the focus goes to the `<nav>` element which is fine

Blacktiger:
> When I click a button which disappears as a result of the click (for instance clicking ... until I get to the end of the list), both the focus and the ChromeVox nav appear to get "lost". Trying to go to the next element results it is wrapping back to the top of the entire page.

This appears to just be a problem with ChromeVox. In VoiceOver on both Chrome and Safari when the ellipsis button disappears the screen reader just focuses on the nav element instead (though it doesn't read out the aria label, it just tells the user they are on a nav element and then you can navigate back onto the pagination buttons).

lib/src/components/NxPagination/NxPagination.tsx
```diff
@@ -78,33 +78,50 @@ export default function NxPagination({ className, pageCount, currentPage, onChan
...
+          }
+          else {
+            ariaLabel = `goto page ${num + 1}`;
+          }
+
           return (
-            <NxButton onClick={selected ? undefined : handleBtnClick(num)}
-                      tabIndex={selected ? -1 : 0 /* selected page is not focusable */}
+            <NxButton onClick={selected ? () => {} : handleBtnClick(num)}
+                      tabIndex={0}
```
rpokorny, line 94: Isn't this the default for buttons? Is this actually necessary?

lib/src/components/NxPagination/NxPagination.tsx
```diff
@@ -78,33 +78,50 @@ export default function NxPagination({ className, pageCount, currentPage, onChan
...
+            ariaLabel = 'goto last page';
+          }
+          else {
+            ariaLabel = `goto page ${num + 1}`;
+          }
+
           return (
-            <NxButton onClick={selected ? undefined : handleBtnClick(num)}
-                      tabIndex={selected ? -1 : 0 /* selected page is not focusable */}
+            <NxButton onClick={selected ? () => {} : handleBtnClick(num)}
```
rpokorny, line 93: Was this actually needed? I would think it makes no difference.

Blacktiger: Hmm, it probably isn't. I'll revert this.

lib/src/components/NxPagination/__tests__/NxPagination.test.tsx
```diff
@@ -302,16 +334,14 @@ describe('NxPagination', function() {
     testWithProps({ pageCount: 6, currentPage: 5 });
   });
 
-  it('adds tabIndex=-1 to the selected page and either tabIndex=0 or no tabIndex to all other buttons', function() {
+  it('adds either tabIndex=0 or no tabIndex to all other buttons', function() {
```
rpokorny, line 337: More accurate to what you have at the moment: `adds tabIndex=0 to all buttons`.  However as I mentioned above I'm not sure you even need to do that.  If you don't, this test can just be removed.

gallery/src/components/NxPagination/NxPaginationExample.tsx
```diff
@@ -10,5 +10,12 @@ import { NxPagination } from '@sonatype/react-shared-components';
 export default function NxPaginationExample() {
   const [page, setPage] = useState(17);
 
-  return <NxPagination onChange={setPage} pageCount={41} currentPage={page} />;
+  return (
+    <div>
```
glenhunter, line 14: What is the purpose of this `<div>`? The example is already within `<div class="gallery-example-live">` so I don't think that we need to add anymore `<div>`'s

gallery/src/components/NxPagination/NxPaginationSinglePageExample.tsx
```diff
@@ -8,5 +8,10 @@ import React from 'react';
 import { NxPagination } from '@sonatype/react-shared-components';
 
 export default function NxPaginationSinglePageExample() {
-  return <NxPagination onChange={() => {}} pageCount={1} currentPage={0} />;
+  return (
+    <div>
```
glenhunter, line 12: See: https://github.com/sonatype/sonatype-react-shared-components/pull/220/files#r527894215

glenhunter:
I found it worked as I expected in VoiceOver, and aside from a  couple of comments about the HTML it looks good to me.

lib/src/components/NxPagination/NxPagination.tsx
Blacktiger: 2bd2e8af0fa88ea99a9e63845b2374bcf22e17cb

lib/src/components/NxPagination/__tests__/NxPagination.test.tsx
Blacktiger: I removed this test entirely in e9672ddf911cb578dece43342b812ef76719e189. I don't see any reason to keep it since we don't need to have the tabIndex set on any of the buttons.

gallery/src/components/NxPagination/NxPaginationExample.tsx
Blacktiger: I guess I could have used a fragment instead. I just needed to have two components in here.

rpokorny (approved):

santdelv (approved):

lib/src/components/NxPagination/__tests__/NxPagination.test.tsx
glenhunter: > I guess I could have used a fragment instead. I just needed to have two components in here.

To me a fragment is always preferable to just a `<div>`, especially because some devs are likely to cut and paste the example code and we could end up with app code littered with unnecessary bare `<div>` tags.

PR #222, opened 2020-11-19
nxFontAwesomeIcon a11y

glenhunter:
Added an example of an icon only button and a bare icon with accessibility features. Created documentation to support.


gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconExample.tsx
```diff
@@ -27,6 +27,12 @@ const NxFontAwesomeIconExample = () =>
         */}
       <NxFontAwesomeIcon icon={faBatteryEmpty} color="red" />
     </button>
+    {/* Icon only variant button with aria-label on the icon, note that aria-hidden has been set to false */}
+    <NxButton variant="icon-only">
+      <NxFontAwesomeIcon icon={faEdit} aria-hidden="false" aria-label="Edit text" />
```
Blacktiger, line 32: Should we automatically set aria-hidden to false in NxFontAwesomeIcon when an aria-label prop is set instead of manually doing it here?

rpokorny, line 32: We have icon-only button examples in the `NxButton` page which are in fact written differently than this - they put the `aria-label` directly on the button.  We don't want a conflicting example here.

gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconPage.tsx
```diff
@@ -33,12 +33,26 @@ const NxFontAwesomeIconPage = () => {
...
+          By default Font Awesome icons are set to <code className="nx-code">aria-hidden="true"</code> which means that
+          they are not "seen" by screen readers. Normally this is acceptable because icons are usually presentational
+          in nature, however if the icon has information that is important to convey to the user (e.g. an icon only
+          button) the value of <code className="nx-code">aria-hidden</code> should be set to
+          {' '}<code className="nx-code">false</code> and an <code className="nx-code">aria-label</code> with
+          explanatory text should also be employed. The last two examples demonstrate this.
+        </p>
+        <p className="nx-p">
+          If you are in doubt about whether an icon should be made accessible consider whether the user could perform
+          their given task, or understand an explaination if that icon was not there.
```
rpokorny, line 47: May want to also point out the converse: that if adding a label to the button would result in duplicitous announcements (e.g. due to adjacent text content) then the label should not be added.

gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconExample.tsx
jflinchbaugh: if we set aria-hidden automatically, we can revisit lots of other style and component code to remove aria-hidden.

glenhunter: > We have icon-only button examples in the `NxButton` page which are in fact written differently than this - they put the `aria-label` directly on the button. We don't want a conflicting example here.

I thought about this a bunch and IMO `aria-label` should be applied to the icon because that is what it is labelling, not the button. I had avoided modifying the NxButton examples (for now) because i didn't want to overlap with anybody else's work but I think that they should be modified to follow this pattern. I'm happy to take that on as a part of this ticket.

glenhunter: > Should we automatically set aria-hidden to false in NxFontAwesomeIcon when an aria-label prop is set instead of manually doing it here?

That's an interesting idea and would streamline things, I'll take a look at that.

rpokorny: > I thought about this a bunch and IMO aria-label should be applied to the icon because that is what it is labelling, not the button. I had avoided modifying the NxButton examples (for now) because i didn't want to overlap with anybody else's work but I think that they should be modified to follow this pattern. I'm happy to take that on as a part of this ticket.

TBH I have the opposite philosophy about it.  I think the semantics are most accurately associated with the button, and the icon is a presentational concern within the button.  Think of it this way: you may have two buttons in different places which use the same icon to mean different things, especially if you craft that meaning in a more specific way (For instance, "delete waiver" rather than just "delete").  It's not that there's an icon that inherently means "delete waiver", it's that there's a particular button on that particular page that has that meaning.

The fact that doing it that way avoids the need for the `aria-hidden` override is an added bonus.

Another, bigger thought: should icon-only buttons always have tooltips? Should we ideally make a separate `NxIconOnlyButton` component that renders an `NxButton` wrapped in an `NxTooltip`?  That would be an alternative way to label the buttons and avoid the question above.

glenhunter: > > I thought about this a bunch and IMO aria-label should be applied to the icon because that is what it is labelling, not the button. I had avoided modifying the NxButton examples (for now) because i didn't want to overlap with anybody else's work but I think that they should be modified to follow this pattern. I'm happy to take that on as a part of this ticket.
> 
> TBH I have the opposite philosophy about it. I think the semantics are most accurately associated with the button, and the icon is a presentational concern within the button. Think of it this way: you may have two buttons in different places which use the same icon to mean different things, especially if you craft that meaning in a more specific way (For instance, "delete waiver" rather than just "delete"). It's not that there's an icon that inherently means "delete waiver", it's that there's a particular button on that particular page that has that meaning.
> 
> The fact that doing it that way avoids the need for the `aria-hidden` override is an added bonus.
> 
> Another, bigger thought: should icon-only buttons always have tooltips? Should we ideally make a separate `NxIconOnlyButton` component that renders an `NxButton` wrapped in an `NxTooltip`? That would be an alternative way to label the buttons and avoid the question above.

I think that your point is valid, however I would prefer to do that work in a  separate `NxButton` PR.

rpokorny: > I would prefer to do that work in a separate NxButton PR.

I assume you are referring here to the very last thing I said?  Everything I said describes the way the NxButton examples already are.

gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconPage.tsx
glenhunter: > May want to also point out the converse: that if adding a label to the button would result in duplicitous announcements (e.g. due to adjacent text content) then the label should not be added.

I felt like that was implied, but I have added some (hopefully) clarifying text here: https://github.com/sonatype/sonatype-react-shared-components/pull/222/commits/a905b20509beab7ef707a886a7e134983c514869

gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconExample.tsx
glenhunter: I have created https://issues.sonatype.org/browse/RSC-398 for the button work.

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.9.14",
+  "version": "2.9.16",
```
rpokorny, line 3: Looks like a double-bump.  Double check that it's right after merging from master.

gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconPage.tsx
```diff
@@ -33,12 +33,30 @@ const NxFontAwesomeIconPage = () => {
           <a href="https://github.com/FortAwesome/react-fontawesome#features" target="_blank">documentation</a>
           {' '}for details on available props
         </p>
+        <h3 className="nx-h3">Accessibility</h3>
+        <p className="nx-p">
+          <code className="nx-code">FontAwesomeIcon</code> has a <code className="nx-code">title</code> prop which
+          sets up a <code className="nx-code">&lt;title&gt;</code> element within the rendered SVG and configures is
```
rpokorny, line 39:
```suggestion
          sets up a <code className="nx-code">&lt;title&gt;</code> element within the rendered SVG and configures it
```

gallery/src/components/NxFontAwesomeIcon/NxFontAwesomeIconPage.tsx
```diff
@@ -33,12 +33,30 @@ const NxFontAwesomeIconPage = () => {
...
+          as the accessible name for the icon. Use this attribute when an icon itself needs to be read by a
+          screenreader. Note however that this will also create a native tooltip with the title, which is not ideal
+          due to our preference for <code className="nx-code">NxTooltip</code>. This technique should therefore be
+          used sparingly, and labels on parent elements should be preferred where appropriate. For instance,
+          the accessible name for an icon-only button should be placed on the button itself, not on the icon.
+          See the <a href="#/pages/NxButton"><code className="nx-code">NxButton</code></a> page for an example.
+        </p>
+        <p className="nx-p">
+          If you are in doubt about whether an icon should be made accessible consider whether the user could perform
+          their given task, or understand an explaination if that icon was not there. Take care that adding an
```
rpokorny, line 49:
```suggestion
          their given task, or understand an explanation if that icon was not there. Take care that adding an
```

rpokorny (approved):

gallery/package.json
glenhunter: > Looks like a double-bump. Double check that it's right after merging from master.

I see master as 2.9.15

PR #224, opened 2020-11-19
i-week nx-alert examples for a11y

jflinchbaugh:
match the style examples to what the NxAlert component will do for a11y: roles and aria props.

build: https://jenkins.ci.sonatype.dev/job/uxui/job/sonatype-react-shared-components/job/iweek-nx-alert-a11y/

### DIFF
--- a/gallery/src/styles/NxAlert/NxAlertErrorExample.tsx
+++ b/gallery/src/styles/NxAlert/NxAlertErrorExample.tsx
@@ -10,7 +10,7 @@ import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
 
 const NxAlertErrorExample = () =>
   <div className="nx-alert nx-alert--error" role="alert" aria-atomic={true}>
-    <NxFontAwesomeIcon icon={faExclamationCircle} aria-label="error" />
+    <NxFontAwesomeIcon icon={faExclamationCircle} aria-label="error" aria-hidden={false} />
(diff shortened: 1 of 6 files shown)


gallery/src/styles/NxAlert/NxAlertErrorExample.tsx
```diff
@@ -9,8 +9,8 @@ import { NxFontAwesomeIcon, NxCloseButton } from '@sonatype/react-shared-compone
 import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
 
 const NxAlertErrorExample = () =>
-  <div className="nx-alert nx-alert--error">
-    <NxFontAwesomeIcon icon={faExclamationCircle}/>
+  <div className="nx-alert nx-alert--error" role="alert" aria-atomic={true}>
```
rpokorny, line 12: `role="alert"` implies `aria-atomic`.  The component just sets `aria-atomic` universally since some alerts don't have a role.  But in some cases it is redundant.

jflinchbaugh: what about role="status"? does that imply it as well?

jflinchbaugh: ^^ it does, so i'll remove it from there as well.

rpokorny (approved):

santdelv (approved):

PR #225, opened 2020-11-20
IWeek: NxBinaryDonutChart A11y

rpokorny:
https://trello.com/c/qWNKtm0g/27-nxbinarydonutchart


Blacktiger (approved):

santdelv (approved):

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
```diff
@@ -10,12 +10,12 @@ import { NxBinaryDonutChart } from '@sonatype/react-shared-components';
...
     <>
-      <NxBinaryDonutChart percent={0} />
+      <NxBinaryDonutChart percent={0} aria-label="An empty donut chart" />
       <NxBinaryDonutChart percent={15} />
       <NxBinaryDonutChart percent={25} />
-      <NxBinaryDonutChart percent={50} />
+      <NxBinaryDonutChart percent={50} aria-label="A half-full donut chart" />
       <NxBinaryDonutChart percent={90} />
-      <NxBinaryDonutChart percent={100} />
+      <NxBinaryDonutChart percent={100} aria-label="A full donut chart" />
```
jflinchbaugh, line 18: I'm surprised we show an example with squishy english descriptions (empty, half, full), since I'd expect it's always going to be fed some dynamic value in any real use. Why not "100% donut chart", etc?

jflinchbaugh (approved):
+1

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
rpokorny: > I'm surprised we show an example with squishy english descriptions (empty, half, full), since I'd expect it's always going to be fed some dynamic value in any real use. Why not "100% donut chart", etc?

It should be text that explains the context of the situation.  I've added a more illustrative example in d556d6c2

PR #226, opened 2020-11-20
Improving accessibility on nx-list examples.

santdelv:
I added the list role even if it should not been added, it is because of a Safari + VO oddity (https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html)
I did not add aria-selected because it was not working inside of a list.
aria disabled inside of an li element works only for Safari.

### Additions as of June 11, 2021:

- Converted the clickable example to use `<button>`'s which are semantically more correct
- Removed `tabindex` and `role` attributes because they are redundant
- Used `aria-pressed` to announce the "selected" state for the example that uses `<button>`'s https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role 
- there is no equivalent for `<a>` tags, so when a list requires a "sticky" selected state `<button>`'s should be used

There is a CSS bug (fixed in Chrome and FF) that causes elements that appear with a parent that has `display: contents` to be removed from the accessibility tree. Hopefully this will be fixed soon.


rpokorny:
> I did not add aria-selected because it was not working inside of a list.

Hmm by the spec, it seems that `aria-selected` is not defined for use within lists, so I'm not sure what to do there.

gallery/src/styles/NxList/NxListClickableExample.tsx
```diff
@@ -22,7 +22,7 @@ const NxListClickableExample = () =>
       <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
     </li>
     <li className="nx-list__item selected" tabIndex={0}>
-      <span className="nx-list__text">This list item is selected</span>
+      <span className="nx-list__text" aria-selected="true">This list item is selected</span>
```
rpokorny, line 25: PR description says you didn't add this?

Blacktiger: I'd recommend putting aria-selected on the element with the tabIndex set to ensure that screen readers will stop on that element to read it.

santdelv: > PR description says you didn't add this?

I tested by adding it in the span and li, I missed deleting it from the span.

santdelv: @rpokorny 34ddd91819afb9e23d3790004adc1354165df210
@Blacktiger the aria-selected doesn't seem necessary for the screen reader to stop and read it, and it does not make the screen reader say anything about it being selected.

rpokorny:
> Hmm by the spec, it seems that aria-selected is not defined for use within lists, so I'm not sure what to do there.

Maybe we should try `aria-current` instead

santdelv:
> > Hmm by the spec, it seems that aria-selected is not defined for use within lists, so I'm not sure what to do there.
> 
> Maybe we should try `aria-current` instead

That worked well 782252b99231c1fe448c680c81a0babc8e8d7883

rpokorny:
> That worked well 782252b

I don't understand the way you added that.  Why is it on the disabled list item not the one that is selected?  Also I see you are still using `aria-selected` on the actual selected list item, despite the conversations above about how you weren't because it doesn't work.  By the way, `aria-selected` does actually work there in ChromeVox - it gets called out as selected.

santdelv:
> > That worked well [782252b](https://github.com/sonatype/sonatype-react-shared-components/commit/782252b99231c1fe448c680c81a0babc8e8d7883)
> 
> I don't understand the way you added that. Why is it on the disabled list item not the one that is selected? Also I see you are still using `aria-selected` on the actual selected list item, despite the conversations above about how you weren't because it doesn't work. By the way, `aria-selected` does actually work there in ChromeVox - it gets called out as selected.

I got confused when testing because it said "This item list is selected" and that is what the text field says. Neither aria-current or aria-selected work for me, but if you can confirm aria-selected works for chrome I think there is no harm on leaving it.

Blacktiger (approved):

gallery/src/styles/NxList/NxListClickableExample.tsx
```diff
@@ -10,31 +10,41 @@ import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
 
 const NxListClickableExample = () =>
   <ul className="nx-list nx-list--clickable">
-    <li className="nx-list__item" tabIndex={0}>
-      <span className="nx-list__text">Action 1</span>
-      <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+    <li className="nx-list__item nx-list__item--link">
```
rpokorny, line 13: I feel like the `.nx-list__item--link` class name should be changed if it's being used for things other than links.  Probably `.nx-list__item--clickable`.  Of course we'll still need to keep `.nx-list__item--link` for backwards compat too

rpokorny:
Since we don't want this to be a breaking change, we need to make sure that the old way of doing button lists still works (though it would be deprecated).  So we need an example of that way too

gallery/src/styles/NxList/NxListClickableExample.tsx
```diff
@@ -10,31 +10,41 @@ import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
...
+      </button>
     </li>
-    <li className="nx-list__item" tabIndex={0}>
-      <span className="nx-list__text">Action 4</span>
-      <span className="nx-list__subtext">
-        This is a fourth line. It includes text that might relate to the top line or might not.
-      </span>
-      <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+    <li className="nx-list__item nx-list__item--link">
+      <button className="nx-list__btn selected" aria-pressed="true">
```
rpokorny, line 35: I think we need to look a little bit more into the difference between "pressed" and "selected", which element should have whichever one we want to use (I assume we won't need both) and which item in this example actually demonstrates that state. ChromeVox seems to support both the way this item uses `aria-pressed` and the way the one above uses `aria-selected` - they both get read as pressed or selected respectively.

gallery/src/styles/NxList/NxListClickableExample.tsx
```diff
@@ -10,31 +10,41 @@ import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
...
+          This list item demonstrates the selected styles
+        </span>
+        <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+      </button>
     </li>
-    <li className="nx-list__item disabled">
-      <span className="nx-list__text">This list item is disabled</span>
-      <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+    <li className="nx-list__item nx-list__item--link" aria-disabled="true">
+      <button className="nx-list__btn disabled">
```
rpokorny, line 44: Would the `disabled` attribute work here? I'd imagine that might knock out both the styling and the screen reading with a single value rather than needing both the `aria` attr and the class name.

gallery/src/styles/NxList/NxListClickableLinksExample.tsx
```diff
@@ -11,13 +11,13 @@ import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
...
     <li className="nx-list__item nx-list__item--link">
-      <a href="#/pages/nx-list" className="nx-list__link selected">
+      <a href="#/pages/nx-list" className="nx-list__link">
         <span className="nx-list__text">nx-list page</span>
         <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
       </a>
     </li>
     <li className="nx-list__item nx-list__item--link">
-      <a href="#/pages/nx-table" className="nx-list__link">
+      <a href="#/pages/nx-table" className="nx-list__link selected">
```
rpokorny, line 20: Any reason this got moved?

gallery/src/styles/NxList/NxListDefaultExample.html
```diff
@@ -9,7 +9,7 @@
 <h3 class="nx-h3">
   List heading
 </h3>
-<ul class="nx-list">
+<ul class="nx-list" role="list">
```
rpokorny, line 12: Is this necessary? It's not on the other examples and seems like it'd be redundant

rpokorny:
ChromeVox is still less than ideal on some of the examples lower down on the page:

* On the bulleted example, it is inconsistent about the nested lists.  When arriving at the top level list, it announces it as a list of x items but then only reads the first item. You can then navigate down and it'll read the basic text of the second item, but not (immediately) the nested list in the second item.  Then if you navigate down again it goes into the second list, giving no indication that it is a sublist.  In the second list the behavior is the same - it reads the items only one at a time and it doesn't read the third (double-nested) list along with the other content of its third item.  However, then when you get to that third list, it behaves differently! It highlights and reads the entire list all at once.  The pattern I'm starting to see is that it'll only read the entire list if none of the items contain any "rich" content of their own.  It still doesn't however clarify that it's a sublist within the third item of the list above. It also doesn't indicate when you've exited the bottom of a nested list and are back in a higher-level list.  I tend to doubt there's much we can do about any of this.

* In the action menu example, each list item is again read out one at a time, but it doesn't read the whole item.  It reads the first (main) bit of text, identifying _that_ as the list-item.  Then navigating down again causes it to read the subtext (identified as nothing in particular) and continuing to navigate down causes it to read the action buttons, one at a time.  It doesn't make it clear that all this content is still part of the same list item.  Again, I'm not sure this is something we can do anything about.  As I said above, it seems to want to break up any list items that have rich content.

* The multi-line example is broken up similarly, with the subtext read separately from the rest of the list item.

FWIW I tried `aria-atomic` to fix the breaking-up of list items, but it didn't help.

gallery/src/styles/NxList/NxListClickableExample.tsx
rpokorny: > I think we need to look a little bit more into the difference between "pressed" and "selected", which element should have whichever one we want to use (I assume we won't need both) and which item in this example actually demonstrates that state. ChromeVox seems to support both the way this item uses aria-pressed and the way the one above uses aria-selected - they both get read as pressed or selected respectively.

Also see the older discussions in the PR about these and `aria-current`.

glenhunter: > I think we need to look a little bit more into the difference between "pressed" and "selected", which element should have whichever one we want to use (I assume we won't need both) and which item in this example actually demonstrates that state. ChromeVox seems to support both the way this item uses `aria-pressed` and the way the one above uses `aria-selected` - they both get read as pressed or selected respectively.

I picked "pressed" specifically because I felt that it matched best with the use of `<button>`'s. However there is a use case where "selected" may make more sense and that is when a list is used as a way to show chunks of content on a page (I think APL does this). In that case I think that "selected" is more appropriate.

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > Would the `disabled` attribute work here? I'd imagine that might knock out both the styling and the screen reading with a single value rather than needing both the `aria` attr and the class name.

It does not:
```
Property 'disabled' does not exist on type 'DetailedHTMLProps<LiHTMLAttributes<HTMLLIElement>, HTMLLIElement>
```

gallery/src/styles/NxList/NxListDefaultExample.html
glenhunter: > Is this necessary? It's not on some of the other examples and seems like it'd be redundant

It is not. I should have removed it: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/8a50d3e2699f67a312bdbaaaaa284d1760d5183e

gallery/src/styles/NxList/NxListClickableLinksExample.tsx
glenhunter: > Any reason this got moved?

I felt that the styling in the example was clearer when the selected state was on the second item rather than the first

gallery/src/styles/NxList/NxListClickableExample.tsx
rpokorny: > It does not:

The button; not the li

gallery/src/styles/NxList/NxListDefaultExample.html
rpokorny: > It is not. I should have removed it: 8a50d3e

It's present is several other examples as well

gallery/src/styles/NxList/NxListClickableExample.tsx
rpokorny: > I picked "pressed" specifically because I felt that it matched best with the use of <button>'s. However there is a use case where "selected" may make more sense and that is when a list is used as a way to show chunks of content on a page (I think APL does this). In that case I think that "selected" is more appropriate.

I'm not sure I understand the distinction you're making, but surely it doesn't make sense to use them both in the same example as we have here right?  And if they do each make sense in different situations, we should document when to use each.  And consider carefully which element it should be used on (the `<li>` vs the `<button`>, and also what about the lists of links?)

lib/src/base-styles/_nx-lists.scss
```diff
@@ -20,7 +20,7 @@
 }
 
 .nx-list__item {
-  &, > .nx-list__link {
+  &, > .nx-list__link, & > .nx-list__btn {
```
rpokorny, line 23: Maybe a different class name should be chosen that would be common to both of them

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > I feel like the `.nx-list__item--link` class name should be changed if it's being used for things other than links. Probably `.nx-list__item--clickable`. Of course we'll still need to keep `.nx-list__item--link` for backwards compat too

Addressed here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/035f4534eb8c5c58eaa029563f146b39d9acc559

gallery/src/styles/NxList/NxListExamples.tsx
```diff
@@ -101,6 +103,15 @@ const NxListsExamples = () =>
       Basic <code className="nx-code">nx-list--description</code> with a heading. Wrapping is demonstrated in both
       the terms and the descriptions.
     </GalleryExampleTile>
+
+    <GalleryExampleTile title="Deprecated NX Clickable List Example"
+                        id="nx-list-deprecated-clickable-example"
+                        liveExample={NxListDeprecatedClickableExample}
+                        codeExamples={NxListDeprecatedClickableCode}>
+      This method of created clickable lists is deprecated. It is strongly recommended that one of the above clickable
```
glenhunter, line 111:
```suggestion
      This method of creating clickable lists is deprecated. It is strongly recommended that one of the above clickable
```

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > > I picked "pressed" specifically because I felt that it matched best with the use of 's. However there is a use case where "selected" may make more sense and that is when a list is used as a way to show chunks of content on a page (I think APL does this). In that case I think that "selected" is more appropriate.
> 
> I'm not sure I understand the distinction you're making, but surely it doesn't make sense to use them both in the same example as we have here right? And if they do each make sense in different situations, we should document when to use each. And consider carefully which element it should be used on (the `<li>` vs the `<button>`, and also what about the lists of links?)

Changed over to `aria-selected` here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/62c077029f3259c2c04e2f525505a2e7e5d5f006

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > > It does not:
> 
> The button; not the li

Ah. Required a SCSS tweak, but works nicely (confirmed in VO): https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/8390a450712248fa2df174655777e24386d22c2c

gallery/src/styles/NxList/NxListClickableExample.tsx
```diff
@@ -10,31 +10,41 @@ import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
 
 const NxListClickableExample = () =>
   <ul className="nx-list nx-list--clickable">
-    <li className="nx-list__item" tabIndex={0}>
-      <span className="nx-list__text">Action 1</span>
-      <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+    <li className="nx-list__item nx-list__item--clickable">
```
rpokorny, line 13: The top border here is thicker than it used to be

lib/src/base-styles/_nx-lists.scss
glenhunter: > Maybe a different class name should be chosen that would be common to both of them

The best name I could think of was `__click-target` since both are within `--clickable`. I'm actually OK with keeping them distinct, especially since I can imagine ways in which the link and button examples could diverge in use or style. But if you feel strongly and can suggest a good name we can change them.

gallery/src/styles/NxList/NxListClickableExample.tsx
rpokorny: > Changed over to aria-selected here: 62c0770

Needed on the links example too

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > The top border here is thicker than it used to be

Yeah, I noticed that too, fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/24e0af1dec552b42d7c08a2b7cc8a3a486359a52

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > > Changed over to aria-selected here: [62c0770](https://github.com/sonatype/sonatype-react-shared-components/commit/62c077029f3259c2c04e2f525505a2e7e5d5f006)
> 
> Needed on the links example too

I purposefully did not put it on the link example because I don't think that the `<a>` should be used when we have a "selected" state (it should not be sticky in that way), the `<a>` should only be used for navigation.

glenhunter: Per discussion in Slack moved to `aria-current` rather than `aria-selected`: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/436eac785f56aa512bf85bfdf6668af0913ba9b9

gallery/src/styles/NxList/NxListClickableExample.tsx
rpokorny: > Yeah, I noticed that too, fixed here: 24e0af1

Pretty sure that's wrong.  I think you were trying to remove the border from the first list item there but you're actually removing the top border on the list itself, when the list is the first item within its parent.

gallery/src/styles/NxList/NxListClickableExample.tsx
rpokorny: > I purposefully did not put it on the link example because I don't think that the <a> should be used when we have a "selected" state (it should not be sticky in that way), the <a> should only be used for navigation.

I don't follow.  It is very common for one of the links in the list to be to the current page.  Think about the gallery side nav, or the IQ Violation Details sidebar.  In that case the current link is styled in a highlighted way and IMO it should be called out differently too

gallery/src/styles/NxList/NxListClickableExample.tsx
glenhunter: > > Yeah, I noticed that too, fixed here: [24e0af1](https://github.com/sonatype/sonatype-react-shared-components/commit/24e0af1dec552b42d7c08a2b7cc8a3a486359a52)
> 
> Pretty sure that's wrong. I think you were trying to remove the border from the first list item there but you're actually removing the top border on the list itself, when the list is the first item within its parent.

Oops, got my indentation wrong: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/dab89e9c8dc90fc2c1be4ef0db1b3746fd3c27f0

glenhunter:
> Since we don't want this to be a breaking change, we need to make sure that the old way of doing button lists still works (though it would be deprecated). So we need an example of that way too

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/6b50a39361217c624e71ec6559bd14c1503472db

glenhunter:
> Since we don't want this to be a breaking change, we need to make sure that the old way of doing button lists still works (though it would be deprecated). So we need an example of that way too

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/6b50a39361217c624e71ec6559bd14c1503472db

rpokorny:
> Fixed here: 6b50a39

I don't think we should bother with aria attrs at all in the deprecated example.

rpokorny:
I think there should be a follow-up ticket to make more substantial react components for nx-list so that downstream projects don't need to manage all the details present here.

glenhunter:
> I think there should be a follow-up ticket to make more substantial react components for nx-list so that downstream projects don't need to manage all the details present here.

https://issues.sonatype.org/browse/RSC-609

glenhunter:
> > Fixed here: [6b50a39](https://github.com/sonatype/sonatype-react-shared-components/commit/6b50a39361217c624e71ec6559bd14c1503472db)
> 
> I don't think we should bother with aria attrs at all in the deprecated example.

https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/07457add3743c3651a530814842d4a05eee72c3c

gallery/src/styles/NxList/NxListDeprecatedClickableExample.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { faAngleRight } from '@fortawesome/free-solid-svg-icons';
+import { NxFontAwesomeIcon } from '@sonatype/react-shared-components';
+
+const NxListClickableExample = () =>
```
rpokorny, line 11:
```suggestion
const NxListDeprecatedClickableExample = () =>
```

gallery/src/styles/NxList/NxListDeprecatedClickableExample.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+    </li>
+    <li className="nx-list__item">
+      <span className="nx-list__text">Action 4</span>
+      <span className="nx-list__subtext">
+        This list item demonstrates the selected styles
+      </span>
+      <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
+    </li>
+    <li className="nx-list__item">
```
rpokorny, line 35: This is missing disabled styling

rpokorny:
Class List table in the gallery needs updating

rpokorny:
Needs visual test for the deprecated example.  I assume the others already have visual tests

glenhunter:
> Needs visual test for the deprecated example. I assume the others already have visual tests

They do, and here's the deprecated test: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/3911a92760a520cad907636907353890ff7a7a4d

glenhunter:
> Class List table in the gallery needs updating

Ah yes, updated: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/22625098aa6d29418c31cb78f9d663f5da6c32d5

gallery/src/styles/NxList/NxListDeprecatedClickableExample.tsx
glenhunter: > This is missing disabled styling

https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/aa997b08ac9e3b4e50d3919dd46ccce0da944e3a

gallery/src/styles/NxList/NxListPage.tsx
```diff
@@ -124,6 +124,32 @@ const NxListPage = () =>
...
             </td>
           </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-list__item--clickable</code></td>
+            <td className="nx-cell"/>
+            <td className="nx-cell">Modifier of <code className="nx-code">.nx-list__item</code></td>
+            <td className="nx-cell">
+              Applied to <code className="nx-code">.nx-list__item</code> this allows for the correct styling of
+              clickable lists. Ensure that the <code className="nx-code">.nx-list</code> modifier
+              {' '}<code className="nx-code">.nx-list--clickable</code> is also used.
```
rpokorny, line 134: Should probably elaborate on how it must contain a single `.nx-list__btn` or `.nx-list__link` as a child

gallery/src/styles/NxList/NxListPage.tsx
```diff
@@ -124,6 +124,32 @@ const NxListPage = () =>
...
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-list__btn</code></td>
+            <td className="nx-cell"/>
```
rpokorny, line 139: Should these two have convenience components?

glenhunter: > Should these two have convenience components?

I'm not certain if they _should_, but they certainly can: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/bcbc7f97a37439bead92cdc57ec15a87f4e6ac29

gallery/src/styles/NxList/NxListPage.tsx
glenhunter: > Should probably elaborate on how it must contain a single `.nx-list__btn` or `.nx-list__link` as a child

https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/f54fbde52ee8d895cbb4b916ab17a599dfa4f056

gallery/src/styles/NxList/NxListPage.tsx
```diff
@@ -124,6 +124,34 @@ const NxListPage = () =>
               would only be a single list item which contains the error alert.
             </td>
           </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-list__item--clickable</code></td>
```
rpokorny, line 128: I feel like either this class or `.nx-list--clickable` shouldn't be necessary.  One or the other should be enough to provide the necessary styling.

glenhunter: > I feel like either this class or `.nx-list--clickable` shouldn't be necessary. One or the other should be enough to provide the necessary styling.

~The deprecated clickable lists require `.nx-list--clickable` to differentiate the <li>’s from the non-clickable ones, and we need `.nx-list__item--link` and `.nx-list__item--clickable` to differentiate those <li>’s from the deprecated ones.~

I figured out a solution, though in some respects I like the intentionality of adding `nx-list--clickable` to lists. https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/a24817c3fb8609eb0f60d86509a1bb46fc1c99e0

rpokorny:
Top of the focus border on the first item is janky: 
image

gallery/src/styles/NxList/NxListPage.tsx
rpokorny: > I figured out a solution, though in some respects I like the intentionality of adding nx-list--clickable to lists. a24817c
>
> WDYT?

I like it

lib/src/base-styles/_nx-lists.scss
```diff
@@ -171,59 +167,84 @@
...
-      }
+.nx-list__item--link, .nx-list__item--clickable {
+  display: contents;
+}
 
-      &:hover {
-        background-color: var(--nx-color-interactive-background-hover);
-        cursor: pointer;
+.nx-list__item--clickable, // preferred
+.nx-list__item--link,
```
rpokorny, line 176: This one is deprecated too right?

lib/src/base-styles/_nx-lists.scss
```diff
@@ -171,59 +167,84 @@
...
+    }
 
-      &:focus {
-        box-shadow: var(--nx-box-shadow-focus);
-        border-color: var(--nx-color-interactive-border-focus);
-        outline: none;
-      }
+    &:active {
+      background-color: var(--nx-color-interactive-background-active);
+      color: inherit;
```
rpokorny, line 193: I think we need this all the time not just on `:active`.  It seems that inactive ones get pure black as their text color (user-agent default button style)

glenhunter: > I think we need this all the time not just on `:active`. It seems that inactive ones get pure black as their text color (user-agent default button style)

Moving `color: inherit;` to the top of these rules so that it covers links and buttons fixed it as well as for all interactive states. https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/f7894c0f9939a5543d848d6e4f8abb25b89b030b

lib/src/base-styles/_nx-lists.scss
glenhunter: > This one is deprecated too right?

Yeah, in my mind this was sort of a soft deprecation, but noted here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/0df257c287abae5db9943f4590e9739269cac2a0

glenhunter:
> Top of the focus border on the first item is janky:
> image

The fix simplified the code, but makes me suspicious about why it was the way it was: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/b1c3ba5d14be074d7b76861526463d75df102220

However I can find no fault with the change

rpokorny:
> The fix simplified the code, but makes me suspicious about why it was the way it was: b1c3ba5
>
> However I can find no fault with the change

It does beg the question of what an empty `.nx-list` should look like

rpokorny:
> It does beg the question of what an empty .nx-list should look like

Well I guess we've got the special styles for the row that's supposed to tell you its empty, so it shoudl never truly be empty, DOM wise

lib/src/base-styles/_nx-lists.scss
```diff
@@ -42,7 +41,7 @@
   }
 
   &:first-child {
-    border-top: none;
+    border-top: var(--nx-border-default);
```
rpokorny, line 44: Checking whether this is redundant made me realize that our default border and subsection border variables are currently identical.  Is that right?

rpokorny (approved):

lib/src/base-styles/_nx-lists.scss
glenhunter: > Checking whether this is redundant made me realize that our default border and subsection border variables are currently identical. Is that right?

So the subsection border in `nx-tile` does match the subsection border variable. However the lists should not have been using that variable, so I have changed it here: https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/bfd417b8b2109d39bc5b389337c5f0c31970475b

rpokorny: > So the subsection border in nx-tile does match the subsection border variable.

Not sure how you figure that, looking at the code it's using the variable for it.

> However the lists should not have been using that variable, so I have changed it here

What makes you say that? I seem to remember @dturner32 saying this was an example of a subsection border

glenhunter: > > So the subsection border in nx-tile does match the subsection border variable.
> 
> Not sure how you figure that, looking at the code it's using the variable for it.
> 
> > However the lists should not have been using that variable, so I have changed it here
> 
> What makes you say that? I seem to remember @dturner32 saying this was an example of a subsection border

I'm looking at the colours in Sketch, and it's grey, not the blue-grey of an `nx-tile` subsection border. The top border of the list is the same blue-grey as the subsection border

rpokorny: > I'm looking at the colours in Sketch, and it's grey, not the blue-grey of an nx-tile subsection border. The top border of the list is the same blue-grey as the subsection border

@dturner32 please clarify the following:
* what color "subsection borders" in general are supposed to be
* what color the tile subsection border is supposed to be
* what color the top border of a list is supposed to be
* what color the borders between rows of a list are supposed to be

dturner32: (related to the current color palette, not the new variables)
• subsection border color is #D1D6F0
• tile subsection border is also #D1D6F0 
• top border of list in #D1D6F0
• border color between rows is #E9E9E9
• yes, list borders will fall under subsection borders

glenhunter: > border color between rows is #E9E9E9
> • yes, list borders will fall under subsection borders

Aren't these mutually exclusive?

glenhunter: > (related to the current color palette, not the new variables)
> • subsection border color is #D1D6F0
> • tile subsection border is also #D1D6F0
> • top border of list in #D1D6F0
> • border color between rows is #E9E9E9
> • yes, list borders will fall under subsection borders

I think that it would be best if you confirmed what is in Sketch. If Sketch is incorrect then correct and post here, if Sketch is correct, post here. I don't want to move forward on this until I know that what I'm doing will match Sketch.

dturner32: I checked this against the Sketch file and these colors are coming directly from the components which are correct. That being said, these colors are different for the v2 variable updates. If those are the colors you'd rather work with I can provide those. Otherwise, the colors above match what's designed in sketch.

glenhunter: > I checked this against the Sketch file and these colors are coming directly from the components which are correct. That being said, these colors are different for the v2 variable updates. If those are the colors you'd rather work with I can provide those. Otherwise, the colors above match what's designed in sketch.

I think we'll leave that for a separate ticket

rpokorny: > (related to the current color palette, not the new variables)

To be clear, that is of no value since RSC is no longer in the old color palette

rpokorny: > yes, list borders will fall under subsection borders

Given that you specified that there are two different colors of list borders, I have to assume this comment was only referring to one of them.  From what it sounds like, the top border of a list should be a "subsection border", that is, should match the color of tile subsections.  The borders between list items, on the other hand, should be some other kind of border that we perhaps don't have a name for.  I think I remember reading or hearing something to the effect that this latter color is also supposed to be used between table rows, is that correct?  Note that that is not currently the case - right now table borders use the "subsection" border styling.

Adding to this complication is the fact that at the moment, we have subsection borders defined equivalently to default borders.  I have to think that was a mistake at some point.

rpokorny: But anyway this isn't terribly relevant to this PR.  I'm fine with straightening out the borders in another ticket so long as this PR doesn't make them worse

rpokorny: > But anyway this isn't terribly relevant to this PR. I'm fine with straightening out the borders in another ticket so long as this PR doesn't make them worse

This PR does currently have it differing from master which seems like a problem

glenhunter: > > But anyway this isn't terribly relevant to this PR. I'm fine with straightening out the borders in another ticket so long as this PR doesn't make them worse
> 
> This PR does currently have it differing from master which seems like a problem

Because the colour values need to be HSL I have used the v2 colours supplied by @dturner32, which is our "normal" `indigo-95` colour for the top border and `indigo-95` for the internal borders. https://github.com/sonatype/sonatype-react-shared-components/pull/226/commits/8b33df4bef0b92b8e22d0313b24a5709ecffb3e3

Also based on discussions with @rpokorny I have created  https://issues.sonatype.org/browse/RSC-614 to look at rationalizing our border colours.

rpokorny: > I have used the v2 colours supplied by @dturner32, which is our "normal" indigo-95 colour for the top border and indigo-95 for the internal borders

So they're both supposed to be indigo-95? The top border on this branch is currently indigo-90

glenhunter: > > I have used the v2 colours supplied by @dturner32, which is our "normal" indigo-95 colour for the top border and indigo-95 for the internal borders
> 
> So they're both supposed to be indigo-95? The top border on this branch is currently indigo-90

Crap, no that was a typo, `indigo-90` for the top border, `indigo-95` for the inside borders

rpokorny (approved):

PR #227, opened 2020-11-24
update padding around buttons in nx alert RSC-377

glenhunter:
https://issues.sonatype.org/browse/RSC-377

This ticket was described as an issue with the buttons but actually it's the default alert padding which has changed. I adjusted that to match the Zeplin designs (16px) on all sides. 

After checking for other changes in Zeplin I added an `NxModal` example demonstrating `NxAlert` in the `NxModal` footer, and a selector that provides the correct bottom margin when an `NxAlert` appears at the top of the page (above the page title).


rpokorny:
Missing 24px between button and text:
image

Also long unbroken words can overflow and push the button out of the alert.  Probably need `overflow-wrap: break-word`.

rpokorny:
Edge case to consider: When an alert with a close button and multiple lines of text is present, the spacing above and below the text is uneven. This is because the top spacing is 16px to the top of the button, which is a little bit higher than the top of the text.  The bottom spacing on the other hand is to the bottom of the text, as the button doesn't go down far enough to affect it:
image

lib/src/base-styles/_nx-alert.scss
```diff
@@ -82,3 +82,7 @@
     color: #007623;
   }
 }
+
+.nx-page-main > .nx-alert {
+  margin-bottom: 24px;
```
rpokorny, line 87: I see that alerts have a default top/bottom margin of 8px.  Where does that value come from? It seems awfully small and I don't see it anywhere in the mockups. I wonder if it's a holdover from 0.x and if we should just set the default top/bottom margin to 24px

glenhunter: > I see that alerts have a default top/bottom margin of 8px. Where does that value come from? It seems awfully small and I don't see it anywhere in the mockups. I wonder if it's a holdover from 0.x and if we should just set the default top/bottom margin to 24px

I went ahead and applied our standard `16px` above, `24px` below component margin to the alert.https://github.com/sonatype/sonatype-react-shared-components/pull/227/commits/d41d9571814c91fbd73b9138e6107347264675fa

glenhunter:
> Edge case to consider: When an alert with a close button and multiple lines of text is present, the spacing above and below the text is uneven. This is because the top spacing is 16px to the top of the button, which is a little bit higher than the top of the text. The bottom spacing on the other hand is to the bottom of the text, as the button doesn't go down far enough to affect it:
> image

I doubt that most people would notice this.

lib/src/base-styles/_nx-alert.scss
glenhunter: > > I see that alerts have a default top/bottom margin of 8px. Where does that value come from? It seems awfully small and I don't see it anywhere in the mockups. I wonder if it's a holdover from 0.x and if we should just set the default top/bottom margin to 24px
> 
> I went ahead and applied our standard `16px` above, `24px` below component margin to the alert.[d41d957](https://github.com/sonatype/sonatype-react-shared-components/commit/d41d9571814c91fbd73b9138e6107347264675fa)

@dturner32 ^^

glenhunter:
> Missing 24px between button and text:
> image

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/227/commits/3d554e6b04d9bb269954fa6e034fb492f24ee3f5
 
> Also long unbroken words can overflow and push the button out of the alert. Probably need `overflow-wrap: break-word`.

Apparently it was `word-wrap: break-word;` that did the trick here: https://github.com/sonatype/sonatype-react-shared-components/pull/227/commits/8a153eec55c0d43a5a325be45341865399a98612

rpokorny:
> I doubt that most people would notice this.

FWIW it caught my eye immediately

rpokorny:
> Apparently it was word-wrap: break-word; that did the trick here: 8a153ee

You set `word-break: break-word`, not `word-wrap: break-word;`.  This family of properties is ridiculously confusing, but here's the summary:

* `word-wrap` is the original non-standard name for `overflow-wrap`.  For the options that they both support, they should be equivalent.  It shouldn't be necessary anymore now that we are dropping IE11.  `word-wrap` and `overflow-wrap` control whether to break up words at all, but on their own still always try to keep words together where possible.

* `overflow-wrap` however also supports a new `anywhere` option that `word-wrap` didn't.  The difference between `anywhere`, and `break-word` is that `anywhere` allows wrapping to affect min instrinsic widths.  Therefore `anywhere` is often the option to use when within an auto-width element like a flex child or table cell.

* `word-break` basically controls whether the browser tries to avoid splitting words (the `normal` setting) or always fills each line box to the greatest extent possible, splitting wherever it fills up (the `break-all` setting).  It also has two other settings which we can currently ignore: `keep-all` which has some effect on CJK text that I don't understand, and `break-word` which is a deprecated setting that's equivalent to `normal` plus `overflow-wrap: anywhere`.

To summarize, you used `word-break: break-word` which is deprecated.  The standard equivalent is `overflow-wrap: anywhere` plus keeping `word-break` set to `normal` as it already was.

glenhunter:
> > Apparently it was word-wrap: break-word; that did the trick here: [8a153ee](https://github.com/sonatype/sonatype-react-shared-components/commit/8a153eec55c0d43a5a325be45341865399a98612)
> 
> You set `word-break: break-word`, not `word-wrap: break-word;`. This family of properties is ridiculously confusing, but here's the summary:
> 
> * `word-wrap` is the original non-standard name for `overflow-wrap`.  For the options that they both support, they should be equivalent.  It shouldn't be necessary anymore now that we are dropping IE11.  `word-wrap` and `overflow-wrap` control whether to break up words at all, but on their own still always try to keep words together where possible.
> * `overflow-wrap` however also supports a new `anywhere` option that `word-wrap` didn't.  The difference between `anywhere`, and `break-word` is that `anywhere` allows wrapping to affect min instrinsic widths.  Therefore `anywhere` is often the option to use when within an auto-width element like a flex child or table cell.
> * `word-break` basically controls whether the browser tries to avoid splitting words (the `normal` setting) or always fills each line box to the greatest extent possible, splitting wherever it fills up (the `break-all` setting).  It also has two other settings which we can currently ignore: `keep-all` which has some effect on CJK text that I don't understand, and `break-word` which is a deprecated setting that's equivalent to `normal` plus `overflow-wrap: anywhere`.
> 
> To summarize, you used `word-break: break-word` which is deprecated. The standard equivalent is `overflow-wrap: anywhere` plus keeping `word-break` set to `normal` as it already was.

Thanks for the summary

https://github.com/sonatype/sonatype-react-shared-components/pull/227/commits/9798a84d44145efeac87f5b9941c98a2d2a4626c

dturner32:
> > Edge case to consider: When an alert with a close button and multiple lines of text is present, the spacing above and below the text is uneven. This is because the top spacing is 16px to the top of the button, which is a little bit higher than the top of the text. The bottom spacing on the other hand is to the bottom of the text, as the button doesn't go down far enough to affect it:
> > image
> 
> I doubt that most people would notice this. @dturner32 WDYT?

its noticeable but i don't know if it's a big enough thing to really warrant altering it. The content in the alert (when its just one line) is centered creating a 20px top and bottom padding. In a case where there are multiple lines we could have the bottom padding of the text be 20px to match the top padding. It's not a standard measurement we use but its a suggestion. Otherwise I'm ok with leaving it as is if want to leave it

rpokorny (approved):

PR #229, opened 2020-11-27
NxPolicyViolationIndicator RSC-339

glenhunter:
https://issues.sonatype.org/browse/RSC-339


gallery/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicatorPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        <code className="nx-code">NxPolicyViolationIndicator</code> is an element used to indicate IQ policy threat
+        level via color and text.
+      </p>
+      <p className="nx-p">
+        There are two scales to choose from: threat level by category, and threat
+        level by number. When using this component, it is expected that just one of the props will be passed. If both
+        are passed, <code className="nx-code">threatLevelCategory</code> takes precedence. If neither are passed,
+        the <code className="nx-code">unspecified</code> category is used
```
rpokorny, line 30:
```suggestion
        the <code className="nx-code">unspecified</code> category is used.
```

gallery/visualtests/nxPolicyViolationIndicator.js
```diff
@@ -0,0 +1,22 @@
...
+    await browser.url('#/pages/NxPolicyViolationIndicator');
+  });
+
+  const policyViolationIndicatorCategoryExampleSelector =
+      '#nx-policy-violation-indicator-category-example .gallery-example-live',
+      policyViolationIndicatorNumberExampleSelector =
+        '#nx-policy-violation-indicator-number-example .gallery-example-live';
+
+  it('looks right', simpleTest(policyViolationIndicatorCategoryExampleSelector));
+  it('looks right', simpleTest(policyViolationIndicatorNumberExampleSelector));
```
rpokorny, line 21: I don't think this second one is necessary.  Keep in mind visual tests are just about different visual variations.  Different data or usage variations would be covered by unit tests.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-variables';
+@import '../../scss-shared/nx-text-helpers';
+
+.nx-policy-violation-indicator {
+  align-items: center;
+  background-color: #fff;
```
rpokorny, line 12: Zeplin shows them as transparent

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+  border: 1px solid;
+  border-radius: $nx-border-radius;
+  display: flex;
+  height: 30px;
+  margin: 0 0 $nx-spacing-xs $nx-spacing-md;
+  max-width: 860px;
```
rpokorny, line 18: I'm wondering whether there's really value in making this explicit.  It was reverse engineered from what we think the max width of a valid policy name could be, so it might be better to just leave it up to that.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+  width: fit-content;
```
rpokorny, line 19: How well did you test this across browsers?  I always have trouble with it, given that the specs have gone back and forth over whether it should be a keyword or a function (e.g. `fit-content(100px)`).  Looks like the way it's turned out is CSS 3 only defines the function syntax and the keyword syntax is deferred to CSS 4.

Moreover, I don't think the content of this component (an icon followed by non-wrapping text) has any difference between the min and max widths, so `min-content`, `max-content,` and `fit-content` all end up doing the same thing, and either of the other two are simpler to understand and probably more widely supported.

rpokorny: > Moreover, I don't think the content of this component (an icon followed by non-wrapping text) has any difference between the min and max widths, so min-content, max-content, and fit-content all end up doing the same thing, and either of the other two are simpler to understand and probably more widely supported.

This does beg the question of what should happen when the indicator wants to be wider than its container.  Right now it overflows. Should it use ellipsis truncation?

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.nx-policy-violation-indicator__badge {
+  border-top-left-radius: $nx-border-radius;
+  border-bottom-left-radius: $nx-border-radius;
```
rpokorny, line 24: if the `.nx-policy-violation-indicator` had `overflow: hidden` this wouldn't be necessary.  That would be the preferred solution as doing it this way creates a gap:
image

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+  height: 30px;
```
rpokorny, line 25: Maybe the parent should use `align-items: stretch`? (or `align-self: stretch` on this item)?

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+
+  .nx-icon {
+    color: #fff;
+    vertical-align: -5px;
```
rpokorny, line 29: Rather than this I recommend making `.nx-policy-violation-indicator__badge` a flex container and using `align-items`.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+.nx-policy-violation-indicator__text {
+  @include semi-bold();
+
+  border-top-right-radius: $nx-border-radius;
+  border-bottom-right-radius: $nx-border-radius;
```
rpokorny, line 37: Same as https://github.com/sonatype/sonatype-react-shared-components/pull/229/files#r532758384

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+  display: inline-block;
```
rpokorny, line 38: I don't think this does anything

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+  font-size: $nx-font-size-s;
+  margin: 0 $nx-spacing-md;
+  text-transform: uppercase;
+  vertical-align: middle;
```
rpokorny, line 42: Doesn't do anything; this should be controlled with flex alignment.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.nx-policy-violation-indicator--unspecified {
+  border-color: $nx-threat-color-unspecified;
+
+  .nx-policy-violation-indicator__badge {
+    background-color: $nx-threat-color-unspecified;
```
rpokorny, line 49: If you wanted to, these could be a good time to play with CSS custom properties/variables. The border color and background color would be defined just once in the non-modifier CSS blocks (using a custom property), and then these modifier blocks would just set the custom property.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+import React, { FunctionComponent, ReactElement } from 'react';
+
+import { ThreatLevelCategory, categoryByPolicyThreatLevel } from '../../util/threatLevels';
+import { Props, propTypes } from './types';
+export { Props } from './types';
+
+import './NxPolicyViolationIndicator.scss';
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
```
rpokorny, line 15: Minor: organizing imports

```suggestion
import React, { FunctionComponent, ReactElement } from 'react';
import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';

import { ThreatLevelCategory, categoryByPolicyThreatLevel } from '../../util/threatLevels';
import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';

import { Props, propTypes } from './types';
export { Props } from './types';

import './NxPolicyViolationIndicator.scss';
```

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const NxPolicyViolationIndicator: FunctionComponent<Props> =
+  function NxPolicyViolationIndicator({ threatLevelCategory, policyThreatLevel, children }): ReactElement<Props> {
```
rpokorny, line 18: If we would ever want the ability to add a tooltip to one of these you'll need to use `forwardRef`.  See `NxButton` as an example.

rpokorny, line 18: This should be redundant:
```suggestion
  function NxPolicyViolationIndicator({ threatLevelCategory, policyThreatLevel, children }) {
```

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const category: ThreatLevelCategory =
+          threatLevelCategory ? threatLevelCategory :
+          policyThreatLevel != null ? categoryByPolicyThreatLevel[policyThreatLevel] : 'unspecified',
```
rpokorny, line 21:
```suggestion
        threatLevelCategory ? threatLevelCategory :
        policyThreatLevel != null ? categoryByPolicyThreatLevel[policyThreatLevel] : 'unspecified',
```

rpokorny, line 21: I like to format nested ternaries so that each line is basically an if-then, with a final line for the base case.
```suggestion
    const category: ThreatLevelCategory =
          threatLevelCategory ? threatLevelCategory :
          policyThreatLevel != null ? categoryByPolicyThreatLevel[policyThreatLevel] : 
          'unspecified',
```

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    return (
+      <div className={className}>
+        <div className="nx-policy-violation-indicator__badge">
+          <NxFontAwesomeIcon aria-hidden={false}
+                             aria-label={label}
+                             icon={faExclamationCircle}/>
+        </div>
+        <div className="nx-policy-violation-indicator__text">
+          <span>{children || category}</span>
```
rpokorny, line 34: Unnecessary span?  Additionally would the `.nx-policy-violation-indicator__badge` and `nx-policy-violation-indicator__text` be better as spans than divs?

glenhunter: > Unnecessary span? Additionally would the `.nx-policy-violation-indicator__badge` and `nx-policy-violation-indicator__text` be better as spans than divs?

Removing the span: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/3da1f0356c771ee48cfe2d33ec036362524de1c8

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > > Moreover, I don't think the content of this component (an icon followed by non-wrapping text) has any difference between the min and max widths, so min-content, max-content, and fit-content all end up doing the same thing, and either of the other two are simpler to understand and probably more widely supported.
> 
> This does beg the question of what should happen when the indicator wants to be wider than its container. Right now it overflows. Should it use ellipsis truncation?

Truncation: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/3da1f0356c771ee48cfe2d33ec036362524de1c8

gallery/visualtests/nxPolicyViolationIndicator.js
glenhunter: > I don't think this second one is necessary. Keep in mind visual tests are just about different visual variations. Different data or usage variations would be covered by unit tests.

Ah, removed: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/f2ced0972a54e081b9882d395dffbd79bfa59f0c

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > Zeplin shows them as transparent

I missed that: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/61198f4e3a2082ed38d352e79ed376ccab2a2d6c

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > I'm wondering whether there's really value in making this explicit. It was reverse engineered from what we think the max width of a valid policy name could be, so it might be better to just leave it up to that.

I only had it there bc it was explicitly mentioned in the Zeplin file, I'd prefer that max-widths be handled by the container wherever possible and that should be the case here and so I have removed it: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/77791f37d04ce6a2ab9d195907412b2c1fadb211

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > How well did you test this across browsers? I always have trouble with it, given that the specs have gone back and forth over whether it should be a keyword or a function (e.g. `fit-content(100px)`). Looks like the way it's turned out is CSS 3 only defines the function syntax and the keyword syntax is deferred to CSS 4.
> 
> Moreover, I don't think the content of this component (an icon followed by non-wrapping text) has any difference between the min and max widths, so `min-content`, `max-content,` and `fit-content` all end up doing the same thing, and either of the other two are simpler to understand and probably more widely supported.

I used `min-content` in the end: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/3da1f0356c771ee48cfe2d33ec036362524de1c8

glenhunter: > > How well did you test this across browsers? I always have trouble with it, given that the specs have gone back and forth over whether it should be a keyword or a function (e.g. `fit-content(100px)`). Looks like the way it's turned out is CSS 3 only defines the function syntax and the keyword syntax is deferred to CSS 4.
> > Moreover, I don't think the content of this component (an icon followed by non-wrapping text) has any difference between the min and max widths, so `min-content`, `max-content,` and `fit-content` all end up doing the same thing, and either of the other two are simpler to understand and probably more widely supported.
> 
> I used `min-content` in the end: [3da1f03](https://github.com/sonatype/sonatype-react-shared-components/commit/3da1f0356c771ee48cfe2d33ec036362524de1c8)

And then I read the spec some more and decided that `max-content` was better: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/694452a627541181b8abcb3a14fa5eec174586db

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > if the `.nx-policy-violation-indicator` had `overflow: hidden` this wouldn't be necessary. That would be the preferred solution as doing it this way creates a gap:
> image

Yes, I never remember that: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/85320ea38d108b6339d28c5b4e3dba332ad1c6c4

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > Maybe the parent should use `align-items: stretch`? (or `align-self: stretch` on this item)?

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/cf4b075d9ef14be82fc2383067caec93e3200dba

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > Rather than this I recommend making `.nx-policy-violation-indicator__badge` a flex container and using `align-items`.

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/cf4b075d9ef14be82fc2383067caec93e3200dba

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > I don't think this does anything

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/cf4b075d9ef14be82fc2383067caec93e3200dba

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > Doesn't do anything; this should be controlled with flex alignment.

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/cf4b075d9ef14be82fc2383067caec93e3200dba

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > Same as https://github.com/sonatype/sonatype-react-shared-components/pull/229/files#r532758384

https://github.com/sonatype/sonatype-react-shared-components/commit/85320ea38d108b6339d28c5b4e3dba332ad1c6c4

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
glenhunter: > If we would ever want the ability to add a tooltip to one of these you'll need to use `forwardRef`. See `NxButton` as an example.

I find it hard to think of a reason why we'd want to wrap one of these in a tooltip but I added `forwardRef` anyway: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/891cfe3686e33de89c180984dbae252e804cb014

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > If you wanted to, these could be a good time to play with CSS custom properties/variables. The border color and background color would be defined just once in the non-modifier CSS blocks (using a custom property), and then these modifier blocks would just set the custom property.

Tricky but fun, thanks for the tip that got this across the finish line: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/22cfc350d2aa4b3010f9bf38ead213a04e9dab04

gallery/visualtests/nxPolicyViolationIndicator.js
rpokorny: > Ah, removed: f2ced09

I believe you've got an unused variable now.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -37,8 +36,8 @@ const NxPolicyViolationIndicator: FunctionComponent<Props> =
         </div>
       </div>
     );
-  };
+  );
 
-NxPolicyViolationIndicator.propTypes = propTypes;
+NxPolicyViolationIndicator.propTypes = propTypes
```
rpokorny, line 41: Lost a semi-colon here

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.nx-policy-violation-indicator__text {
+  @include semi-bold();
+  @include nx-truncate-ellipsis;
+
+  font-size: $nx-font-size-s;
+  margin: 0 $nx-spacing-md;
+  text-transform: uppercase;
+  // max width of element from design is 859px, subtract the badge, the padding, and the borders.
+  max-width: 793px;
```
rpokorny, line 41: I'm thinking about this more and wondering if what we should really do is make the indicators inline-block elements, considering their typical placement, and also give them a max-width of 100%.  What you have now doesn't really address the issue at hand, which is the possibility of the indicator being wider than its container.  For example (using a hardcoded 400px width on the `.gallery-example-live`):
image

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
glenhunter: > Lost a semi-colon here

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/c2b3c1e85b2ed618fd9f73970b8c32a7f37aa499

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-variables';
+@import '../../scss-shared/nx-text-helpers';
+
+.nx-policy-violation-indicator {
+  --policy-indicator-highlight-colour: #{$nx-threat-color-unspecified};
```
rpokorny, line 11: We use the american spelling of "color"

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { ThreatLevelCategory, categoryByPolicyThreatLevel } from '../../util/threatLevels';
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+
+import { Props, propTypes } from './types';
+export { Props } from './types';
+
+import './NxPolicyViolationIndicator.scss';
+
+const NxPolicyViolationIndicator = forwardRef<HTMLDivElement, Props>(
+  function NxPolicyViolationIndicator({ threatLevelCategory, policyThreatLevel, children }, ref ) {
```
rpokorny, line 19:
```suggestion
  function NxPolicyViolationIndicator({ threatLevelCategory, policyThreatLevel, children }, ref) {
```

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <NxFontAwesomeIcon aria-hidden={false}
+                             aria-label={label}
+                             icon={faExclamationCircle}/>
+        </div>
+        <div className="nx-policy-violation-indicator__text">
+          {children || category}
+        </div>
+      </div>
+    );
+  });
```
rpokorny, line 40: For symmetry
```suggestion
  }
);
```

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
rpokorny, line 19: We need to extract the rest of the props (the native div attributes) and apply them to the top div

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
```diff
@@ -0,0 +1,44 @@
...
+import './NxPolicyViolationIndicator.scss';
+
+const NxPolicyViolationIndicator = forwardRef<HTMLDivElement, Props>(
+  function NxPolicyViolationIndicator({ threatLevelCategory, policyThreatLevel, children }, ref ) {
+    const category: ThreatLevelCategory =
+          threatLevelCategory ? threatLevelCategory :
+          policyThreatLevel != null ? categoryByPolicyThreatLevel[policyThreatLevel] : 
+          'unspecified',
+
+        className = `nx-policy-violation-indicator nx-policy-violation-indicator--${category}`,
```
rpokorny, line 25: Should allow custom classnames as well

lib/src/components/NxPolicyViolationIndicator/types.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as PropTypes from 'prop-types';
+import { HTMLAttributes, WeakValidationMap } from 'react';
+
+import { ThreatLevelNumber, ThreatLevelCategory, allThreatLevelCategories, allThreatLevelNumbers }
+  from '../../util/threatLevels';
+
+export interface Props extends HTMLAttributes<HTMLDivElement> {
+  threatLevelCategory?: ThreatLevelCategory;
+  policyThreatLevel?: ThreatLevelNumber;
+}
```
rpokorny, line 16: Given that we really only want strings for the children, we can signify that by adding `children?: string | null;` here. And then `children: PropTypes.string` in the propTypes below.

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > I'm thinking about this more and wondering if what we should really do is make the indicators inline-block elements, considering their typical placement, and also give them a max-width of 100%. What you have now doesn't really address the issue at hand, which is the possibility of the indicator being wider than its container. For example (using a hardcoded 400px width on the `.gallery-example-live`):

Just adding `max-width: 100%;` solves the issue. I'm loathe to refactor the whole thing to be an `inline-block` element the layout is much cleaner and easier with flex box.

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/3b9a69d92a6a8704a2ecdcfa56c3cc12ce1e40a0

Screen Shot 2020-12-01 at 11 04 13 AM

gallery/visualtests/nxPolicyViolationIndicator.js
glenhunter: > > Ah, removed: [f2ced09](https://github.com/sonatype/sonatype-react-shared-components/commit/f2ced0972a54e081b9882d395dffbd79bfa59f0c)
> 
> I believe you've got an unused variable now.

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/4417f2cb8012dcea9fea8fdbc3f37675b97209dd

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > We use the american spelling of "color"

Damn these Canadian fingers of mine: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/9253b5a25ae5d96ad83f1f22af1d820ddfbacab9

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
rpokorny: > I'm loathe to refactor the whole thing to be an inline-block element the layout is much cleaner and easier with flex box.

I guess I meant `inline-flex`.  Externally an inline element was my point, I wasn't suggesting changing the way the internals lay out.  They're meant to commonly be used in headers where they are supposed to lay out adjacent to the main header text.  Though I'll admit, the vertical alignment that's intended there isn't clear to me from the zeplin mockups.

lib/src/components/NxPolicyViolationIndicator/types.ts
glenhunter: > Given that we really only want strings for the children, we can signify that by adding `children?: string | null;` here. And then `children: PropTypes.string` in the propTypes below.

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/bec43678f47e6988702a7b94e7b826ff7643b26d

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
glenhunter: > Should allow custom classnames as well

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/cc14aa0f2408a3148e653be8c45c748163c82150

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.tsx
glenhunter: > We need to extract the rest of the props (the native div attributes) and apply them to the top div

https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/cc14aa0f2408a3148e653be8c45c748163c82150

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
glenhunter: > > I'm loathe to refactor the whole thing to be an inline-block element the layout is much cleaner and easier with flex box.
> 
> I guess I meant `inline-flex`. Externally an inline element was my point, I wasn't suggesting changing the way the internals lay out. They're meant to commonly be used in headers where they are supposed to lay out adjacent to the main header text. Though I'll admit, the vertical alignment that's intended there isn't clear to me from the zeplin mockups.

I wish that I had remembered `inline-flex` existed: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/7aca8373cc7d26a3c5ce19e22ee959749e1da968

rpokorny:
I think we need examples of these indicators placed into headers as shown in the mockups.  I'm not confident that they will lay out as prescribed, particularly in regards to the 8px from the bottom to the subheader.

rpokorny:
Missing 8px above subheader
image

glenhunter:
> I think we need examples of these indicators placed into headers as shown in the mockups. I'm not confident that they will lay out as prescribed, particularly in regards to the 8px from the bottom to the subheader.

Tile: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/e9e254e46494d2ba373fb7b1faeef0b729fc785a

Page Title:
https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/9bbc58be4cd5123a5b7e064f074ef042ffccb36a

glenhunter:
> Missing 8px above subheader
> image

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/229/commits/9331949b1bfd7fa6263a0ad42b706c4f7f6206f8

and I added examples both with and without sub-titles

rpokorny:
> Fixed here: 9331949
>
> and I added examples both with and without sub-titles

That fix changes the spacing between the title and subtitle even when the indicator isn't present.

glenhunter:
> > Fixed here: [9331949](https://github.com/sonatype/sonatype-react-shared-components/commit/9331949b1bfd7fa6263a0ad42b706c4f7f6206f8)
> > and I added examples both with and without sub-titles
> 
> That fix changes the spacing between the title and subtitle even when the indicator isn't present.

Yep, though I'm stumped on how to fix that, margin or padding on the indicator container only serve to push it up so its alignment with the title text goes off. Any top spacing applied to the sub-title itself would be global and descendent selectors won't work either. To me it's an issue that @dturner32 may have to reconcile to himself: either having 8px of spacing there all of the time or none of the time. 

8px of spacing
Screen Shot 2020-12-07 at 9 36 33 AM

0px px of spacing
Screen Shot 2020-12-07 at 9 36 49 AM

Note that if the sub-title text didn't wrap we'd have exactly the same issue with the buttons to the right.

dturner32:
> > > Fixed here: [9331949](https://github.com/sonatype/sonatype-react-shared-components/commit/9331949b1bfd7fa6263a0ad42b706c4f7f6206f8)
> > > and I added examples both with and without sub-titles
> > 
> > 
> > That fix changes the spacing between the title and subtitle even when the indicator isn't present.
> 
> Yep, though I'm stumped on how to fix that, margin or padding on the indicator container only serve to push it up so its alignment with the title text goes off. Any top spacing applied to the sub-title itself would be global and descendent selectors won't work either. To me it's an issue that @dturner32 may have to reconcile to himself: either having 8px of spacing there all of the time or none of the time.
> 
> 8px of spacing
> Screen Shot 2020-12-07 at 9 36 33 AM
> 
> 0px px of spacing
> Screen Shot 2020-12-07 at 9 36 49 AM
> 
> Note that if the sub-title text didn't wrap we'd have exactly the same issue with any buttons to the right.

I prefer having the 8px of spacing there consistently

rpokorny:
Zeplin shows an example with an nx-page-title that includes an NxPolicyViolationIndicator, followed by an nx-tile 24px below the page header (32px below the indicator).  Your page-title example, on the other hand, includes a page-title description element after the header.  This raises two questions?

1. What are the specs for a page title with an indicator and a description.  This scenario is not shown in zeplin.
2. For the example that is in zeplin, why is there so much spacing between the header and the tile? Our existing page headers only have 16px of space there.  Is that wrong, or is this mockup wrong, or is the presence of the indicator supposed to increase that space?

glenhunter:
> Zeplin shows an example with an nx-page-title that includes an NxPolicyViolationIndicator, followed by an nx-tile 24px below the page header (32px below the indicator). Your page-title example, on the other hand, includes a page-title description element after the header. This raises two questions.
> 
> 1. What are the specs for a page title with an indicator and a description.  This scenario is not shown in zeplin.
> 2. For the example that is in zeplin, why is there so much spacing between the header and the tile? Our existing page headers only have 16px of space there.  Is that wrong, or is this mockup wrong, or is the presence of the indicator supposed to increase that space?

1. My expectation is that the presence of the indicator should not change the existing spacing, however the lack of reference in explain is problematic.

2. The page layout example in zeplin shows 24px between the title and the page content, in this case I think RSC is incorrect.

@dturner32 ^^

dturner32:
> > Zeplin shows an example with an nx-page-title that includes an NxPolicyViolationIndicator, followed by an nx-tile 24px below the page header (32px below the indicator). Your page-title example, on the other hand, includes a page-title description element after the header. This raises two questions.
> > 
> > 1. What are the specs for a page title with an indicator and a description.  This scenario is not shown in zeplin.
> > 2. For the example that is in zeplin, why is there so much spacing between the header and the tile? Our existing page headers only have 16px of space there.  Is that wrong, or is this mockup wrong, or is the presence of the indicator supposed to increase that space?
> 
> 1. My expectation is that the presence of the indicator should not change the existing spacing, however the lack of reference in Zeplin is a problem.
> 2. The page layout example in Zeplin shows 24px between the title and the page content, in this case I think RSC is incorrect.
> 
> @dturner32 ^^

1. that expectation is correct. The existing spacing shouldn't be effected. I added an example in the zeplin mock up for reference

2. There should be 24px between the page header and the tile

glenhunter:
> > > Zeplin shows an example with an nx-page-title that includes an NxPolicyViolationIndicator, followed by an nx-tile 24px below the page header (32px below the indicator). Your page-title example, on the other hand, includes a page-title description element after the header. This raises two questions.
> > > 
> > > 1. What are the specs for a page title with an indicator and a description.  This scenario is not shown in zeplin.
> > > 2. For the example that is in zeplin, why is there so much spacing between the header and the tile? Our existing page headers only have 16px of space there.  Is that wrong, or is this mockup wrong, or is the presence of the indicator supposed to increase that space?
> > 
> > 
> > 
> > 1. My expectation is that the presence of the indicator should not change the existing spacing, however the lack of reference in Zeplin is a problem.
> > 2. The page layout example in Zeplin shows 24px between the title and the page content, in this case I think RSC is incorrect.
> > 
> > @dturner32 ^^
> 
> 1. that expectation is correct. The existing spacing shouldn't be effected. I added an example in the zeplin mock up for reference
> 2. There should be 24px between the page header and the tile

Ok, I have created https://issues.sonatype.org/browse/RSC-407 to deal with changing the existing `16px` of spacing to `24px`

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -60,11 +65,29 @@ const NxPageTitlePage = () =>
...
+                        codeExamples={nxPageTitleCode}>
+      A simple example of an <code className="nx-code">nx-page-title</code>.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="NX Page Title with Actions Example"
+                        id="nx-page-title-actions-example"
+                        defaultCheckeredBackground={true}
+                        htmlExample={nxPageTitleActionsCode}
+                        codeExamples={nxPageTitleActionsCode}>
+      An example of <code className="nx-code">nx-page-title</code> with actions in the form of a drop-down.
```
rpokorny, line 81: Why is this change showing up in this PR? This is separate and already merged to master, right?

glenhunter: > Why is this change showing up in this PR? This is separate and already merged to master, right?

When I added the new example I updated the two other examples to use the new gallery example tile

lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.nx-policy-violation-indicator__text {
+  @include semi-bold();
+  @include nx-truncate-ellipsis;
+
+  font-size: $nx-font-size-s;
+  margin: 0 $nx-spacing-md;
+  text-transform: uppercase;
+  // max width of element from design is 859px, subtract the badge, the padding, and the borders.
+  max-width: 793px;
```
rpokorny, line 42: Is this still needed?

glenhunter: > Is this still needed?

Yes because the max-width for the component as specified by Zeplin is less than the maximum width of some of the element's potential containers.

rpokorny: Maybe I wasn't clear, but part of what I was trying to say in the other thread on this line (https://github.com/sonatype/sonatype-react-shared-components/pull/229/files#r533511093) was that I don't think the max as specified in zeplin is valuable to have.   The important thing is that the indicator doesn't overflow it's container, which we now handle via `max-width: 100%`.  Beyond that, I'm not sure it's helpful to limit the displayable size of its contents.

glenhunter: > Maybe I wasn't clear, but part of what I was trying to say in the other thread on this line (https://github.com/sonatype/sonatype-react-shared-components/pull/229/files#r533511093) was that I don't think the max as specified in zeplin is valuable to have. The important thing is that the indicator doesn't overflow it's container, which we now handle via `max-width: 100%`. Beyond that, I'm not sure it's helpful to limit the displayable size of its contents.

IMO this is a question for @dturner32 and he needs to be tagged or he won't see it.

dturner32: > > Maybe I wasn't clear, but part of what I was trying to say in the other thread on this line (https://github.com/sonatype/sonatype-react-shared-components/pull/229/files#r533511093) was that I don't think the max as specified in zeplin is valuable to have. The important thing is that the indicator doesn't overflow it's container, which we now handle via `max-width: 100%`. Beyond that, I'm not sure it's helpful to limit the displayable size of its contents.
> 
> IMO this is a question for @dturner32 and he needs to be tagged or he won't see it.

I'd argue for keeping the max width setting to keep the content in the indicator from stretching all the way across the page.

rpokorny (approved):

PR #230, opened 2020-12-01
Grey colour palette RSC-261

glenhunter:
https://issues.sonatype.org/browse/RSC-400

Create variables from the supplied swatches. Apply those variables in the SCSS, and when commonalities are found create broader variables for those.

### DIFF
--- a/lib/src/scss-shared/_nx-color-swatches.scss
+++ b/lib/src/scss-shared/_nx-color-swatches.scss
@@ -17,3 +17,17 @@ $nx-blue-300: #c5deff;
 $nx-blue-200: #d3e6ff;
 $nx-blue-100: #e8f2ff;
 $nx-blue-50: #f1f7ff;
+
+// Greys
+$nx-grey-1100: #2c2c2c;
+$nx-grey-1000: #3b3b3b;
+$nx-grey-900: #474747;
+$nx-grey-800: #555;
+$nx-grey-700: #686868;
+$nx-grey-600: #828282;
+$nx-grey-500: #949494;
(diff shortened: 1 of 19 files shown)


lib/src/scss-shared/_nx-color-swatches.scss
```diff
@@ -17,3 +17,17 @@ $nx-blue-300: #c5deff;
...
+$nx-grey-900: #474747;
+$nx-grey-800: #555;
+$nx-grey-700: #686868;
+$nx-grey-600: #828282;
+$nx-grey-500: #949494;
+$nx-grey-400: #a7a7a7;
+$nx-grey-300: #c7c7c7;
+$nx-grey-200: #e9e9e9;
+$nx-grey-100: #f1f3fa;
+$nx-grey-50: #f4f5f9;
```
rpokorny, line 33: Are we going to get a variable for white at some point?  Would that count as a grey?

rpokorny (approved):

lib/src/scss-shared/_nx-color-swatches.scss
glenhunter: > Are we going to get a variable for white at some point? Would that count as a grey?

My assumption was that black and white would be separate from the greys, but that's up to @dturner32

dturner32: > > Are we going to get a variable for white at some point? Would that count as a grey?
> 
> My assumption was that black and white would be separate from the greys, but that's up to @dturner32

that is a correct assumption

PR #231, opened 2020-12-01
Add buttons to page title RSC-400

glenhunter:
https://issues.sonatype.org/browse/RSC-400

The purpose of this ticket was to add support for displaying buttons inside the page title. I used our standard `nx-btn-bar` container and a custom class on the page title, this allowed me to use CSS grid for positioning.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.13.4",
+  "version": "2.14.0",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.13.3",
+  "version": "2.13.4",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
+++ b/gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
@@ -14,8 +14,8 @@
     Page Title
   </h1>
   <div class="nx-btn-bar">
-    <button class="nx-btn">First</button>
-    <button class="nx-btn">Second</button>
+    <button class="nx-btn nx-btn--tertiary">First</button>
+    <button class="nx-btn nx-btn--tertiary">Second</button>
   </div>
   <div class="nx-page-title__description">
     <p class="nx-p">
--- a/gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
+++ b/gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
@@ -1,0 +1,28 @@
+<!--
+
+    Copyright (c) 2019-present Sonatype, Inc.
+    This program and the accompanying materials are made available under
+    the terms of the Eclipse Public License 2.0 which accompanies this
+    distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+
+-->
+<div class="nx-page-title nx-page-title__actions">
+  <h1 class="nx-h1">
+    <svg class="nx-icon nx-page-title__page-icon" style="height: 1em; width: 1em;" viewBox="-1 -1 2 2">
+      <circle r="1" />
+    </svg>
+    Page Title
+  </h1>
+  <div class="nx-btn-bar">
+    <button class="nx-btn">First</button>
+    <button class="nx-btn">Second</button>
+  </div>
+  <div class="nx-page-title__description">
+    <p class="nx-p">
+      Tokyo alcohol marketing girl vehicle jeans sign papier-mache assassin San Francisco rifle physical 3D-printed
+      denim tanto courier concrete dolphin rebar free-market. tank-traps papier-mache dead free-market tanto drone
+      concrete dolphin sunglasses weathered dead jeans office vehicle nodal point. motion film meta- monofilament knife
+      vinyl post- bridge jeans city
+    </p>
+  </div>
+</div>
--- a/gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
+++ b/gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
@@ -11,7 +11,7 @@
     <svg class="nx-icon nx-page-title__page-icon" style="height: 1em; width: 1em;" viewBox="-1 -1 2 2">
       <circle r="1" />
     </svg>
-    Page Title
+    Long page title text that will cause ellipsis truncation to occur
   </h1>
   <div class="nx-btn-bar">
     <button class="nx-btn nx-btn--tertiary">First</button>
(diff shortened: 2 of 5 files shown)


rpokorny:
I would have expected this to be done in a way that's more consistent with how tile headers work - that it's always a grid and the action button container may or may not be present.

glenhunter:
> I would have expected this to be done in a way that's more consistent with how tile headers work - that it's always a grid and the action button container may or may not be present.

I forgot that's how we did it there and it's a much better solution: https://github.com/sonatype/sonatype-react-shared-components/pull/231/commits/c309a1d0b2af5c726d3368cdcc69e5e01162cc26

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.13.3",
+  "version": "2.13.4",
```
rpokorny, line 3: Probably should be a minor version since it's a new capability.

gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
```diff
@@ -0,0 +1,28 @@
...
+-->
+<div class="nx-page-title nx-page-title__actions">
+  <h1 class="nx-h1">
+    <svg class="nx-icon nx-page-title__page-icon" style="height: 1em; width: 1em;" viewBox="-1 -1 2 2">
+      <circle r="1" />
+    </svg>
+    Page Title
+  </h1>
+  <div class="nx-btn-bar">
+    <button class="nx-btn">First</button>
```
rpokorny, line 17: I think we need design input on what types of buttons should be used here.  I think in most mockups I've seen it's the tertiary buttons.  The blue secondary ones can look a little out of place. @dturner32

rpokorny:
Page title truncates prematurely
image

rpokorny:
The buttons seems a bit high to me.  @dturner32 can you comment on how they should be aligned, vertically?
image

glenhunter:
> Page title truncates prematurely

I think that's correct (according to the code), both the h1 and the sub-title have a max-width of 562px. I'm not sure why they have that set, but I didn't add them @dturner32 ^^

gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
glenhunter: > I think we need design input on what types of buttons should be used here. I think in most mockups I've seen it's the tertiary buttons. The blue secondary ones can look a little out of place. @dturner32

I changed them to tertiary, but IMO that's going to be up to the designers: https://github.com/sonatype/sonatype-react-shared-components/pull/231/commits/875b8850526b34ec9f9d9e4dd017d39532dbf51c

glenhunter:
> > Page title truncates prematurely
> 
> I think that's correct (according to the code), both the h1 and the sub-title have a max-width of 562px. I'm not sure why they have that set, but I didn't add them @dturner32 ^^

I did tweak the grid a bit to allow the title more space: https://github.com/sonatype/sonatype-react-shared-components/pull/231/commits/875b8850526b34ec9f9d9e4dd017d39532dbf51c

gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
dturner32: > I think we need design input on what types of buttons should be used here. I think in most mockups I've seen it's the tertiary buttons. The blue secondary ones can look a little out of place. @dturner32

I'd advise using tertiary buttons in this case

dturner32:
> The buttons seems a bit high to me. @dturner32 can you comment on how they should be aligned, vertically?
> image

The buttons do look a little high, they should be centered with the page header. They look a little off center now

dturner32:
> > Page title truncates prematurely
> 
> I think that's correct (according to the code), both the h1 and the sub-title have a max-width of 562px. I'm not sure why they have that set, but I didn't add them @dturner32 ^^

I don't recall where that 562px max width came from, I don't even remember setting that honestly. In general the page header should extend until it hits the 24px padding at the edge of the page at which point it truncates. If there are buttons in line with the page header then the header truncates 24px from the edge of a button

rpokorny:
> I don't recall where that 562px max width came from, I don't even remember setting that honestly. 

I'm fairly sure it predates the design studio

> In general the page header should extend until it hits the 24px padding at the edge of the page at which point it truncates. If there are buttons in line with the page header then the header truncates 24px from the edge of a button

This makes sense to me and is consistent with how our other headers work.

glenhunter:
> > > Page title truncates prematurely
> > 
> > 
> > I think that's correct (according to the code), both the h1 and the sub-title have a max-width of 562px. I'm not sure why they have that set, but I didn't add them @dturner32 ^^
> 
> I don't recall where that 562px max width came from, I don't even remember setting that honestly. In general the page header should extend until it hits the 24px padding at the edge of the page at which point it truncates. If there are buttons in line with the page header then the header truncates 24px from the edge of a button

Implemented here: https://github.com/sonatype/sonatype-react-shared-components/pull/231/commits/c7fe597b152c81f38ca2cbcd91b69b73472adb5b

Note that I left the max-width restriction on the sub-title text, if that should also be removed please let me know @dturner32

glenhunter:
> > The buttons seems a bit high to me. @dturner32 can you comment on how they should be aligned, vertically?
> 
> The buttons do look a little high, they should be centered with the page header. They look a little off center now

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/231/commits/c7fe597b152c81f38ca2cbcd91b69b73472adb5b

gallery/package.json
glenhunter: > Probably should be a minor version since it's a new capability.

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/231/commits/67bc53e2af8fa4d64ea0ef5a5c5c4b9c98ea5c77

gallery/src/styles/NxPageTitle/NxPageTitleActionsExample.html
```diff
@@ -0,0 +1,28 @@
...
+    the terms of the Eclipse Public License 2.0 which accompanies this
+    distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+
+-->
+<div class="nx-page-title nx-page-title__actions">
+  <h1 class="nx-h1">
+    <svg class="nx-icon nx-page-title__page-icon" style="height: 1em; width: 1em;" viewBox="-1 -1 2 2">
+      <circle r="1" />
+    </svg>
+    Page Title
```
rpokorny, line 14: I'd recommend demonstrating the overflow in this example.

rpokorny (approved):
+1 aside from my last comment and the open question about the description max-width.

dturner32:
> > > > Page title truncates prematurely
> > > 
> > > 
> > > I think that's correct (according to the code), both the h1 and the sub-title have a max-width of 562px. I'm not sure why they have that set, but I didn't add them @dturner32 ^^
> > 
> > 
> > I don't recall where that 562px max width came from, I don't even remember setting that honestly. In general the page header should extend until it hits the 24px padding at the edge of the page at which point it truncates. If there are buttons in line with the page header then the header truncates 24px from the edge of a button
> 
> Implemented here: [c7fe597](https://github.com/sonatype/sonatype-react-shared-components/commit/c7fe597b152c81f38ca2cbcd91b69b73472adb5b)
> 
> Note that I left the max-width restriction on the sub-title text, if that should also be removed please let me know @dturner32

I would remove it there too

PR #234, opened 2020-12-03
NxOverflowTooltip - RSC-161

rpokorny:
https://issues.sonatype.org/browse/RSC-161

While the ticket is about dropdown menu items specifically, I felt the best approach was to add a general component that handles overflow tooltips, and then to demonstrate it on a dropdown.


glenhunter (approved):
+1 for the HTML/SCSS I'd be more comfortable with somebody else looking at the React code

Blacktiger (approved):

PR #235, opened 2020-12-03
Update tooltip styles - RSC-148

rpokorny:
https://issues.sonatype.org/browse/RSC-148

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.13.3",
+  "version": "2.13.4",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "2.17.0",
+  "version": "2.17.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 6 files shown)


glenhunter:
The Zeplin file is showing the tooltips as semi-bold

glenhunter:
We should probably have an example that shows long content wrapping as in the Zeplin file

rpokorny:
> The Zeplin file is showing the tooltips as semi-bold

It was also in the wrong font face. 2f87ea5d

rpokorny:
> We should probably have an example that shows long content wrapping as in the Zeplin file

0a8577aa

glenhunter:
Safari problems:
Screen Shot 2020-12-07 at 1 23 10 PM
Screen Shot 2020-12-07 at 1 22 59 PM
Screen Shot 2020-12-07 at 1 22 55 PM
Seems like it has something to do with the radio button. Might be better in general to have these laid out in "correct" form layout styles to make sure there are no issues there especially since that's how we expect the form elements to be used?

rpokorny:
> Safari problems:

Are you referring to the fact that the tooltip alignment doesn't work the same for the filter input vs the normal input?  That's the case in Chrome as well, and I doubt it's specifically to do with the non-standardness of that page.  I'll look into it.

glenhunter:
> > Safari problems:
> 
> Are you referring to the fact that the tooltip alignment doesn't work the same for the filter input vs the normal input? That's the case in Chrome as well, and I doubt it's specifically to do with the non-standardness of that page. I'll look into it.

See the way in the screenshots how the tooltip is slightly indented and partially covers the text input

rpokorny:
> See the way in the screenshots how the tooltip is slightly indented and partially covers the text input

Pre-existing issue.  I've made a but ticket for it. https://issues.sonatype.org/browse/RSC-408

glenhunter (approved):

PR #237, opened 2020-12-04
Description Lists - RSC-275

rpokorny:
https://issues.sonatype.org/browse/RSC-275

Note that while we've traditionally called these definition lists, the current WHATWG spec refers to them as description lists, which is probably a more appropriately broad name.


glenhunter:
It's possible to have multiple descriptions `<dd>` following a single `<dt>`. Should we style that or are we explicitly not supporting that? @dturner32

rpokorny:
> It's possible to have multiple descriptions <dd> following a single <dt>. Should we style that or are we explicitly not supporting that? @dturner32

I believe the docs I wrote explicitly say we don't currently support that.  I assumed that was our stance for these styles, given that we didn't have designs for it

glenhunter:
> > It's possible to have multiple descriptions  following a single . Should we style that or are we explicitly not supporting that? @dturner32
> 
> I believe the docs I wrote explicitly say we don't currently support that. I assumed that was our stance for these styles, given that we didn't have designs for it

I saw that I just wanted to confirm with @dturner32 since I wasn't sure that he knew that description lists can support multiple `<dd>`'s.

glenhunter (approved):

dturner32:
> > It's possible to have multiple descriptions  following a single . Should we style that or are we explicitly not supporting that? @dturner32
> 
> I believe the docs I wrote explicitly say we don't currently support that. I assumed that was our stance for these styles, given that we didn't have designs for it

is there any other reason we don't support multiple descriptions other than there just not being any designs for it? I'd be in favor of supporting having multiple descriptions

glenhunter:
> > > It's possible to have multiple descriptions  following a single . Should we style that or are we explicitly not supporting that? @dturner32
> > 
> > 
> > I believe the docs I wrote explicitly say we don't currently support that. I assumed that was our stance for these styles, given that we didn't have designs for it
> 
> is there any other reason we don't support multiple descriptions other than there just not being any designs for it? I'd be in favor of supporting having multiple descriptions

In this layout multiple descriptions are difficult to lay out. The only reason is when you want to semantically have two descriptions.

I'm actually sorry that I brought this up, I think that we should leave this as it is unless we discover a solid reason to have multiple descriptions.

dturner32:
> > > > It's possible to have multiple descriptions  following a single . Should we style that or are we explicitly not supporting that? @dturner32
> > > 
> > > 
> > > I believe the docs I wrote explicitly say we don't currently support that. I assumed that was our stance for these styles, given that we didn't have designs for it
> > 
> > 
> > is there any other reason we don't support multiple descriptions other than there just not being any designs for it? I'd be in favor of supporting having multiple descriptions
> 
> In this layout multiple descriptions are difficult to lay out. The only reason is when you want to semantically have two descriptions.
> 
> I'm actually sorry that I brought this up, I think that we should leave this as it is unless we discover a solid reason to have multiple descriptions.

if there is no solid reason to have multiple descriptions then we can leave it as is for now

rpokorny:
> if there is no solid reason to have multiple descriptions then we can leave it as is for now

Some examples of places where multiple descriptions would be semantically meaningful:

* A dictionary where a single word has multiple separate meanings
* A list of authors, each author would ideally be a separate `<dd>`

Here's what the HTML spec says: 
> The values within a group are alternatives; multiple paragraphs forming part of the same value must all be given within the same dd element.

Note that HTML also allows multiple terms to be defined by the same description (or group of descriptions).  I previously misunderstood the semantics here.  For instance consider the following structure:

```
<dl>
  <dt>Authors</dt>
  <dt>Editors</dt>
  <dd>John Doe</dd>
  <dd>Will Smith</dd>
</dl>
```

I used to think this meant that there are no authors and that John Doe and Will Smith are both editors.  After re-reading the spec however, I found the the meaning intended by this structure is actually that John Doe and Will Smith and both authors and editors.

Again, I am fine with us saying that we don't support more than 1:1 mappings for now.  I'm just pointing out the kinds of cases where they can come up.

PR #240, opened 2020-12-08
nx-viewport-sized - RSC-366

rpokorny:
https://issues.sonatype.org/browse/RSC-366


rpokorny:
Class names adjusted at @glenhunter 's suggestion.

gallery/src/styles/NxViewportSized/NxViewportSizedExample.tsx
```diff
@@ -0,0 +1,624 @@
...
+          </p>
+          <p className="nx-p" style={{ width: '400px' }}>
+            This paragraph has an explicit, non-100% width. It is still left-aligned within the tile just like it
+            would be in block layout.
+          </p>
+          <p className="nx-p">
+            Single line paragraph.
+          </p>
+          <NxInfoAlert>
+            This alert is stretched as expected.
```
glenhunter, line 34: I don't understand this sentence, the alert is not stretched in any way that I can see. It appears correctly.

glenhunter:
One thing that users might find confusing about these examples is that in the first case (if I understand correctly) they will probably need to use the `--expanding` modifier as well since they won't always know if the contents of the scrollable container will be capable of expanding the page to full height. I'm concerned that the way the examples are written it seem a little bit either/or.

rpokorny:
> One thing that users might find confusing about these examples is that in the first case (if I understand correctly) they will probably need to use the --expanding modifier as well since they won't always know if the contents of the scrollable container will be capable of expanding the page to full height. I'm concerned that the way the examples are written it seem a little bit either/or.

I did try to spell out in the docs, in multiple places, that the expanding behavior is in addition to the standard behavior of these classes and doesn't undo any of it.  I do however think it's an open question whether the expanding behavior should be the default, with non-expanding being a separate option, or if we should maybe even have the expanding behavior be the only option.

>  they will probably need to use the --expanding modifier as well since they won't always know if the contents of the scrollable container will be capable of expanding the page to full height.

While I think I got the gist of what your comment was about, I'm afraid I don't really understand what you're saying in this part.

glenhunter:
> > One thing that users might find confusing about these examples is that in the first case (if I understand correctly) they will probably need to use the --expanding modifier as well since they won't always know if the contents of the scrollable container will be capable of expanding the page to full height. I'm concerned that the way the examples are written it seem a little bit either/or.
> 
> I did try to spell out in the docs, in multiple places, that the expanding behavior is in addition to the standard behavior of these classes and doesn't undo any of it. I do however think it's an open question whether the expanding behavior should be the default, with non-expanding being a separate option, or if we should maybe even have the expanding behavior be the only option.
> 
> > they will probably need to use the --expanding modifier as well since they won't always know if the contents of the scrollable container will be capable of expanding the page to full height.
> 
> While I think I got the gist of what your comment was about, I'm afraid I don't really understand what you're saying in this part.

In the first example if the table was only a few rows and they didn't use `--expanding` the table doesn't expand. Since we often don't know the height of tables it seems to me that we'll almost always want to use `--expanding`. 
Screen Shot 2020-12-14 at 9 27 18 AM

glenhunter:
I keep expecting to find this example page in the "Layout" section of the Nav. I understand why it isn't there but I think that we should provide a mention and a link in the Page Layout example.

rpokorny:
> I keep expecting to find this example page in the "Layout" section of the Nav. I understand why it isn't there but I think that we should provide a mention and a link in the Page Layout example.

89065d11

rpokorny:
> I do however think it's an open question whether the expanding behavior should be the default, with non-expanding being a separate option, or if we should maybe even have the expanding behavior be the only option.

After more discussion we've decided to drop the non-expanding behavior and have it expand by default. 
https://github.com/sonatype/sonatype-react-shared-components/pull/240/files/89065d115224bcd0d92ced89af79c2a08765777a..b6f643a702b6ec7a529e6b528a8c87ce57f72d2e

glenhunter (approved):

PR #241, opened 2020-12-08
implement teal colour palette RSC-362

glenhunter:
https://issues.sonatype.org/browse/RSC-362

The main colour used in these swatches was the teal for the focus borders. I created a variable for that, otherwise there were very few of these colours used in the code.

### DIFF
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -39,7 +39,7 @@
     .nx-radio-checkbox__control {
       border-color: $nx-focus-border-color;
       stroke: $nx-focus-border-color;
-      filter: drop-shadow(0 0 3px rgba(2, 166, 217, 0.4));
+      filter: drop-shadow(0 0 3px rgba($nx-focus-border-color, 0.4));
     }
 
     &:hover {
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -37,8 +37,8 @@
 
   &:focus-within {
     .nx-radio-checkbox__control {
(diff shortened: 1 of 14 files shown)


lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -37,8 +37,8 @@
 
   &:focus-within {
     .nx-radio-checkbox__control {
-      border-color: #02a6d9;
-      stroke: #02a6d9;
+      border-color: $nx-focus-border-color;
+      stroke: $nx-focus-border-color;
       filter: drop-shadow(0 0 3px rgba(2, 166, 217, 0.4));
```
rpokorny, line 42: It's not written in hex format but I believe this is the same color.

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -17,7 +17,7 @@ $nx-sonatype-orange: #f89100;
 // Tints based on HSL lightness value
 $nx-color-blue-41: #3A4D95;
 $nx-color-blue-42: #2B66AB;
-$nx-color-blue-44: #00A1E2;
+$nx-color-blue-44:$nx-teal-400;
```
rpokorny, line 20:
```suggestion
$nx-color-blue-44: $nx-teal-400;
```

rpokorny (approved):
+1 aside from the missed rgba value

lib/src/base-styles/_nx-radio-checkbox.scss
glenhunter: > It's not written in hex format but I believe this is the same color.

https://github.com/sonatype/sonatype-react-shared-components/pull/241/commits/3e0e378d3293dcee864b59be5b66b959b2725845

PR #244, opened 2020-12-10
NxFormGroup React component - RSC-219

rpokorny:
https://issues.sonatype.org/browse/RSC-219


gallery/src/components/NxFormGroup/NxFormGroupPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <NxTableCell>JSX content to render as the sublabel for the group.</NxTableCell>
+          </NxTableRow>
+          <NxTableRow>
+            <NxTableCell>isRequired</NxTableCell>
+            <NxTableCell>boolean</NxTableCell>
+            <NxTableCell>No</NxTableCell>
+            <NxTableCell>false</NxTableCell>
+            <NxTableCell>
+              Sets whether the input should display the optional flag – the flag is present by default and
+              setting <code className="nx-code">isRequired</code> to true removes the flag.
```
glenhunter, line 66: I wonder if we should default to "required". I know that most of the fields in IQ are required, and Michael Martz thinks RM is probably ~50/50

rpokorny: > I wonder if we should default to "required". I know that most of the fields in IQ are required, and Michael Martz thinks RM is probably ~50/50

I asked myself that, but I decided to go this way because it would be consistent with the following two things:

1. `NxTextInput` defaults to not required in the sense that if you want it to validate that it isn't empty, you have to go out of your way to supply a validator function
2. HTML `<input>`s have a `required` attribute which you have to add if you want them to be considered required by HTML form validation

gallery/src/components/NxFormGroup/NxFormGroupPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+      an <code className="nx-code">NxTextInput</code> which uses the isRequired flag to remove the "Optional"
+      indicator.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Sublabel Example"
+                        liveExample={NxFormGroupSublabelExample}
+                        codeExamples={nxFormGroupSublabelExampleCode}>
+      An example of an <code className="nx-code">NxFormGroup</code> which includes a sublabel. This example also
+      shows that <code className="nx-code">NxFormGroup</code> can wrap a
+      <code className="nx-code">&lt;select&gt;</code> in addition to
```
glenhunter, line 109:
```suggestion
      {' '}<code className="nx-code">&lt;select&gt;</code> in addition to
```

glenhunter (approved):
+1 except for idUtil which I don't understand

rpokorny:
Requesting review from other JS devs for the idUtil changes

raveneyex (approved):
Very clever!

Blacktiger (approved):

PR #247, opened 2020-12-14
Improve nx-btn accessibility RSC-398

glenhunter:
https://issues.sonatype.org/browse/RSC-398

Added tooltips to icon only buttons, added examples of "normal" buttons with only icons as content. Updated/improved descriptions and examples.

### DIFF
--- a/gallery/src/components/NxButton/NxButtonIconExample.tsx
+++ b/gallery/src/components/NxButton/NxButtonIconExample.tsx
@@ -5,7 +5,7 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React from 'react';
-import { faSync, faSave, faEdit } from '@fortawesome/free-solid-svg-icons';
+import { faSync, faSave, faEdit, faSmile } from '@fortawesome/free-solid-svg-icons';
 
 import { NxButton, NxFontAwesomeIcon, NxTooltip } from '@sonatype/react-shared-components';
 
@@ -17,11 +17,15 @@ const NxButtonIconExample = () =>
     </NxButton>
 
     <NxTooltip title="Save">
-      <NxButton><NxFontAwesomeIcon icon={faSave} aria-label="save" /></NxButton>
+      <NxButton><NxFontAwesomeIcon icon={faSave} /></NxButton>
     </NxTooltip>
 
     <NxTooltip title="Edit">
-      <NxButton variant="tertiary"><NxFontAwesomeIcon icon={faEdit} aria-label="edit" /></NxButton>
+      <NxButton variant="tertiary"><NxFontAwesomeIcon icon={faEdit} /></NxButton>
+    </NxTooltip>
+
+    <NxTooltip title="Smile">
+      <NxButton variant="primary"><NxFontAwesomeIcon icon={faSmile} /></NxButton>
     </NxTooltip>
   </div>;
 
--- a/gallery/src/components/NxButton/NxButtonIconExample.tsx
+++ b/gallery/src/components/NxButton/NxButtonIconExample.tsx
@@ -5,16 +5,24 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React from 'react';
-import {faSync} from '@fortawesome/free-solid-svg-icons';
+import { faSync, faSave, faEdit } from '@fortawesome/free-solid-svg-icons';
 
-import { NxButton, NxFontAwesomeIcon } from '@sonatype/react-shared-components';
+import { NxButton, NxFontAwesomeIcon, NxTooltip } from '@sonatype/react-shared-components';
 
 const NxButtonIconExample = () =>
   <div className="nx-btn-bar">
-    <NxButton>
+    <NxButton aria-label="sync">
       <NxFontAwesomeIcon icon={faSync}/>
       <span>Icons in buttons</span>
     </NxButton>
+
+    <NxTooltip title="Save">
+      <NxButton><NxFontAwesomeIcon icon={faSave} aria-label="save" /></NxButton>
+    </NxTooltip>
+
+    <NxTooltip title="Edit">
+      <NxButton variant="tertiary"><NxFontAwesomeIcon icon={faEdit} aria-label="edit" /></NxButton>
+    </NxTooltip>
   </div>;
 
 export default NxButtonIconExample;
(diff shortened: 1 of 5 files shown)


gallery/src/components/NxButton/NxButtonIconExample.tsx
```diff
@@ -5,16 +5,24 @@
...
-import {faSync} from '@fortawesome/free-solid-svg-icons';
+import { faSync, faSave, faEdit } from '@fortawesome/free-solid-svg-icons';
 
-import { NxButton, NxFontAwesomeIcon } from '@sonatype/react-shared-components';
+import { NxButton, NxFontAwesomeIcon, NxTooltip } from '@sonatype/react-shared-components';
 
 const NxButtonIconExample = () =>
   <div className="nx-btn-bar">
-    <NxButton>
+    <NxButton aria-label="sync">
```
rpokorny, line 14: Typically in a button containing both an icon and text, the icon is semantically redundant and should not have a label.

gallery/src/components/NxButton/NxButtonIconExample.tsx
```diff
@@ -5,16 +5,24 @@
...
       <NxFontAwesomeIcon icon={faSync}/>
       <span>Icons in buttons</span>
     </NxButton>
+
+    <NxTooltip title="Save">
```
rpokorny, line 19: I thought the idea with this ticket was to create an `NxIconOnlyButton` component that encapsulated the need for a tooltip.

rpokorny:
Are there zeplin mockups that you can link to for the now-multiple kinds of icon-only buttons?

gallery/src/components/NxButton/NxButtonPage.tsx
```diff
@@ -68,15 +68,18 @@ export default function NxButtonPage() {
                           id="nx-button-icon-example"
                           liveExample={NxButtonIconExample}
                           codeExamples={nxButtonIconCode}>
-        An example of a button containing an icon in addition to text.
+        An example of buttons containing icons. When an icon's only content consists of an icon it should be wrapped
+        in an <code className="nx-code">NxTooltip</code> and its <code className="nx-code">aria-label</code> set.
```
rpokorny, line 72: This isn't true.  If the button has a tooltip, you shouldn't need to set the aria-label on anything.

gallery/src/components/NxButton/NxButtonPage.tsx
```diff
@@ -68,15 +68,18 @@ export default function NxButtonPage() {
...
       <GalleryExampleTile title="Icon only buttons"
                           id="nx-button-icon-only-example"
                           liveExample={NxButtonIconOnlyExample}
                           codeExamples={nxButtonIconOnlyCode}>
-        An example of a button containing only an icon. For accessibility purposes it is important to add an
-        aria-label for the screen reader to interpret the content correctly.
+        The <code className="nx-code">icon-only</code> variant is meant to be used in places like lists or tables where
+        a slightly smaller button size is desirable. For accessibility purposes it is important to wrap it in
+        an <code className="nx-code">NxTooltip</code> and an <code className="nx-code">aria-label</code> for the screen
+        reader to interpret the content correctly.
```
rpokorny, line 82: I think it's confusing to have two different kinds of "icon only buttons" only one of which is the `icon-only` variant.  This should be renamed in a way that reflects it's intended usage.

glenhunter: > I think it's confusing to have two different kinds of "icon only buttons" only one of which is the `icon-only` variant. This should be renamed in a way that reflects it's intended usage.

The difference is that one is just a "regular" button whose content is nothing more than an icon, and the `icon-only` variant which has design differences and intended usage that are different. I agree that the naming sucks.

glenhunter:
> Are there zeplin mockups that you can link to for the now-multiple kinds of icon-only buttons?

No. They aren't necessary (at this time @dturner32 might decide to tweak them at some point), the idea was just to provide examples of the regular button variants with content that is just an icon.

rpokorny:
> No. They aren't necessary (at this time @dturner32 might decide to tweak them at some point), the idea was just to provide examples of the regular button variants with content that is just an icon.

I mean, I'm finding them necessary as otherwise I don't know whether or not the left and right padding around the icon is correct.  I suspect it isn't but I can't say for sure without a spec.  Additionally I was hoping the spec would help clarify when exactly one kind of icon-only button is supposed to be used vs the other.

glenhunter:
> > No. They aren't necessary (at this time @dturner32 might decide to tweak them at some point), the idea was just to provide examples of the regular button variants with content that is just an icon.
> 
> I mean, I'm finding them necessary as otherwise I don't know whether or not the left and right padding around the icon is correct. I suspect it isn't but I can't say for sure without a spec. Additionally I was hoping the spec would help clarify when exactly one kind of icon-only button is supposed to be used vs the other.

IIRC the icon-only variant was only intended to be used in lists and tables, that's why it's smaller. However IQ and RM both have some places where there is an icon-only button next to a form field and the smaller `icon-only` variants look bad there because the height of the button and the form fields don't match.

glenhunter:
> > No. They aren't necessary (at this time @dturner32 might decide to tweak them at some point), the idea was just to provide examples of the regular button variants with content that is just an icon.
> 
> I mean, I'm finding them necessary as otherwise I don't know whether or not the left and right padding around the icon is correct. I suspect it isn't but I can't say for sure without a spec. 

My expectation was that they would just follow the spacing of the other buttons.

glenhunter:
Closing until @dturner32 returns and we have better designs and clearer direction.

gallery/src/components/NxButton/NxButtonIconExample.tsx
glenhunter: > Typically in a button containing both an icon and text, the icon is semantically redundant and should not have a label.

https://github.com/sonatype/sonatype-react-shared-components/pull/247/commits/dedb255a7ff9c5c641fd7bfc853d02d008d45534

gallery/src/components/NxButton/NxButtonPage.tsx
glenhunter: > This isn't true. If the button has a tooltip, you shouldn't need to set the aria-label on anything.

https://github.com/sonatype/sonatype-react-shared-components/pull/247/commits/e562b598f7b399b6a6f123cac9a02451fbd78726

PR #249, opened 2020-12-15
nx-btn-bar alignment - RSC-415

rpokorny:
https://issues.sonatype.org/browse/RSC-415

Causing: https://issues.sonatype.org/browse/CLM-17541

### DIFF
--- a/lib/src/base-styles/_nx-btn-bar.scss
+++ b/lib/src/base-styles/_nx-btn-bar.scss
@@ -8,6 +8,8 @@
 .nx-btn-bar {
   @include container-horizontal;
 
+  align-items: first baseline;
+  display: flex;
   margin: $nx-spacing-sm 0;
-  text-align: right;
+  justify-content: flex-end;
 }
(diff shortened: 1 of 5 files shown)


glenhunter:
Pagination buttons are no longer centred. 
Screen Shot 2020-12-15 at 10 41 37 AM

lib/src/base-styles/_nx-btn-bar.scss
```diff
@@ -8,6 +8,8 @@
 .nx-btn-bar {
   @include container-horizontal;
 
+  align-items: first baseline;
```
glenhunter, line 11: I'd prefer `center` since it's clearer in its effect/intent

rpokorny:
> Pagination buttons are no longer centred.

065aac57

lib/src/base-styles/_nx-btn-bar.scss
rpokorny: > I'd prefer center since it's clearer in its effect/intent

8d20874

glenhunter (approved):

PR #254, opened 2020-12-16
NxPolicyThreatSlider style update - RSC-338

rpokorny:
https://issues.sonatype.org/browse/RSC-338

NOTE: The zeplin mockups here seem to be missing some changes that I recall being agreed to:

* In the mockups the track goes from the furthest left that the left edge of the thumb can be, to the furthest right that the right edge of the thumb can be.  I thought we agreed to keep the track as essentially covering the extent that the center of the thumb can travel.
* The "Select Range" text included in the mockup is not part of the component, it is just part of the gallery example.

Additionally, the mockup does not include focus styles, and I can't remember what was said about them.  I've updated them to match our typical focus styling.

I also went against the mockup regarding the overall width of the component.  In order to keep the track size divisible by 10 so that it renders well (without gaps) using our 5% and 10% widths, I increased it by 4px from the spec'd 228 to 232.

### DIFF
--- a/lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
+++ b/lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
@@ -51,34 +51,39 @@
     &.MuiSlider-active, &:hover {
       box-shadow: none;
     }
+
+    &:focus {
+      border-color: $nx-focus-border-color;
+      box-shadow: $nx-focus-box-shadow;
+    }
   }
 
   .nx-policy-threat-slider__value-label--none {
-    .MuiSlider-thumb {
+    .MuiSlider-thumb:not(:focus) {
       border-color: $nx-threat-color-none;
     }
   }
 
   .nx-policy-threat-slider__value-label--low {
-    .MuiSlider-thumb {
+    .MuiSlider-thumb:not(:focus) {
       border-color: $nx-threat-color-low;
     }
   }
 
   .nx-policy-threat-slider__value-label--moderate {
-    .MuiSlider-thumb {
+    .MuiSlider-thumb:not(:focus) {
       border-color: $nx-threat-color-moderate;
     }
   }
 
   .nx-policy-threat-slider__value-label--severe {
-    .MuiSlider-thumb {
+    .MuiSlider-thumb:not(:focus) {
       border-color: $nx-threat-color-severe;
     }
   }
 
   .nx-policy-threat-slider__value-label--critical {
-    .MuiSlider-thumb {
+    .MuiSlider-thumb:not(:focus) {
       border-color: $nx-threat-color-critical;
     }
   }
--- a/lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
+++ b/lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
@@ -15,7 +15,6 @@
 
   // flex so line-height doesn't get involved in the element overall height
   display: inline-flex;
-  font-size: 13px;
   margin: $nx-spacing-xs 0;
 
   // the .MuiSlider-thumb elements actually stick out of the sides of the .MuiSlider-root at their min and max
--- a/lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
+++ b/lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
@@ -7,19 +7,22 @@
 @import '../../scss-shared/nx-variables';
 
 .nx-policy-threat-slider {
-  display: inline-block;
+  $rail-height: 8px;
+  $thumb-size: 32px;
+
+  // distance from top of thumb to top of rail
+  $thumb-vertical-overflow: ($thumb-size - $rail-height) / 2;
 
   font-size: 13px;
 
   // the .MuiSlider-thumb elements actually stick out of the sides of the .MuiSlider-root at their min and max
   // positions.  This padding ensures that they don't stick out of the overall box for this component
-  padding: 0 9px;
+  padding: 0 ($thumb-size / 2);
 
   .MuiSlider-root {
-    display: inline-block;
-    height: 6px;
-    padding: 9px 0;
-    width: 180px;
+    height: $rail-height;
+    padding: $thumb-vertical-overflow 0;
+    width: 200px;
   }
 
   .MuiSlider-markActive {
@@ -31,15 +34,18 @@
   }
 
   .MuiSlider-thumb {
-    align-items: stretch;
-    border-radius: 4px;
-
-    color: $nx-off-white;
-
-    height: 17px;
-    width: 18px;
-
-    margin-left: -9px;
+    box-sizing: border-box;
+    background-color: $nx-white;
+    border: 2px solid;
+    border-radius: $nx-border-radius;
+    color: $nx-text-color-dark;
+    font-size: $nx-font-size-s;
+    font-weight: bold;
+    height: $thumb-size;
+    width: $thumb-size;
+    line-height: 1;
+    margin-left: $thumb-size / -2;
+    margin-top: -($thumb-vertical-overflow);
 
     &.MuiSlider-active, &:hover {
       box-shadow: none;
@@ -48,40 +54,42 @@
 
   .nx-policy-threat-slider__value-label--none {
     .MuiSlider-thumb {
-      background-color: $nx-threat-color-none;
+      border-color: $nx-threat-color-none;
     }
   }
 
   .nx-policy-threat-slider__value-label--low {
     .MuiSlider-thumb {
-      background-color: $nx-threat-color-low;
+      border-color: $nx-threat-color-low;
     }
   }
 
   .nx-policy-threat-slider__value-label--moderate {
     .MuiSlider-thumb {
-      background-color: $nx-threat-color-moderate;
+      border-color: $nx-threat-color-moderate;
     }
   }
 
   .nx-policy-threat-slider__value-label--severe {
     .MuiSlider-thumb {
-      background-color: $nx-threat-color-severe;
+      border-color: $nx-threat-color-severe;
     }
   }
 
   .nx-policy-threat-slider__value-label--critical {
     .MuiSlider-thumb {
-      background-color: $nx-threat-color-critical;
+      border-color: $nx-threat-color-critical;
     }
   }
 
   .MuiSlider-mark {
-    height: 6px;
+    $border-radius: $rail-height / 2;
+
+    height: $rail-height;
 
     // threat level 0
-    border-radius: 4px 0 0 4px;
-    color: $nx-threat-color-none-light;
+    border-radius: $border-radius 0 0 $border-radius;
+    color: $nx-threat-color-none;
 
     // the first mark only goes from 0 - 0.5, so is only 5% wide.
     // The rest (except the last) each go from (n - 0.5) to (n + 0.5), so are 10% wide with a -5% left margin.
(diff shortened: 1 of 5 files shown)


glenhunter:
There's a bit of inconsistent weirdness when the indicators overlap:

Screen Shot 2020-12-17 at 1 18 46 PM
Screen Shot 2020-12-17 at 1 18 51 PM
Screen Shot 2020-12-17 at 1 18 58 PM

rpokorny:
> There's a bit of inconsistent weirdness when the indicators overlap:

Considering that I thought I got rid of that translucent blue-grey thing, and can't figure out how to make it appear anymore, could you give some steps to reproduce what's in those screenshots?

glenhunter:
> > There's a bit of inconsistent weirdness when the indicators overlap:
> 
> Considering that I thought I got rid of that translucent blue-grey thing, and can't figure out how to make it appear anymore, could you give some steps to reproduce what's in those screenshots?

I just tabbed onto the indicators and then used the arrow keys to move them left and right.

lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
```diff
@@ -7,19 +7,24 @@
...
+  padding: 0 ($thumb-size / 2);
 
   .MuiSlider-root {
-    display: inline-block;
-    height: 6px;
-    padding: 9px 0;
-    width: 180px;
+    height: $rail-height;
+    padding: $thumb-vertical-overflow 0;
+    width: 200px;
```
glenhunter, line 27: I'm curious about where this value comes from

glenhunter:
The widths of the colours is different in the mockup and implementation, one of them needs to be adjusted (I suspect it's the mockup)

rpokorny:
> The widths of the colours is different in the mockup and implementation, one of them needs to be adjusted (I suspect it's the mockup)

See the notes in the PR description

lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
rpokorny: > I'm curious about where this value comes from

See the notes in the PR description.  If this isn't divisible by 10 it renders poorly on my monitor, with visible gaps between some segments.

glenhunter: > > I'm curious about where this value comes from
> 
> See the notes in the PR description. If this isn't divisible by 10 it renders poorly on my monitor, with visible gaps between some segments.

I understand but the values you reference above are `I increased it by 4px from the spec'd 228 to 232` and I don't see any width values at all specified in Zeplin

glenhunter:
> > The widths of the colours is different in the mockup and implementation, one of them needs to be adjusted (I suspect it's the mockup)
> 
> See the notes in the PR description

I did read the notes and I don't see anything that explains why the 0 value is 1/4, and the 1 value is 1/2 the width of the Zeplin mockup

glenhunter:
To expand on the width question I think that if the designed width is purely a result of the width of IQ's Manage Filters sidebar that isn't how we should design a component. There is no guarantee that the only place that it will be used in every future application will only be that width. At the very least we might want to specify a few potential widths.

rpokorny:
> I just tabbed onto the indicators and then used the arrow keys to move them left and right.

MUI was doing something weird with an extra class that has something to do with focus.  Fixed in ae8a627a

rpokorny:
> I did read the notes and I don't see anything that explains why the 0 value is 1/4, and the 1 value is 1/2 the width of the Zeplin mockup

1. The mockup doesn't have the correct number of track segments necessary to make sense.  The "Low" segment is 32px wide, and only threat level 1 is considered low.  So we get a ratio of 32px per threat level.  The way it's designed, with the ends going all the way to the edges of the sliders, it needs 11 equal-width segments (0 - 10 inclusive) which would be 352px.  Yet the track is only 228px wide (as far as I can tell, it's hard to measure without a mockup where both ends are visible)!  Another way you can see this is if you visualize in your head sliding the "4" slider in the bottom-left example up one level at a time, you'll see that when you get to the position where the mockup shows "10", you'll really only be at 8.
2. The center of the slider sits on the position it is representing, so I'd prefer to keep the ends of the track tucked in as I have them, so that the track shows the extent from 0 to 10, not from -0.5 to 10.5.  I'd be willing to change that if others feel strongly about it though.  This halves the width of the first and last segments.

rpokorny:
> To expand on the width question I think that if the designed width is purely a result of the width of IQ's Manage Filters sidebar that isn't how we should design a component. There is no guarantee that the only place that it will be used in every future application will only be that width. At the very least we might want to specify a few potential widths.

Perhaps, but I think we do want to make sure that it fits there, or more generally, that it fits as a tree view child in a sidebar.

lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
rpokorny: > I understand but the values you reference above are I increased it by 4px from the spec'd 228 to 232 and I don't see any width values at all specified in Zeplin

Well not written out, no.  I mean from measuring things in zeplin.

glenhunter:
Here's something interesting, when using the arrow keys to move the indicators around the focused indicator always falls behind the unfocused indicator when they cross.

rpokorny:
> Here's something interesting, when using the arrow keys to move the indicators around the focused indicator always falls behind the unfocused indicator when they cross.

I noticed that too.  I think that's within the details of the third-party component and I doubt we could adjust it easily.

glenhunter:
> > Here's something interesting, when using the arrow keys to move the indicators around the focused indicator always falls behind the unfocused indicator when they cross.
> 
> I noticed that too. I think that's within the details of the third-party component and I doubt we could adjust it easily.

That's my guess as well, interesting that the behaviour is exactly the same regardless of which indictor is moving, I'd assumed that one would be consistently "top".

lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
rpokorny: > Well not written out, no. I mean from measuring things in zeplin.

And I just realized I misinterpreted what zeplin was saying.  I think the actual spec'd width is something like 288px or 291px.  But as I mention elsewhere, it's actually missing two segments worth of track, and adding those in would push the width up to 352px (11 * 32) which is too wide for the place that we know this needs to be usable.  Any thoughts on how we should address that?

rpokorny: > But as I mention elsewhere, it's actually missing two segments worth of track, and adding those in would push the width up to 352px (11 * 32) which is too wide for the place that we know this needs to be usable

In fact, even the spec'd width is too much.  So this design really is problematic.

image

glenhunter: > > Well not written out, no. I mean from measuring things in zeplin.
> 
> And I just realized I misinterpreted what zeplin was saying. I think the actual spec'd width is something like 288px or 291px. But as I mention elsewhere, it's actually missing two segments worth of track, and adding those in would push the width up to 352px (11 * 32) which is too wide for the place that we know this needs to be usable. Any thoughts on how we should address that?

My thinking was that because we know that the bar length must be divisible by 10 that we either come up with 2 or 3 default sizes and hard code them, or that we allow the user to specify a value, that must be divisible by 10.

rpokorny: > My thinking was that because we know that the bar length must be divisible by 10 that we either come up with 2 or 3 default sizes and hard code them, or that we allow the user to specify a value, that must be divisible by 10.

After out-of-band discussion with @glenhunter we agreed to stick to a single size but to increase it a bit while still taking care to make it fit in the existing usage. 54c99c30

dturner32:
> https://issues.sonatype.org/browse/RSC-338
> 
> NOTE: The zeplin mockups here seem to be missing some changes that I recall being agreed to:
> 
> * In the mockups the track goes from the furthest left that the left edge of the thumb can be, to the furthest right that the right edge of the thumb can be.  I thought we agreed to keep the track as essentially covering the extent that the center of the thumb can travel.
> * The "Select Range" text included in the mockup is not part of the component, it is just part of the gallery example.
> 
> Additionally, the mockup does not include focus styles, and I can't remember what was said about them. I've updated them to match our typical focus styling.
> 
> I also went against the mockup regarding the overall width of the component. In order to keep the track size divisible by 10 so that it renders well (without gaps) using our 5% and 10% widths, I increased it by 4px from the spec'd 228 to 232.

to your first point @rpokorny, I think that is where we landed I just didn't update it in the mockup but I can make that correction.

I remember that we agreed to removing the "select range" text as part of the component and that it would just be part of the gallery example. Updating that as well 

As for focus styles, they aren't reflected in the mockup but should match the standard focus styles. Will add those to the mockup for reference. 

In regards to the width discussion, the set width of the entire component is '290px' but after reading through the thread here it seems that that width is still too long to fit in the sidebar. I'm inclined to agree with whatever most feasible solution to implement is in this case and update the mockup accordingly

lib/src/components/NxPolicyThreatSlider/NxPolicyThreatSlider.scss
dturner32: the width of the component is '290 px' but it sounds like that width won't work in this case with the included segments. I'm open to whatever solution was reached here

dturner32:
> The widths of the colours is different in the mockup and implementation, one of them needs to be adjusted (I suspect it's the mockup)

this will more than likely need to be adjusted in the mockup

dturner32:
> > To expand on the width question I think that if the designed width is purely a result of the width of IQ's Manage Filters sidebar that isn't how we should design a component. There is no guarantee that the only place that it will be used in every future application will only be that width. At the very least we might want to specify a few potential widths.
> 
> Perhaps, but I think we do want to make sure that it fits there, or more generally, that it fits as a tree view child in a sidebar.

I agree with both points here, there could be future use of this component in areas other than the sidebar so it may warrant consideration of playing around with a more variable width. To Ross' point, I think we do need to make sure it fits where it is for now

dturner32:
> > I did read the notes and I don't see anything that explains why the 0 value is 1/4, and the 1 value is 1/2 the width of the Zeplin mockup
> 
> 1. The mockup doesn't have the correct number of track segments necessary to make sense.  The "Low" segment is 32px wide, and only threat level 1 is considered low.  So we get a ratio of 32px per threat level.  The way it's designed, with the ends going all the way to the edges of the sliders, it needs 11 equal-width segments (0 - 10 inclusive) which would be 352px.  Yet the track is only 228px wide (as far as I can tell, it's hard to measure without a mockup where both ends are visible)!  Another way you can see this is if you visualize in your head sliding the "4" slider in the bottom-left example up one level at a time, you'll see that when you get to the position where the mockup shows "10", you'll really only be at 8.
> 2. The center of the slider sits on the position it is representing, so I'd prefer to keep the ends of the track tucked in as I have them, so that the track shows the extent from 0 to 10, not from -0.5 to 10.5.  I'd be willing to change that if others feel strongly about it though.  This halves the width of the first and last segments.

I'd like to discuss a little more in WW tomorrow, with some visual aid, just to get a better idea of the consideration here before agreeing to or updating anything

glenhunter:
What if it's disabled? @dturner32

glenhunter (approved):

rpokorny:
> What if it's disabled? @dturner32

In WW we determined that a separate ticket will be created for disabled `NxPolicyThreatSlider`s

PR #260, opened 2020-12-23
NxFieldset - RSC-416

rpokorny:
https://issues.sonatype.org/browse/RSC-416

### DIFF
--- a/gallery/src/components/NxFieldset/NxFieldsetPage.tsx
+++ b/gallery/src/components/NxFieldset/NxFieldsetPage.tsx
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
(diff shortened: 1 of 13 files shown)


gallery/src/components/NxFieldset/NxFieldsetPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+    nxFieldsetSublabelExampleCode = require('!raw-loader!!./NxFieldsetSublabelExample').default,
+    nxFieldsetRichLabelExampleCode = require('!raw-loader!!./NxFieldsetRichLabelExample').default,
+    nxFieldsetRequiredExampleCode = require('!raw-loader!!./NxFieldsetRequiredExample').default;
+
+const NxFieldsetPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        An <code className="nx-code">NxFieldset</code> is a wrapper around a collection of checkboxes, radios, or
+        similar form elements which should be displayed to be user as a group with an overall label. When used
```
glenhunter, line 26: Is there any other form element a fields could apply to?

glenhunter, line 26:
```suggestion
        similar form elements which should be displayed to the user as a group with an overall label. When used
```

glenhunter (approved):

gallery/src/components/NxFieldset/NxFieldsetPage.tsx
rpokorny: > Is there any other group of form elements a fieldset could be used with?

Toggles, and maybe other things in the future.  I didn't want to exclude prematurely.

PR #263, opened 2021-01-04
NxFilterInput refactor - RSC-165

rpokorny:
https://issues.sonatype.org/browse/RSC-165

The API changes here would be breaking changes, so this would be part of the 3.0 bump that it's parent branch creates.

### DIFF
--- a/gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
+++ b/gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
@@ -75,7 +75,7 @@ const NxFilterInputPage = () =>
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell">Input HTML Attributes | Textarea HTML Attributes</td>
+            <td className="nx-cell">Input HTML Attributes</td>
             <td className="nx-cell">
               <a target="_blank"
                  rel="noopener"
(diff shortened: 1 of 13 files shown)


gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
```diff
@@ -40,33 +44,64 @@ const NxFilterInputPage = () =>
...
+                   href="https://www.w3.org/TR/uievents-key/#named-key-attribute-values">
+                  named keys
+                </a>
+              </p>
+            </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell">onChange</td>
-            <td className="nx-cell">Function</td>
+            <td className="nx-cell">Input HTML Attributes | Textarea HTML Attributes</td>
```
glenhunter, line 78: Unless there's a place where an `NxFilterInput` could be a `textarea` I'd recommend removing references to `Textarea`'s here and below, to me it's unnecessary and a little confusing.

glenhunter (approved):

gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
rpokorny: > Unless there's a place where an NxFilterInput could be a textarea I'd recommend removing references to Textarea's here and below, to me it's unnecessary and a little confusing.

2839e94e

PR #265, opened 2021-01-05
Set explicit box-sizing where necessary - RSC-84

rpokorny:
https://issues.sonatype.org/browse/RSC-84

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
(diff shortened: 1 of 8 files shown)


glenhunter:
While this fixes existing components that break when `box-sizing: border-box` is set globally it doesn't do anything to prevent future components from having the same issue. Do we want to set it globally for the gallery? Or create a toggle of some sort for testing?

rpokorny:
> Do we want to set it globally for the gallery? Or create a toggle of some sort for testing?

I mentioned this in WW.  Setting it globally in the gallery doesn't help as it makes it just as likely (maybe even more) that we'll accidentally create components that are broken when it isn't set.  Making it toggleable in some way would be a first step but without also creating visual tests of everything with it toggled both ways I'm not sure that's all that valuable.  And I'm not sure doubling our visual tests is worth it.

glenhunter (approved):

PR #266, opened 2021-01-05
Safari legend margin bug RSC-406

glenhunter:
https://issues.sonatype.org/browse/RSC-406

The issue is a long-standing bug in Safari where it collapses all margin on `<legend>`. There are a variety of fixes offered and this one seemed to work the best with our layouts. I also made our use of `sub-label` consistent across the layout examples.

### DIFF
--- a/gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
+++ b/gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
@@ -60,11 +60,11 @@ export default function NxFormLayoutExample() {
       <div className="nx-form-group">
         <label htmlFor="long-field" className="nx-label nx-label--optional">
           <span className="nx-label__text">Hostname</span>
-          <div className="nx-sub-label">
-            <NxFontAwesomeIcon icon={faCalendar}/>
-            <span id="long-field-sublabel">The field element below is wider than the default.</span>
-          </div>
         </label>
+        <div className="nx-sub-label">
+          <NxFontAwesomeIcon icon={faCalendar}/>
+          <span id="long-field-sublabel">The field element below is wider than the default.</span>
+        </div>
         <NxStatefulTextInput id="long-field" aria-describedby="long-field-sublabel" className="nx-text-input--long"/>
       </div>
       <fieldset className="nx-fieldset">
@@ -78,8 +78,8 @@ export default function NxFormLayoutExample() {
       <fieldset className="nx-fieldset">
         <legend className="nx-legend nx-legend--optional">
           <span className="nx-legend__text">Primary Color</span>
-          <div className="nx-sub-label">Pick a single color</div>
         </legend>
+        <div className="nx-sub-label">Pick a single color</div>
         <NxRadio name="color"
                  value="red"
                  onChange={setColor}
@@ -117,10 +117,10 @@ export default function NxFormLayoutExample() {
           <span className="nx-legend__text">
             Enable features
           </span>
-          <div className="nx-sub-label">
-            In a form layout toggles are laid out in a <code className="nx-code">&lt;fieldset&gt;</code>
-          </div>
         </legend>
+        <div className="nx-sub-label">
+          In a form layout toggles are laid out in a <code className="nx-code">&lt;fieldset&gt;</code>
+        </div>
         <NxToggle inputId="subscribe-check" onChange={toggleWarp} isChecked={isWarpOn}>
           Enable Warp Drive
         </NxToggle>
@@ -134,8 +134,8 @@ export default function NxFormLayoutExample() {
       <div className="nx-form-group">
         <label className="nx-label">
           <span className="nx-label__text">Comments</span>
+          <NxStatefulTextInput type="textarea" placeholder="placeholder"/>
         </label>
-        <NxStatefulTextInput type="textarea" placeholder="placeholder"/>
       </div>
       <footer className="nx-footer">
         <div className="nx-btn-bar">
--- a/gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
+++ b/gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
@@ -48,8 +48,8 @@ export default function NxFormLayoutExample() {
       <div className="nx-form-group">
         <label className="nx-label">
           <span className="nx-label__text">A Field to Fill in</span>
-          <NxStatefulTextInput aria-required={true} validator={validator}/>
         </label>
+        <NxStatefulTextInput aria-required={true} validator={validator}/>
       </div>
       <div className="nx-form-group">
         <label htmlFor="input-2" className="nx-label nx-label--optional">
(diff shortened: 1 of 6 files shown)


gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -78,8 +78,8 @@ export default function NxFormLayoutExample() {
       <fieldset className="nx-fieldset">
         <legend className="nx-legend nx-legend--optional">
           <span className="nx-legend__text">Primary Color</span>
-          <span className="nx-sub-label">Pick a single color</span>
         </legend>
+        <div className="nx-sub-label">Pick a single color</div>
```
rpokorny, line 82: I was under the impression that this was inside the `legend` on purpose.  It makes sense that way to me.  The recently-merged `NxFieldset` also is implemented with the `.nx-sub-label` within the `.nx-legend`.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -47,6 +47,11 @@
   }
 }
 
+// hack to get Safari to recognize the legend's text margins
+.nx-legend {
+  -webkit-margin-collapse: separate;
```
rpokorny, line 52: Are there any side-effects to this or does it just seem to do what we want here?

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > I was under the impression that this was inside the `legend` on purpose. It makes sense that way to me. The recently-merged `NxFieldset` also is implemented with the `.nx-sub-label` within the `.nx-legend`.

Mostly I think that `.sub-label` should be consistently either inside or outside of labels and legends, I'm not sure that it matters which and I can switch all of the examples to inside.

lib/src/base-styles/_nx-forms.scss
glenhunter: > Are there any side-effects to this or does it just seem to do what we want here?

I checked in other browsers and couldn't find any side effects.

glenhunter: > Are there any side-effects to this or does it just seem to do what we want here?

I couldn't find any side effects

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
rpokorny: > Mostly I think that .sub-label should be consistently either inside or outside of labels and legends, I'm not sure that it matters which and I can switch all of the examples to inside.

Originally, the label text, sublabel, and input were all inside of the label.  Then for screenreader support we changed it so the input and sublabel were outside of the label.  Maybe we didn't have to move the sublabel at that time, I'm not sure.  But as it stands that history means we are already supporting two different ways of doing them, I'd rather not add a third by moving the sublabel back inside of labels.  If you want them consistent I'd rather move legend sublabels to the outside.  But if that's what you do it needs to be changed in `NxFieldset` as well.

glenhunter: > I was under the impression that this was inside the `legend` on purpose. It makes sense that way to me. The recently-merged `NxFieldset` also is implemented with the `.nx-sub-label` within the `.nx-legend`.

Moved back inside here: https://github.com/sonatype/sonatype-react-shared-components/pull/266/commits/3d9261cac84abbe4f371464bff300e6b3c071c9d

glenhunter: > > Mostly I think that .sub-label should be consistently either inside or outside of labels and legends, I'm not sure that it matters which and I can switch all of the examples to inside.
> 
> Originally, the label text, sublabel, and input were all inside of the label. Then for screenreader support we changed it so the input and sublabel were outside of the label. Maybe we didn't have to move the sublabel at that time, I'm not sure. But as it stands that history means we are already supporting two different ways of doing them, I'd rather not add a third by moving the sublabel back inside of labels. If you want them consistent I'd rather move legend sublabels to the outside. But if that's what you do it needs to be changed in `NxFieldset` as well.

I moved them back outside here: https://github.com/sonatype/sonatype-react-shared-components/pull/266/commits/61f6808a091263c2deebf6897a7d716fc1a95e5d and modified `NxFieldset` as well, which was good because now it is consistent with `NxFormGroup`: https://github.com/sonatype/sonatype-react-shared-components/pull/266/commits/7235813987a69b379aaa876904baea86fbb192ff

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -48,8 +48,8 @@ export default function NxFormLayoutExample() {
       <div className="nx-form-group">
         <label className="nx-label">
           <span className="nx-label__text">A Field to Fill in</span>
-          <NxStatefulTextInput aria-required={true} validator={validator}/>
         </label>
+        <NxStatefulTextInput aria-required={true} validator={validator}/>
```
rpokorny, line 52: IIRC we were deliberately keeping some inconsistency in the examples given that we consider both ways to be currently supported.

lib/src/components/NxFieldset/NxFieldset.tsx
```diff
@@ -19,8 +19,8 @@ const NxFieldset = forwardRef<HTMLFieldSetElement, Props>(
         <fieldset className={classNames} ref={ref} { ...attrs }>
           <legend className={legendClassnames}>
             <span className="nx-legend__text">{label}</span>
-            { sublabel && <span className="nx-sub-label">{sublabel}</span> }
           </legend>
+          { sublabel && <span className="nx-sub-label">{sublabel}</span> }
```
rpokorny, line 23: You changed the one in the form layout example to a `<div>`, but this is still a `<span>`

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
glenhunter: > IIRC we were deliberately keeping some inconsistency in the examples given that we consider both ways to be currently supported.

I think that we should settle on a "correct" way, and demonstrate only that. I'm not suggesting that we pull support for the other ways just that we maintain consistency in our examples. I created a ticket to go through and make the other example pages consistent.

lib/src/components/NxFieldset/NxFieldset.tsx
glenhunter: > You changed the one in the form layout example to a `<div>`, but this is still a `<span>`

Drat I missed that: https://github.com/sonatype/sonatype-react-shared-components/pull/266/commits/243d13fe6b6d4c3040f01092a2658888ae84204a

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
rpokorny: > I think that we should settle on a "correct" way, and demonstrate only that. I'm not suggesting that we pull support for the other ways just that we maintain consistency in our examples. I created a ticket to go through and make the other example pages consistent.

The problem is that we can't know for sure that we haven't broken a way that we don't have a demonstration of.

glenhunter: > > I think that we should settle on a "correct" way, and demonstrate only that. I'm not suggesting that we pull support for the other ways just that we maintain consistency in our examples. I created a ticket to go through and make the other example pages consistent.
> 
> The problem is that we can't know for sure that we haven't broken a way that we don't have a demonstration of.

Well we could create a some kind of semi-hidden legacy page, but I don't think that we want to be stuck in a position where we endlessly support every historical variation of HTML layout. We should demonstrate what we consider to be the "correct" way of laying out the code. In this case I'm also resolving an inconsistency between the `fieldset` and the `form-group`.

Also I'm not removing any capabilities, I'm only making the examples consistent.

glenhunter: > > I think that we should settle on a "correct" way, and demonstrate only that. I'm not suggesting that we pull support for the other ways just that we maintain consistency in our examples. I created a ticket to go through and make the other example pages consistent.
> 
> The problem is that we can't know for sure that we haven't broken a way that we don't have a demonstration of.

As discussed in WW I created this ticket to create deprecated example(s): https://issues.sonatype.org/browse/RSC-436

rpokorny (approved):

PR #267, opened 2021-01-05
nx-alert at page level - RSC-434

rpokorny:
https://issues.sonatype.org/browse/RSC-434

While talking to @regoarrarr today we found that it's going to be common for applications to wrap both their `.nx-page-main` and `.nx-page-sidebar` in a single `NxLoadWrapper` - which would place the error alert as a direct child of the `.nx-page-content`.  This PR adds support for that placement.

### DIFF
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -41,6 +41,12 @@
   max-width: $nx-page-width-max;
   overflow-y: hidden;
   width: 100%;
+
+  // page-load errors are often at this level and not encapsulated within a nx-page-sidebar or nx-page-main
+  > .nx-alert {
+    align-self: flex-start;
+    margin: $nx-spacing-xl auto 0 auto;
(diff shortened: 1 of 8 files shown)


lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -79,13 +85,13 @@
...
-    position: fixed;
+    position: sticky;
     top: 0;
     left: 0;
     z-index: 1;
   }
 
   .nx-page-content {
-    margin: $nx-main-header-height auto 0 auto;
+    margin: 0 auto;
```
rpokorny, line 95: While working on this, I discovered that switching between page scrolling and section scrolling was moving the page contents up and down by 1px.  It turns out the margin here was not set to the correct amount as it failed to consider the bottom border on the header.  Setting the position on the `.nx-page-header` to sticky avoids the need for a specific number here.

regoarrarr (approved):

PR #268, opened 2021-01-06
Do not allow NxTreeView to collapse or expand when disabled - RSC-78

rpokorny:
https://issues.sonatype.org/browse/RSC-78

The only issue described in this ticket which still exists is that the various NxTreeView*Select components had inconsistent behavior with regards to whether they can collapse or expand when disabled.  To fix this I have added logic to the underlying `NxTreeView` component to disallow it from collapsing or expanding when disabled.  I am not however certain that this is the desired behavior, so I'd like input from Design on that. @dturner32

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.1.0",
+  "version": "3.1.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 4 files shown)


glenhunter:
The triangle still gets a `:focus` state: 
Screen Shot 2021-01-06 at 2 50 28 PM

glenhunter (approved):
Aside from the `:focus` issue it looks good to me

rpokorny:
> The triangle still gets a :focus state:

The issue isn't so much the styling as the fact that it shouldn't be focusable when disabled.  Fixed in 
2863e52

PR #269, opened 2021-01-06
Rerun NxModal onClose registration when onClose changes - RSC-133

rpokorny:
https://issues.sonatype.org/browse/RSC-133

### DIFF
--- a/lib/src/components/NxModal/NxModal.tsx
+++ b/lib/src/components/NxModal/NxModal.tsx
@@ -42,7 +42,7 @@ const NxModal: FunctionComponent<Props> = ({className, onClose, variant, ...attr
     currentModalCloseHandlers.push(onClose);
 
     return removeCloseHandlerListener;
-  }, []);
+  }, [onClose]);
 
   return (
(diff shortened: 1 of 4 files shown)


lib/src/components/NxModal/NxModal.tsx
```diff
@@ -42,7 +42,7 @@ const NxModal: FunctionComponent<Props> = ({className, onClose, variant, ...attr
     currentModalCloseHandlers.push(onClose);
 
     return removeCloseHandlerListener;
-  }, []);
+  }, [onClose]);
```
jlopez-sonatype, line 45: I remember getting a warning because I wasn't using all received props here, is not present anymore? (memory fails if it was in the IDE or in the console)

rpokorny: > I remember getting a warning because I wasn't using all received props here, is not present anymore? (memory fails if it was in the IDE or in the console)

I haven't gotten that warning nor do I see how it would make sense on this line.

jlopez-sonatype (approved):

PR #270, opened 2021-01-07
Expand tree view section for current page - RSC-160

rpokorny:
https://issues.sonatype.org/browse/RSC-160

This PR makes it so that when you first load the gallery, the correct sidebar section for whatever page you are on will be expanded.  If you are on a page that is not in the sidebar (for instance, the Home page), then the first section will be the expanded one as was previously the case.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.2.1",
+  "version": "3.2.2",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -69,7 +69,7 @@
     "ramda": "0.26.1",
     "react": "16.10.1",
     "react-dom": "16.10.1",
-    "react-router-dom": "=5.0.0",
+    "react-router-dom": "^5.1.0",
     "react-syntax-highlighter": "9.0.1"
   }
 }
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -25,7 +25,8 @@
     "@types/classnames": "=2.2.7",
     "@types/react": "16.9.14",
     "@types/react-dom": "16.9.1",
-    "@types/react-router-dom": "=4.3.2",
+    "@types/react-router": "^5.1.0",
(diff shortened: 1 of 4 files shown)


gallery/package.json
```diff
@@ -69,7 +70,8 @@
     "ramda": "0.26.1",
     "react": "16.10.1",
     "react-dom": "16.10.1",
-    "react-router-dom": "=5.0.0",
+    "react-router": "^5.1.0",
```
rpokorny, line 73: `react-router` should have been an explicit dependency before, but it wasn't and we were relying on it being a transitive dependency of `react-router-dom`.  Additionally, this PR upgrades the react-router packages because the `useLocation` hook that we use was introduced in 5.1.0

gallery/package.json
```diff
@@ -25,7 +25,8 @@
     "@types/classnames": "=2.2.7",
     "@types/react": "16.9.14",
     "@types/react-dom": "16.9.1",
-    "@types/react-router-dom": "=4.3.2",
+    "@types/react-router": "^5.1.0",
```
Blacktiger, line 28: Should we be using =5.1.0 instead? With this syntax each build on CI could have a different version of react-router whether we expect it to or not. That also means the CI build could pick up a new version of the dependency with bad code embedded before we've had a change to review things.

Blacktiger: NVM, this is only in the gallery so I'm not concerned.

rpokorny: > Should we be using =5.1.0 instead? 

No, in fact I'm currently working on opening up all of the dependency versions.

> With this syntax each build on CI could have a different version of react-router whether we expect it to or not. That also means the CI build could pick up a new version of the dependency with bad code embedded before we've had a change to review things.

Not true.  The lockfile prevents that.

Blacktiger (approved):

gallery/package.json
Blacktiger: Good point, CI should only be using the lock file so as long as we review the lock file before merging we're fine.

rpokorny: > Good point, CI should only be using the lock file so as long as we review the lock file before merging we're fine.

Well, on that topic, the lockfile often changes in mysterious ways.  It's more a question of _did_ it change in a given PR (and it shouldn't unless deps in package.json are modified) rather than _how_ it changed.

PR #271, opened 2021-01-08
nx-table background - RSC-417

rpokorny:
https://issues.sonatype.org/browse/RSC-417


glenhunter:
Given what we have seen with what Jeffry did with RSC yesterday I think we should create and use a variable like `$nx-component-background-color` for this (and eventually other components) so that we can future-proof against the need for "dark mode"

rpokorny:
> Given what we have seen with what Jeffry did with RSC yesterday I think we should create and use a variable like $nx-component-background-color for this (and eventually other components) so that we can future-proof against the need for "dark mode"

Sure.  That's bigger than this ticket though

glenhunter:
> > Given what we have seen with what Jeffry did with RSC yesterday I think we should create and use a variable like $nx-component-background-color for this (and eventually other components) so that we can future-proof against the need for "dark mode"
> 
> Sure. That's bigger than this ticket though

It's not that much bigger, I'm just suggesting that instead of using `$nx-white` you create a new variable that has `$nx-white` mapped to it.

rpokorny:
> It's not that much bigger, I'm just suggesting that instead of using $nx-white you create a new variable that has $nx-white mapped to it.

And then use that variable everywhere, including many places that aren't related to this PR

glenhunter:
> > It's not that much bigger, I'm just suggesting that instead of using $nx-white you create a new variable that has $nx-white mapped to it.
> 
> And then use that variable everywhere, including many places that aren't related to this PR

No, I'm saying create it and use it here just for the tables. Now we have it and we can roll it out elsewhere later.

rpokorny:
> No, I'm saying create it and use it here just for the tables. Now we have it and we can roll it out elsewhere later.

I've created another ticket and PR to set up the variable, and am using it in this PR in commit 
0fe63d6

glenhunter (approved):

PR #272, opened 2021-01-08
Add bold italic font face RSC-95

glenhunter:
https://issues.sonatype.org/browse/RSC-95

Add both semi-bold italic and bold italic, now we have normal and italic faces for each of the three font weights that we support. I also created mixins for the two new font faces


lib/src/base-styles/_nx-typography.scss
```diff
@@ -28,8 +42,8 @@
 
 @font-face {
   font-family: "Open Sans";
-  src: url("../assets/fonts/open-sans-v17-latin-italic.woff") format("woff");
-  font-weight: normal;
+  src: url("../assets/fonts/open-sans-v18-latin-700italic.woff") format("woff");
```
rpokorny, line 45: Seems we have a mix of v17 and v18 font files.  Should probably be consistent.

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -16,16 +16,33 @@
   font-weight: normal;
 }
 
+@mixin italic() {
```
rpokorny, line 19: I think the value of these is debatable.

lib/src/base-styles/_nx-typography.scss
glenhunter: > Seems we have a mix of v17 and v18 font files. Should probably be consistent.

I couldn't find any differences but sure: https://github.com/sonatype/sonatype-react-shared-components/pull/272/commits/14273e263e469f4ecd4af9f2296878ea58f87fa0 & https://github.com/sonatype/sonatype-react-shared-components/pull/272/commits/4837e6638527b1045fe5d477b778d0505325bb2e

lib/src/scss-shared/_nx-text-helpers.scss
glenhunter: > I think the value of these is debatable.

I know but we use them and we use them consistently so we may as well have full coverage

rpokorny (approved):
Remove the v17 files, otherwise +1

PR #273, opened 2021-01-08
monospace font RSC-176

glenhunter:
https://issues.sonatype.org/browse/RSC-176

Added Courier Prime font files and modified the various "code" classes


lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -73,7 +73,7 @@
   border: $nx-border;
   border-radius: $nx-border-radius;
   color: #cd0028;
-  font-family: monospace;
+  font-family: Courier Prime, monospace;
```
rpokorny, line 76: Should we make a mixin that encapsulates this line?

glenhunter: > Should we make a mixin that encapsulates this line?

I thought about it but didn't think that it was worth it given how little it's used. Do you think it's worth it? For consistency if nothing else?

rpokorny: > I thought about it but didn't think that it was worth it given how little it's used. Do you think it's worth it? For consistency if nothing else?

I feel like it is worth it (and easy to do) but it's not a big deal either way.

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -90,6 +99,7 @@
 
   // "code" used below to support Markdown text in NxVulnerabilityDetails
   code, .nx-code {
+    @include monospace();
```
rpokorny, line 102: So you want `.nx-code` to be normal weight even when it's in something that's otherwise bold (like a header)? That's not the effect that this PR had before you added the mixins

glenhunter: > So you want `.nx-code` to be normal weight even when it's in something that's otherwise bold (like a header)? That's not the effect that this PR had before you added the mixins

Ah good point. https://github.com/sonatype/sonatype-react-shared-components/pull/273/commits/1b7e411138b982aa98cb77c2652340f48de4ae7a

lib/src/scss-shared/_nx-text-helpers.scss
glenhunter: > > I thought about it but didn't think that it was worth it given how little it's used. Do you think it's worth it? For consistency if nothing else?
> 
> I feel like it is worth it (and easy to do) but it's not a big deal either way.

https://github.com/sonatype/sonatype-react-shared-components/pull/273/commits/97344f92b4bd3a316830578ebab02ba1869e4f4d

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -49,6 +53,10 @@
   white-space: nowrap;
 }
 
+@mixin monospace() {
```
rpokorny, line 56: This is now redundant with the other mixin

glenhunter: > This is now redundant with the other mixin

I kept `@monospace` because the naming was similar to our existing commonly used font mixins: https://github.com/sonatype/sonatype-react-shared-components/pull/273/commits/b92cf1cdf46a95a9a43a1ed5d5e2e5ed2af4cd36

rpokorny (approved):

PR #274, opened 2021-01-11
Add nx-form-group and nx-fieldset docs - RSC-435

rpokorny:
https://issues.sonatype.org/browse/RSC-435
https://issues.sonatype.org/browse/RSC-436

This PR converts all previously-existing examples to use the `NxFormGroup` and `NxFieldset` react components, and makes dedicated doc pages for the bare classes, including explanations of why they should be avoided in favor of the react components, and explanations and examples of deprecated alternate layouts.


glenhunter:
I found other instances of `nx-fieldset` (not the component) on the Form validation example page, and on the form layout example pages.

glenhunter:
I think that the alert on the `nx-form-group` page should be "warning" rather than "info". We probably want to have the same alert on the `nx-fieldset` page.

rpokorny:
> I found other instances of nx-fieldset (not the component) on the Form validation example page, and on the form layout example pages.

Whoops must've lost track of the rest of them over the weekend. 5dd50663

rpokorny:
> I think that the alert on the nx-form-group page should be "warning" rather than "info". 

b2a8c01f

> We probably want to have the same alert on the nx-fieldset page.

The alert is specifically about the tedium of setting up the screenreading support, which doesn't apply to `nx-fieldset`.  While `nx-fieldset` is just as unfavored as, say, `nx-btn`, `nx-form-group` is _especially_ unfavored due to having to connect all the ids and things.

glenhunter (approved):

PR #277, opened 2021-01-12
Add mention of native attributes to component props - RSC-445

rpokorny:
https://issues.sonatype.org/browse/RSC-445


glenhunter:
Per convo on Slack. I think that we should consistently refer to **HTML attributes** as **attributes** (rather than **props**) to make clear the difference between a **prop** which we probably created and a native **HTML attribute** (which while it could be called a **prop**, is not something that we created).

rpokorny:
> Per convo on Slack. I think that we should consistently refer to HTML attributes as attributes (rather than props) to make clear the difference between a prop which we probably created and a native HTML attribute (which while it could be called a prop, is not something that we created).

a98117b8

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -191,7 +197,9 @@ export default function NxTablePage() {
           </header>
           <p className="nx-p">
             Equivalent to the <code className="nx-code">&lt;th&gt;</code> or
-            {' '}<code className="nx-code">&lt;td&gt;</code> element.
+            {' '}<code className="nx-code">&lt;td&gt;</code> element. This component can
+            receive any prop that would be valid on a <code className="nx-code">&lt;td&gt;</code> as well as the
```
glenhunter, line 201:
```suggestion
            receive any attribute that would be valid on a <code className="nx-code">&lt;td&gt;</code> as well as the
```

glenhunter (approved):

PR #278, opened 2021-01-12
Create and use style variable for background color - RSC-447

rpokorny:
https://issues.sonatype.org/browse/RSC-447

Stemming from @glenhunter 's comment here https://github.com/sonatype/sonatype-react-shared-components/pull/271#issuecomment-758913952 and @DarthHater 's recent work, we are starting to think about making it easier to create an alternative color theme for RSC, and/or allowing consuming projects to easily override colors used within RSC.  Towards that end, this PR adds a CSS custom property for the background color of typical components.  Being a CSS custom prop allows it to be easily overridden in consuming projects.


lib/src/base-styles/_nx-btn.scss
```diff
@@ -21,7 +21,7 @@
   @include container-horizontal;
   @extend %nx-btn--disabled;
 
-  background-color: $nx-white;
+  background-color: var(--rsc-component-background-color);
```
glenhunter, line 24: I don't think that we want to use this variable on buttons, we'd want a separate set for the various button types

lib/src/base-styles/_nx-form-select.scss
```diff
@@ -11,7 +11,7 @@
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
-  background: #fff url("data:image/svg+xml;utf8, \
+  background: var(--rsc-component-background-color) url("data:image/svg+xml;utf8, \
```
glenhunter, line 14: I'm not sure that we would want this, the text input, or the policy slider to also be tied to this variable. I think a form element background colour variable might be a better way to go.

lib/src/base-styles/_nx-btn.scss
rpokorny: > I don't think that we want to use this variable on buttons, we'd want a separate set for the various button types

aa490bf3

lib/src/base-styles/_nx-form-select.scss
rpokorny: > I'm not sure that we would want this, the text input, or the policy slider to also be tied to this variable. I think a form element background colour variable might be a better way to go.

aa490bf3

rpokorny:
Renamed the variable to use `nx` as the prefix rather than `rsc`.  For consistency with our other variables. 96696627

glenhunter (approved):

PR #284, opened 2021-01-18
Improve documentation around bundler requirements - RSC-451

rpokorny:
https://issues.sonatype.org/browse/RSC-451

These changes are meant to make the living example of an RSC webpack config, within the gallery, clearer to understand.


doddi (approved):
both tickets +1

PR #285, opened 2021-01-19
POC: Main section scrollbar at edge of viewport - RSC-140

rpokorny:
https://issues.sonatype.org/browse/RSC-140

I admit this change does make things look nicer, but it makes the styling much more complicated, and more inconsistent/asymmetrical.

Before:
image

After:
image

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.3.9",
+  "version": "3.3.10",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "3.3.9",
+  "version": "3.3.10",
   "description": "Sonatype shared UI components and utilities written in React",
   "main": "index.js",
   "repository": {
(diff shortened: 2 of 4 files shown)


glenhunter:
It would be good to have a "before" picture attached to the PR as well for context

glenhunter:
While I am impressed with this from a technical standpoint (well done Ross!), it seems very brittle - very much like the older IQ dashboard which broke every time a number got tweaked. I'd suggest two things:

1. We go back to base assumptions and re-write the page layout code from the ground up with this as a requirement (which may make no difference to its complexity)
2. We abandon this / revisit our required page scrolling behaviours

rpokorny:
> It would be good to have a "before" picture attached to the PR as well for context

Added

rpokorny:
> We go back to base assumptions and re-write the page layout code from the ground up with this as a requirement (which may make no difference to its complexity)

The problem is that I don't think there's a cleaner way to do it without adding elements which makes it a harder change for downstream projects to consume

> We abandon this / revisit our required page scrolling behaviours

I'm in favor of abandoning it.  From what I hear the design preference is for page scrolling anyway and section scrolling is just for certain existing cases (namely IQ)

rpokorny:
>  certain existing cases that will hopefully be redesigned at some point (namely the two column pages in IQ)

And honestly, most of those could probably be switched to page scrolling today and not be any worse for wear.  It may really only be the IQ dashboard that's a problem.

rpokorny:
Given the extra complexity required for this solution, we've decided not to move forward with this.

PR #286, opened 2021-01-19
Update font-weight and font-size for sublabels - RSC-340

rpokorny:
https://issues.sonatype.org/browse/RSC-340

TBH I don't remember what made us decide that we wanted this change.  I consider revisiting whether or not we really do to be part of the review here.

image

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.3.9",
+  "version": "3.3.10",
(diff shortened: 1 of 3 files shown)


glenhunter:
In my opinion the font looks too small. It looks like a mistake.

rpokorny:
> In my opinion the font looks too small. It looks like a mistake.

@dturner32

rpokorny:
After seeing the results we've decided we want to leave these styles as-is and not go through with these changes.

PR #288, opened 2021-01-19
Document supported node version - RSC-452

rpokorny:
https://issues.sonatype.org/browse/RSC-452

Turns out we already do this, but I've updated it and expanded it slightly.


rpokorny:
Also updating mentions of `npm` commands to use `yarn` instead, since I removed `npm` from the Required Software list

glenhunter (approved):

PR #293, opened 2021-01-22
Color variable replacement: pink, green, red, orange, yellow, turquoise, purple - RSC-61

rpokorny:
https://issues.sonatype.org/browse/RSC-61

### DIFF
--- a/lib/src/base-styles/_nx-alert.scss
+++ b/lib/src/base-styles/_nx-alert.scss
@@ -79,9 +79,9 @@
 
 .nx-alert--success {
   background-color: #fafffc;
-  border-color: #007623;
+  border-color: $nx-green-700;
 
   > .nx-icon {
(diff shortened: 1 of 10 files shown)


lib/src/base-styles/_nx-alert.scss
```diff
@@ -61,10 +61,10 @@
 
 .nx-alert--error {
   background-color: #fff6f8;
-  border-color: #cd0028;
+  border-color: $nx-red-800;
```
glenhunter, line 64: Should we make `$nx-error` common variables for here and the buttons (and any other error states)?

rpokorny: > We should make $nx-error common variables for here and other error states.

I definitely think we should go through and improve and cleanup the more semantic color variables, but not in this ticket.

PR #297, opened 2021-01-26
Fix doc for NxForm

Blacktiger:
Not sure if I need to bump the package.json versions for this.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.4.0",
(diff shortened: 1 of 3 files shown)


rpokorny:
> Not sure if I need to bump the package.json versions for this.

You do simply because that's how our CI works.  Every commit/merge to master it tries to release as a new version.  If you don't bump the versions it'll try to publish a dupe to the npm registry which will fail.  To head that off we have a script in the dev branch builds which will fail _those_ if you don't have the version bump.

rpokorny (approved):

PR #299, opened 2021-01-28
NxSegmentedButton - RSC-449

rpokorny:
https://issues.sonatype.org/browse/RSC-449

### DIFF
--- a/gallery/visualtests/testUtils.js
+++ b/gallery/visualtests/testUtils.js
@@ -36,11 +36,19 @@ module.exports = {
 
       // make sure mouse is not on element
       await targetElement.moveTo({ xOffset: -10, yOffset: -10 });
-      await browser.execute(function(el) {
-        el.focus();
-      }, focusElement);
 
-      await browser.eyesRegionSnapshot(null, Target.region(targetElement));
+      try {
+        await browser.execute(function(el) {
+          el.focus();
+        }, focusElement);
+
+        await browser.eyesRegionSnapshot(null, Target.region(targetElement));
+      }
+      finally {
+        await browser.execute(function(el) {
+          el.blur();
+        }, focusElement);
+      }
     };
   },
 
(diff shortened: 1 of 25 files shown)


gallery/visualtests/testUtils.js
```diff
@@ -36,11 +36,19 @@ module.exports = {
...
+      try {
+        await browser.execute(function(el) {
+          el.focus();
+        }, focusElement);
+
+        await browser.eyesRegionSnapshot(null, Target.region(targetElement));
+      }
+      finally {
+        await browser.execute(function(el) {
+          el.blur();
```
rpokorny, line 49: Remove focus after test finishes to avoid impacting other tests

lib/src/base-styles/_nx-btn.scss
```diff
@@ -56,7 +58,7 @@
   @extend %nx-btn--disabled;
 
   background-color: $nx-blue-600;
-  border-color: $nx-blue-600;
+  border-color: transparent;
```
rpokorny, line 61: This part of the primary button styles was wrong.  There should be no visible border according to zeplin

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -94,4 +96,8 @@
   .nx-page-content {
     margin: 0 auto;
   }
+
+  .nx-page-main {
+    overflow: visible;
```
rpokorny, line 101: Added because the dropdowns happened to be going beyond the edge of `.nx-page-main` and I discovered that they get cut off there.

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -66,6 +67,7 @@
 
   box-sizing: border-box;
   flex: 1 0 $nx-content-width;
+  min-width: 0;
```
rpokorny, line 70: Added to mitigate the `overflow` change below.  Without this, long unbroken text content would make the `nx-page-main` exceed its proper width and smoosh the sidebar

glenhunter (approved):

PR #304, opened 2021-02-01
Indigo palette - RSC-479

rpokorny:
https://issues.sonatype.org/browse/RSC-479

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.5.2",
+  "version": "3.5.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.5.1",
+  "version": "3.5.2",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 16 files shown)


glenhunter:
It seem like this is a pattern that repeats a lot: `1px solid $nx-indigo-100;` do we want to look and see if there's a better way to group these with variables?

rpokorny:
> It seem like this is a pattern that repeats a lot: 1px solid $nx-indigo-100; do we want to look and see if there's a better way to group these with variables?

I think it's kind of our standard border.  Could make it an `$nx-standard-border` variable or something.  Would that be in-scope here do you think?

glenhunter:
> > It seem like this is a pattern that repeats a lot: 1px solid $nx-indigo-100; do we want to look and see if there's a better way to group these with variables?
> 
> I think it's kind of our standard border. Could make it an `$nx-standard-border` variable or something. Would that be in-scope here do you think?

I would not be opposed to having it in this ticket, but it would effect the scope (possibly by a lot) so I also wouldn't be opposed to having it as a separate ticket.

I don't think that we should replace all instances with a single variable. I think we should break it down into a few different variables based on use. Probably the versions used inside tables and lists as one variable. Another for borders that separate a part of a larger container from another like headers, footers, etc. Then if it makes sense another for borders that wrap an entire component like buttons, or the borders around checkboxes and radio buttons.

rpokorny:
> I think it's kind of our standard border. Could make it an $nx-standard-border variable or something. Would that be in-scope here do you think?

821a5af

> I don't think that we should replace all instances with a single variable. I think we should break it down into a few different variables based on use. Probably the versions used inside tables and lists as one variable. Another for borders that separate a part of a larger container from another like headers, footers, etc. Then if it makes sense another for borders that wrap an entire component like buttons, or the borders around checkboxes and radio buttons.

After talking in slack we decided not to split it up in that way.  If some of these borders deviate from the default in the future then they are free to stop using the new variable at that point.

glenhunter (approved):

PR #307, opened 2021-02-09
nx-card examples RSC-467

glenhunter:
https://issues.sonatype.org/browse/RSC-467

Updates to `nx-card` based on experience modifying them for use in IQ.

Per conversation in Wall Walk on Feb 17, the decision was made to remove the Horizontal cards altogether as there is too much duplication between them and `nx-tile`.


gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
```diff
@@ -8,93 +8,87 @@ import React from 'react';
 import { faShapes, faEdit, faTrash } from '@fortawesome/free-solid-svg-icons';
 
 import { NxFontAwesomeIcon, NxButton } from '@sonatype/react-shared-components';
+const chart = require('./chart.png');
 
 export default function NxCardColumnLayoutExample() {
   return (
     <div className="nx-card-container nx-card-container--column">
-      <section className="nx-card" aria-label="card 1 in column layout">
+      <section className="nx-card nx-card--horizontal" aria-label="card 1 in column layout">
```
rpokorny, line 16: Wasn't the idea previously that `.nx-card--horizontal` should be redundant because of `.nx-card-container--column`?

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -8,21 +8,23 @@ import React from 'react';
...
   <>
     <GalleryDescriptionTile>
       <p className="nx-p">
-        Cards can be displayed in two orientations: rows and columns. Cards can be transformed between rows and columns
-        with a simple class change on their container. There is also a row variant supplied which displays the call out
-        and content side-by-side, this variant cannot be transformed into column oriented cards.
+        Cards can be displayed in two orientations: vertically or horizontally. At its most basic a card is simply a
+        container like a smaller version of <code className="nx-code">nx-tile</code> with header, content, and footer
+        content areas. Because it's expected that cards will have many uses the examples below do not represent an
+        exhasutive list, instead the intent is to display some common patterns that we have seen to date, and provide a
```
rpokorny, line 26:
```suggestion
        exhaustive list, instead the intent is to display some common patterns that we have seen to date, and provide a
```

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -65,36 +65,30 @@ const NxCardPage = () =>
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-card__header</code></td>
-            <td className="nx-cell">Element</td>
+            <td className="nx-cell"><code className="nx-code">.nx-card--horizontal</code></td>
+            <td className="nx-cell">Modifer of <code className="nx-code">.nx-card</code></td>
             <td className="nx-cell">
-              Card header element, typically wraps an <code className="nx-code">&lt;h3&gt;</code>.
+              Used for cards that are laid out horizontally rather than vertically.
```
rpokorny, line 71: As it was originally, I think the horizontal/vertical nomenclature is very ambiguous.  Since these cards stack vertically they could just as easily (or perhaps even more easily) be thought of as "vertical cards"

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -65,36 +65,30 @@ const NxCardPage = () =>
...
+            <td className="nx-cell"><code className="nx-code">.nx-card__footer</code></td>
+            <td className="nx-cell">Element</td>
             <td className="nx-cell">
-              Used when the content of <code className="nx-code">.nx-card__call-out</code> consists of text.
+              In a vertically oriented card the footer. Typically provides a link to more information.
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-card__text</code></td>
+            <td className="nx-cell"><code className="nx-code">.nx-card__actions</code></td>
```
rpokorny, line 89: Should `.nx-card__footer` be merged with this?  It seems like they do the same things just in the two different kinds of cards.

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -103,9 +97,27 @@ const NxCardPage = () =>
...
               {' '}<code className="nx-code">&lt;div&gt;</code> wraps
               {' '}<code className="nx-code">.nx-card__call-out</code> and
-              {' '}<code className="nx-code">.nx-card__call-out-text</code> in the alternate row card layout.
+              {' '}<code className="nx-code">.nx-card__text</code> or any other card content.
             </td>
           </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-card__call-out</code></td>
+            <td className="nx-cell">
+              Element, typically a child of <code className="nx-code">.nx-card__content</code> in a vertically oriented
```
rpokorny, line 106: Why does this say "in a vertically oriented card"?  Seem to be used in horizontal cards as well.

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+// Reveal problems with box-sizing set to border-box, see https://issues.sonatype.org/browse/RSC-84
```
rpokorny, line 7: Is this comment still valid?

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
+/**
```
rpokorny, line 1: This file needs to be included as a code example

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+// Reveal problems with box-sizing set to border-box, see https://issues.sonatype.org/browse/RSC-84
+
+.nx-card__content--2-column-example {
```
rpokorny, line 9: Typically our example classes start with `.gallery` to make it clear that they aren't part of the RSC lib

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
...
+  align-items: center;
+  display: flex;
+
+  >:first-child {
```
rpokorny, line 13:
```suggestion
  > :first-child {
```

gallery/src/styles/NxCard/NxCardPageExamples.scss
rpokorny, line 9: Is this capability not something that we want in RSC itself?  It was in the original nx-card mockups

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
...
+    margin-right: 16px;
+  }
+}
+
+.nx-card--align-left {
+  text-align: left;
```
rpokorny, line 19: This one feels weird to me.  Like left aligning the title and left aligning the content text are two separate, and rather custom, concerns.  I'm not sure it would generally make sense to do them both with a single class and I'm not sure it isn't too arbitrary for us to be demonstrating.

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.nx-card-container--row .nx-card--75wide {
+  max-width: 75%;
+}
+
+.nx-card-container--row .nx-card--50wide {
+  max-width: 50%;
```
rpokorny, line 27: Why max-width and not width for these?

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.nx-icon--indicator {
```
rpokorny, line 30: This seems out of scope.  Also they aren't properly aligned

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
```diff
@@ -0,0 +1,266 @@
...
+              <NxFontAwesomeIcon icon={faShapes} className="nx-icon--card-call-out" />
+            </div>
+            <div className="nx-card__text">
+              Sentient post-singularity
+            </div>
+          </div>
+        </section>
+      </div>
+      <div className="nx-card-container nx-card-container--row">
+        <section className="nx-card nx-card--50wide" aria-label="card 10 in row">
```
rpokorny, line 239: Is there an intent that these two would fit side-by-side? They don't

lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "3.6.1",
+  "version": "4.0.0",
```
rpokorny, line 3: Could you clarify where you were unable to keep things backwards compatible and why?

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > As it was originally, I think the horizontal/vertical nomenclature is very ambiguous. Since these cards stack vertically they could just as easily (or perhaps even more easily) be thought of as "vertical cards"

I also thought of saying "wide rather than tall". I even considered removing the "wide" cards entirely since there is no demand for them at this point at all. I was trying to make a distinction in the language between how a card is oriented vs how cards are arrayed on the page.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Why does this say "in a vertically oriented card"? Seem to be used in horizontal cards as well.

It is used in horizontal cards but it is not a child of `nx-card__content` in horizontal cards. This is one of the breaking changes.

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > Is this comment still valid?

No: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/f4fbce248894a3a44d33b33bb3ed5a4ed0d8dac3

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > Is this capability not something that we want in RSC itself? It was in the original nx-card mockups

I wavered on this, but have moved it back to the cards: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/17a7663a804fa121012ada6de25f21f4a2a1837c

glenhunter: > Typically our example classes start with `.gallery` to make it clear that they aren't part of the RSC lib

Right: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/17a7663a804fa121012ada6de25f21f4a2a1837c

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > This seems out of scope. Also they aren't properly aligned

It was just meant to show how to do what has become a common layout in the mockups, when we have the status indicators in the RSC these should be replaced. I fixed the alignment issues here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/17a7663a804fa121012ada6de25f21f4a2a1837c

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
```diff
@@ -0,0 +1,266 @@
...
+ */
+import React from 'react';
+import { faShapes, faChartLine, faEdit, faTrash, faCircle } from '@fortawesome/free-solid-svg-icons';
+
+import { NxFontAwesomeIcon, NxButton } from '@sonatype/react-shared-components';
+
+export default function NxCardRowLayoutExample() {
+  return (
+    <>
+      <div className="nx-card-container nx-card-container--row">
```
rpokorny, line 15: This example has a mix of separate `.nx-card-container--row`s and `.nx-card-container--row`s that wrap.  I think we should clarify that difference in some way.  Either by splitting it into multiple examples or by putting some label text above each `nx-card-container--row` describing it.

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
+  padding: $nx-spacing-l;
+  text-align: center;
 }
 
 .nx-card-container--row {
+  @include container-horizontal;
   display: flex;
-  -ms-flex-wrap: wrap;
   flex-wrap: wrap;
+  justify-content: space-between;
```
rpokorny, line 25: When there is a multi-line flex container, this doesn't work as expected on the last item of the non-last line(s).  I suggest removing the margin from the cards and using `gap` instead.

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
   border-radius: $nx-border-radius;
   box-sizing: border-box;
   margin-bottom: $nx-spacing-md;
```
rpokorny, line 16: The ways the top and bottom margins of the cards and containers interact is probably not ideal.  Containers have a 16px top margin.  Cards have a 16px bottom margin.  These two margins do not collapse into each other for adjacent containers, resulting in a total space of 32px.  Then of course the space between rows of a single, multi-line container is only the 16px card bottom margin.  My recommendation: remove the bottom margin from the cards, add a `row-gap` to the container, and add a bottom margin to the container

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
-    text-align: center;
-    min-width: 230px;
-    max-width: 330px;
-  }
-
-  .nx-card__header {
-    @include container-vertical;
-    margin-bottom: $nx-spacing-md;
-  }
+.nx-card {
```
rpokorny, line 29: You've got multiple `nx-card` blocks in this file

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
+  @include container-vertical;
+  display: grid;
+  grid-template-rows: auto 1fr auto;
```
rpokorny, line 32: I feel like it might be simpler and more future-proof to use `flex` for this.  By using grid in this way, you're basically assuming the first element will always be the header.  If you use flex, the sizings would be specified by the children, not here, so there wouldn't have to be assumptions about which children are present.

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
+  margin-right: $nx-spacing-md;
+  min-width: 230px;
+  max-width: 330px;
```
rpokorny, line 35: We'll probably want to make it clear in the documentation that this is a good default but which the consumer would often override for deliberately-wide cards.  Or maybe we should even make a `--wide` variant?

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
glenhunter: > Is there an intent that these two would fit side-by-side? They don't

Fixed here though it feels clunky: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/8f5da678aea8c9dbaeb0dae27bd601b2b3379311

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
+}
 
-  .nx-card__call-out--text-only {
-    + .nx-card__content {
-      margin-top: $nx-spacing-xs;
-    }
-  }
+.nx-card__content {
+  align-content: start;
+  align-items: start;
```
rpokorny, line 45: I'm fairly sure this doesn't do anything since the grid box heights are fit to the content anyway.

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
glenhunter: > This example has a mix of separate `.nx-card-container--row`s and `.nx-card-container--row`s that wrap. I think we should clarify that difference in some way. Either by splitting it into multiple examples or by putting some label text above each `nx-card-container--row` describing it.

This is something that I was going to bring up in WW today, but unfortunately @dturner32 won't be there. If we're moving to a broader responsive model then things like this are going to wrap more frequently at places we can't always anticipate.

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
+  display: grid;
```
rpokorny, line 46: What's the reason that this is a grid? I'd think it would work just as well as a normal block container using `container-vertical` and appropriate margins on its children

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
-  .nx-card__footer {
-    border-top: $nx-border-default;
-    margin-top: $nx-spacing-l;
-    padding-top: $nx-spacing-l;
-    width: 100%;
+.nx-card__footer {
+  border-top: $nx-border-default;
+  margin-top: $nx-spacing-l;
+  padding-top: $nx-spacing-l;
+  width: 100%;
```
rpokorny, line 54: Not sure this does anything

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
-    -ms-grid-rows: auto auto;
-    grid-template-areas:
-      "card-callout card-title card-footer"
-      "card-callout card-text card-footer";
-  }
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  grid-template-areas:
+    "card-callout card-title card-actions"
+    "card-callout card-content card-actions";
```
rpokorny, line 67: What's the purpose of prefixing all of these with "card"?

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
+  margin-right: 0;
+  text-align: left;
+  max-width: 100%;
```
rpokorny, line 70: Should this just be `width`? Aren't these cards always supposed to stretch all the way across their container?

rpokorny: > Should this just be width? Aren't these cards always supposed to stretch all the way across their container?

Oh I see, this is just to cancel out the `max-width: 330px` on vertical cards.  Might be clearer to use `unset` or `auto` then.

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
   }
 
   .nx-card__content {
-    -ms-grid-row: 2;
-    -ms-grid-column: 2;
-    grid-area: card-text;
+    align-self: center;
```
rpokorny, line 77: What is the expected behavior when there is a header, content, and a callout large enough that there is extra vertical space amidst the header and content? Is this correct/acceptable:
image

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
 
   .nx-card__call-out {
-    -ms-grid-row-align: center;
     align-self: center;
-    -ms-grid-row: 1;
-    -ms-grid-row-span: 2;
-    -ms-grid-column: 1;
     grid-area: card-callout;
-    -ms-grid-column-align: center;
     justify-self: center;
```
rpokorny, line 87: Doesn't do anything since this element's width is dictated by its content

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
+    align-self: center;
+    grid-area: card-content;
     margin: 0;
     padding: 0;
+    row-gap: 0;
   }
 
   .nx-card__call-out {
-    -ms-grid-row-align: center;
     align-self: center;
```
rpokorny, line 85: This is redundant with the `height: 100%`.  One or the other can be removed

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
-    -ms-grid-row: 1;
-    -ms-grid-row-span: 2;
-    -ms-grid-column: 1;
     grid-area: card-callout;
-    -ms-grid-column-align: center;
     justify-self: center;
 
     align-items: center;
```
rpokorny, line 89: If `height: 100%` is removed I believe this would be redundant too

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
     border-right: $nx-border-default;
     display: flex;
```
rpokorny, line 91: Also probably redundant if `height: 100%` is removed

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
     box-sizing: content-box;
-    -ms-grid-row-align: center;
     align-self: center;
-    -ms-grid-row: 1;
-    -ms-grid-row-span: 2;
-    -ms-grid-column: 3;
-    grid-area: card-footer;
-    -ms-grid-column-align: start;
+    grid-area: card-actions;
     justify-self: start;
```
rpokorny, line 101: Redundant

lib/src/base-styles/_nx-card.scss
```diff
@@ -14,91 +14,62 @@
...
-      justify-content: center;
-    }
+  > .nx-btn-bar {
+    justify-content: center;
   }
 }
 
-.nx-card-container--column {
+.nx-card--horizontal {
   @include container-vertical;
```
rpokorny, line 58: `.nx-card` already has this mixin

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
     align-self: center;
-    -ms-grid-row: 1;
-    -ms-grid-row-span: 2;
-    -ms-grid-column: 3;
-    grid-area: card-footer;
-    -ms-grid-column-align: start;
+    grid-area: card-actions;
     justify-self: start;
 
     @include container-horizontal;
```
rpokorny, line 103: Mixins typically are at the top of a code block

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
     align-items: center;
-    border: none;
     border-left: $nx-border-default;
     display: flex;
     height: 100%;
```
rpokorny, line 107: This probably has all the same redundancies as the same property in `.nx-card__header`

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
-    margin: 0 0 0 $nx-spacing-xxl;
-    padding: 0 0 0 $nx-spacing-xxl;
+    margin: 0 0 0 $nx-spacing-l;
+    padding: 0 0 0 $nx-spacing-l;
```
rpokorny, line 109: This isn't in sync with the mockups.  IIRC we agreed with @dturner32 to decrease these spacings?  He should update zeplin then.

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
+.nx-card__call-out {
+  @include bold();
+
+  font-size: 32px;
+
+  .nx-icon--card-call-out {
+    font-size: 48px;
+  }
+
+  .nx-icon--card-call-out-xl {
```
rpokorny, line 125: I feel like these would be better called `.nx-card__call-out-icon`  and `.nx-card__call-out-icon--xl`

lib/src/base-styles/_nx-card.scss
```diff
@@ -107,75 +78,64 @@
...
+    font-size: 64px;
+  }
+}
+
+.nx-card__content--2-column {
```
rpokorny, line 130: If `.nx-card__content` was flex-based this could probably be as simple as overriding the `flex-direction` with no other changes

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > I also thought of saying "wide rather than tall". I even considered removing the "wide" cards entirely since there is no demand for them at this point at all. I was trying to make a distinction in the language between how a card is oriented vs how cards are arrayed on the page.

I'm wondering if we can avoid the issue by simply getting rid of this class, and having all the styles that it provides attached to any `.nx-card` that is a child of a `.nx-card-container--column`

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > It is used in horizontal cards but it is not a child of nx-card__content in horizontal cards. This is one of the breaking changes.

Ok, I think this wording could be improved then; as it is now it suggests that this element is only used in vertically oriented cards

gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
glenhunter: > Wasn't the idea previously that `.nx-card--horizontal` should be redundant because of `.nx-card-container--column`?

I think that making the card's appearance/layout dependent on their container was a mistake. That's why I moved over to a modifier class on `nx-card`

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Should `.nx-card__footer` be merged with this? It seems like they do the same things just in the two different kinds of cards.

I purposely separated them because in the horizontal layout they appear at the right side, not below. They had so little in common I just didn't see the point.

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > This file needs to be included as a code example

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/3b7359d1d9654d0acb9a78749a333b4892173239

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > I also thought of saying "wide rather than tall". I even considered removing the "wide" cards entirely since there is no demand for them at this point at all. I was trying to make a distinction in the language between how a card is oriented vs how cards are arrayed on the page.
> 
> I'm wondering if we can avoid the issue by simply getting rid of this class, and having all the styles that it provides attached to any `.nx-card` that is a child of a `.nx-card-container--column`

That's how it was originally. I moved away from it because I didn't like the container dictating the styling of the cards. I think that it's cleaner if the cards are responsible for their own internal layout and the container is responsible for the card's layout on the page, nothing more.

Initially it was based on my assumption that the ability to transform vertical cards to horizontal cards and back was a good idea. I think that was a bad assumption which made the cards overly complicated and too rigid in their layout.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > It is used in horizontal cards but it is not a child of nx-card__content in horizontal cards. This is one of the breaking changes.
> 
> Ok, I think this wording could be improved then; as it is now it suggests that this element is only used in vertically oriented cards

How's this? https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/9074805fa28bd78b150190e7170a1410c77c4819

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > Why max-width and not width for these?

They have to override `max-width: 330px;` set on `.nx-card`

lib/package.json
glenhunter: > Could you clarify where you were unable to keep things backwards compatible and why?

- removed `<footer>` and `.nx-footer` from the horizontal cards and replaced it with `.nx-card__actions`.
- removed the alternate vertical layout.
- all vertical layouts now have all content within `.nx-card__content`.

It's possible that I could have made the changes that I felt were necessary without causing breaking changes, but my concern was that we'd end up with convoluted and difficult to maintain (and modify) code. Better in my opinion to try and take what we learned from Expedition teams' work and do it right.

lib/src/base-styles/_nx-card.scss
glenhunter: > We'll probably want to make it clear in the documentation that this is a good default but which the consumer would often override for deliberately-wide cards. Or maybe we should even make a `--wide` variant?

We'll never get a `--wide` variant right. Some design will always need it narrower or wider than whatever magic number we choose. I'd rather just leave it custom and demonstrate how to set that custom value.

lib/src/base-styles/_nx-card.scss
glenhunter: > What's the reason that this is a grid? I'd think it would work just as well as a normal block container using `container-vertical` and appropriate margins on its children

"appropriate margins on its children" is the reason that I used `grid` it allows us to just use `row-gap` and not worry about applying that margin to every child object (which you know people will forget to do). I think that this will be easier for others implementing custom content.

lib/src/base-styles/_nx-card.scss
glenhunter: > I'm fairly sure this doesn't do anything since the grid box heights are fit to the content anyway.

Yep: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/c8367f79be644db6e71d67289f6e749e99d519b9

lib/src/base-styles/_nx-card.scss
glenhunter: > What is the expected behavior when there is a header, content, and a callout large enough that there is extra vertical space amidst the header and content? Is this correct/acceptable:

My expectation would be that if a designer specifies something like this that they will also design the rest of the layout appropriately to their needs (i.e. move the content text upwards, change its size, or whatever). I don't want to get too caught up in details of what happens to the example layouts in various scenarios because I think the example layouts are not prescriptive - as in "here are some examples of what you can do and how it's done", not "here are the layouts that we have made available".

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > This one feels weird to me. Like left aligning the title and left aligning the content text are two separate, and rather custom, concerns. I'm not sure it would generally make sense to do them both with a single class and I'm not sure it isn't too arbitrary for us to be demonstrating.

First pass I had it as separate values for each area but it looked so ugly having one centred and another left justified that I figured that no designer would ever do that. In the unlikely event that a designer does request that the developer would be welcome to create their own modifier.

It may be too arbitrary, I added it in because there was a mockup that appeared with left aligned content.

lib/src/base-styles/_nx-card.scss
glenhunter: > When there is a multi-line flex container, this doesn't work as expected on the last item of the non-last line(s). I suggest removing the margin from the cards and using `column-gap` instead.

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/14e55445ecb4714a15553f7625a7549a8c9efe1a

lib/src/base-styles/_nx-card.scss
glenhunter: > The ways the top and bottom margins of the cards and containers interact is probably not ideal. Containers have a 16px top margin. Cards have a 16px bottom margin. These two margins do not collapse into each other for adjacent containers, resulting in a total space of 32px. Then of course the space between rows of a single, multi-line container is only the 16px card bottom margin. My recommendation: remove the bottom margin from the cards, add a `row-gap` to the container, and add a bottom margin to the container

Good suggestion: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/14e55445ecb4714a15553f7625a7549a8c9efe1a

lib/src/base-styles/_nx-card.scss
glenhunter: > This is redundant with the `height: 100%`. One or the other can be removed

Here and below `height: 100%` is necessary to get the border to display full height.

lib/src/base-styles/_nx-card.scss
glenhunter: > If `.nx-card__content` was flex-based this could probably be as simple as overriding the `flex-direction` with no other changes

It wasn't quite that easy, but yeah: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > I feel like these would be better called `.nx-card__call-out-icon` and `.nx-card__call-out-icon--xl`

Better: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > This probably has all the same redundancies as the same property in `.nx-card__header`

many redundancies removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > Mixins typically are at the top of a code block

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > `.nx-card` already has this mixin

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > Redundant

many redundancies removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > Also probably redundant if `height: 100%` is removed

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > If `height: 100%` is removed I believe this would be redundant too

many redundancies removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > > This is redundant with the `height: 100%`. One or the other can be removed
> 
> Here and below `height: 100%` is necessary to get the border to display full height.

However changing the cards to flexbox solved that: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > Doesn't do anything since this element's width is dictated by its content

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > > Should this just be width? Aren't these cards always supposed to stretch all the way across their container?
> 
> Oh I see, this is just to cancel out the `max-width: 330px` on vertical cards. Might be clearer to use `unset` or `auto` then.

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/d7f3021fc33ed24d93b07e9313b516682735c42e

lib/src/base-styles/_nx-card.scss
glenhunter: > What's the purpose of prefixing all of these with "card"?

No purpose, removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/848980e874c7c3a447e1b31ebe5da962a4789218

lib/src/base-styles/_nx-card.scss
glenhunter: > Not sure this does anything

It doesn't https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/950eff29507a4b95fa04cda1424f7e36a340131f

lib/src/base-styles/_nx-card.scss
glenhunter: > I feel like it might be simpler and more future-proof to use `flex` for this. By using grid in this way, you're basically assuming the first element will always be the header. If you use flex, the sizings would be specified by the children, not here, so there wouldn't have to be assumptions about which children are present.

That's a good point https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1a904c393841bca80421915a60e0b8e1d7f4305f

lib/src/base-styles/_nx-card.scss
glenhunter: > You've got multiple `nx-card` blocks in this file

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/14e55445ecb4714a15553f7625a7549a8c9efe1a

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > I purposely separated them because in the horizontal layout they appear at the right side, not below. They had so little in common I just didn't see the point.

To me they're conceptually the same, and it would be one less class to deal with.  Per your point though, if they were merged I'd say they should be called `.nx-card__actions` not `.nx-card__footer`

gallery/src/styles/NxCard/NxCardPageExamples.scss
```diff
@@ -4,32 +4,25 @@
...
   max-width: 75%;
 }
 
-.nx-card-container--row .nx-card--50wide {
+.nx-card-container--row .gallery-card--50wide {
   max-width: 50%;
 }
 
-.nx-icon--indicator {
+.gallery--indicator {
```
rpokorny, line 20: Did you mean `.gallery-icon`?

gallery/src/styles/NxCard/NxCardPageExamples.scss
rpokorny: > It was just meant to show how to do what has become a common layout in the mockups, when we have the status indicator component in the RSC these should be replaced. I fixed the alignment issues here: 17a7663

I'd rather they just be removed until they are proper components.  They're just inline content so they don't really demonstrate anything special here, and they still aren't properly aligned.  In my experimentation on the Firewall page PR, I had to use a magic number to align them properly.

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
rpokorny: > Fixed here though it feels clunky: 8f5da67

That's now how I'd do it, I'd have a class for equal-width cards (probably on the container, not the cards themselves), which would set `flex: 1; max-width: auto` on the cards.

gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
```diff
@@ -8,93 +8,87 @@ import React from 'react';
...
+        <div className="nx-card__call-out">
```
rpokorny, line 17: If we do care at all about trying to keep the structure the same between the two types of cards, I think it would be possible using `display: contents`.  You could have them always structured the way the vertical cards are currently, and then horizontal cards would set `display: contents` on the `.nx-card__content` in order to position the callout and text independently.

rpokorny: > If we do care at all about trying to keep the structure the same between the two types of cards, I think it would be possible using display: contents. You could have them always structured the way the vertical cards are currently, and then horizontal cards would set display: contents on the .nx-card__content in order to position the callout and text independently.

This would also avoid the currently situation where in horizontal cards, `.nx-card__content` and `.nx-card__text` are kind of redundant.

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -103,18 +98,37 @@ const NxCardPage = () =>
...
               {' '}<code className="nx-code">.nx-card__call-out</code> and
-              {' '}<code className="nx-code">.nx-card__call-out-text</code> in the alternate row card layout.
+              {' '}<code className="nx-code">.nx-card__text</code> or any other card content.
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">nx-icon--card-call-out</code></td>
+            <td className="nx-cell"><code className="nx-code">.nx-card__call-out</code></td>
+            <td className="nx-cell">
+              Element which contains a very small piece of content, frequently graphical, meant for visual impact
```
rpokorny, line 107: Much of the content in this cell seems like it should be in the details cell instead.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > I purposely separated them because in the horizontal layout they appear at the right side, not below. They had so little in common I just didn't see the point.
> 
> To me they're conceptually the same, and it would be one less class to deal with. Per your point though, if they were merged I'd say they should be called `.nx-card__actions` not `.nx-card__footer`

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/307a84fed051a9b3fb6edaf35150c47d3a22fa65

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
```diff
@@ -0,0 +1,266 @@
...
+        <section className="nx-card nx-card--align-left gallery-card--66wide" aria-label="card 9 in row">
+          <header className="nx-card__header">
+            <h3 className="nx-h3">A wide card with left aligned content</h3>
+          </header>
+          <div className="nx-card__content nx-card__content--2-column">
+            <div className="nx-card__call-out">
+              <NxFontAwesomeIcon icon={faChartLine} className="nx-card__call-out-icon" />
+            </div>
+            <div className="nx-card__text">
+              Lots of content to force the card to a wider size sentient noodles sprawl tower corrupted cartel
```
rpokorny, line 219: "Lots of content to force the card to a wider size" isn't true here anymore

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
```diff
@@ -0,0 +1,266 @@
...
+        </section>
+      </div>
+      <div className="nx-card-container nx-card-container--row">
+        <section className="nx-card gallery-card--49wide" aria-label="card 10 in row">
+          <header className="nx-card__header">
+            <h3 className="nx-h3 nx-h3--card-header">A card that's ~50% of available width</h3>
+          </header>
+          <div className="nx-card__content">
+            <div className="nx-card__text">
+              Lots of content to force the card to a wider size alcohol realism tower cardboard network sunglasses
```
rpokorny, line 243: "Lots of content to force the card to a wider size" isn't true here anymore

lib/src/base-styles/_nx-card.scss
```diff
@@ -9,173 +9,121 @@
...
-    font-size: 64px;
-  }
+.nx-card__header {
+  @include container-vertical;
+  margin-bottom: $nx-spacing-md;
 }
 
-.nx-card-container--row {
+.nx-card__content {
+  align-content: flex-start;
```
rpokorny, line 46: Doesn't do anything

lib/src/base-styles/_nx-card.scss
rpokorny: > It wasn't quite that easy, but yeah: 1a904c3

It could be made that easy.  The `align-items: center` here could (and maybe even should) be moved to `.nx-card__content`.  And then if `.nx-card__content` specified `gap` instead of `row-gap`, we could get rid of `column-gap` here.

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
glenhunter: > > Fixed here though it feels clunky: [8f5da67](https://github.com/sonatype/sonatype-react-shared-components/commit/8f5da678aea8c9dbaeb0dae27bd601b2b3379311)
> 
> That's now how I'd do it, I'd have a class for equal-width cards (probably on the container, not the cards themselves), which would set `flex: 1; max-width: auto` on the cards.

I have tweaked the examples here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/2bfda09a9ab7682780dc05476eaea76a98e7581b

`max-width: auto` is not valid. I've used `unset` instead. The issue with having the cards appear side-by-side was `flex-wrap: wrap` which always favoured wrapping. I created a custom class to handle this and tweaked the examples further here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/367617f0ee1bbccfefecb3a81c23913471eb2aeb and https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/26d7fd068a9f7201927fd0283b3911d5ab00b543

gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
glenhunter: > > If we do care at all about trying to keep the structure the same between the two types of cards, I think it would be possible using display: contents. You could have them always structured the way the vertical cards are currently, and then horizontal cards would set display: contents on the .nx-card__content in order to position the callout and text independently.
> 
> This would also avoid the currently situation where in horizontal cards, `.nx-card__content` and `.nx-card__text` are kind of redundant.

I did have that before, it caused some other (minor) issues. I still think that making the cards convert from one layout to another was a mistake.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Much of the content in this cell seems like it should be in the details cell instead.

I like it better with minimal text, highlights that the SCSS is effecting the width, not the content: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/38859601e2652ef2c8fd2d2d283de7d9654cafb2

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
glenhunter: > "Lots of content to force the card to a wider size" isn't true here anymore

As above: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/38859601e2652ef2c8fd2d2d283de7d9654cafb2

lib/src/base-styles/_nx-card.scss
glenhunter: > Doesn't do anything

https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/1ac0539a4a8f13e7f292c67f126f06e52a1af6ba

gallery/src/styles/NxCard/NxCardPageExamples.scss
glenhunter: > > It was just meant to show how to do what has become a common layout in the mockups, when we have the status indicator component in the RSC these should be replaced. I fixed the alignment issues here: [17a7663](https://github.com/sonatype/sonatype-react-shared-components/commit/17a7663a804fa121012ada6de25f21f4a2a1837c)
> 
> I'd rather they just be removed until they are proper components. They're just inline content so they don't really demonstrate anything special here, and they still aren't properly aligned. In my experimentation on the Firewall page PR, I had to use a magic number to align them properly.

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/fc2308d74662eb8447c2eabfa5446fca1b879174

lib/src/base-styles/_nx-card.scss
glenhunter: > > It wasn't quite that easy, but yeah: [1a904c3](https://github.com/sonatype/sonatype-react-shared-components/commit/1a904c393841bca80421915a60e0b8e1d7f4305f)
> 
> It could be made that easy. The `align-items: center` here could (and maybe even should) be moved to `.nx-card__content`. And then if `.nx-card__content` specified `gap` instead of `row-gap`, we could get rid of `column-gap` here.

That does work: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/c8ff12e1387070f401c2c4f332b097c233bd3fbc

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > I like it better with minimal text, highlights that the SCSS is effecting the width, not the content: 3885960

That commit isn't relevant.  This comment is about the documentation table in the page

gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
rpokorny: > I did have that before, it caused some other (minor) issues. I still think that making the cards convert from one layout to another was a mistake.

Could you elaborate on the issues you saw?  To me it isn't about the convertibility, it's about decreasing the mental load.  It would be easier on developers if the structure was the same between the two cards.

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
rpokorny: > The issue with having the cards appear side-by-side was flex-wrap: wrap which always favoured wrapping

So that's only an issue because you specified `width: 50%`.  What I suggested above was to not do that, and to instead just give them the same flex factor.

lib/src/base-styles/_nx-card.scss
rpokorny: > We'll never get a --wide variant right. Some design will always need it narrower or wider than whatever magic number we choose. I'd rather just leave it custom and demonstrate how to set that custom value.

I see we've now got an example of a card that expands as much as possible which is great, but do you think we also need an example of a card with a specific custom max-width, or do you think consumers will be able to figure that out on their own?

gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
rpokorny: > Could you elaborate on the issues you saw? To me it isn't about the convertibility, it's about decreasing the mental load. It would be easier on developers if the structure was the same between the two cards.

I've tried it myself and it seems to work.  Are there issues that I'm not noticing? See https://github.com/sonatype/sonatype-react-shared-components/pull/311

gallery/src/styles/NxCard/NxCardVerticalLayoutExample.tsx
glenhunter: > > The issue with having the cards appear side-by-side was flex-wrap: wrap which always favoured wrapping
> 
> So that's only an issue because you specified `width: 50%`. What I suggested above was to not do that, and to instead just give them the same flex factor.

Ah, my mistake lay in not applying both `shrink` and `grow`. I liked this method so much I created a modifier class for it and some further examples: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/f7f3dee65e5f1b5d4edb6166456e6886e6fd0e41

lib/src/base-styles/_nx-card.scss
glenhunter: > This isn't in sync with the mockups. IIRC we agreed with @dturner32 to decrease these spacings? He should update zeplin then.

@dturner32 has adjusted the mockups

gallery/src/styles/NxCard/NxCardHorizontalLayoutExample.tsx
glenhunter: > > Could you elaborate on the issues you saw? To me it isn't about the convertibility, it's about decreasing the mental load. It would be easier on developers if the structure was the same between the two cards.
> 
> I've tried it myself and it seems to work. Are there issues that I'm not noticing? See #311

No longer applicable as we have removed the horizontal layouts entirely.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > I like it better with minimal text, highlights that the SCSS is effecting the width, not the content: [3885960](https://github.com/sonatype/sonatype-react-shared-components/commit/38859601e2652ef2c8fd2d2d283de7d9654cafb2)
> 
> That commit isn't relevant. This comment is about the documentation table in the page

oops: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/21c7cf5d9c42a40dd9fbedd90ddaa78bd3ffaa3b

gallery/src/styles/NxCard/NxCardLayoutExample.tsx
```diff
@@ -0,0 +1,296 @@
...
+            </div>
+            <div className="nx-status-indicator">
+              <span>Server two</span>
+            </div>
+          </div>
+          <footer className="nx-card__footer">
+            <a href="#" className="nx-text-link">Link</a>
+          </footer>
+        </section>
+        <section className="nx-card" aria-label="card 2 in row">
```
rpokorny, line 106: Many of the `aria-label`s from here down no longer make sense

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -8,21 +8,22 @@ import React from 'react';
...
+    NxCardCustomSCSSCode = require('!!raw-loader!./NxCardPageExamples.scss').default;
 
 const NxCardPage = () =>
   <>
     <GalleryDescriptionTile>
       <p className="nx-p">
-        Cards can be displayed in two orientations: rows and columns. Cards can be transformed between rows and columns
-        with a simple class change on their container. There is also a row variant supplied which displays the call out
-        and content side-by-side, this variant cannot be transformed into column oriented cards.
+        Cards can be displayed in two orientations: vertically or horizontally. At its most basic a card is simply a
```
rpokorny, line 22: Various places in the docs need to be updated to remove horizontal card and column container mentions

gallery/src/styles/NxCard/NxCardPage.tsx
```diff
@@ -37,31 +38,33 @@ const NxCardPage = () =>
...
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-card</code></td>
-            <td className="nx-cell">Element</td>
+            <td className="nx-cell"><code className="nx-code">.nx-card--equal</code></td>
+            <td className="nx-cell">Modifier of <code className="nx-code">.nx-card</code></td>
             <td className="nx-cell">
-              Basic card element.
+              <code className="nx-code">.nx-card</code>s have a default <code className="nx-code">min-width</code>
```
rpokorny, line 64: This class doesn't currently remove the `min-width`

lib/src/base-styles/_nx-card.scss
```diff
@@ -6,176 +6,74 @@
...
-  .nx-icon--card-call-out-xl {
-    font-size: 64px;
-  }
+.nx-card__header {
+  @include container-vertical;
+  margin-bottom: $nx-spacing-md;
 }
 
-.nx-card-container--row {
+.nx-card__content {
```
rpokorny, line 39: I think this should have `container-vertical`.  Look at the margins in the nx-read-only example to see why

lib/src/base-styles/_nx-card.scss
```diff
@@ -6,176 +6,74 @@
...
+  flex-direction: column;
+  flex-grow: 1;
+  gap: $nx-spacing-md;
 }
 
-.nx-card-container--column {
-  @include container-vertical;
+.nx-card__footer {
+  border-top: $nx-border-default;
+  margin-top: $nx-spacing-l;
```
rpokorny, line 49: IMO this margin would be better placed on the bottom of `nx-card__content`.  As it is here, if you for some reason had a card with only a header and a footer, the margins of the two double up in the middle.

lib/src/base-styles/_nx-card.scss
```diff
@@ -6,176 +6,74 @@
...
-      }
-    }
-
-    .nx-card__content {
-      flex-grow: 1;
-      margin-top: 0;
-      text-align: left;
-    }
-  }
+.nx-card__content--2-column {
```
rpokorny, line 77: I'm mildly concerned about the naming of this class in light of the fact that we explicitly support using other things as the `.nx-card__content` children besides `.nx-card__callout` and `.nx-card__text`.  It's only really a 2-column layout when the `.nx-card__content` happens to have two children.

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > This class doesn't currently remove the `min-width`

I did that on purpose. I didn't think that it was a good idea to let the cards get crunched below their minimum

lib/src/base-styles/_nx-card.scss
glenhunter: > I'm mildly concerned about the naming of this class in light of the fact that we explicitly support using other things as the `.nx-card__content` children besides `.nx-card__callout` and `.nx-card__text`. It's only really a 2-column layout when the `.nx-card__content` happens to have two children.

I'm not sure where the concern lays. Are you saying that it's named poorly, or that you'd like to see an example that uses something other than call-out and text?

Technically it would support a number of columns equal to the number of children of `content`, should we describe that instead?

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > Various places in the docs need to be updated to remove horizontal card and column container mentions

I only found this one (stupidly obvious) instance that I missed: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/cccad606804a64d48ec59e77a1a47db856ef371f

lib/src/base-styles/_nx-card.scss
glenhunter: > IMO this margin would be better placed on the bottom of `nx-card__content`. As it is here, if you for some reason had a card with only a header and a footer, the margins of the two double up in the middle.

Unlikely but worth doing: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/14b0b0f44eacfc4d9dfa3f9b703f2a046ee0ef63

lib/src/base-styles/_nx-card.scss
glenhunter: > I think this should have `container-vertical`. Look at the margins in the nx-read-only example to see why

Yes: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/14b0b0f44eacfc4d9dfa3f9b703f2a046ee0ef63

gallery/src/styles/NxCard/NxCardLayoutExample.tsx
glenhunter: > Many of the `aria-label`s from here down no longer make sense

Cleaned up here: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/7db7311c2ebc66b585cbd3c79f11add0ec040e84

lib/src/base-styles/_nx-card.scss
glenhunter: > > I'm mildly concerned about the naming of this class in light of the fact that we explicitly support using other things as the `.nx-card__content` children besides `.nx-card__callout` and `.nx-card__text`. It's only really a 2-column layout when the `.nx-card__content` happens to have two children.
> 
> I'm not sure where the concern lays. Are you saying that it's named poorly, or that you'd like to see an example that uses something other than call-out and text?
> 
> Technically it would support a number of columns equal to the number of children of `content`, should we describe that instead?

I changed it to the more generic `--columns`: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/182df9d6645a4b1d575b7f9113a29399bc8c5e83

gallery/src/styles/NxCard/NxCardLayoutExample.tsx
rpokorny: > Cleaned up here: 7db7311

That's an improvement but the numbers assigned to each card ("card 7" etc) still don't seem to make sense; they neither match the cards index in its container nor its index overall

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > I only found this one (stupidly obvious) instance that I missed: cccad60

I thought I had seen one in the class table but maybe you already got it, can't find it now.

gallery/src/styles/NxCard/NxCardPage.tsx
rpokorny: > I did that on purpose. I didn't think that it was a good idea to let the cards get crunched below their minimum

The way this description is written implies (to me) that both the min and max are removed by this class

gallery/src/styles/NxCard/NxCardLayoutExample.tsx
```diff
@@ -0,0 +1,300 @@
...
+        </section>
+        <section className="nx-card" aria-label="Server card">
+          <header className="nx-card__header">
+            <h3 className="nx-h3">Card header</h3>
+          </header>
+          <div className="nx-card__content">
+            <div className="nx-status-indicator">
+              <span>Server one</span>
+            </div>
+            <div className="nx-status-indicator">
```
rpokorny, line 98: Probably not a good idea to be using currently made-up but soon-to-be-real `nx-` classes in the example.

rpokorny (approved):
+1 aside from small outstanding comments

gallery/src/styles/NxCard/NxCardPage.tsx
glenhunter: > > I did that on purpose. I didn't think that it was a good idea to let the cards get crunched below their minimum
> 
> The way this description is written implies (to me) that both the min and max are removed by this class

Re-written: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/e8c4351e9005851c43777f251641c1bdbf6a112e

gallery/src/styles/NxCard/NxCardLayoutExample.tsx
glenhunter: > Probably not a good idea to be using currently made-up but soon-to-be-real `nx-` classes in the example.

I just converted them to `<p className="nx-p">`: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/6ec8ec9daf8d17faea5f865409f9ec997663e2bd

gallery/src/styles/NxCard/NxCardLayoutExample.tsx
glenhunter: > > Cleaned up here: [7db7311](https://github.com/sonatype/sonatype-react-shared-components/commit/7db7311c2ebc66b585cbd3c79f11add0ec040e84)
> 
> That's an improvement but the numbers assigned to each card ("card 7" etc) still don't seem to make sense; they neither match the cards index in its container nor its index overall

Everything has a better `aria-label` now: https://github.com/sonatype/sonatype-react-shared-components/pull/307/commits/6ec8ec9daf8d17faea5f865409f9ec997663e2bd

PR #310, opened 2021-02-11
NxCodeSnippet - RSC-402

rpokorny:
https://issues.sonatype.org/browse/RSC-402

In addition to adding NxCodeSnippet, this PR has a lot of adjustments to the visual testing setup which were required in order to get the clipboard tests to work:

1. In order to have access to the clipboard in javascript on a non-https web page I have to add the `--unsafely-treat-insecure-origin-as-secure` to the visual testing browser instances.
2. Chrome has a bug where that flag has no effect when run in headless mode, which is what we were doing. See https://bugs.chromium.org/p/chromium/issues/detail?id=1176255
3. Removing the headless flag broke many visual tests, as we run tests in parallel and a single non-headless browser is going to be sharing a single global mouse position between concurrent tests, causing them to affect each other
4. Removing the parallelism was a non-starter since it caused our CI to take about 40 minutes
5. Therefore the solution was to start up a whole "selenium grid" rather than just a single selenium server.  This grid consists of a "hub" in one docker container, which receives the webdriver requests from the tests, and then multiple "node" docker containers which each run a single browser session and execute tests one at a time.  The level of parallelism is determined by the number of nodes (which I've increased to 10 by the way, our parallelism factor was previously 4).
6. In order to get selenium hub working I had to upgrade to a new major version of the selenium docker images
7. This upgrade revealed some issues in other visual tests, where their state was based on other tests having run before.  These tests are also fixed in this PR.


gallery/visualtests/nxDropdown.js
```diff
@@ -27,10 +27,8 @@ describe('NxDropdown', function() {
     beforeEach(async function() {
       const button = await browser.$(defaultSelector + ' .nx-dropdown__toggle');
 
-      // for some weird reason the test infra requires you to click it twice
```
rpokorny, line 30: This went away with the selenium upgrade

gallery/visualtests/nxTextInput.js
```diff
@@ -134,11 +134,11 @@ describe('NxTextInput', function() {
   });
 
   describe('Long NxTextInput', function() {
-    it('looks right', simpleTest(textareaComponentSelector));
+    it('looks right', simpleTest(longComponentSelector));
```
rpokorny, line 137: These selectors were just straight up wrong, and I happened to notice due to the applitools diffs that came from the selenium upgrade.

lib/src/components/NxCodeSnippet/NxCodeSnippet.scss
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-variables';
+.nx-code-snippet {
```
Blacktiger, line 8: I think this component is missing some margin - two of them next to each other don't have any space in between.

gallery/src/components/NxCodeSnippet/NxCodeSnippetPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+                          id="nx-code-snippet-example"
+                          liveExample={NxCodeSnippetComplexExample}
+                          codeExamples={nxCodeSnippetComplexCode}>
+        An <code className="nx-code">NxCodeSnippet</code> demonstrating optional
+        props: <code className="nx-code">sublabel</code> and <code className="nx-code">onCopyUsingBtn</code>.
+      </GalleryExampleTile>
+    </>
+  );
+}
+
```
Blacktiger, line 114: We might consider adding a test that has two of these next to each other since I know that will happen in nxrm in more than one location.

lib/src/components/NxCodeSnippet/NxCodeSnippet.tsx
```diff
@@ -0,0 +1,78 @@
...
+      // document.execCommand works outside of https, but is deprecated and might be deactivated some day
+      copyWithExecCommand();
+    }
+  }
+
+  return (
+    <div className={classes} { ...otherProps }>
+      <NxButton type="button" variant="tertiary" onClick={copyToClipboard}>Copy to Clipboard</NxButton>
+      <NxFormGroup label={label} sublabel={sublabel}>
+        <NxTextInput ref={textInputRef} type="textarea" value={content} isPristine={true} readOnly />
```
Blacktiger, line 72: Since I don't see any way to pass in props to this NxTextInput element I'm thinking if I wanted to adjust the size of the textarea I'd end up using css to select it - it might be nice to ensure that we can pass the rows prop through maybe?

lib/src/components/NxCodeSnippet/NxCodeSnippet.tsx
```diff
@@ -0,0 +1,78 @@
...
+        if (copySuccessful && onCopyUsingBtn) {
+          onCopyUsingBtn();
+        }
+      }
+    }
+  }
+
+  function copyToClipboard() {
+    // the clipboard object is the modern API, but it is only available in secure contexts (ie https or localhost)
+    if (window.navigator.clipboard) {
```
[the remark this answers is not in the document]
rpokorny: > On http window.navigator.clipboard is false-like? I would have guessed that it would be there but throw an exception or something. I assume you already tested this, but since it didn't match my expectations I figured I'd note it here.

Yeah the entire object is missing when in an insecure context

lib/src/components/NxCodeSnippet/NxCodeSnippet.scss
rpokorny: > I think this component is missing some margin - two of them next to each other don't have any space in between.

Added margin matching `NxFormGroup` 31c19012

Blacktiger (approved):
Going to leave an approval on what's here as I can probably use it as-is and I'll be out half of next week.

gallery/src/components/NxCodeSnippet/NxCodeSnippetPage.tsx
rpokorny: > We might consider adding a test that has two of these next to each other since I know that will happen in nxrm in more than one location.

8b33296b

lib/src/components/NxCodeSnippet/NxCodeSnippet.tsx
rpokorny: > Since I don't see any way to pass in props to this NxTextInput element I'm thinking if I wanted to adjust the size of the textarea I'd end up using css to select it - it might be nice to ensure that we can pass the rows prop through maybe?

934da191

Blacktiger (approved):

lib/src/components/NxCodeSnippet/NxCodeSnippet.scss
```diff
@@ -0,0 +1,53 @@
...
+  align-items: end;
+  display: grid;
+  grid-template:
+    "label    copy-btn" 1fr
+    "sublabel copy-btn" auto
+    "textarea textarea" auto /
+      1fr      auto;
+
```
glenhunter, line 16: Speaking for myself I really dislike this shorthand. I'd prefer to see this broken out into `grid-template-areas`, `grid-template-rows`, and `grid-template-columns` for clarity.

glenhunter (approved):

lib/src/components/NxCodeSnippet/NxCodeSnippet.scss
rpokorny: > Speaking for myself I really dislike this shorthand. I'd prefer to see this broken out into grid-template-areas, grid-template-rows, and grid-template-columns for clarity.

Honestly I find it clearest this way, at least when using line breaks in the way that I have.  To me it makes it easy to visualize the template spec as a table, where for instance the `1fr` at the bottom can be easily understood as the width of the column above it.  When people use this syntax but don't align things though it can get very confusing.

glenhunter:
@dturner32 looking at the Zelin files I see that the spacing for this component does not match the spacing for other field elements. Specifically the header and sub-header spacing. This component likely will end up in a form field so I think that probably it should have the same spacing values.

dturner32:
> @dturner32 looking at the Zelin files I see that the spacing for this component does not match the spacing for other field elements. Specifically the header and sub-header spacing. This component likely will end up in a form field so I think that probably it should have the same spacing values.

so Ross and I actually talked about this and how this particular component might deviate slightly from the current standard. Reason being is due to the button padding being higher than is the standard for the header and sub-header. This results in the alignment of the text and the button looking off and rather odd-looking. What I mentioned to Ross as, for this case, is that the spacing between the header/sub-header and the text box be increased to more accurately match the height of the button

dturner32:
@glenhunter chatted with Ross again and decided to adjust the button padding instead of the header/sub-header padding, so I will update that in the mockups/zeplin and let you guys know when that change has been made

dturner32:
@rpokorny @glenhunter just updated the zeplin mockup

rpokorny:
> @rpokorny @glenhunter just updated the zeplin mockup

@dturner32 that's still inconsistent with how labels usually are: the sublabel is usually 16px from the input, but you have it at 8px there

dturner32:
> > @rpokorny @glenhunter just updated the zeplin mockup
> 
> @dturner32 that's still inconsistent with how labels usually are: the sublabel is usually 16px from the input, but you have it at 8px there

16 is correct, I may have moved some things while I was updating and just didn't move it back. Will update

dturner32:
> > > @rpokorny @glenhunter just updated the zeplin mockup
> > 
> > 
> > @dturner32 that's still inconsistent with how labels usually are: the sublabel is usually 16px from the input, but you have it at 8px there
> 
> 16 is correct, I may have moved some things while I was updating and just didn't move it back. Will update

updated in zeplin

glenhunter (approved):
I think that looks much better overall

PR #311, opened 2021-02-17
nx-cards with uniform structure - RSC-467

rpokorny:
Changes to make nx-card structure the same in both vertical and horizontal cards.  Re this thread: https://github.com/sonatype/sonatype-react-shared-components/pull/307#discussion_r576901741


rpokorny:
Horizontal cards have been removed in the parent PR so this is no longer relevant

PR #312, opened 2021-02-18
useToggle - Improvement Day

rpokorny:
There have been numerous (> 50 it seems) places in the examples where we have had a piece of boolean react state that we want to do nothing but toggle back and forth.  This has required, in each instance, both the `useState` invocation and then an additional function that wraps the actual toggling along with a call to the state setter function returned by `useState`.

While this isn't terribly inconvenient, the sheer number of times this comes up made decide to streamline it a bit by adding a new hook: `useToggle`.

`useToggle` simply wraps `useState` and create the toggling wrapper function itself so that consumers don't have to.


glenhunter:
I think the example page should use RSC components.

rpokorny:
> I think the example page should use RSC components.

I considered that but deliberately avoided it to ensure that it was clear that this hook isn't some magic that's only compatible with RSC components

glenhunter:
> > I think the example page should use RSC components.
> 
> I considered that but deliberately avoided it to ensure that it was clear that this hook isn't some magic that's only compatible with RSC components

I'd leave it up to a developer to understand whether or not it is magic that can exist outside the RSC. To me it looks like we've made a mistake. If you strongly feel that it shouldn't use RSC components then I think that you should call that out in the examples.

rpokorny:
> I'd leave it up to a developer to understand whether or not it is magic that can exist outside the RSC. To me it looks like we've made a mistake. If you strongly feel that it shouldn't use RSC components then I think that you should call that out in the examples.

de1aa8cc

glenhunter (approved):

PR #313, opened 2021-02-18
Colour documentation RSC-143

glenhunter:
https://issues.sonatype.org/browse/RSC-143

Screen Shot 2021-02-18 at 1 34 25 PM
Screen Shot 2021-02-18 at 1 34 33 PM


rpokorny:
There's also `$nx-white`, should that be documented?  It is still used but I guess didn't make it into any of the swatches due to its lack of hue

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "3.7.0",
+  "version": "3.8.0",
```
rpokorny, line 3: This is just a documentation update so should be a patch version

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React from 'react';
+
+import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
+import Swatcher from './ColorPaletteExample';
+
+const ColorPalettePage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        The design group have created a palette of colors for use in RSC. The colors are organized into swatches
```
rpokorny, line 16:
```suggestion
        The design group has created a palette of colors for use in RSC. The colors are organized into swatches
```

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
```diff
@@ -0,0 +1,57 @@
...
+        available in RSC along with their hex value and corresponding RSC variable.
+      </p>
+      <section className="nx-tile-subsection">
+        <header className="nx-tile-subsection__header">
+          <h3 className="nx-h3">Color Variables</h3>
+        </header>
+        <p className="nx-p">
+          Where a color is used commonly across multiple components a custom variable is created. For example many
+          components with a disabled state have the same background color so we have created a color variable:
+          {' '}<code className="nx-code">$nx-disabled-background: $nx-grey-200;</code>. If the disabled background color
```
rpokorny, line 31: I'm not sure we should mention this unless we are going to document what additional variables are supported.  Especially since `_nx-colors.scss` currently contains a mix of up-to-date semantic color variables, and deprecated color variables that aren't part of the current system and which we are theoretically working on removing.

gallery/src/guidelines/ColorPalette/Swatcher.scss
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ @import '~@sonatype/react-shared-components/scss-shared/nx-colors';
```
rpokorny, line 7: Redundant; `nx-variables` imports this

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <h3 className="nx-h3">Using Colors</h3>
+        </header>
+        <p className="nx-p">
+          It is expected that designers and consumers of the RSC will limit themselves to using the colors that have
+          been made available in <code className="nx-code">_nx-color-swatches.scss</code>. It is preferable to use the
+          {' '}<code className="nx-code">$color-###</code> variable rather than the hex value. Doing so makes it easier
+          to replace colors in the future if the palette changes.
+        </p>
+      </section>
+    </GalleryDescriptionTile>
```
rpokorny, line 48: Do we want to describe the preferred way of importing the color variables?

gallery/src/guidelines/ColorPalette/Swatcher.scss
```diff
@@ -0,0 +1,43 @@
...
+ @import '~@sonatype/react-shared-components/scss-shared/nx-variables';
+ @import '~@sonatype/react-shared-components/scss-shared/nx-text-helpers';
+
+.gallery-swatches {
```
rpokorny, line 11: I wonder if these would be good candidates to make as `.nx-card`s

gallery/src/guidelines/ColorPalette/Swatcher.scss
```diff
@@ -0,0 +1,43 @@
...
+  height: 64px;
+  width: 64px;
+}
+
+.gallery-swatch__variable, .gallery-swatch__hex {
+  @include semi-bold();
+}
+
+.gallery-swatch__hex {
+  flex-basis: 7ch;
```
rpokorny, line 41: What an odd choice of unit.  Why?

rpokorny: > What an odd choice of unit. Why?

Oh I guess because a full width color code (assuming no transparency) is 7 characters wide? Keep in mind that's only precise in a fixed-width font

gallery/src/guidelines/ColorPalette/Swatcher.scss
glenhunter: > I wonder if these would be good candidates to make as `.nx-card`s

You know I did think about that, I didn't just because the `NxCard` PR is still in review. We could hold on this until it's merged.

gallery/src/guidelines/ColorPalette/Swatcher.scss
glenhunter: > > What an odd choice of unit. Why?
> 
> Oh I guess because a full width color code (assuming no transparency) is 7 characters wide? Keep in mind that's only precise in a fixed-width font

yeah, seemed to work well here though

gallery/src/guidelines/ColorPalette/Swatcher.scss
glenhunter: > Redundant; `nx-variables` imports this

Ooops: https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/2add318c9ea7af6b13b90cacc20d0452876b0014

gallery/package.json
glenhunter: > This is just a documentation update so should be a patch version

De-bumped: https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/d03f60260740793e3b6c4fd01e41d55ead98257f

glenhunter:
> There's also `$nx-white`, should that be documented? It is still used but I guess didn't make it into any of the swatches due to its lack of hue

I added black and white here: https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/d3882e547f5086a0d86f06cc4cead2c0b3de18c5

When I have the cards set up we can see how it works and whether we want to keep them. I suspect we won't want to

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
glenhunter: > Do we want to describe the preferred way of importing the color variables?

Good idea: https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/b40c16b8122633991ea811a49a9bd48050d5abfd

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
glenhunter: > I'm not sure we should mention this unless we are going to document what additional variables are supported. Especially since `_nx-colors.scss` currently contains a mix of up-to-date semantic color variables, and deprecated color variables that aren't part of the current system and which we are theoretically working on removing.

My goal was to explain why sometimes we use `$nx-disabled-background` instead of `$nx-grey-200` (for example), I agree that we don't want to list all of the named variables, but I do think that the explanation has value.

rpokorny: > My goal was to explain why sometimes we use $nx-disabled-background instead of $nx-grey-200 (for example), I agree that we don't want to list all of the named variables, but I do think that the explanation has value.

I'm worried that it prompts users to use colors from `_nx-colors.scss` and that's not currently a safe idea.

rpokorny:
> I added black and white here: d3882e5
>
> When I have the cards set up we can see how it works and whether we want to keep them. I suspect we won't want to

You know those aren't getting included in the gallery page, right?

lib/src/scss-shared/_nx-color-swatches.scss
```diff
@@ -150,3 +150,9 @@ $nx-indigo-300: #bbc5f6;
 $nx-indigo-200: #cad1f3;
 $nx-indigo-100: #d1d6f0;
 $nx-indigo-50: #e7eaf9;
+
+// Black
+$nx-black: #000;
+
+// White
+$nx-white: #fff;
```
rpokorny, line 158: `$nx-white` and `$nx-black` were already present in `_nx-colors.scss`, so if you add them here you should remove them there.  I would think it wouldn't compile with them present in both places, actually.

glenhunter: > `$nx-white` and `$nx-black` were already present in `_nx-colors.scss`, so if you add them here you should remove them there. I would think it wouldn't compile with them present in both places, actually.

Sure it's not problem it just overrides the value of the variable based on load order. Regardless I removed them, the more I thought about it the less value that I saw in having just black and white. https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/7d76da37609656f5400fcb600c1dd7586dd77efe

glenhunter:
> > I added black and white here: [d3882e5](https://github.com/sonatype/sonatype-react-shared-components/commit/d3882e547f5086a0d86f06cc4cead2c0b3de18c5)
> > When I have the cards set up we can see how it works and whether we want to keep them. I suspect we won't want to
> 
> You know those aren't getting included in the gallery page, right?

hah, I removed them: https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/7d76da37609656f5400fcb600c1dd7586dd77efe

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
glenhunter: > > My goal was to explain why sometimes we use $nx-disabled-background instead of $nx-grey-200 (for example), I agree that we don't want to list all of the named variables, but I do think that the explanation has value.
> 
> I'm worried that it prompts users to use colors from `_nx-colors.scss` and that's not currently a safe idea.

The problem is that there are many variables in `_color.scss` that we _do_ want people to use rather than a `nx-color-###` variable. I think that this isn't a problem with this documentation but rather a problem with `_colors.scss`. I know that we had a discussion a little while ago about moving some of those variables down to the bottom of the file into a "deprecated" section. Maybe we should up the priority of that task.

rpokorny: > The problem is that there are many variables in _color.scss that we do want people to use rather than a nx-color-### variable. I think that this isn't a problem with this documentation but rather a problem with _colors.scss. I know that we had a discussion a little while ago about moving some of those variables down to the bottom of the file into a "deprecated" section. Maybe we should up the priority of that task.

Right.  But until we clean that up, I think we shouldn't document it in the way you have here.  We can make sure we document the desirable variables later once they are distinguishable.

gallery/src/guidelines/ColorPalette/Swatcher.scss
glenhunter: > > I wonder if these would be good candidates to make as `.nx-card`s
> 
> You know I did think about that, I didn't just because the `NxCard` PR is still in review. We could hold on this until it's merged.

`.nx-card` implementation. I tweaked the `max-width` a bit so that the longer variables didn't wrap.

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
glenhunter: > > The problem is that there are many variables in _color.scss that we do want people to use rather than a nx-color-### variable. I think that this isn't a problem with this documentation but rather a problem with _colors.scss. I know that we had a discussion a little while ago about moving some of those variables down to the bottom of the file into a "deprecated" section. Maybe we should up the priority of that task.
> 
> Right. But until we clean that up, I think we shouldn't document it in the way you have here. We can make sure we document the desirable variables later once they are distinguishable.

Removed here and I created: https://issues.sonatype.org/browse/RSC-496

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "4.1.0",
+  "version": "4.1.10",
```
rpokorny, line 3: .10??

rpokorny:
Minor: I'm finding it a little strange that the cards aren't all the same width.  Probably not worth worrying about though. 
image

rpokorny:
> > > I added black and white here: d3882e5
> > > When I have the cards set up we can see how it works and whether we want to keep them. I suspect we won't want to
> > 
> > You know those aren't getting included in the gallery page, right?
>
> hah, I removed them: 7d76da3

Ok, well that brings us back to the beginning of the discussion: https://github.com/sonatype/sonatype-react-shared-components/pull/313#issuecomment-782181260

gallery/src/guidelines/ColorPalette/ColorPalettePage.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <h3 className="nx-h3">Using Colors</h3>
+        </header>
+        <p className="nx-p">
+          It is expected that designers and consumers of the RSC will limit themselves to using the colors that have
+          been made available in <code className="nx-code">_nx-color-swatches.scss</code>. It is preferable to use the
+          {' '}<code className="nx-code">$color-###</code> variable rather than the hex value. Doing so makes it easier
+          to replace colors in the future if the palette changes.
+        </p>
+        <p className="nx-p">
+          To access to the color variables you'll need to load
```
rpokorny, line 35:
```suggestion
          To access the color variables you'll need to load
```

glenhunter:
> > > > I added black and white here: [d3882e5](https://github.com/sonatype/sonatype-react-shared-components/commit/d3882e547f5086a0d86f06cc4cead2c0b3de18c5)
> > > > When I have the cards set up we can see how it works and whether we want to keep them. I suspect we won't want to
> > > 
> > > 
> > > You know those aren't getting included in the gallery page, right?
> > 
> > 
> > hah, I removed them: [7d76da3](https://github.com/sonatype/sonatype-react-shared-components/commit/7d76da37609656f5400fcb600c1dd7586dd77efe)
> 
> Ok, well that brings us back to the beginning of the discussion: [#313 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/313#issuecomment-782181260)

I don't think that we need to show black or white here.

glenhunter:
> Minor: I'm finding it a little strange that the cards aren't all the same width. Probably not worth worrying about though.

Ah, just adding `flex-grow:1` to the cards made it nearly perfect while still respecting the `max-width` and working responsively.

gallery/package.json
glenhunter: > .10??

Weird: https://github.com/sonatype/sonatype-react-shared-components/pull/313/commits/ac12d51610a479191198d08428b95d92cf96c7c9

rpokorny (approved):

PR #316, opened 2021-02-19
SidebarNav component - RSC-485

raveneyex:
[Jira](https://issues.sonatype.org/browse/RSC-485)
[Zeplin](https://zpl.io/bL6QeyQ)

Putting up this _draft_ PR to get feedback before I start investing into writing the component tests. I'm specifically looking for feedback regarding the SCSS and the API of the component.
Also note that I've hijacked the main gallery page — This is just a quick & dirty _temporary_ change meant to provide a playground for the new sidebar. Working on actually creating an example of this component in the gallery is an entirely different ticket.

### Expanded
Expanded

### Collapsed
Collapsed2

### DIFF
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -76,9 +76,7 @@ function Application() {
             <p className="example-attribution">Powered by Nexux IQ Server</p>
             <p className="example-attribution">Created by Sonatype</p>
           </div>
-          <div className="nx-sidebar__collapsed-content">
-            release 105
-          </div>
+          <p className="nx-sidebar__collapsed-content">release 105</p>
         </NxNavigationSidebarContent>
       </NxStatefulNavigationSidebar>
       <div className="nx-page-main">
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -4,92 +4,82 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React, { useEffect } from 'react';
-import { RouteChildrenProps } from 'react-router';
-import { HashRouter as Router, Route, Redirect, Switch } from 'react-router-dom';
-import { mergeAll, values } from 'ramda';
-
-// polyfill Array.prototype.includes which is used in query-string
-import 'core-js/features/array/includes';
-import queryString from 'query-string';
-
-import { PageMapping } from './pageConfigTypes';
-import pageConfig from './pageConfig';
-import PageHeader from './PageHeader';
-import GalleryNav from './GalleryNav';
-import Home from './pages/Home';
-import handleQueryParams from './handleQueryParams';
-
-import NxLoadWrapperPageLevelExample from './components/NxLoadWrapper/NxLoadWrapperPageLevelExample';
-import NxViewportSizedExample from './styles/NxViewportSized/NxViewportSizedExample';
-import NxViewportSizedExpandingExample
-  from './styles/NxViewportSized/NxViewportSizedExpandingExample';
-import SectionScrollingWrapper from './styles/NxViewportSized/SectionScrollingWrapper';
-
-const pageMappings: PageMapping = mergeAll(values(pageConfig));
-
-function Page({ match, location }: RouteChildrenProps<{ pageName: string }>) {
-  const pageName = match ? match.params.pageName : null,
-      pageHeader = pageName || 'Home',
-      Content = pageName ? pageMappings[pageName] : Home;
-
-  useEffect(function() {
-    handleQueryParams(queryString.parse(location.search));
-  }, [location.search]);
-
-  if (Content) {
-    // Put a key on <main> so that it re-renders entirely on route change, resetting scroll position
-    return (
-      <main key={pageName || 'home'} className="nx-page-main">
-        <div className="nx-page-title">
-          <h1 className="nx-h1">
-            {pageHeader}
-          </h1>
-        </div>
-        <Content/>
-      </main>
-    );
-  }
-  else {
-    // unknown page
-    return <Redirect to="/" />;
-  }
-}
+import React from 'react';
+import {
+  faAtom,
+  faBiohazard,
+  faBolt,
+  faCrow,
+  faPlaceOfWorship
+} from '@fortawesome/free-solid-svg-icons';
+import { NxStatefulSidebarNavigation } from '@sonatype/react-shared-components';
 
 function Application() {
-  return (
-    <Router>
-      <PageHeader />
-      <div className="nx-page-content">
-        <Switch>
-          <Route exact path="/PageLevelAlertExample">
-            <NxLoadWrapperPageLevelExample/>
-          </Route>
-          <Route>
-            <aside className="nx-page-sidebar" id="gallery-sidebar">
-              <GalleryNav />
-            </aside>
-            <Switch>
-              <Route path="/pages/:pageName" component={Page} />
-              <Route exact path="/" component={Page} />
+  const navLinks = [
+    {
+      name: 'Link Name Alpha',
+      href: '/alpha',
+      icon: faAtom,
+      current: true
+    }, {
+      name: 'Link Name Beta',
+      href: '/beta',
+      icon: faBiohazard
+    }, {
+      name: 'Link Name Gamma',
+      href: '/gamma',
+      icon: faBolt
+    }, {
+      name: 'Link Name Delta',
+      href: '/delta',
+      icon: faCrow
+    }, {
+      name: 'Link Name Epsilon',
+      href: '/epsilon',
+      icon: faPlaceOfWorship
+    }, {
+      name: 'Link Name Zeta',
+      href: '/zeta',
+      icon: faAtom
+    }, {
+      name: 'Link Name Etha',
+      href: '/etha',
+      icon: faPlaceOfWorship
+    }, {
+      name: 'Link Name Theta',
+      href: '/theta',
+      icon: faBiohazard
+    }
+  ];
+  const attributions = [
+    'Powered by Nexux IQ Server',
+    'Created by Sonatype'
+  ];
+  const logoImg = require('./resources/logos/SON_hexagon_cropped.svg');
 
-              {/* Special cases, these examples need their own page separate from their documentation */}
-              <Route exact path="/NxViewportSizedExample">
-                <SectionScrollingWrapper>
-                  <NxViewportSizedExample />
-                </SectionScrollingWrapper>
-              </Route>
-              <Route exact path="/NxViewportSizedExpandingExample">
-                <SectionScrollingWrapper>
-                  <NxViewportSizedExpandingExample />
-                </SectionScrollingWrapper>
-              </Route>
-              <Redirect to="/" />
-            </Switch>
-          </Route>
-        </Switch>
+  return (
+    <>
+      <NxStatefulSidebarNavigation isDefaultOpen={true}
+                                   logoImg={logoImg}
+                                   logoText="nexus lifecycle"
+                                   logoLink="/home"
+                                   links={navLinks}
+                                   helpLink="http://google.com"
+                                   helpText="google help"
+                                   collapsedReleaseText="release 105"
+                                   expandedReleaseText="lifecycle release 105"
+                                   attributions={attributions} />
+      <div className="nx-page-content">
+        <div className="nx-page-main">
+          <p>
+            Cupidatat pariatur mollit sit pariatur mollit qui.{' '}
+            Est sint sunt ea laborum minim officia excepteur nostrud deserunt proident officia.{' '}
+            Ipsum nostrud do ipsum nisi occaecat reprehenderit aliquip in occaecat culpa consequat pariatur et.{' '}
+            Non adipisicing qui officia nisi do aliquip.
+          </p>
+        </div>
       </div>
-    </Router>
+    </>
   );
 }
 
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -10,9 +10,14 @@ import {
   faBiohazard,
   faBolt,
   faCrow,
-  faPlaceOfWorship
+  faPlaceOfWorship,
+  faQuestionCircle
 } from '@fortawesome/free-solid-svg-icons';
-import { NxStatefulNavigationSidebar } from '@sonatype/react-shared-components';
+import {
+  NxStatefulNavigationSidebar,
+  NxNavigationSidebarContent,
+  NxFontAwesomeIcon
+} from '@sonatype/react-shared-components';
 
 function Application() {
   const navLinks = [
@@ -51,24 +56,35 @@ function Application() {
       icon: faBiohazard
     }
   ];
-  const attributions = [
-    'Powered by Nexux IQ Server',
-    'Created by Sonatype'
-  ];
   const logoImg = require('./resources/logos/SON_hexagon_cropped.svg');
 
   return (
     <div className="nx-page-content nx-page-content--full-width">
       <NxStatefulNavigationSidebar isDefaultOpen={true}
+                                   toggleOpenIcon={faBiohazard}
+                                   toggleCloseIcon={faCrow}
                                    logoImg={logoImg}
                                    logoText="nexus lifecycle"
                                    logoLink="/home"
-                                   links={navLinks}
-                                   helpLink="http://google.com"
-                                   helpText="google help"
-                                   collapsedReleaseText="release 105"
-                                   expandedReleaseText="lifecycle release 105"
-                                   attributions={attributions} />
+                                   links={navLinks}>
+        <NxNavigationSidebarContent>
+          <a rel="noreferrer"
+             target="_blank"
+             className="nx-sidebar__help-link"
+             href="www.google.com">
+            <NxFontAwesomeIcon icon={faQuestionCircle}/>
+            <span className="nx-sidebar__expanded-content">Help and Support</span>
+          </a>
+          <div className="nx-sidebar__expanded-content">
+            <p>lifecycle release 105</p>
+            <p>Powered by Nexux IQ Server</p>
+            <p>Created by Sonatype</p>
+          </div>
+          <div className="nx-sidebar__collapsed-content">
+            release 105
+          </div>
+        </NxNavigationSidebarContent>
+      </NxStatefulNavigationSidebar>
       <div className="nx-page-main">
         <p>
           Cupidatat pariatur mollit sit pariatur mollit qui.{' '}
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -12,7 +12,7 @@ import {
   faCrow,
   faPlaceOfWorship
 } from '@fortawesome/free-solid-svg-icons';
-import { NxStatefulSidebarNavigation } from '@sonatype/react-shared-components';
+import { NxStatefulNavigationSidebar } from '@sonatype/react-shared-components';
 
 function Application() {
   const navLinks = [
@@ -58,8 +58,8 @@ function Application() {
   const logoImg = require('./resources/logos/SON_hexagon_cropped.svg');
 
   return (
-    <>
-      <NxStatefulSidebarNavigation isDefaultOpen={true}
+    <div className="nx-page-content nx-page-content--full-width">
+      <NxStatefulNavigationSidebar isDefaultOpen={true}
                                    logoImg={logoImg}
                                    logoText="nexus lifecycle"
                                    logoLink="/home"
@@ -69,17 +69,15 @@ function Application() {
                                    collapsedReleaseText="release 105"
                                    expandedReleaseText="lifecycle release 105"
                                    attributions={attributions} />
-      <div className="nx-page-content">
-        <div className="nx-page-main">
-          <p>
-            Cupidatat pariatur mollit sit pariatur mollit qui.{' '}
-            Est sint sunt ea laborum minim officia excepteur nostrud deserunt proident officia.{' '}
-            Ipsum nostrud do ipsum nisi occaecat reprehenderit aliquip in occaecat culpa consequat pariatur et.{' '}
-            Non adipisicing qui officia nisi do aliquip.
-          </p>
-        </div>
+      <div className="nx-page-main">
+        <p>
+          Cupidatat pariatur mollit sit pariatur mollit qui.{' '}
+          Est sint sunt ea laborum minim officia excepteur nostrud deserunt proident officia.{' '}
+          Ipsum nostrud do ipsum nisi occaecat reprehenderit aliquip in occaecat culpa consequat pariatur et.{' '}
+          Non adipisicing qui officia nisi do aliquip.
+        </p>
       </div>
-    </>
+    </div>
   );
 }
 
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -20,54 +20,48 @@ import {
 } from '@sonatype/react-shared-components';
 
 function Application() {
-  const navLinks = [
-    {
-      name: 'Link Name Alpha',
-      href: '/alpha',
-      icon: faAtom,
-      current: true
-    }, {
-      name: 'Link Name Beta',
-      href: '/beta',
-      icon: faBiohazard
-    }, {
-      name: 'Link Name Gamma',
-      href: '/gamma',
-      icon: faBolt
-    }, {
-      name: 'Link Name Delta',
-      href: '/delta',
-      icon: faCrow
-    }, {
-      name: 'Link Name Epsilon',
-      href: '/epsilon',
-      icon: faPlaceOfWorship
-    }, {
-      name: 'Link Name Zeta',
-      href: '/zeta',
-      icon: faAtom
-    }, {
-      name: 'Link Name Etha',
-      href: '/etha',
-      icon: faPlaceOfWorship
-    }, {
-      name: 'Link Name Theta',
-      href: '/theta',
-      icon: faBiohazard
-    }
-  ];
-  const logoImg = require('./resources/logos/SON_hexagon_cropped.svg');
-
   return (
     <div className="nx-page-content nx-page-content--full-width">
       <NxStatefulNavigationSidebar isDefaultOpen={true}
                                    toggleOpenIcon={faBiohazard}
                                    toggleCloseIcon={faCrow}
-                                   logoImg={logoImg}
                                    logoText="nexus lifecycle"
-                                   logoLink="/home"
-                                   links={navLinks}>
-        <NxNavigationSidebarContent>
+                                   logoLink="/home">
+        <NxNavigationSidebarContent navigation>
+          <a href="/alpha" className="selected">
+            <NxFontAwesomeIcon icon={faAtom} fixedWidth />
+            <span>Link Name Alpha</span>
+          </a>
+          <a href="/beta">
+            <NxFontAwesomeIcon icon={faBiohazard} fixedWidth />
+            <span>Link Name Beta</span>
+          </a>
+          <a href="/gamma">
+            <NxFontAwesomeIcon icon={faPlaceOfWorship} fixedWidth />
+            <span>Link Name Gamma</span>
+          </a>
+          <a href="/delta">
+            <NxFontAwesomeIcon icon={faCrow} fixedWidth />
+            <span>Link Name delta</span>
+          </a>
+          <a href="/epsilon">
+            <NxFontAwesomeIcon icon={faBolt} fixedWidth />
+            <span>Link Name Epsilon</span>
+          </a>
+          <a href="/Zeta">
+            <NxFontAwesomeIcon icon={faAtom} fixedWidth />
+            <span>Link Name Zeta</span>
+          </a>
+          <a href="/Etha">
+            <NxFontAwesomeIcon icon={faPlaceOfWorship} fixedWidth />
+            <span>Link Name Etha</span>
+          </a>
+          <a href="/Theta">
+            <NxFontAwesomeIcon icon={faCrow} fixedWidth />
+            <span>Link Name Theta</span>
+          </a>
+        </NxNavigationSidebarContent>
+        <NxNavigationSidebarContent className="example-sidebar-footer">
           <a rel="noreferrer"
              target="_blank"
              className="nx-sidebar__help-link"
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -16,6 +16,7 @@ import {
 import {
   NxStatefulNavigationSidebar,
   NxNavigationSidebarContent,
+  NxNavigationSidebarLinks,
   NxFontAwesomeIcon
 } from '@sonatype/react-shared-components';
 
@@ -27,7 +28,7 @@ function Application() {
                                    toggleCloseIcon={faCrow}
                                    logoText="nexus lifecycle"
                                    logoLink="/home">
-        <NxNavigationSidebarContent navigation>
+        <NxNavigationSidebarLinks>
           <a href="/alpha" className="selected">
             <NxFontAwesomeIcon icon={faAtom} fixedWidth />
             <span>Link Name Alpha</span>
@@ -60,7 +61,7 @@ function Application() {
             <NxFontAwesomeIcon icon={faCrow} fixedWidth />
             <span>Link Name Theta</span>
           </a>
-        </NxNavigationSidebarContent>
+        </NxNavigationSidebarLinks>
         <NxNavigationSidebarContent className="example-sidebar-footer">
           <a rel="noreferrer"
              target="_blank"
--- a/gallery/src/Application.tsx
+++ b/gallery/src/Application.tsx
@@ -22,7 +22,7 @@ import {
 
 function Application() {
   return (
-    <div className="nx-page-content nx-page-content--full-width">
+    <div id="temporary-navigation-sidebar-example" className="nx-page-content nx-page-content--full-width">
       <NxStatefulNavigationSidebar isDefaultOpen={true}
                                    toggleOpenIcon={faBiohazard}
                                    toggleCloseIcon={faCrow}
@@ -62,18 +62,19 @@ function Application() {
             <span>Link Name Theta</span>
           </a>
         </NxNavigationSidebarLinks>
-        <NxNavigationSidebarContent className="example-sidebar-footer">
+        <NxNavigationSidebarContent className="navigation-sidebar-custom-content-1">
           <a rel="noreferrer"
              target="_blank"
-             className="nx-sidebar__help-link"
              href="www.google.com">
-            <NxFontAwesomeIcon icon={faQuestionCircle}/>
+            <NxFontAwesomeIcon fixedWidth icon={faQuestionCircle}/>
             <span className="nx-sidebar__expanded-content">Help and Support</span>
           </a>
+        </NxNavigationSidebarContent>
+        <NxNavigationSidebarContent className="navigation-sidebar-custom-content-2">
           <div className="nx-sidebar__expanded-content">
-            <p>lifecycle release 105</p>
-            <p>Powered by Nexux IQ Server</p>
-            <p>Created by Sonatype</p>
+            <p className="example-release-note">Lifecycle Release 105</p>
+            <p className="example-attribution">Powered by Nexux IQ Server</p>
+            <p className="example-attribution">Created by Sonatype</p>
           </div>
           <div className="nx-sidebar__collapsed-content">
             release 105
--- a/gallery/src/index.html
+++ b/gallery/src/index.html
@@ -14,7 +14,7 @@
     <link rel="stylesheet" href="gallery.css"/>
   </head>
   <body class="nx-body">
-    <div id="ui" class="nx-page"></div>
+    <div id="ui" class="nx-page--sidebar"></div>
     <script src="bundle.js"></script>
   </body>
 </html>
--- a/gallery/src/index.html
+++ b/gallery/src/index.html
@@ -7,14 +7,14 @@
 
 -->
 <!DOCTYPE html>
-<html class="nx-html nx-html--page-scrolling">
+<html class="nx-html">
   <head>
     <meta charset="utf-8">
     <title>Sonatype shared component gallery</title>
     <link rel="stylesheet" href="gallery.css"/>
   </head>
   <body class="nx-body">
-    <div id="ui" class="nx-page nx-page--sidebar"></div>
+    <div id="ui" class="nx-page"></div>
     <script src="bundle.js"></script>
   </body>
 </html>
--- a/gallery/src/index.html
+++ b/gallery/src/index.html
@@ -14,7 +14,7 @@
     <link rel="stylesheet" href="gallery.css"/>
   </head>
   <body class="nx-body">
-    <div id="ui" class="nx-page--sidebar"></div>
+    <div id="ui" class="nx-page nx-page--sidebar"></div>
     <script src="bundle.js"></script>
   </body>
 </html>
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -32,6 +32,14 @@
   min-width: $nx-page-width-min;
 }
 
+.nx-page--sidebar {
+  @include container-vertical;
+
+  display: flex;
+  height: 100%;
+  min-width: $nx-page-width-min;
+}
+
 .nx-page-content {
   @include container-horizontal;
 
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -33,11 +33,10 @@
 }
 
 .nx-page--sidebar {
-  @include container-vertical;
+  @include container-horizontal;
 
-  display: flex;
-  height: 100%;
-  min-width: $nx-page-width-min;
+  align-items: flex-start;
+  flex-direction: row;
 }
 
 .nx-page-content {
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -32,13 +32,6 @@
   min-width: $nx-page-width-min;
 }
 
-.nx-page--sidebar {
-  @include container-horizontal;
-
-  align-items: flex-start;
-  flex-direction: row;
-}
-
 .nx-page-content {
   @include container-horizontal;
 
@@ -54,6 +47,10 @@
     align-self: flex-start;
     margin: $nx-spacing-xl auto 0 auto;
   }
+}
+
+.nx-page-content--full-width{
+  max-width: none;
 }
 
 .nx-page-sidebar {
--- a/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
+++ b/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
@@ -22,7 +22,7 @@ const NxSidebarNavLink = function NxSidebarNavLink({ name, href, icon, current }
 
   return (
     <a href={href} className={navigationLinkClassnames} key={`nav-link-${name}`}>
-      <NxFontAwesomeIcon icon={icon} />
+      <NxFontAwesomeIcon icon={icon} fixedWidth />
       <span className="nx-sidebar__link-name">{name}</span>
     </a>
   );
--- a/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
+++ b/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
@@ -13,7 +13,9 @@ import NxButton from '../NxButton/NxButton';
 import NxCloseButton from '../NxCloseButton/NxCloseButton';
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import { Props, propTypes } from './types';
+
 import './NxSidebarNavitation.scss';
+
 
 const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
   const { children, className, isOpen, helpLink, onToggleOpen } = props;
@@ -21,7 +23,6 @@ const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
     'nx-sidebar-nav__open': isOpen,
     'nx-sidebar-nav__closed': !isOpen
   });
-
   const toggleButtonClasses = 'nx-sidebar-nav--icon nx-sidebar-nav--toggle';
   const closeBtn = (
     <NxCloseButton className={toggleButtonClasses}
@@ -35,10 +36,19 @@ const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
     </NxButton>
   );
   const toggleButton = isOpen ? closeBtn : openBtn;
-
+  const logoImg = require('../../assets/img/SON_hexagon_cropped.svg');
+  const logo = <img src={logoImg} className="nx-sidebar-nav--logo" />;
+  const logoLink = 'http://google.com';
+  
   return (
     <nav className={classes}>
       { toggleButton }
+      { isOpen &&
+        <div className="nx-sidebar-nav--product-info">
+          { logoLink ? <a href={logoLink} aria-label="logo">{logo}</a> : logo }
+          <span className="nx-sidebar-nav--product-name">nexus lifecycle</span>
+        </div>
+      }
       { children }
       { isOpen && <div className="nx-sidebar-nav--separator" /> }
       <footer className="nx-sidebar-nav--footer">
--- a/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
+++ b/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
@@ -9,6 +9,7 @@ import React from 'react';
 import classnames from 'classnames';
 import { faBars, faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
 
+import Close from '../../icons/Close';
 import NxButton from '../NxButton/NxButton';
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import { Props, propTypes } from './types';
@@ -22,15 +23,17 @@ const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
     'nx-sidebar-nav__closed': !isOpen
   });
 
+  const toggleButton = (
+    <NxButton variant="icon-only"
+              onClick={onToggleOpen}
+              className='nx-sidebar-nav--icon nx-sidebar-nav--toggle'>
+      { isOpen ? <Close /> : <NxFontAwesomeIcon icon={faBars} /> }
+    </NxButton>
+  )
+
   return (
     <nav className={classes}>
-      { !isOpen &&
-        <NxButton variant="icon-only"
-                  className="nx-sidebar-nav--icon nx-closed-menu-icon"
-                  onClick={() => onToggleOpen()}>
-          <NxFontAwesomeIcon icon={faBars}/>
-        </NxButton>
-      }
+      { toggleButton }
       { children }
       <footer className="nx-sidebar-nav--footer">
         { !!helpLink &&
--- a/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
+++ b/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
@@ -16,17 +16,15 @@ import { Props, SidebarNavLinkProps, propTypes } from './types';
 import './NxSidebarNavitation.scss';
 
 const NxSidebarNavLink = function NxSidebarNavLink({ name, href, icon, current }: SidebarNavLinkProps) {
-  const navigationLinkClassnames = classnames('nx-sidebar-nav--item', {
+  const navigationLinkClassnames = classnames('nx-sidebar-nav--item', 'nx-sidebar-nav--link', {
     'selected': current
   });
 
   return (
-    <li className={navigationLinkClassnames} key={`nav-link-${name}`}>
-      <a href={href} className="nx-sidebar-nav--link">
-        <NxFontAwesomeIcon icon={icon} />
-        <span className="nx-sidebar-nav--link-name">{name}</span>
-      </a>
-    </li>
+    <a href={href} className={navigationLinkClassnames} key={`nav-link-${name}`}>
+      <NxFontAwesomeIcon icon={icon} />
+      <span className="nx-sidebar-nav--link-name">{name}</span>
+    </a>
   );
 };
 
@@ -69,15 +67,15 @@ const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
   const toggleButton = isOpen ? closeBtn : openBtn;
 
   const navigationLinks = (
-    <ul className="nx-sidebar-nav--items">
+    <nav className="nx-sidebar-nav--items">
       { links.map(NxSidebarNavLink)}
-    </ul>
+    </nav>
   );
 
   const logo = <img src={logoImg} className="nx-sidebar-nav--logo" />;
 
   return (
-    <nav className={classes}>
+    <aside className={classes}>
       { toggleButton }
       { isOpen &&
         <div className="nx-sidebar-nav--product-info">
@@ -112,7 +110,7 @@ const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
           attributions.map((attribution, index) => <p key={`attribution-${index}`}>{ attribution }</p>)
         }
       </footer>
-    </nav>
+    </aside>
   );
 };
 
--- a/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
+++ b/lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
@@ -9,12 +9,11 @@ import React from 'react';
 import classnames from 'classnames';
 import { faBars, faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
 
-import Close from '../../icons/Close';
 import NxButton from '../NxButton/NxButton';
+import NxCloseButton from '../NxCloseButton/NxCloseButton';
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import { Props, propTypes } from './types';
 import './NxSidebarNavitation.scss';
-
 
 const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
   const { children, className, isOpen, helpLink, onToggleOpen } = props;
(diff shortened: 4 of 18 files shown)


lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,180 @@
...
+  background-color: $nx-sidebar-nav-background-color;
+  color: $nx-sidebar-nav-text-color;
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+  min-width: 250px;
+  padding: 0;
+  padding-top: $nx-spacing-l;
+  position: relative;
+  // transition: all 0.5s ease;
```
raveneyex, line 35: Do we want to add any sort of transition for the toggling of the sidebar? If so I could use some help smoothing it out: with the links' text suddenly appearing the transition looks _jumpy_.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const navigationLinks = (
+    <ul className="nx-sidebar-nav--items">
+      { links.map(NxSidebarNavLink)}
+    </ul>
+  );
+
+  const logo = <img src={logoImg} className="nx-sidebar-nav--logo" />;
+
+  return (
+    <nav className={classes}>
```
glenhunter, line 80: I don't think that you should be using the `<nav>` element here, as it it it wraps many items which aren't navigational. I'd suggest that it should only wrap the links.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-variables';
+@import '../../scss-shared/nx-text-helpers';
+@import '../../scss-shared/nx-container-helpers';
+
+$nx-sidebar-width__open: 250px;
+$nx-sidebar-width__closed: 69px;
```
glenhunter, line 12: Values like `69px` and `33px` seem very _magical_

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import './NxSidebarNavitation.scss';
+
+const NxSidebarNavLink = function NxSidebarNavLink({ name, href, icon, current }: SidebarNavLinkProps) {
+  const navigationLinkClassnames = classnames('nx-sidebar-nav--item', {
+    'selected': current
+  });
+
+  return (
+    <li className={navigationLinkClassnames} key={`nav-link-${name}`}>
```
glenhunter, line 24: I don't see any reason why this needs to be a list. `<a>`s wrapped with `<nav>` should be sufficient and better for accessibility.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+      { toggleButton }
+      { isOpen &&
+        <div className="nx-sidebar-nav--product-info">
+          { logoLink ? <a href={logoLink} aria-label="logo">{logo}</a> : logo }
+          <span className="nx-sidebar-nav--product-name">{logoText}</span>
+        </div>
+      }
+      { navigationLinks }
+      { isOpen && <div className="nx-sidebar-nav--separator" /> }
+      <footer className="nx-sidebar-nav--footer">
```
glenhunter, line 90: This shouldn't be `--footer` it should be `__footer`, as in not a modifier of but rather an element of `nx-sidebar-nav`

glenhunter:
I think that the class naming could use a bit of attention. `nx-sidebar-nav` seems to act as a container for many things which are not nav related at all. I'd suggest that `nx-sidebar-nav` should be limited to the navigation portions and a new class cleated for the container that wraps the whole thing. We could just use `.nx-sidebar__` for the various elements like the product name, footer, divider lines, etc.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
glenhunter: Though see: https://github.com/sonatype/sonatype-react-shared-components/pull/316#issuecomment-782336954

glenhunter:
There's a strange flash on both the "X" and hamburger icons when they are clicked.
Screen Shot 2021-02-19 at 3 47 36 PM

glenhunter:
I **VERY VERY** strongly dislike the "X" and hamburger as the open/close UX. 

The "X" does not look _intentionally_ placed, it looks like a mistake. Stuck weirdly up and to the right of the title. Just floating in space. "Why's that X up there?"

Other places in the UI that we use an "X" icon it means *close*, as in gone and not coming back, modals, alerts, etc. But here it means something different. It's not consistent with the rest of the UI.

The hamburger is not quite as problematic but still has a meaning that's _not quite_ the same as what folks are used to.

I think that we should use arrows or a clearer indication that we are sliding, or shrinking and growing a part of the UI.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+              onClick={onToggleClick}
+              className={toggleButtonClasses}>
+      <NxFontAwesomeIcon icon={faBars} />
+    </NxButton>
+  );
+
+  const toggleButton = isOpen ? closeBtn : openBtn;
+
+  const navigationLinks = (
+    <ul className="nx-sidebar-nav--items">
```
glenhunter, line 72: This is where I think the `<nav>` element should be used.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+.nx-sidebar-nav {
+  @include container-vertical;
+
+  box-sizing: border-box;
+  background-color: $nx-sidebar-nav-background-color;
+  color: $nx-sidebar-nav-text-color;
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+  min-width: 250px;
```
glenhunter, line 31: I thought there was already a variable for the sidebar width?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+}
+
+.nx-sidebar-nav--product-info {
+  margin: $nx-spacing-sm $nx-spacing-l 0;
+  height: $nx-sidebar-logo-height;
+}
+
+.nx-sidebar-nav--product-name {
+  display: inline-block;
+  font-size: 16px;
```
glenhunter, line 45: We have a variable for this

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+}
+
+.nx-sidebar-nav--items {
+  list-style-type: none;
+  margin: 36px 0;
+  padding: 0;
+  width: 100%;
+}
+
+.nx-sidebar-nav--item {
```
glenhunter, line 80: This needs `container-horizontal`, if you look you'll see that the icons have left margin we want to remove

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  margin-bottom: $nx-spacing-sm;
+  margin-top: $nx-spacing-xxl;
+}
+
+.nx-sidebar-nav--link {
+  @include regular;
+
+  color: inherit;
+  display: inline-block;
+  font-size: 16px;
```
glenhunter, line 113: Use variable

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  height: 32px;
+  line-height: 32px;
+  text-decoration: none;
+
+  .nx-icon {
+    width: 21px;
```
glenhunter, line 119: Why this? Looking at Zeplin the icons look like they're just standard 16px icons.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  }
+}
+
+.nx-sidebar-nav--help-link {
+  color: inherit;
+  font-size: 14px;
```
glenhunter, line 125: Variable

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+
+.nx-sidebar-nav--items {
+  margin-bottom: $nx-spacing-xxl;
+}
+
+.nx-sidebar-nav__open {
+  .nx-sidebar-nav--toggle {
+    height: 16px;
+    position: absolute;
+    right: $nx-spacing-l;
```
glenhunter, line 145: I wouldn't use the spacing variable here, we generally just use them for margin/padding. I'm not sure that we'd want this to move if we adjusted our standard margin/padding values.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+    padding-left: $nx-spacing-l;
+  }
+
+  .nx-sidebar-nav--link-name {
+    margin-left: $nx-spacing-sm;
+  }
+}
+
+.nx-sidebar-nav__closed {
+  min-width: 69px;
```
glenhunter, line 159: You created a variable for this up above

glenhunter:
Another design comment, the nav links in the sidebar seem cramped now that I'm actually using them. I'm wondering if we should maybe give them a bit more height? @stephenim

gallery/src/index.html
```diff
@@ -14,7 +14,7 @@
     <link rel="stylesheet" href="gallery.css"/>
   </head>
   <body class="nx-body">
-    <div id="ui" class="nx-page"></div>
+    <div id="ui" class="nx-page--sidebar"></div>
```
rpokorny, line 17: The base Block class should still be present in addition to the modifier

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -32,6 +32,14 @@
   min-width: $nx-page-width-min;
 }
 
+.nx-page--sidebar {
+  @include container-vertical;
```
rpokorny, line 36: But isn't in a horizontal container?

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+    </li>
+  );
+};
+
+const NxSidebarNavigation = function NxSidebarNavigation(props: Props) {
+  const {
+    isOpen,
+    className,
+    onToggleClick,
+    logoImg,
```
rpokorny, line 38: Do we want the general Sonatype hexagon to be a default or do we want the caller to always have to specify an icon?

lib/src/base-styles/_nx-page-layout.scss
raveneyex: > But isn't in a horizontal container?

I don't understand this comment.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+    helpLink,
+    helpText,
+    collapsedReleaseText,
+    expandedReleaseText,
+    attributions
+  } = props;
+
+  const classes = classnames(className, 'nx-sidebar-nav', {
+    'nx-sidebar-nav__open': isOpen,
+    'nx-sidebar-nav__closed': !isOpen
```
rpokorny, line 51: These look like they are supposed to be Modifiers, not Elements

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+      { links.map(NxSidebarNavLink)}
+    </ul>
+  );
+
+  const logo = <img src={logoImg} className="nx-sidebar-nav--logo" />;
+
+  return (
+    <nav className={classes}>
+      { toggleButton }
+      { isOpen &&
```
rpokorny, line 82: I'd try to do as much of the open/closed difference as possible in CSS

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <div className="nx-sidebar-nav--product-info">
+          { logoLink ? <a href={logoLink} aria-label="logo">{logo}</a> : logo }
+          <span className="nx-sidebar-nav--product-name">{logoText}</span>
+        </div>
+      }
+      { navigationLinks }
+      { isOpen && <div className="nx-sidebar-nav--separator" /> }
+      <footer className="nx-sidebar-nav--footer">
+        { !!helpLink &&
+          <a rel="noreferrer"
```
rpokorny, line 92: Do we want `noreferrer` on what would typically be a link to a page we control?

gallery/src/Application.tsx
```diff
@@ -4,92 +4,82 @@
...
+    <>
+      <NxStatefulSidebarNavigation isDefaultOpen={true}
+                                   logoImg={logoImg}
+                                   logoText="nexus lifecycle"
+                                   logoLink="/home"
+                                   links={navLinks}
+                                   helpLink="http://google.com"
+                                   helpText="google help"
+                                   collapsedReleaseText="release 105"
+                                   expandedReleaseText="lifecycle release 105"
```
rpokorny, line 70: I wonder if the difference between these two should be worked out automatically by the component

raveneyex: > I wonder if the difference between these two should be worked out automatically by the component

That's an assumption based on the instance of the mock, but it may well be a possibility that the two texts are entirely different.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
rpokorny, line 90: An idea: would it be better in some way (more flexible maybe) if the sidebar footer was a separate component of its own which the caller was expected to include as a child of the sidebar?

marushkevych:
> Other places in the UI that we use an "X" icon it means _close_, as in gone and not coming back, modals, alerts, etc. But here it means something different. It's not consistent with the rest of the UI.

Agree. I think we should be consistent with this:

image

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny, line 31: You have a variable for this which isn't getting used here.  Additionally, is this really supposed to be a min or should it just be `width`?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+
+  box-sizing: border-box;
+  background-color: $nx-sidebar-nav-background-color;
+  color: $nx-sidebar-nav-text-color;
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+  min-width: 250px;
+  padding: 0;
+  padding-top: $nx-spacing-l;
```
rpokorny, line 33: Less re-definey:
```suggestion
  padding: $nx-spacing-l 0 0 0;
```

glenhunter:
> > Other places in the UI that we use an "X" icon it means _close_, as in gone and not coming back, modals, alerts, etc. But here it means something different. It's not consistent with the rest of the UI.
> 
> Agree. I think we should be consistent with this:
> 
That icon is much clearer in its intent/effect

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny: > If so I could use some help smoothing it out: with the links' text suddenly appearing the transition looks jumpy.

Could have the text slide in from the right as the sidebar expands.  Just an idea, not sure if it would be good.  Also I don't recommend using `all` for the transition property.  I feel like that could very easily effect properties we don't intend it to

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  position: relative;
+  // transition: all 0.5s ease;
+}
+
+.nx-sidebar-nav--product-info {
+  margin: $nx-spacing-sm $nx-spacing-l 0;
```
rpokorny, line 39: This ties into @glenhunter 's comments about the positioning of the close button (comments that I agree with) but typically we don't want top margin on the first child in a container.  Even if the button position doesn't get changed, what I'd probably prefer would be for the close button and the product info to be wrapped up in a sidebar header or something, which does not have external top margin, and then the positioning between the close button and the product info would be handled within that element.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+  );
+
+  const logo = <img src={logoImg} className="nx-sidebar-nav--logo" />;
+
+  return (
+    <nav className={classes}>
+      { toggleButton }
+      { isOpen &&
+        <div className="nx-sidebar-nav--product-info">
+          { logoLink ? <a href={logoLink} aria-label="logo">{logo}</a> : logo }
```
rpokorny, line 82: Should the link be just the icon or also the text?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  height: $nx-sidebar-logo-height;
+}
+
+.nx-sidebar-nav--product-name {
+  display: inline-block;
```
rpokorny, line 44: What's the purpose of this property?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  flex-direction: column;
+  min-height: 100vh;
+  min-width: 250px;
+  padding: 0;
+  padding-top: $nx-spacing-l;
+  position: relative;
+  // transition: all 0.5s ease;
+}
+
+.nx-sidebar-nav--product-info {
```
rpokorny, line 38: The way the logo sticks out of the anchor tag makes me uncomfortable: 
image

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  display: inline-block;
+  font-size: 16px;
+  line-height: $nx-sidebar-logo-height;
+  vertical-align: top;
+}
+
+.nx-sidebar-nav--logo {
+  height: $nx-sidebar-logo-height;
+  margin-right: $nx-spacing-xs;
+  width: 30px;
```
rpokorny, line 53: Should probably only specify one dimension and let the logo's aspect ratio determine the other

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+}
+
+.nx-sidebar-nav--icon {
```
rpokorny, line 56: This class is poorly named.  It should be an element not a modifier, and it is confusing because you use it not on icons, but on specific buttons (that only contain icons).

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  color: $nx-sidebar-nav-text-color;
+  margin: 0;
+}
+
+.nx-footer-help-icon {
```
rpokorny, line 61: Unused?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  color: $nx-sidebar-nav-help-color;
+}
+
+.nx-sidebar-nav--footer {
```
rpokorny, line 65: Another modifier that should be an element (and more below)

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  align-self: center;
+  bottom: $nx-spacing-xxl;
+  font-size: $nx-font-size-xs;
+  position: absolute;
```
rpokorny, line 69: I don't recommend absolute positioning for this.  There will be no way for it to adapt to the presence of other nearby content that way:
image

Make it a normal flex child pushed to the bottom via margin

rpokorny: > I don't recommend absolute positioning for this. There will be no way for it to adapt to the presence of other nearby content that way:

Speaking of which, what is the expected scrolling behavior within the sidebar?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Could have the text slide in from the right as the sidebar expands

How can I do that?

> I don't recommend using all for the transition property

Which properties do you think we should transition?

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > I don't think that you should be using the `<nav>` element here, as it it it wraps many items which aren't navigational. I'd suggest that it should only wrap the links.

Using `<nav>` only for links and `<aside>` for the whole sidebar in 81e0139d95d4864157bcc88e144d66c34ba93f81

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  text-align: center;
+}
+
+.nx-sidebar-nav--items {
+  list-style-type: none;
+  margin: 36px 0;
```
rpokorny, line 74: Use a `$nx-spacing-...` variable

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > I don't see any reason why this needs to be a list. `<a>`s wrapped with `<nav>` should be sufficient and better for accessibility.

Done in 81e0139d95d4864157bcc88e144d66c34ba93f81

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  padding: 0;
+  width: 100%;
```
rpokorny: IMO it'd be cleanest if the sidebar used `align-items: stretch` which would avoid the need for this

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+}
+
+.nx-sidebar-nav--item {
+  height: 32px;
```
rpokorny, line 79: One or the other of the height here and the height on the `<a>` is redundant

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  &.selected {
+    background-color: $nx-sidebar-nav-selected-color;
+
+    .nx-sidebar-nav--link {
+      @include semi-bold;
+    }
+  }
+}
+
+.nx-sidebar-nav--separator {
```
rpokorny, line 94: We try to avoid making separate elements for these sorts of things.  This could be a bottom border on the list

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  background-color: $nx-sidebar-separator-color;
+  height: 1px;
+  width: 100%;
+}
+
+.nx-sidebar-nav--release {
+  height: 36px;
```
rpokorny, line 101: It seems like this height is being used to just add more space underneath the text which could be bottom margin

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  margin-bottom: $nx-spacing-sm;
+  margin-top: $nx-spacing-xxl;
+}
+
+.nx-sidebar-nav--link {
```
rpokorny, line 106: Long text problems:
image

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny: > Why this? Looking at Zeplin the icons look like they're just standard 16px icons.

Yeah you shouldn't mess with the width of icons.  If you need more space around them use margin.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  height: 32px;
+  line-height: 32px;
+  text-decoration: none;
+
+  .nx-icon {
+    width: 21px;
+  }
+}
+
+.nx-sidebar-nav--help-link {
```
rpokorny, line 120: Should use `container-horizontal` which would avoid the need for the `.nx-icon` custom margin below.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+    margin-left: 0;
+  }
+}
+
+.nx-sidebar-nav--help-text {
+  font-size: 14px;
+}
+
+.nx-sidebar-nav--items {
+  margin-bottom: $nx-spacing-xxl;
```
rpokorny, line 135: Merge this into the margin declaration in the other `.nx-sidebar-nav--items` block above

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  font-size: 14px;
+  text-decoration: none;
+
+  .nx-icon {
+    margin-left: 0;
+  }
+}
+
+.nx-sidebar-nav--help-text {
+  font-size: 14px;
```
rpokorny, line 131: Redundant

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+}
+
+.nx-sidebar-nav--items {
+  margin-bottom: $nx-spacing-xxl;
+}
+
+.nx-sidebar-nav__open {
+  .nx-sidebar-nav--toggle {
+    height: 16px;
```
rpokorny, line 140: This is probably the cause of the oddly shaped white box on click (well, the cause of the oddly-shaped aspect of it anyway)

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny: > I wouldn't use the spacing variable here, we generally just use them for margin/padding. I'm not sure that we'd want this to move if we adjusted our standard margin/padding values.

I don't see why this wouldn't be considered spacing.  It's the distance from other things, just like margin or padding.  

> I'm not sure that we'd want this to move if we adjusted our standard margin/padding values.

That could be equally said about any other place we use those variables.  If we were ever to change them it would be a large and subjective undertaking

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+    position: absolute;
+    right: $nx-spacing-l;
+    width: 16px;
```
rpokorny, line 143: Why is this custom?

rpokorny: > Why is this custom?

I suspect you should get mockups of what the x and hamburger icons are supposed to look like when hovered, focused, and clicked, to determine what (if any) size customizations would be correct here.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  .nx-sidebar-nav--item {
+    padding-left: $nx-spacing-l;
```
rpokorny, line 147: I tend to this this should be margin

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  .nx-sidebar-nav--link-name {
+    margin-left: $nx-spacing-sm;
```
rpokorny, line 151: This block doesn't really need to be nested under here

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+  .nx-sidebar-nav--logo {
+    display: none;
+  }
+
+  .nx-sidebar-nav--items {
+    text-align: center;
+  }
+
+  .nx-sidebar-nav--release {
+    width: 44px;
```
rpokorny, line 171: magic number.  It appears this is here to force the release number to wrap before it gets too close to the edge.  Id' say that should be the responsibility of padding on the sidebar itself.

lib/src/components/NxSidebarNavigation/stateful/NxStatefulSidebarNavigation.tsx
```diff
@@ -0,0 +1,40 @@
...
+const NxStatefulSidebarNavigation = function(props: Props) {
+  const {
+    isDefaultOpen,
+    onToggleClick,
+    ...otherProps
+  } = props;
+
+  const [isOpen, setIsOpen] = useState(isDefaultOpen);
+
+  const toggleHandler = function toggleHandler() {
```
rpokorny, line 21: Once my `useToggle` improvement day branch is merged this would be a place to use that

lib/src/components/NxSidebarNavigation/stateful/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  logoLink: string;
+  links: SidebarNavLinkProps[];
+  helpLink?: string | null;
+  helpText?: string | null;
+  collapsedReleaseText?: string | null;
+  expandedReleaseText?: string | null;
+  attributions?: string[] | null;
+}
+
+export const propTypes: WeakValidationMap<Props> = {
```
rpokorny, line 27: Use `ValidationMap` if possible.  `WeakValidationMap` doesn't really work right in terms of enforcing that the TS props and react propTypes match

lib/src/components/NxSidebarNavigation/types.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IconProp } from '@fortawesome/fontawesome-svg-core';
+
+export interface SidebarNavLinkProps {
+  name: string;
+  href: string;
+  icon: IconProp;
+  current?: boolean | null;
+}
+
+export interface Props {
```
rpokorny, line 18: It's generally preferred to base the stateful component props on the stateless ones using `extends` and potentially [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)

rpokorny: > It's generally preferred to base the stateful component props on the stateless ones using extends and potentially Omit

And similarly, to base the stateful propTypes on the stateless ones using object spread and ramda's `omit`.

lib/src/base-styles/_nx-page-layout.scss
rpokorny: > I don't understand this comment.

I fixed my spelling if that was the source of the confusion.  The `.nx-page` contains two elements side-by-side, so it is a horizontal container not a vertical container, correct?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny: > Which properties do you think we should transition?

That depends on the animation which is ultimately up to the designer.  But only the exact ones that the animation actually needs to transition.  So most likely `width` and maybe a small number of others.

> How can I do that?

It would take some fiddling; I wouldn't spend time on it unless its decided upon as the desired effect.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > This is where I think the `<nav>` element should be used.

Done in https://github.com/sonatype/sonatype-react-shared-components/commit/81e0139d95d4864157bcc88e144d66c34ba93f81

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > We have a variable for this

Using variable in ea52a08b7388fd58d6bba5cbe10c1014fa3af2b4

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Use variable

Done in ea52a08b7388fd58d6bba5cbe10c1014fa3af2b4

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Variable

Done in ea52a08b7388fd58d6bba5cbe10c1014fa3af2b4

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > > Which properties do you think we should transition?
> 
> That depends on the animation which is ultimately up to the designer. But only the exact ones that the animation actually needs to transition. So most likely `width` and maybe a small number of others.
> 
> > How can I do that?
> 
> It would take some fiddling; I wouldn't spend time on it unless its decided upon as the desired effect.

Removed transition for now. 55e1df8284ca6704b9e34378dcd7b38f214fd082

gallery/src/index.html
raveneyex: > The base Block class should still be present in addition to the modifier

Done in e09f5d2899a1cf58b68b50eccda0a4319473b46b

lib/src/base-styles/_nx-page-layout.scss
raveneyex: > > I don't understand this comment.
> 
> I fixed my spelling if that was the source of the confusion. The `.nx-page` contains two elements side-by-side, so it is a horizontal container not a vertical container, correct?

Yes. You are right. Changed in dee00bb9e1484775afa1f6a0c9164206120d8358

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > This shouldn't be `--footer` it should be `__footer`, as in not a modifier of but rather an element of `nx-sidebar-nav`

Fixed (I hope) in 232611c992543e74499f2aab8381c27efdf6e960

raveneyex:
> I think that the class naming could use a bit of attention. `nx-sidebar-nav` seems to act as a container for many things which are not nav related at all. I'd suggest that `nx-sidebar-nav` should be limited to the navigation portions and a new class cleated for the container that wraps the whole thing. We could just use `.nx-sidebar__` for the various elements like the product name, footer, divider lines, etc.

I hope this fixes it. 232611c992543e74499f2aab8381c27efdf6e960

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Another modifier that should be an element (and more below)

Changed in 232611c992543e74499f2aab8381c27efdf6e960

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Merge this into the margin declaration in the other `.nx-sidebar-nav--items` block above

Done as part of fb8c6551007cf0bd70c9d0bd0b2437f1544155cf

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Use a `$nx-spacing-...` variable

Done in fb8c6551007cf0bd70c9d0bd0b2437f1544155cf

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > This class is poorly named. It should be an element not a modifier, and it is confusing because you use it not on icons, but on specific buttons (that only contain icons).

I got rid of it entirely in 232611c992543e74499f2aab8381c27efdf6e960

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Less re-definey:

Done as part of fb8c6551007cf0bd70c9d0bd0b2437f1544155cf

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Values like `69px` and `33px` seem very _magical_

I got them directly out of what Zeplin indicated. Should I define them in `_nx-variables.scss` or ... ?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > You created a variable for this up above

Changed in fb8c6551007cf0bd70c9d0bd0b2437f1544155cf

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > These look like they are supposed to be Modifiers, not Elements

Changed in 232611c992543e74499f2aab8381c27efdf6e960

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Should probably only specify one dimension and let the logo's aspect ratio determine the other

Removed `width` in fb8c6551007cf0bd70c9d0bd0b2437f1544155cf

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > I tend to this this should be margin

Using `margin` prevents the highlights for hover and selected from extending all the way to the border.

raveneyex:
> There's a strange flash on both the "X" and hamburger icons when they are clicked.
> Screen Shot 2021-02-19 at 3 47 36 PM

I fixed the weird aspect ratio of the "flash" in f1fa55d896f6ffae67174724752e9e604d3f1d54 but it is still there: that strange flash is the `:active` modifier of the `icon-only` variation of `nx-button`.
Should I disable that style modification?

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Why is this custom?

I was just using the dimensions that were shown in Zeplin. I removed them in f1fa55d896f6ffae67174724752e9e604d3f1d54.

> I suspect you should get mockups of what the x and hamburger icons are supposed to look like when hovered, focused, and clicked,

@stephenim, @ChurchRyan can you folks help me with this? Right now the `icon-only` variation of `NxButton` has an `:active` style that looks like a _flash_ in here. I think we need specific styles for this case.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > This block doesn't really need to be nested under here

Moved in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > This is probably the cause of the oddly shaped white box on click (well, the cause of the oddly-shaped aspect of it anyway)

Yes. I removed it in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Redundant

Removed in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Should use `container-horizontal` which would avoid the need for the `.nx-icon` custom margin below.

Done in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Long text problems:
> image

Spoke about this today at wall walk and we (@stephenim and I) believe this to be irrelevant because we at Sonatype —and specifically the design org— controls the copy that will go into these links.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > It seems like this height is being used to just add more space underneath the text which could be bottom margin

Changed in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > We try to avoid making separate elements for these sorts of things. This could be a bottom border on the list

Cleaned up in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > One or the other of the height here and the height on the `<a>` is redundant

Cleaned up as part of f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > IMO it'd be cleanest if the sidebar used `align-items: stretch` which would avoid the need for this

This `width` is gone as of f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > Unused?

Removed in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > What's the purpose of this property?

Removed in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > You have a variable for this which isn't getting used here

Changed to use variable in fb8c6551007cf0bd70c9d0bd0b2437f1544155cf

> is this really supposed to be a min or should it just be `width`?

Maybe you could help me understand something here: 
if I use `width` the sidebar's width ends up being _less_ than 250px. If I use `min-width` the sidebar's width is _exactly_ 250px.

rpokorny: > if I use width the sidebar's width ends up being less than 250px; If I use min-width the sidebar's width is exactly 250px. I don't understand why.

The sidebar probably needs `flex: none`.  `flex-shrink` defaults to 1, not 0, which I always find counter-intuitive

gallery/src/Application.tsx
rpokorny: > but it may well be a possibility that the two texts are entirely different.

I think it would be up to design to say whether or not that's true.  Given that we've only got the one example so far, I can only guess.

rpokorny:
> That icon is much clearer in its intent/effect

But I believe it's only available in FA Pro

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny: > Spoke about this today at wall walk and we (@stephenim and I) believe this to be irrelevant because we at Sonatype —and specifically the design org— controls the copy that will go into these links.

Sounds good.  Remember to specifically mention that limitation when writing the docs for this component.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
rpokorny: > Id' say that should be the responsibility of padding on the sidebar itself.

Or maybe padding on the sidebar footer, if that would work better

lib/src/components/NxSidebarNavigation/stateful/NxStatefulSidebarNavigation.tsx
rpokorny: > Once my useToggle improvement day branch is merged this would be a place to use that

It has been merged

rpokorny:
> I fixed the weird aspect ratio of the "flash" in f1fa55d but it is still there: that strange flash is the :active modifier of the icon-only variation of nx-button.
Should I disable that style modification?

I assume we still want these buttons to have active, hover, and focus styles.  But being as they are customized in terms of their icon color and background color, they will inevitably need to also be customized in terms of their border and background color in each of those states.  The flash here isn't really "strange", it's just the normal background hover color of a button.

rpokorny:
I'm noticing that in your implementation the links are slightly higher up when the sidebar is collapsed than when it is open.  In the mockup they appear to maintain a consistent vertical position.

rpokorny:
In the mockup the link icons change size when the sidebar is collapsed.  That does not happen in your impl

rpokorny:
Looking at the distance from the footer to the bottom of the viewport, the mockups are a bit confusing but at any rate your impl does not match them.  In the mockups, the lowest line of text in the footer is 48px from the bottom when the sidebar is expanded (seems fine, that is one of our spacing variables). Yet when the footer is collapsed, the lowever line of text in the footer is 56px from the bottom (which is _not_ one of our standard spacing variables).

In the implementation, the lowest line of text is 60px from the bottom when expanded, and 80px from the bottom when collapsed.

rpokorny:
In the mockup the border under the links is not visible when the sidebar is collapsed

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -32,6 +32,13 @@
   min-width: $nx-page-width-min;
 }
 
+.nx-page--sidebar {
+  @include container-horizontal;
+
+  align-items: flex-start;
```
rpokorny, line 38: Probably want `align-items` stretch.  I'd expect both the sidebar and the `.nx-page-content` to be full height at all times. As it is now the `.nx-page-content` is only as tall as its contents.  This would also allow you to get rid of the `vh` height on the sidebar.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,119 @@
...
+    helpLink,
+    helpText,
+    collapsedReleaseText,
+    expandedReleaseText,
+    attributions
+  } = props;
+
+  const classes = classnames(className, 'nx-sidebar', {
+    'nx-sidebar--open': isOpen,
+    'nx-sidebar--closed': !isOpen
```
rpokorny, line 49: As described on the Styling Components page in the gallery, we use un-BEMed utility classes for certain common states, notably `open` and `closed` which are applicable here.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,119 @@
...
+    logoLink,
+    links,
+    helpLink,
+    helpText,
+    collapsedReleaseText,
+    expandedReleaseText,
+    attributions
+  } = props;
+
+  const classes = classnames(className, 'nx-sidebar', {
```
rpokorny, line 47: The overall class naming here is inevitably going to be tricky, since we've already got another sidebar - `.nx-page-sidebar`.  But I've just realized, do those really need to be two separate sidebars? Or could this be a modifier of that one (e.g. `.nx-page-sidebar--navigation`).  Here's what I'm picturing:

What we have today: `.nx-page` which wraps `.nx-page-header` and `.nx-page-contents`, and `.nx-page-contents` wraps `.nx-page-sidebar` and `.nx-page-main`.

What we could have going forward, still supporting the old way as well:
`.nx-page` still contains `.nx-page-contents` but `.nx-page-header` is optional.  `.nx-page-contents` has  modifier that allows it to go full-width as opposed to the old centered, constrained width style. `.nx-page-sidebar` is still the sidebar of choice, but has a modifier and specific substructure by which it can be the sidebar being implemented in this PR.

I feel like this would maximize consistency/compatibility with what we already have, and may cut down on some of the styling needed on the sidebar element itself.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > This needs `container-horizontal`, if you look you'll see that the icons have left margin we want to remove

Done in f1fa55d896f6ffae67174724752e9e604d3f1d54

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > > if I use width the sidebar's width ends up being less than 250px; If I use min-width the sidebar's width is exactly 250px. I don't understand why.
> 
> The sidebar probably needs `flex: none`. `flex-shrink` defaults to 1, not 0, which I always find counter-intuitive

Fixed in 6c27955c7c30914aa6bcc37bd14edd4a4c8ddc65

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > > Why this? Looking at Zeplin the icons look like they're just standard 16px icons.
> 
> Yeah you shouldn't mess with the width of icons. If you need more space around them use margin.

Fixed in ebfa41bc7a1e4ba95261b7c6645c3c2fca021740

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
glenhunter: > > Values like `69px` and `33px` seem very _magical_
> 
> I got them directly out of what Zeplin indicated. Should I define them in `_nx-variables.scss` or ... ?

We have been using a base-8 grid for most of our spacing so I would have expected the sizing to embrace that.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
glenhunter: ^^ @stephenim

gallery/src/Application.tsx
raveneyex: > > but it may well be a possibility that the two texts are entirely different.
> 
> I think it would be up to design to say whether or not that's true. Given that we've only got the one example so far, I can only guess.

@stephenim

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
raveneyex: > > > Values like `69px` and `33px` seem very _magical_
> > 
> > 
> > I got them directly out of what Zeplin indicated. Should I define them in `_nx-variables.scss` or ... ?
> 
> We have been using a base-8 grid for most of our spacing so I would have expected the sizing to embrace that.

A lot of the dimensions in the mock are definitely _not_ standard. That includes margins/paddings and widths.
@stephenim ^^

rpokorny: > That includes margins/paddings and widths.

Width and height often won't be; they usually depend on content.  It's mainly about padding and margins.

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
glenhunter: > > I wouldn't use the spacing variable here, we generally just use them for margin/padding. I'm not sure that we'd want this to move if we adjusted our standard margin/padding values.
> 
> I don't see why this wouldn't be considered spacing. It's the distance from other things, just like margin or padding.
> 
> > I'm not sure that we'd want this to move if we adjusted our standard margin/padding values.
> 
> That could be equally said about any other place we use those variables. If we were ever to change them it would be a large and subjective undertaking

I agree that it's a fine distinction but I see it as the difference between spacing (margin/padding) which should always be consistent values based on our variables, and positioning which tends (though it probably shouldn't) to be more arbitrary.

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > The overall class naming here is inevitably going to be tricky, since we've already got another sidebar - `.nx-page-sidebar`. But I've just realized, do those really need to be two separate sidebars? Or could this be a modifier of that one (e.g. `.nx-page-sidebar--navigation`). Here's what I'm picturing:
> 
> What we have today: `.nx-page` which wraps `.nx-page-header` and `.nx-page-contents`, and `.nx-page-contents` wraps `.nx-page-sidebar` and `.nx-page-main`.
> 
> What we could have going forward, still supporting the old way as well:
> `.nx-page` still contains `.nx-page-contents` but `.nx-page-header` is optional. `.nx-page-contents` has modifier that allows it to go full-width as opposed to the old centered, constrained width style. `.nx-page-sidebar` is still the sidebar of choice, but has a modifier and specific substructure by which it can be the sidebar being implemented in this PR.
> 
> I feel like this would maximize consistency/compatibility with what we already have, and may cut down on some of the styling needed on the sidebar element itself.

Is this more or less what you had in mind? b9b047801218fedc6d13234b0795b2c66b9c2238

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
```diff
@@ -0,0 +1,119 @@
...
+import React from 'react';
+import classnames from 'classnames';
+import { faBars, faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
+
+import NxButton from '../NxButton/NxButton';
+import NxCloseButton from '../NxCloseButton/NxCloseButton';
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import { Props, SidebarNavLinkProps, propTypes } from './types';
+
+import './NxSidebarNavitation.scss';
```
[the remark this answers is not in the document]
raveneyex: > Navitation?

Renamed in b9b047801218fedc6d13234b0795b2c66b9c2238

lib/src/base-styles/_nx-page-layout.scss
raveneyex: > Probably want `align-items: stretch`. I'd expect both the sidebar and the `.nx-page-content` to be full height at all times. As it is now the `.nx-page-content` is only as tall as its contents. This would also allow you to get rid of the `vh` height on the sidebar.

No longer relevant as of b9b047801218fedc6d13234b0795b2c66b9c2238

lib/src/components/NxSidebarNavigation/NxSidebarNavitation.scss
```diff
@@ -0,0 +1,184 @@
...
+
+.nx-sidebar-nav {
+  @include container-vertical;
+
+  box-sizing: border-box;
+  background-color: $nx-sidebar-nav-background-color;
+  color: $nx-sidebar-nav-text-color;
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
```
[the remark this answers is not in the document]
raveneyex: > I think a better way to do this would be to use `align-items: stretch` on `.nx-page--sidebar`

No longer relevant as of b9b047801218fedc6d13234b0795b2c66b9c2238

lib/src/components/NxSidebarNavigation/types.ts
raveneyex: > > It's generally preferred to base the stateful component props on the stateless ones using extends and potentially Omit
> 
> And similarly, to base the stateful propTypes on the stateless ones using object spread and ramda's `omit`.

Done in 0f25e721c9505e11759b66031eb242387ef74dd1

lib/src/components/NxSidebarNavigation/NxSidebarNavigation.tsx
raveneyex: > As described on the Styling Components page in the gallery, we use un-BEMed utility classes for certain common states, notably `open` and `closed` which are applicable here.

Changed in eafc84f4ef5fc29a4776b1838f1eeb18f18ef294

PR #317, opened 2021-02-22
Example of page-level tabs - RSC-433

rpokorny:
https://issues.sonatype.org/browse/RSC-433

NxTabs already supported what's in the mockups without problem, so this just adds an example


gallery/src/components/NxTabs/NxTabsPage.tsx
```diff
@@ -130,6 +132,15 @@ export default function NxTabsPage() {
         {' '}<code className="nx-code">NxTile</code> where there is no tile header and the tabs are top most within
         the tile.
       </GalleryExampleTile>
+      <GalleryExampleTile title="NxTabs outside of NxTile Example"
+                          id="nx-tab-outside-tile-example"
+                          defaultCheckeredBackground={true}
+                          liveExample={NxTabsOutsideTileExample}
+                          codeExamples={NxTabsOutsideTileExampleCode}>
+        A basic example of how to use the <code className="nx-code">NxTabs</code> family of components in an
```
glenhunter, line 140: The title says "Outside of NxTile" but the description says "in an NxTile". I think the former is correct, but that makes the whole description incorrect. Also shouldn't it be `nx-tile`?

glenhunter:
This is what we should strive with for with all our components, when we need to use it somewhere new it just works.

gallery/src/components/NxTabs/NxTabsPage.tsx
rpokorny: > The title says "Outside of NxTile" but the description says "in an NxTile". I think the former is correct, but that makes the whole description incorrect

Right the description was copypasta that I forgot to update. 9ff954cd

> Also shouldn't it be nx-tile?

Probably, but I was just going for consistency with what's already on this page.

rpokorny:
> This is what we should strive with for with all our components, when we need to use it somewhere new it just works.

Of course, and we do.  I wouldn't say we did a specially good job of that here, nor that we do a bad job of it elsewhere.  Rather this ticket was mainly a design question of deciding whether we wanted this scenario to be a special case styling-wise.  Turns out the answer was no.

glenhunter:
I think that it would be worthwhile making one of the tabs a standalone `NxTable`.

gallery/src/components/NxTabs/NxTabsPage.tsx
```diff
@@ -130,6 +132,14 @@ export default function NxTabsPage() {
         {' '}<code className="nx-code">NxTile</code> where there is no tile header and the tabs are top most within
         the tile.
       </GalleryExampleTile>
+      <GalleryExampleTile title="NxTabs outside of NxTile Example"
+                          id="nx-tab-outside-tile-example"
+                          defaultCheckeredBackground={true}
+                          liveExample={NxTabsOutsideTileExample}
+                          codeExamples={NxTabsOutsideTileExampleCode}>
+        A basic example of how to use the <code className="nx-code">NxTabs</code> family of components at the page
+        level: outside of and typically in control of some tiles or tables.
```
glenhunter, line 141:
```suggestion
        level, outside of tiles or tables.
```

rpokorny:
> I think that it would be worthwhile making one of the tabs a standalone NxTable.

9b228da9

glenhunter (approved):

PR #318, opened 2021-02-23
Safari card fixes RSC-467

glenhunter:
https://issues.sonatype.org/browse/RSC-467

Safari does not support the use of `gap` for `flex` layouts. I added a media query to target Safari which allows us to use `margin` for both internal and external spacing in Safari and `gap` for other browsers.

I also tweaked the examples because the last card in the first row _always_ wrapped but what I wanted to demonstrate was the first row wrapping at smaller viewport sizes.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "4.1.0",
+  "version": "4.1.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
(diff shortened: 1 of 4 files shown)


rpokorny:
I'm seeing double space between adjacent card containers. 
image

glenhunter:
> I'm seeing double space between adjacent card containers.

Ah, thought I'd fixed that, done here: https://github.com/sonatype/sonatype-react-shared-components/pull/318/commits/f407376d3b730984db376c0214b23b75d862f8ac

rpokorny (approved):

PR #319, opened 2021-02-24
NxColorPicker - RSC-476

rpokorny:
https://issues.sonatype.org/browse/RSC-476


glenhunter:
When using Voice Over it identifies the selected colours as "1 of 9", "2 of 9" etc.

Screen Shot 2021-02-25 at 9 38 03 AM

rpokorny:
> When using Voice Over it identifies the selected colours as "1 of 9", "2 of 9" etc. There's also no indication that you are in a color picker or the effect of making a selection.

Hmm that's a shame, it works pretty well in ChromeVox.  Says things like "Pink. Radio button selected". Can you try to play around with it on Mac and see if you can improve it? One idea would be to add an aria-label to the `<input>` itself.  The `<label>` already has one thanks to `NxTooltip`.  Oh also what browser are you in?

> There's also no indication that you are in a color picker or the effect of making a selection.

So for me it's working equally as well as any other group of radio buttons in that regard: If you keyboard nav to the fieldset legend it'll read it, but if you're just tab-navigating it'll miss it.

glenhunter:
Adding this: `aria-label="Color Picker"` to the `<fieldset>` gets VO to identify the group. The colour is being identified correctly now.

glenhunter:
Voice Over 
Safari:
Screen Shot 2021-03-05 at 10 21 14 AM
Firefox:
Screen Shot 2021-03-05 at 10 19 20 AM
Chrome:
Screen Shot 2021-03-05 at 10 19 20 AM

glenhunter (approved):
I'm not sure what to do to fix the accessibility in Chrome. I think we'll have to accept it as-is for now and revisit later.

rpokorny:
> I'm not sure what to do to fix the accessibility in Chrome. I think we'll have to accept it as-is for now and revisit later.

rpokorny:
think

rpokorny:
> I'm not sure what to do to fix the accessibility in Chrome. I think we'll have to accept it as-is for now and revisit later.

Try it now; I've made some significant changes.

glenhunter:
> > I'm not sure what to do to fix the accessibility in Chrome. I think we'll have to accept it as-is for now and revisit later.
> 
> Try it now; I've made some significant changes.

It's better in Safari and Firefox, but unfortunately Chrome is still basically useless.

Screen Shot 2021-03-08 at 9 48 57 AM
Screen Shot 2021-03-08 at 9 48 40 AM
Screen Shot 2021-03-08 at 9 48 15 AM

rpokorny:
> but unfortunately Chrome is still basically useless.

Well if they're in Chrome I guess they can use ChromeVox (which works well here) so IMO this is as good as we can get it

glenhunter:
> > but unfortunately Chrome is still basically useless.
> 
> Well if they're in Chrome I guess they can use ChromeVox (which works well here) so IMO this is as good as we can get it

I agree and my initial Approval stands.

PR #320, opened 2021-02-25
Add custom classes to nxform cancel and submit buttons RSC-495

glenhunter:
https://issues.sonatype.org/browse/RSC-495

I debated going even more specific with the class names and using `nx-btn--nxform-cancel` and `nx-btn--nxform-submit`, but in the end I kept it simple.


rpokorny:
I would have made them Elements under the form Block, e.g. `.nx-form__cancel-btn`

glenhunter:
> I would have made them Elements under the form Block, e.g. `.nx-form__cancel-btn`

Interesting. Can you expand on why?

glenhunter:
FWIW I also considered using `test__nxform-btn-submit` as we have done in IQ in the past

rpokorny:
> Interesting. Can you expand on why?

Because these classes only have meaning within the .nx-form, as children.  That seems exactly like they'd be elements to me.

> FWIW I also considered using test__nxform-btn-submit as we have done in IQ in the past

I'm not comfortable assuming this class is only useful in tests.  I can imagine an app which consumes RSC adding some custom styling to one of these classes.

glenhunter:
> > Interesting. Can you expand on why?
> 
> Because these classes only have meaning within the .nx-form, as children. That seems exactly like they'd be elements to me.
> 
> > FWIW I also considered using test__nxform-btn-submit as we have done in IQ in the past
> 
> I'm not comfortable assuming this class is only useful in tests. I can imagine an app which consumes RSC adding some custom styling to one of these classes.

Renamed classes as suggested above: https://github.com/sonatype/sonatype-react-shared-components/pull/320/commits/013434e299b7a544b0500e5dcacf36d4f55deda8

gallery/src/components/NxForm/NxFormPage.tsx
```diff
@@ -247,6 +247,44 @@ const NxFormPage = () =>
...
+              <NxTableCell><code className="nx-code">nx-btn--form-cancel</code></NxTableCell>
+              <NxTableCell>
+                Applied to the default <code className="nx-code">NxForm</code> Cancel button.
+              </NxTableCell>
+              <NxTableCell>
+                A CSS class that can be used to identify the Cancel button on the page for testing or other purposes.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell><code className="nx-code">nx-btn--form-submit</code></NxTableCell>
```
rpokorny, line 277: Need to update these

glenhunter: > Need to update these

Right: https://github.com/sonatype/sonatype-react-shared-components/pull/320/commits/6c9a620d88fa6f452c18fa00a2e12aaa4e742255

rpokorny (approved):

PR #323, opened 2021-03-01
Colour docs safari fixes RSC-143

glenhunter:
https://issues.sonatype.org/browse/RSC-143

Safari doesn't support `gap` in flex layouts so I had to employ the same fixes to this issue as we did with the `.nx-card` layouts.

### DIFF
--- a/gallery/src/guidelines/ColorPalette/Swatcher.scss
+++ b/gallery/src/guidelines/ColorPalette/Swatcher.scss
@@ -52,7 +52,7 @@
 @media not all and (min-resolution:.001dpcm) {
   @supports (-webkit-appearance:none) and (display:flow-root) {
     // override a default .nx-card style
-    .nx-card__content > :not(:last-child) {
+    .nx-card__content > .gallery-swatch {
       margin-bottom: 0;
     }
 
--- a/gallery/src/guidelines/ColorPalette/Swatcher.scss
+++ b/gallery/src/guidelines/ColorPalette/Swatcher.scss
@@ -19,12 +19,16 @@
   .nx-card__content {
     align-items: stretch;
     gap: 0;
+
+    // override Safari rule from .nx-card
+    & > .gallery-swatch {
+      margin-bottom: 0;
+    }
   }
 }
 
 .gallery-swatch {
   align-items: center;
-  column-gap: $nx-spacing-l;
   display: flex;
   flex-direction: row;
   text-align: left;
@@ -33,6 +37,7 @@
 .gallery-swatch__thumb {
   flex-shrink: 0;
   height: 64px;
+  margin-right: $nx-spacing-l;
   width: 64px;
 }
 
@@ -46,22 +51,5 @@
 
 .gallery-swatch__hex {
   flex-basis: 7ch;
-}
-
-// Safari does not support gap in flex layouts so we need to use this query and margin for Safari.
-@media not all and (min-resolution:.001dpcm) {
-  @supports (-webkit-appearance:none) and (display:flow-root) {
-    // override a default .nx-card style
-    .nx-card__content > .gallery-swatch {
-      margin-bottom: 0;
-    }
-
-    .gallery-swatch__thumb {
-      margin-right: $nx-spacing-l;
-    }
-
-    .gallery-swatch__hex {
-      margin-right: $nx-spacing-l;
-    }
-  }
+  margin-right: $nx-spacing-l;
 }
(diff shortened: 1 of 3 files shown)


gallery/src/guidelines/ColorPalette/Swatcher.scss
```diff
@@ -47,3 +47,21 @@
 .gallery-swatch__hex {
   flex-basis: 7ch;
 }
+
+// Safari does not support gap in flex layouts so we need to use this query and margin for Safari.
+@media not all and (min-resolution:.001dpcm) {
+  @supports (-webkit-appearance:none) and (display:flow-root) {
+    // override a default .nx-card style
+    .nx-card__content > :first-child {
+      margin-bottom: 0;
```
rpokorny, line 56: This makes me very suspicious of whether the default `.nx-card` style here is correct

glenhunter: > This makes me very suspicious of whether the default `.nx-card` style here is correct

I'm overriding the Safari override styles here which is annoying. Basically in a situation where we have an `.nx-content` where the contents are stacked we want the first child to have `margin-bottom: 16px`, but not the last since that would double up the object and container margins. In situations where there are more than 2 objects stacked it's up to the dev to solve that since I was trying to avoid `.nx-card__content * { margin-bottom: $nx-spacing-md;}`.

rpokorny: Here's the problem with the existing styles.  If you go into the card example on master and add a third status indicator to this card, you'll see that the margins between the indicators are different.  Since `.nx-card__content` can have `n` children, adjusting the space between only the first and second children doesn't make sense.

image

rpokorny: > Here's the problem with the existing styles. If you go into the card example on master and add a third status indicator to this card, you'll see that the margins between the indicators are different. Since .nx-card__content can have n children, adjusting the space between only the first and second children doesn't make sense.

What I think we probably need instead is to apply a 16px bottom margin to all `.nx-card__content` children except the last one.

rpokorny: > Here's the problem with the existing styles

This is now causing an issue on a downstream project:
image

glenhunter: > > Here's the problem with the existing styles
> 
> This is now causing an issue on a downstream project:

The fix for this is here: https://github.com/sonatype/sonatype-react-shared-components/pull/324

When that is merged I'll merge Master into this branch

glenhunter: > > > Here's the problem with the existing styles
> > 
> > 
> > This is now causing an issue on a downstream project:
> 
> The fix for this is here: #324
> 
> When that is merged I'll merge Master into this branch

Merged here: https://github.com/sonatype/sonatype-react-shared-components/pull/323/commits/2ab328cb98b0d5f8ac6b02a636cabb8f60b3615a

gallery/src/guidelines/ColorPalette/Swatcher.scss
```diff
@@ -47,3 +47,21 @@
 .gallery-swatch__hex {
   flex-basis: 7ch;
 }
+
+// Safari does not support gap in flex layouts so we need to use this query and margin for Safari.
+@media not all and (min-resolution:.001dpcm) {
+  @supports (-webkit-appearance:none) and (display:flow-root) {
+    // override a default .nx-card style
+    .nx-card__content > :not(:last-child) {
```
rpokorny, line 55: Doesn't need to be quite this complicated
```suggestion
    .nx-card__content > .gallery-swatch {
```

rpokorny (approved):

rpokorny (commented):
Actually, on second thought I doubt we need the ugly media queries and stuff, as they don't add to the specificity.  We can probably just zero out the margin in _all_ browsers.

glenhunter:
> Actually, on second thought I doubt we need the ugly media queries and stuff, as they don't add to the specificity. We can probably just zero out the margin in _all_ browsers.

Bit of shuffling involved but I'm in favour of removing the hack: https://github.com/sonatype/sonatype-react-shared-components/pull/323/commits/a3bb81a5df29d1c27399c27f3e81e50886690913

gallery/src/guidelines/ColorPalette/Swatcher.scss
```diff
@@ -19,12 +19,16 @@
   .nx-card__content {
     align-items: stretch;
     gap: 0;
+
+    // override Safari rule from .nx-card
+    & > .gallery-swatch {
```
rpokorny, line 24:
```suggestion
    > .gallery-swatch {
```

rpokorny (approved):

PR #327, opened 2021-03-04
Add workaround for webpack watch issue

rpokorny:
Workaround for https://github.com/webpack/webpack/issues/12810 which is preventing changes in the lib from triggering a rebuild of the gallery.


glenhunter (approved):
It worked here

PR #328, opened 2021-03-09
Use css variables for page size values RSC-506 

glenhunter:
https://issues.sonatype.org/browse/RSC-506

I created CSS variables to replace the SCSS variables used for page layout.

In doing so I learned that CSS variables are emitting code, so to get around the repetition problem I moved the CSS variables into the `_page-layout.scss` file.

I also replaced instances of the SCSS variables with the appropriate CSS variables.

### DIFF
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -29,7 +29,7 @@
   display: flex;
   flex-direction: column;
   height: 100%;
-  min-width: $nx-page-width-min;
+  min-width: var(--page-width-max);
 }
 
 .nx-page-content {
@@ -38,7 +38,7 @@
   box-sizing: border-box;
   display: flex;
   flex-grow: 1;
-  max-width: $nx-page-width-max;
+  max-width: var(--page-width-max);
   overflow-y: hidden;
   width: 100%;
 
@@ -54,7 +54,7 @@
 
   background-color: $nx-sidebar-background-color;
   box-sizing: border-box;
-  flex-basis: $nx-sidebar-width;
+  flex-basis: var(--sidebar-width);
   min-width: 0;
   overflow-x: hidden;
   padding: $nx-spacing-xl $nx-spacing-l $nx-spacing-xs $nx-spacing-l;
@@ -66,7 +66,7 @@
   @include container-vertical;
 
   box-sizing: border-box;
-  flex: 1 0 $nx-content-width;
+  flex: 1 0 var(--content-width);
   min-width: 0;
   overflow-x: hidden;
   overflow-y: auto;
@@ -83,7 +83,7 @@
   }
 
   .nx-page {
-    min-width: $nx-page-width-min - $nx-scrollbar-width;
+    min-width: var(--page-width-max) - $nx-scrollbar-width;
   }
 
   .nx-page-header {
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -8,6 +8,13 @@
 /*
   #Layout
 */
+:root {
+  --page-width-max: 1600px;
+  --page-width-min: 1366px;
+  --content-width: 76%;
+  --sidebar-width: 100% - var(--content-width);
+}
+
 .nx-html, .nx-body {
   box-sizing: border-box;
   height: 100%;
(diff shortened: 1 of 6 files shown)


lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -8,6 +8,13 @@
 /*
   #Layout
 */
+:root {
```
rpokorny, line 11: We already have a section for these at the bottom of `_nx-globals.scss`

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -8,6 +8,13 @@
...
+  --page-width-max: 1600px;
+  --page-width-min: 1366px;
+  --content-width: 76%;
+  --sidebar-width: 100% - var(--content-width);
```
rpokorny, line 15: I'm surprised this doesn't need to be a calc

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -83,7 +90,7 @@
   }
 
   .nx-page {
-    min-width: $nx-page-width-min - $nx-scrollbar-width;
+    min-width: var(--page-width-max) - $nx-scrollbar-width;
```
rpokorny, line 93: I'd also expect this to need to be a calc

lib/src/base-styles/_nx-page-layout.scss
glenhunter: > We already have a section for these at the bottom of `_nx-globals.scss`

https://github.com/sonatype/sonatype-react-shared-components/pull/328/commits/9d6f28ce5dadad7cebf9d9e225d93582fed82f95

lib/src/base-styles/_nx-page-layout.scss
glenhunter: > I'm surprised this doesn't need to be a calc

In this case the units are the same, it's `100% - 76%`.

lib/src/base-styles/_nx-page-layout.scss
glenhunter: > I'd also expect this to need to be a calc

Same as above the units are the same. `1600px - 15px`

glenhunter: FYI I fixed the min/max issue above here: https://github.com/sonatype/sonatype-react-shared-components/pull/328/commits/484bdde0327df578e0d550f1200762378eb06c7b

rpokorny: > Same as above the units are the same. 1600px - 15px

That doesn't matter; CSS only does math in `calc` expressions.  If you try to do math without one that would be SCSS math, which of course can't work with runtime values like a `var()`.  See this screenshot for proof that the `min-width` is not working:

image

glenhunter: > > Same as above the units are the same. 1600px - 15px
> 
> That doesn't matter; CSS only does math in `calc` expressions. If you try to do math without one that would be SCSS math, which of course can't work with runtime values like a `var()`. See this screenshot for proof that the `min-width` is not working:
> 

I came to the same conclusion this weekend but for different reasons, namely what would happen if a consumer set one of the variables to a different unit. So I made the change here: https://github.com/sonatype/sonatype-react-shared-components/pull/328/commits/92071b851ee12d9c3139fe32eb6c39b80aa3c008

Note that I had to move the scrollbar width into a CSS var or the calc for page scrolling wouldn't work in real time.

rpokorny (approved):

lib/src/base-styles/_nx-page-layout.scss
rpokorny: > Note that I had to move the scrollbar width into a CSS var or the calc for page scrolling wouldn't work in real time.

That doesn't sound right to me, but I'm not against move that to a CSS variable.

glenhunter: > > Note that I had to move the scrollbar width into a CSS var or the calc for page scrolling wouldn't work in real time.
> 
> That doesn't sound right to me, but I'm not against move that to a CSS variable.

It's weird, I couldn't see any reason that it shouldn't work but in the inspector the `min-width` (the result of the  `calc()`) was always `0`

PR #329, opened 2021-03-09
Create nx-vulnerability-indicator RSC-482

glenhunter:
https://issues.sonatype.org/browse/RSC-482

Because the only changing part of the indicator is the number I decided that `nx-vulnerability-indicator` is most similar to `nx-counter` and built it as a "style" component.

### DIFF
--- a/gallery/src/styles/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
+++ b/gallery/src/styles/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
+
+const nxVulnerabilityIndicatorCode = require('./NxVulnerabilityIndicatorExample.html');
+
+const NxVulnerabilityIndicatorPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        Basic style for small indicator token which typically displays a single #, a '# of #' string, or a short text
+        string.
+      </p>
+      <p className="nx-p">
+        Some basic positioning CSS examples have been provided. To right justify the Vulnerability Indicator within its
+        container use <code className="nx-code">nx-pull-right</code>.
+      </p>
+      <table className="nx-table nx-table--gallery-props">
+        <thead>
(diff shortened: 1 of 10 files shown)


rpokorny:
I think this should be a component.  It's a very large amount of boilerplate HTML to write out when an encapsulating component could be something like 
```
<NxVulnerabilityIndicator critical={5} severe={12} moderate={3} low={0} />
```

Additionally we want to force `nx-vulnerability-indicator--critical` and the text "Critical" to always be used together (and likewise for the other levels).

rpokorny:
Could the name of this component be more indicative of it being counts?  Something like `NxSeverityCounts`?  As I write that, I realize that it doesn't even have to always be counting vulnerabilities.  The same UI could be used to count violations or anything else that uses the standard severities categories.

gallery/src/styles/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-vulnerability-indicator-container</code></td>
+            <td className="nx-cell">Container</td>
+            <td className="nx-cell">Basic container class. It's always used with one of the modifiers below.</td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-vulnerability-indicator-container--row</code></td>
```
rpokorny, line 39: I thought we were going to try to implement it so that it automatically lays out based on the available width.  Granted I know we weren't sure that would be possible (to get 2x2 without allows 3x2 for instance).

glenhunter: > I thought we were going to try to implement it so that it automatically lays out based on the available width. Granted I know we weren't sure that would be possible (to get 2x2 without allows 3x2 for instance).

I recall that being mentioned, I didn't recall it being made a requirement. It might be possible with a media query.

glenhunter:
> Could the name of this component be more indicative of it being counts? Something like `NxSeverityCounts`? As I write that, I realize that it doesn't even have to always be counting vulnerabilities. The same UI could be used to count violations or anything else that uses the standard severities categories.

This is also a reason that I didn't make it a component. It seemed likely that somebody would see this and want to use it for something else where they might also want to change the text. I get a bit worried with this type of design because while it seems very specific in its intent right now the design is open enough that it could be used for many things and I don't want to overly constrain it.

rpokorny:
> This is also a reason that I didn't make it a component. It seemed likely that somebody would see this and want to use it for something else where they might also want to change the text. I get a bit worried with this type of design because while it seems very specific in its intent right now the design is open enough that it could be used for many things and I don't want to overly constrain it.

Given that these colors are inherently tied to these severity category names in our apps I'm not sure that's something to worry about

PR #330, opened 2021-03-11
Background mask mixin RSC-401

glenhunter:
https://issues.sonatype.org/browse/RSC-401

I chose to treat the `nx-modal` mask as the "default" figuring that full screen masks are a _slightly_ more common pattern than container masks.


rpokorny:
Why a mixin rather than a common class name?

lib/src/base-styles/_base.scss
```diff
@@ -20,6 +20,7 @@
 @import '../scss-shared/nx-container-helpers';
 @import '../scss-shared/nx-text-helpers';
 @import '../scss-shared/nx-flexbox-helpers';
+@import '../scss-shared/nx-mixins';
```
rpokorny, line 23: All of our other mixins are in more specific files.  I'd think we'd rather continue with that pattern than make a generic file for mixins.

glenhunter:
> Why a mixin rather than a common class name?

That was the title of the JIRA ticket: "Create mixin for background mask attributes used in nxModal and nxSubmitMask". It occurs to me now that on of the differences between the two uses is the `z-index` and the `mixin` could have an input for that value which would make it more useful.

lib/src/base-styles/_base.scss
glenhunter: > All of our other mixins are in more specific files. I'd think we'd rather continue with that pattern than make a generic file for mixins.

It didn't seem to fit in any of the other existing scss files (the closest IMO was `container-helpers`) so rather than create a very specific `_mask.scss` I created a generic one in case we end up with other hard to place mixins. I am open to suggestions though.

rpokorny:
> That was the title of the JIRA ticket: "Create mixin for background mask attributes used in nxModal and nxSubmitMask".

Just because you wrote the title that way doesn't mean it has to be implemented that way 🙂  I think a common class makes more sense, similar to how we have the `.nx-footer` class which is used in multiple places.

glenhunter:
> > That was the title of the JIRA ticket: "Create mixin for background mask attributes used in nxModal and nxSubmitMask".
> 
> Just because you wrote the title that way doesn't mean it has to be implemented that way 🙂 I think a common class makes more sense, similar to how we have the `.nx-footer` class which is used in multiple places.

I agree, but did you see the addition of the `z-index` parameter? I think that makes it worth a `mixin`

rpokorny:
> I agree, but did you see the addition of the z-index parameter? I think that makes it worth a mixin

Alright.  I'd still lean towards a common class personally and just overriding the z-index where necessary, but if you'd rather it be a mixin I can be ok with that

lib/src/base-styles/_base.scss
rpokorny: > I created a generic one in case we end up with other hard to place mixins. I am open to suggestions though.

I'd rather stick to our existing pattern which is to have pretty specific mixin files.  I'd call this one `_nx-mask-helpers.scss` to fit the pattern.

glenhunter: > > I created a generic one in case we end up with other hard to place mixins. I am open to suggestions though.
> 
> I'd rather stick to our existing pattern which is to have pretty specific mixin files. I'd call this one `_nx-mask-helpers.scss` to fit the pattern.

ok: https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/3b3e39459dc735a871ec7543b3e066540d3f8891

lib/src/base-styles/_base.scss
```diff
@@ -20,6 +20,7 @@
 @import '../scss-shared/nx-container-helpers';
 @import '../scss-shared/nx-text-helpers';
 @import '../scss-shared/nx-flexbox-helpers';
+@import '../scss-shared/nx-mask-helpers';
```
rpokorny, line 23: I don't think there's actually a need to import this here.  There only would be if it was being used somewhere within the base styles.

lib/src/scss-shared/_nx-mask-helpers.scss
```diff
@@ -0,0 +1,22 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+ @import 'nx-colors';
+
+ @mixin background-mask($z-index) {
+  align-items: center;
+  background-color: transparentize($nx-grey-800, 0.25);
+  display: flex;
+  flex-direction: column;
```
rpokorny, line 14: The properties that pertain to laying out the mask's children probably shouldn't be part of the mixin since they are specific to how the children are laid out and structured.  So for instance `display: flex`, `flex-direction`, and `align-items` probably shouldn't be in the mixin, and I'm not sure `overflow-y` is necessary at all.

Or.. we can say that a standard way of laying out children _is_ part of the mask, and adjust `.nx-submit-mask` to follow that way instead of using flex sideways as it currently does.

lib/src/base-styles/_base.scss
glenhunter: > I don't think there's actually a need to import this here. There only would be if it was being used somewhere within the base styles.

You're right: https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/3f9802b29f8688333f596400318d9b09feb2d7bb

lib/src/scss-shared/_nx-mask-helpers.scss
glenhunter: > The properties that pertain to laying out the mask's children probably shouldn't be part of the mixin since they are specific to how the children are laid out and structured. So for instance `display: flex`, `flex-direction`, and `align-items` probably shouldn't be in the mixin, and I'm not sure `overflow-y` is necessary at all.
> 
> Or.. we can say that a standard way of laying out children _is_ part of the mask, and adjust `.nx-submit-mask` to follow that way instead of using flex sideways as it currently does.

I prefer that the mask has a default way of layout out its children. I have cleaned up the `flex` use here: https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/a63884e68d58a5960e387e582a14c8c27cbfb7f8

rpokorny:
Modals are broken:
image

glenhunter:
> Modals are broken:

I restored `flex-direction: column` to `nx-modal`: https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/6f3e93e093f6c2b97366ff02b52163718f761546

lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -7,30 +7,21 @@
...
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
-
-  z-index: $nx-submit-mask-z-index;
 }
 
 .nx-submit-mask--fullscreen {
+  @include background-mask($nx-submit-mask-z-index);
```
rpokorny, line 19: Rather than re-invoking the whole mixin this should probably just be an override of `z-index`.

lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -7,30 +7,21 @@
...
+
   position: fixed;
+  justify-content: left;
```
rpokorny, line 22: `start` would be less confusing than `left` here

lib/src/scss-shared/_nx-mask-helpers.scss
```diff
@@ -0,0 +1,23 @@
...
+  justify-content: center;
+  overflow-y: hidden;
```
rpokorny, line 16: Still not sure this property is necessary

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > Rather than re-invoking the whole mixin this should probably just be an override of `z-index`.

I actually did it like this to demonstrate changing the `z-index` value in the mixin.

lib/src/scss-shared/_nx-mask-helpers.scss
glenhunter: > Still not sure this property is necessary

Seems that it isn't, maybe a holdover from IE11 days? https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/c15fb39292d5247bd885798ff4bb8bd6700429eb

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > `start` would be less confusing than `left` here

"less confusing"? https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/c15fb39292d5247bd885798ff4bb8bd6700429eb

lib/src/components/NxSubmitMask/NxSubmitMask.scss
glenhunter: > > Rather than re-invoking the whole mixin this should probably just be an override of `z-index`.
> 
> I actually did it like this to demonstrate changing the `z-index` value in the mixin.

Though since the SCSS doesn't appear in the examples that's pointless: https://github.com/sonatype/sonatype-react-shared-components/pull/330/commits/57c71a5b3459fab6b5cb16e417d6671fce40fb2c

rpokorny (approved):

PR #331, opened 2021-03-16
Cleanup nx colors RSC-496

glenhunter:
https://issues.sonatype.org/browse/RSC-496

Moved unused and old colour variables into a deprecated section at the bottom of `_colors.scss`.

Looked through the existing styles in RSC and where we had hex values that could be replaced with variables I did so. Where the colours were unique I created tickets assigned to @dturner32 to come up with colour options within the RSC palettes.


lib/src/scss-shared/_nx-color-swatches.scss
```diff
@@ -111,6 +111,7 @@ $nx-red-300: #ff8286;
 $nx-red-200: #ff979a;
 $nx-red-100: #ffb6b8;
 $nx-red-50: #ffcbcc;
+$nx-red-25: #fff6f8;
```
rpokorny, line 114: Shouldn't this be checked with @dturner32 ?

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -131,3 +76,66 @@ $nx-selectable-colors: (
   yellow: ($nx-yellow-700, $nx-yellow-50),
   lime: ($nx-green-200, $nx-green-50)
 );
+
+/*
+  WARNING
```
rpokorny, line 81: I think this section would be a lot better as a separate file than a "section" at the bottom of this file

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -131,3 +76,66 @@ $nx-selectable-colors: (
...
+
+$nx-fail-color-active: #5c0000;
+$nx-fail-color-disabled: #dd9595;
+$nx-fail-color-hover: #5d0000;
+
+// Border rows are used by lists and tables
+$nx-border-row: 1px solid $nx-grey-1;
+$nx-border-row-hover: 1px solid $nx-grey-4;
+$nx-border-row-disabled: 1px solid $nx-grey-1;
+$nx-border-row-dark: 1px solid $nx-grey-5;
```
rpokorny, line 141: These aren't colors and so I would think should not be in this section.

lib/src/scss-shared/_nx-color-swatches.scss
glenhunter: > Shouldn't this be checked with @dturner32 ?

I confirmed all the colour changes with @dturner32. He came up with the colour above.

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -94,18 +46,12 @@ $nx-disabled-border: $nx-grey-500;
...
-$nx-border-row: 1px solid $nx-grey-1;
-$nx-border-row-hover: 1px solid $nx-grey-4;
-$nx-border-row-disabled: 1px solid $nx-grey-1;
-$nx-border-row-dark: 1px solid $nx-grey-5;
+$nx-border-darker: 1px solid $nx-grey-800;
+$nx-border-dark: 1px solid $nx-grey-300;
+$nx-border: 1px solid $nx-grey-200; // default
+$nx-border-light: 1px solid $nx-grey-100;
+$nx-border-lighter: 1px solid $nx-grey-50;
+$nx-border-error: 1px solid $nx-red-800;
```
rpokorny, line 54: Should these also be moved to `_nx-variables` while we're at it?

rpokorny (approved):

PR #333, opened 2021-03-17
Add muse.dev configuration.

tneer:
muse.dev is currently complaining that it doesn't have access to the typescript-eslint plugin so seeing if this resolves that issue.


.muse/install_eslint_plugins.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+npm install @typescript-eslint/eslint-plugin@latest --save-dev
```
rpokorny, line 3: Is the issue here that it doesn't like that our `package.json`s aren't in the top-level directory?

rpokorny, line 3: Minor: we use `yarn` everywhere in this repo not `npm`

tneer: Not really sure.  I was just looking at the logs at https://console.muse.dev/result/sonatype/sonatype-react-shared-components/01F1097FRGAF2MSMP8JJAW6RKE?tab=logs and trying to play with things a little without necessarily bugging the muse.dev folks too much.

The analysis for this PR didn't have a build log.  Maybe because I didn't change any typescript.  I may try to introduce an error.

tneer: I can try `yarn`, but Javascript/Typescript projects don't seem to be built by muse.dev so it may just be running eslint via npm irregardless of what the project uses.  https://docs.muse.dev/docs/repository-configuration/ lists the project types that use a build system.

rpokorny:
I feel like I should mention that this is all a bit redundant since eslint already runs as part of CI

PR #342, opened 2021-03-29
WithClass and Convenience Components - RSC-505

rpokorny:
https://issues.sonatype.org/browse/RSC-505

This came from Improvement Day originally (original PR: https://github.com/sonatype/sonatype-react-shared-components/pull/326) but was upgraded at some point to an actual ticket.


gallery/src/jsUtilPages/WithClass/WithClassPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+    withClassExampleScss = require('./WithClassExample.scss?raw');
+
+const WithClassPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <NxP>
+        <NxCode>withClass</NxCode> is a higher-order component that allows for the easy creation of
+        new React Component types that consist simply of a native DOM element with a particular class name.
+        The resulting component supports all props that the underlying native component supports, including
+        additionaly class names that are added in addition to the one specified when <NxCode>withClass</NxCode>
```
glenhunter, line 25:
```suggestion
        class names that are added in addition to the one specified when <NxCode>withClass</NxCode>
```

gallery/src/styles/NxList/NxListPage.tsx
```diff
@@ -129,6 +140,7 @@ const NxListPage = () =>
             </NxTableRow>
             <NxTableRow>
               <NxTableCell><code className="nx-code">.nx-list__item</code></NxTableCell>
```
glenhunter, line 142: Tedious I know, but when Master gets merged into this branch you'll need to update the `.nx-list__item` entry in the top table.

glenhunter:
I think that there is value in expanding the `WithClass` page docs to include a slightly more complicated example with `.notation` as used on `nx-list`, `nx-card` and other places.

gallery/src/jsUtilPages/WithClass/WithClassExample.scss
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@font-face {
+  font-family: precious;
+  src: url(../../assets/Precious.ttf) format("truetype");
```
glenhunter, line 9: Is this font OSS?

gallery/src/pages/Home.tsx
```diff
@@ -5,81 +5,83 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React from 'react';
+import { NxP, NxH3, NxH4, NxCode } from '@sonatype/react-shared-components';
```
glenhunter, line 8: I wonder if we should group these in some way so that a single import gets you all of the "convenience components"

lib/src/components/SimpleComponents.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import withClass from '../util/withClass';
+
+// Simple convenience components that are just a particular tag with a particular classname, with the ability
+// to add all attrs that you would normally be able to add to that tag (including more classnames)
+export const NxP = withClass('nx-p', 'p');
```
glenhunter, line 11: One thing that I find odd about this is that the class is declared before the HTML tag, is that by design or required? I would expect to to look like: `export const NxP = withClass('p', 'nx-p');`

lib/src/components/SimpleComponents.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const NxPageMain = withClass('nx-page-main', 'main');
+export const NxPageSidebar = withClass('nx-page-sidebar', 'aside');
+
+export const NxH1 = withClass('nx-h1', 'h1');
+export const NxH2 = withClass('nx-h2', 'h2');
+export const NxH3 = withClass('nx-h3', 'h3');
+export const NxH4 = withClass('nx-h4', 'h4');
+
+export const NxTile = Object.assign(withClass('nx-tile', 'section'), {
+  Header: withClass('nx-tile-header', 'header'),
```
Blacktiger, line 29: If we're attaching these as child properties on the parent component, should we consider doing something similar for other components like the `NxTable` family of components? Or should these become NxTileHeader etc to match the established pattern in the RSC library?

lib/src/util/withClass.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type ElementType<E extends keyof JSX.IntrinsicElements> = ReactElement<ComponentProps<E>, E>;
+type ComponentConstructorType<E extends keyof JSX.IntrinsicElements> = (p: ComponentProps<E>) => ElementType<E>;
+
+export default function withClass<E extends keyof JSX.IntrinsicElements>(
+  withClassName: string,
+  El: E
+): ComponentConstructorType<E> {
+  return function WithClassWrapper({ className, ...otherProps }: ComponentProps<E>): ElementType<E> {
+    // There should be a safer way but I can't figure it out; TS keeps thinking that
```
Blacktiger, line 18: The top answers to this question look promising, but I'm not sure if they'll help or not here. https://stackoverflow.com/questions/55969769/typing-a-dynamic-tag-in-react-with-typescript

lib/src/components/SimpleComponents.tsx
Blacktiger: Since both arguments are required I think it would be better to have the element name first.

gallery/src/jsUtilPages/WithClass/WithClassExample.scss
rpokorny: > Is this font OSS?

Here's the license file.  When I read it the first time it seemed like we were fine, but reading it again (particularly the Destribution section), I think including it in a public git repo may be a problem. I'll look for a different one.

> 1001Fonts Free For Commercial Use License (FFC)
>
> Preamble
> In this license, 'Precious' refers to the given .zip file, which may contain one or numerous fonts. These fonts can be of any type (.ttf, .otf, ...) and together they form a 'font family' or in short a 'typeface'.
>
> 1. Copyright
> Precious is the intellectual property of its respective author, provided it is original, and is protected by copyright laws in many parts of the world.
>
> 2. Usage
> Precious may be downloaded and used free of charge for both personal and commercial use. Personal use refers to all usage that does not generate financial income in a business manner, for instance:
>
>  - personal scrapbooking for yourself
>  - recreational websites and blogs for friends and family
>  - prints such as flyers, posters, t-shirts for churches, charities, and non-profit organizations
>
> Commercial use refers to usage in a business environment, including:
>
>  - business cards, logos, advertising, websites for companies
>  - t-shirts, books, apparel that will be sold for money
>  - flyers, posters for events that charge admission
>  - freelance graphic design work
>  - anything that will generate direct or indirect income
>
> 3. Modification
> Precious may not be modified, altered, adapted or built upon without written permission by its respective author. This pertains all files within the downloadable font zip-file.
>
> 4. Distribution
> While Precious may freely be copied and passed along to other individuals for private use as its original downloadable zip-file, it may not be sold or published without written permission by its respective author.
>
> 5. Disclaimer
> Precious is offered 'as is' without any warranty. 1001fonts.com and the respective author of Precious shall not be liable for any damage derived from using this typeface. By using Precious you agree to the terms of this license.

rpokorny: > I think including it in a public git repo may be a problem. I'll look for a different one.

451f408a.  This one is released under the Open Font License.

gallery/src/pages/Home.tsx
rpokorny: > I wonder if we should group these in some way so that a single import gets you all of the "convenience components"

What would the value in that be?

glenhunter: > > I wonder if we should group these in some way so that a single import gets you all of the "convenience components"
> 
> What would the value in that be?

It would be tedious to be working on something and keep realizing that you haven't added this convenience component or that convenience component, or even remembering what they are all named. Instead just loading something like 'NxConvenienceComponents` and getting all of them.

rpokorny: > It would be tedious to be working on something and keep realizing that you haven't added this convenience component or that convenience component, or even remembering what they are all named. Instead just loading something like 'NxConvenienceComponents` and getting all of them.

You'd still have to reference them by name though, e.g. `<NxConvenienceComponents.NxP>`.  Additionally from the outside there isn't that much difference between a convenience component and any other component; they just don't do as much

lib/src/components/SimpleComponents.tsx
rpokorny: > Since both arguments are required I think it would be better to have the element name first.

e60d8101

lib/src/components/SimpleComponents.tsx
rpokorny: > If we're attaching these as child properties on the parent component, should we consider doing something similar for other components like the NxTable family of components? Or should these become NxTileHeader etc to match the established pattern in the RSC library?

I've been drifting more towards the style I have here.  `NxAccordion` was done the same way.  We could put in a ticket to set up `NxTable` the same way and deprecate the old top-level names for its children

gallery/src/pages/Home.tsx
glenhunter: > > It would be tedious to be working on something and keep realizing that you haven't added this convenience component or that convenience component, or even remembering what they are all named. Instead just loading something like 'NxConvenienceComponents` and getting all of them.
> 
> You'd still have to reference them by name though, e.g. `<NxConvenienceComponents.NxP>`. Additionally from the outside there isn't that much difference between a convenience component and any other component; they just don't do as much

Right, so it sounds like the answer to my original question is that there isn't a way that we could import _all_ of the convenience components as simply as `import { nxConvenienceComponents }`

gallery/src/jsUtilPages/WithClass/WithClassExample.scss
glenhunter: > > I think including it in a public git repo may be a problem. I'll look for a different one.
> 
>  This one is released under the Open Font License.

I was thinking that you could have just changed the background colour or something, but this is classier.

lib/src/components/SimpleComponents.tsx
Blacktiger: Sounds like it will be a breaking change either way, but I think it makes sense to make the change for consistency. If it's a large burden for existing code to change I suppose we could provide an alias in the index file.

Blacktiger (approved):

lib/src/util/withClass.tsx
rpokorny: > The top answers to this question look promising, but I'm not sure if they'll help or not here.

Thanks that link helpfed me fumble around until I got it.  927d0d67

gallery/src/pages/Home.tsx
rpokorny: > Right, so it sounds like the answer to my original question is that there isn't a way that we could import all of the convenience components as simply as import { nxConvenienceComponents }

Well no matter what, when you go to use one you'd still have to specify _which_ one, which means you'd need to know its name

lib/src/components/SimpleComponents.tsx
rpokorny: > we could provide an alias in the index file.

Right that's what I had in mind with deprecating them.  I've added RSC-543 for this.

gallery/src/styles/NxList/NxListPage.tsx
rpokorny: > Tedious I know, but when Master gets merged into this branch you'll need to update the .nx-list__item entry in the top table.

f31081e3

glenhunter (approved):

PR #343, opened 2021-03-29
Create vulnerability indicator RSC-482

glenhunter:
https://issues.sonatype.org/browse/RSC-482


gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+const nxVulnerabilityIndicatorCode = require('./NxVulnerabilityIndicatorExample?raw'),
+    nxVulnerabilityIndicatorColumnCode = require('./NxVulnerabilityIndicatorColumnExample?raw'),
+    nxVulnerabilityIndicatorGridCode = require('./NxVulnerabilityIndicatorGridExample?raw');
+
+const NxVulnerabilityIndicatorPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        Basic style for small indicator which displays a number representing a count of violations, and a short pre-set
```
rpokorny, line 23: The component is currently named nx _vulnerability_ indicator, but this sentence talks about _violations_?  Which is it intended to work over?  IMO both, as well any anything else that could be categorized in this way.  I'd propose a different name for the component to reflect this: `NxThreatCounts`

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+        text string which displays the vulnerability severity.
+      </p>
+      <p className="nx-p">
+        Three basic layouts have been demonstrated below. It is expected that one of these should satisfy any
+        requirment.
```
rpokorny, line 28:
```suggestion
        requirement.
```

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </p>
+      <table className="nx-table nx-table--gallery-props">
+        <thead>
+          <tr className="nx-table-row">
+            <th className="nx-cell nx-cell--header">Class</th>
+            <th className="nx-cell nx-cell--header">Location</th>
+            <th className="nx-cell nx-cell--header">Details</th>
+          </tr>
+        </thead>
+        <tbody>
```
rpokorny, line 38: As a component, we need to document the props, and probably do not want to document the internal classes.

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <tr className="nx-table-row">
+            <td className="nx-cell">
+              <code className="nx-code">.nx-vulnerability-<br />indicator-container--row</code>
+            </td>
+            <td className="nx-cell">
+              Modifier of <code className="nx-code">.nx-vulnerability-<br />indicator-container</code>
+            </td>
+            <td className="nx-cell">
+              Used when you want the indicators to appear in a row. This is the default. If no layout is specified
+              "row" will be used.
```
rpokorny, line 53: What's that point of having a separate class for this if it is the default?

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,92 @@
...
+.nx-vulnerability-indicator-container {
+  display: flex;
+  margin: $nx-spacing-md 0 $nx-spacing-l 0;
+}
+
+.nx-vulnerability-indicator-container--row {
+  @include container-horizontal;
+
+  .nx-vulnerability-indicator {
+    margin-right: $nx-spacing-md;
```
rpokorny, line 58: If the containers all used `grid` we could use `gap` which would be simpler than messing with these margins

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,92 @@
...
+  flex-wrap: wrap;
+  max-width: 350px;
+
+  .nx-vulnerability-indicator {
+    &:nth-child(1n + 3) {
+      margin-top: $nx-spacing-md;
+    }
+
+    &:nth-child(2n + 2) {
+      margin-left: 16px;
```
rpokorny, line 89: `gap` would be highly preferred to this

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,92 @@
...
+
+    &:last-child {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.nx-vulnerability-indicator-container--grid {
+  flex-wrap: wrap;
+  max-width: 350px;
```
rpokorny, line 81: An arbitrary max-width isn't flexible enough for large-number edge cases:
image

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,92 @@
...
+}
+
+.nx-vulnerability-indicator__count {
+  @include bold();
+}
+
+.nx-vulnerability-indicator__text {
+  margin-left: $nx-spacing-xs;
+  text-transform: uppercase;
+}
```
rpokorny, line 31: Should be semi-bold

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const layoutClasses = classnames('nx-vulnerability-indicator-container', className, {
+    [`nx-vulnerability-indicator-container--${layout}`]: layout,
+    ['nx-vulnerability-indicator-container--row']: !layout
+  });
+
+  return (
+    <div className={layoutClasses} {...attrs}>
+      <div className="nx-vulnerability-indicator nx-vulnerability-indicator--critical">
+        <span className="nx-vulnerability-indicator__count">{criticalCount}</span>
+        <span className="nx-vulnerability-indicator__text">Critical</span>
```
rpokorny, line 27: If you want to simplify the DOM slightly, the text styles for this element could be moved to its parent (and the margin moved over to the count).  It's also fine as-is though

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
```diff
@@ -0,0 +1,45 @@
...
+export default function NxVulnerabilityIndicator({
+  criticalCount, severeCount, moderateCount, lowCount, layout, className, ...attrs }: Props) {
+
+  const layoutClasses = classnames('nx-vulnerability-indicator-container', className, {
+    [`nx-vulnerability-indicator-container--${layout}`]: layout,
+    ['nx-vulnerability-indicator-container--row']: !layout
+  });
+
+  return (
+    <div className={layoutClasses} {...attrs}>
```
rpokorny, line 24: Would it make sense for this whole thing to be a `<dl>` with the count and text as `<dt>` and `<dd>`?

lib/src/components/NxVulnerabilityIndicator/types.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface Props extends Omit<HTMLAttributes<HTMLDivElement>, 'className' | 'children'> {
+  criticalCount: number;
+  severeCount: number;
+  moderateCount: number;
+  lowCount: number;
+  layout?: string;
+  className?: string | null;
+}
+
+export const propTypes: WeakValidationMap<Props> = {
```
rpokorny, line 19: Use `ValidationMap` instead of `WeakValidationMap` if possible

lib/src/components/NxVulnerabilityIndicator/types.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+import { HTMLAttributes, WeakValidationMap } from 'react';
+import * as PropTypes from 'prop-types';
+
+export interface Props extends Omit<HTMLAttributes<HTMLDivElement>, 'className' | 'children'> {
+  criticalCount: number;
+  severeCount: number;
+  moderateCount: number;
+  lowCount: number;
+  layout?: string;
```
rpokorny, line 15: I'd expect this to need to be `string | null`

lib/src/scss-shared/_nx-colors.scss
```diff
@@ -65,3 +65,9 @@ $nx-selectable-colors: (
   yellow: ($nx-yellow-700, $nx-yellow-50),
   lime: ($nx-green-200, $nx-green-50)
 );
+
+// Vulnerability Colours
+$nx-vulnerability-indicator-critical: $nx-red-800;
+$nx-vulnerability-indicator-severe: $nx-orange-700;
+$nx-vulnerability-indicator-moderate: $nx-yellow-700;
+$nx-vulnerability-indicator-low: $nx-blue-600;
```
rpokorny, line 73: These duplicate the `$nx-threat-color` values above

lib/src/scss-shared/_nx-variables.scss
```diff
@@ -13,6 +13,7 @@ $nx-main-header-height: 72px;
 $nx-paragraph-width-maximum: 562px;
 $nx-form-label-width-maximum: 562px;
 
+$nx-scrollbar-width: 15px;
```
rpokorny, line 16: We already have this as a CSS variable.  Also what does it have to do with this PR?

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
glenhunter: > The component is currently named nx _vulnerability_ indicator, but this sentence talks about _violations_? Which is it intended to work over? IMO both, as well any anything else that could be categorized in this way. I'd propose a different name for the component to reflect this: `NxThreatCounts`

https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/bb302d60abf5b3fb11308011c6e1f612dc76c1a1

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,92 @@
...
+  box-sizing: content-box;
+  border: 1px solid;
+  border-left-width: 8px;
+  border-radius: $nx-border-radius;
+  display: inline-flex;
+  font-size: $nx-font-size-s;
+  height: 30px;
+  padding-right: $nx-spacing-md;
+  padding-left: $nx-spacing-md;
+  width: max-content;
```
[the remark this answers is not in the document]
glenhunter: > What's this for?

Without it the boxes expand to fit their containers.

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
glenhunter: > Should be semi-bold

https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e876a57eacad29d4171059706217dd98cfec0631

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
glenhunter: > What's that point of having a separate class for this if it is the default?

There's so much to override between them I prefer a separate class.

lib/src/scss-shared/_nx-variables.scss
glenhunter: > We already have this as a CSS variable. Also what does it have to do with this PR?

Seems to be an artifact from a merge conflict that I resolved incorrectly: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/031e55381c1f87e5bc35c0dff85c0027cfcb6e10

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
glenhunter: > Would it make sense for this whole thing to be a `<dl>` with the count and text as `<dt>` and `<dd>`?

I see what you're thinking but It seems like a bit of a stretch to me. It's a pretty simple change though so if you feel strongly about it I can make it so.

lib/src/components/NxVulnerabilityIndicator/types.ts
glenhunter: > I'd expect this to need to be `string | null`

Ah, I should have caught that: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/ee516df5ef2ccaff199b3c2cb9b519c3fa563fc8

lib/src/components/NxVulnerabilityIndicator/types.ts
glenhunter: > Use `ValidationMap` instead of `WeakValidationMap` if possible

I will confess to not fully understanding the difference: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/ee516df5ef2ccaff199b3c2cb9b519c3fa563fc8

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
glenhunter: > An arbitrary max-width isn't flexible enough for large-number edge cases:
> image

This is fixed with the conversion to grid containers: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/0c49ff6f35705c7d71d2c31f9fccde3215fe17b7 & https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e876a57eacad29d4171059706217dd98cfec0631

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
glenhunter: > `gap` would be highly preferred to this

Yes it's much better:https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/0c49ff6f35705c7d71d2c31f9fccde3215fe17b7 & https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e876a57eacad29d4171059706217dd98cfec0631

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
glenhunter: > If the containers all used `grid` we could use `gap` which would be simpler than messing with these margins

Yep. https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/0c49ff6f35705c7d71d2c31f9fccde3215fe17b7 & https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e876a57eacad29d4171059706217dd98cfec0631

The 2x2 layout differs slightly from the mockup but I have conferred with @dturner32 and he is cool with the change

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
glenhunter: > > What's that point of having a separate class for this if it is the default?
> 
> There's so much to override between them I prefer a separate class.

Though now with the move to grid it makes more sense to remove the custom class: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/d738dd31f8092cde9df5949377cbda7b6f5b1052

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
glenhunter: > As a component, we need to document the props, and probably do not want to document the internal classes.

Right, done here: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/ba74de4ac6db5e7c7249a4abf7f83f110cb4777f

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
glenhunter: > If you want to simplify the DOM slightly, the text styles for this element could be moved to its parent (and the margin moved over to the count). It's also fine as-is though

I get what you're saying, but I prefer having the font styles directly on the thing to which they apply.

lib/src/scss-shared/_nx-colors.scss
glenhunter: > These duplicate the `$nx-threat-color` values above

You're right. De-duped here: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/08e5952fea5f1e1540cb124cbcbe35cb322dc15f

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
rpokorny: > The 2x2 layout differs slightly from the mockup but I have conferred with @dturner32 and he is cool with the change

I guess that does point out the question of how things are supposed to adjust as some numbers get bigger but not others.  Personally I think it looks better in the mockup, and I'd be willing to put up with adding extra DOM elements to get it to match that.

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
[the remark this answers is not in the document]
rpokorny: > Without it the boxes expand to fit their containers.

`justify-items: start` on the container is probably cleaner

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
rpokorny: > I see what you're thinking but It seems like a bit of a stretch to me. It's a pretty simple change though so if you feel strongly about it I can make it so.

I feel strongly enough about it; it seems like a perfect semantic fit to me. That said the discussion about changing the DOM to better match the grid mockup may cause issues with this idea so lets hold off on it for now

lib/src/components/NxVulnerabilityIndicator/types.ts
rpokorny: > I will confess to not fully understanding the difference: ee516df

One actually works (meaning gives errors if your propTypes and TS types don't match) and the other doesn't seem to

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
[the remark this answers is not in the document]
glenhunter: > > Without it the boxes expand to fit their containers.
> 
> `justify-items: start` on the container is probably cleaner

It was, but in playing around with that I found a better solution that provides a cleaner look: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/1c0d8a8da7f415a5bc409b6a2f5f9317ce5bd3a7

I have discussed with @dturner32 and he is going to adjust the mocks accordingly

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
glenhunter: > > The 2x2 layout differs slightly from the mockup but I have conferred with @dturner32 and he is cool with the change
> 
> I guess that does point out the question of how things are supposed to adjust as some numbers get bigger but not others. Personally I think it looks better in the mockup, and I'd be willing to put up with adding extra DOM elements to get it to match that.

It turns out that removing `max-width: content` as suggested below lead to a layout that I think looks better all round. The row counters fit to their natural widths and the column layouts expand to the widest counter in the column. As a result the column & grid layouts have clean right edges and a nice square grid. The cost is some extra white-space on the right interior of the counters. I showed this to @dturner32 and he likes it and will be changing the mockups: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/1c0d8a8da7f415a5bc409b6a2f5f9317ce5bd3a7

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
glenhunter: > > I see what you're thinking but It seems like a bit of a stretch to me. It's a pretty simple change though so if you feel strongly about it I can make it so.
> 
> I feel strongly enough about it; it seems like a perfect semantic fit to me. That said the discussion about changing the DOM to better match the grid mockup may cause issues with this idea so lets hold off on it for now

Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/443c0e20143d2e2638abcac16552510b4a8bc021

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+const NxThreatCounterCode = require('./NxThreatCounterExample?raw'),
+    NxThreatCounterColumnCode = require('./NxThreatCounterColumnExample?raw'),
+    NxThreatCounterGridCode = require('./NxThreatCounterGridExample?raw');
+
+const NxThreatCounterPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <p className="nx-p">
+        Basic style for small counter which displays a number representing a count of violations, and a short pre-set
+        text string which displays the threat severity.
```
rpokorny, line 24: This text could use some updating.  It's now a component not a style, and it represents several numbers together not just one.

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <tr className="nx-table-row">
+            <th className="nx-cell nx-cell--header">Prop</th>
+            <th className="nx-cell nx-cell--header">Type</th>
+            <th className="nx-cell nx-cell--header">Required</th>
+            <th className="nx-cell nx-cell--header">Details</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr className="nx-table-row">
+            <td className="nx-cell">criticalCount | severeCount | moderateCount | lowCount</td>
```
rpokorny, line 41: Each of these props should be a separate row in the table

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <td className="nx-cell">numerical</td>
```
rpokorny, line 42: Where possible we should stick to the actual type names in TS/JS
```suggestion
            <td className="nx-cell">number</td>
```

rpokorny:
@dturner32 appears to have added an example with a large number to the mockup.  In the example the number has comma separators (i.e. "9,999,999").  That does not happen by default/in the current implementation/elsewhere in RSC.  Should we actually add that here or should the mockup be changed not to have that?

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.nx-threat-counter--low {
+  border-color: $nx-threat-color-low;
+}
+
+.nx-threat-counter-container {
+  display: inline-grid;
+  row-gap: $nx-spacing-md;
+  column-gap: $nx-spacing-md;
```
rpokorny, line 53:
```suggestion
  gap: $nx-spacing-md;
```

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.nx-threat-counter-container--column {
+  grid-template-rows: auto auto;
+  grid-template-columns: auto;
+}
+
+.nx-threat-counter-container--grid {
+  grid-template-columns: auto auto;
+  grid-template-rows: auto auto;
```
rpokorny, line 65: Given that the number of items is constant and `grid-template-columns` is set, `grid-template-rows` is redundant.  Also the value for it here in `.nx-thread-counter-container--column` doesn't make sense

lib/src/components/NxThreatCounter/NxThreatCounter.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default function NxThreatCounter({
+  criticalCount, severeCount, moderateCount, lowCount, layout, className, ...attrs }: Props) {
+
+  const layoutClasses = classnames('nx-threat-counter-container', className, {
+    [`nx-threat-counter-container--${layout}`]: layout
+  });
+
+  return (
+    <div className={layoutClasses} {...attrs}>
+      <dl className="nx-threat-counter nx-threat-counter--critical">
```
rpokorny, line 24: I intended that the container would be the `<dl>` and the counters would be `<div>`s

lib/src/components/NxThreatCounter/NxThreatCounter.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <dt className="nx-threat-counter__count">{criticalCount}</dt>
+        <dd className="nx-threat-counter__text">Critical</dd>
+      </dl>
+      <dl className="nx-threat-counter nx-threat-counter--severe">
+        <dt className="nx-threat-counter__count">{severeCount}</dt>
+        <dd className="nx-threat-counter__text">Severe</dd>
```
rpokorny, line 30: You have the meaning of the `<dt>`s and `<dd>`s backwards.  "Severe" needs to be the `<dt>`.  However, the `<dt>` needs to appear in the DOM before the `<dd>` so you'll need to play with `flex-direction` too.

glenhunter:
> @dturner32 appears to have added an example with a large number to the mockup. In the example the number has comma separators (i.e. "9,999,999"). That does not happen by default/in the current implementation/elsewhere in RSC. Should we actually add that here or should the mockup be changed not to have that?

IMO we should not dictate comma (or other) separators. Different countries use different separators and if we dictate them it may come back to bite us.

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
glenhunter: > Each of these props should be a separate row in the table

https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e05d98d23ec6c78941170469b2d5eac8cda9fc20

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
glenhunter: > This text could use some updating. It's now a component not a style, and it represents several numbers together not just one.

https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e05d98d23ec6c78941170469b2d5eac8cda9fc20

lib/src/components/NxThreatCounter/NxThreatCounter.tsx
glenhunter: > I intended that the container would be the `<dl>` and the counters would be `<div>`s

Ah, https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e03e606a5c880e49c5b2be926c41c27df75c36de

lib/src/components/NxThreatCounter/NxThreatCounter.tsx
glenhunter: > You have the meaning of the `<dt>`s and `<dd>`s backwards. "Severe" needs to be the `<dt>`. However, the `<dt>` needs to appear in the DOM before the `<dd>` so you'll need to play with `flex-direction` too.

https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/ed67c2380ec6f9e57c23a09aa7f893e0b765faf3

dturner32:
> > @dturner32 appears to have added an example with a large number to the mockup. In the example the number has comma separators (i.e. "9,999,999"). That does not happen by default/in the current implementation/elsewhere in RSC. Should we actually add that here or should the mockup be changed not to have that?
> 
> IMO we should not dictate comma (or other) separators. Different countries use different separators and if we dictate them it may come back to bite us.

I'm not married to keeping the commas in. We can remove them and I'll update the mock to reflect

dturner32:
> > > @dturner32 appears to have added an example with a large number to the mockup. In the example the number has comma separators (i.e. "9,999,999"). That does not happen by default/in the current implementation/elsewhere in RSC. Should we actually add that here or should the mockup be changed not to have that?
> > 
> > 
> > IMO we should not dictate comma (or other) separators. Different countries use different separators and if we dictate them it may come back to bite us.
> 
> I'm not married to keeping the commas in. We can remove them and I'll update the mock to reflect

just updated the zeplin mock

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -0,0 +1,73 @@
...
+  gap: $nx-spacing-md;
+  grid-template-columns: auto auto auto auto;
+  margin: $nx-spacing-md 0 $nx-spacing-l 0;
+}
+
+.nx-threat-counter-container--column {
+  grid-template-columns: auto;
+
+  .nx-threat-counter {
+    justify-content: flex-end;
```
rpokorny, line 63: This could just be on `.nx-threat-counter` in general rather than duplicated on the two cases where it matters

rpokorny (approved):

lib/src/components/NxThreatCounter/NxThreatCounter.scss
glenhunter: > This could just be on `.nx-threat-counter` in general rather than duplicated on the two cases where it matters

https://github.com/sonatype/sonatype-react-shared-components/pull/343/commits/e11ab4e3f9bc987b86de8262981382aa57e92916

PR #346, opened 2021-03-30
Flexible nx-page-main - RSC-534

rpokorny:
https://issues.sonatype.org/browse/RSC-534

While reviewing/testing, you can locally revert dc8fc5e in order to add an `NxGlobalSidebar` to the gallery to see how that plays into things.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "4.7.4",
+  "version": "4.8.0",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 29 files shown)


glenhunter (approved):

glenhunter:
When I revert the commit as suggested above I see the width of the "regular" sidebar jump in size.

Screen Shot 2021-04-05 at 2 49 42 PM
Screen Shot 2021-04-05 at 2 49 18 PM

rpokorny:
As discussed with @dturner32 in WW, the solution here is to limit the max width of the sidebar to what it would be in a non-full-width `nx-page-content` parent. 744e8d1e

rpokorny:
> Also since the other variables are CSS variables why make this one a SCSS variable?

I guess I was thinking that since it was unitless it couldn't be a CSS variable.  But that's not true. 914d930a

glenhunter (approved):

PR #356, opened 2021-04-05
Remove Nonfullscreen Submit Mask - RSC-538

rpokorny:
https://issues.sonatype.org/browse/RSC-538


lib/src/components/NxSubmitMask/NxSubmitMask.scss
```diff
@@ -12,11 +12,6 @@
 .nx-submit-mask {
   @include background-mask(0);
 
-  position: absolute;
-  justify-content: center;
-}
-
-.nx-submit-mask--fullscreen {
```
glenhunter, line 19: Since all Submit Masks are now fullscreen the attributes below (position, and justify-content) are redundant because they're set in the mixin. Also the `z-index` should be moved up into the `background-mask()`.

rpokorny: > Since all Submit Masks are now fullscreen the attributes below (position, and justify-content) are redundant because they're set in the mixin. Also the z-index should be moved up into the background-mask().

b6fa2499

glenhunter (approved):

PR #357, opened 2021-04-06
Add logoPath and alt props to NxPageHeader RSC-550

DarthHater:
https://issues.sonatype.org/browse/RSC-550

Hello!

This PR adds `logoPath` consistent with `NxNexusPageHeader` to `NxPageHeader`, and also adds a `alt` prop, which if absent defaults to `Sonatype` to ensure existing behavior is maintained!

Figured I'd file this to see if the changes are ok and then add some tests to make sure it's good to go!

Also, I did NOT add `logoPath`, `alt` to `AbstractNxPageHeader` because it appears that you pass `logo` in to that, so I didn't see a reason to move them up there (even though they are now common)

### DIFF
--- a/lib/src/components/NxPageHeader/types.ts
+++ b/lib/src/components/NxPageHeader/types.ts
@@ -18,6 +18,8 @@ export interface ProductInfo {
 
 export type Props = Omit<AbstractNxPageHeaderProps, 'logo' | 'productInfoContent'> & {
   productInfo?: ProductInfo | null;
+  logoPath?: string | null;
+  alt?: string | null;
 };
 
 export const propTypes: ValidationMap<Props> = {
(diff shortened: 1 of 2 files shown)


glenhunter:
What you have here looks good to me, feel free to add your tests.

lib/src/components/NxPageHeader/types.ts
```diff
@@ -18,12 +18,16 @@ export interface ProductInfo {
 
 export type Props = Omit<AbstractNxPageHeaderProps, 'logo' | 'productInfoContent'> & {
   productInfo?: ProductInfo | null;
+  logoPath?: string | null;
+  alt?: string | null;
```
rpokorny, line 22: `logoAlt` would probably be a clearer name

PR #360, opened 2021-04-12
Correct nx-btn-bar Margin - RSC-542

rpokorny:
https://issues.sonatype.org/browse/RSC-542


glenhunter:
You removed the `nx-card` example with the button bar but not the SCSS was that intentional?

rpokorny:
> You removed the nx-card example with the button bar but not the SCSS was that intentional?

No, I hadn't checked to see if there was any.  Is that what you're referring to? efa1ea10

glenhunter:
> > You removed the nx-card example with the button bar but not the SCSS was that intentional?
> 
> No, I hadn't checked to see if there was any. Is this what you're referring to? [efa1ea1](https://github.com/sonatype/sonatype-react-shared-components/commit/efa1ea1084d21e3cad3a8b3c3c705cdc024f945c)

Yes

glenhunter:
For the record I still think that it's a mistake removing the button bar styling from `nx-card`. I won't be in the least surprised if we end up adding it back in very soon.

glenhunter (approved):

PR #362, opened 2021-04-15
Icon-only button sizing and A11Y - RSC-398

rpokorny:
https://issues.sonatype.org/browse/RSC-398

### DIFF
--- a/gallery/src/components/NxButton/NxButtonPage.tsx
+++ b/gallery/src/components/NxButton/NxButtonPage.tsx
@@ -58,9 +58,9 @@ export default function NxButtonPage() {
               <NxTableCell>Empty</NxTableCell>
               <NxTableCell>
                 A string to render as a tooltip and accessibility label for the button. This is generally not necessary
-                for buttons that include text content, but icon-only buttons should use this to make the button's meaning
-                clear in all contexts. Omitting this prop when using an icon-only button is deprecated and will become
-                unsupported in a future release.
+                for buttons that include text content, but icon-only buttons should use this to make the button's
+                meaning clear in all contexts. Omitting this prop when using an icon-only button is deprecated and will
+                become unsupported in a future release.
               </NxTableCell>
             </NxTableRow>
           </NxTableBody>
--- a/gallery/src/components/NxButton/NxButtonPage.tsx
+++ b/gallery/src/components/NxButton/NxButtonPage.tsx
@@ -51,6 +51,18 @@ export default function NxButtonPage() {
               <NxTableCell>secondary</NxTableCell>
               <NxTableCell>The variant of button. See examples of each variant below.</NxTableCell>
             </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>title</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>Required on icon-only buttons</NxTableCell>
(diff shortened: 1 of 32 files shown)


gallery/src/components/NxButton/NxButtonPage.tsx
```diff
@@ -51,6 +51,18 @@ export default function NxButtonPage() {
...
               <NxTableCell>The variant of button. See examples of each variant below.</NxTableCell>
             </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>title</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>Required on icon-only buttons</NxTableCell>
+              <NxTableCell>Empty</NxTableCell>
+              <NxTableCell>
+                A string to render as a tooltip and accessibility label for the button. This is generally not necessary
+                for buttons that include text content, but icon-only buttons should use this to make the buttons meaning
```
glenhunter, line 61:
```suggestion
                for buttons that include text content, but icon-only buttons should use this to make the button's meaning
```

gallery/src/components/NxButton/NxButtonPage.tsx
```diff
@@ -96,8 +108,8 @@ export default function NxButtonPage() {
                           id="nx-button-icon-only-example"
                           liveExample={NxButtonIconOnlyExample}
                           codeExamples={nxButtonIconOnlyCode}>
-        An example of a button containing only an icon. For accessibility purposes it is important to add an
-        aria-label for the screen reader to interpret the content correctly.
+        An example of a button containing only an icon. For accessibility purposes it is important to use the
+        title prop for the screen reader to interpret the content correctly.
```
glenhunter, line 112:
```suggestion
        title prop for a screen reader to interpret the content correctly.
```

lib/src/base-styles/_nx-btn.scss
```diff
@@ -20,6 +20,7 @@
   @include container-horizontal;
   @extend %nx-btn--disabled;
 
+  box-sizing: content-box;
```
glenhunter, line 23: Why this change?

rpokorny:
Remaining question for @dturner32 : which size should icon-only buttons in the page header be? There were several applitools tests of `NxNexusPageHeader` that changed in this PR

lib/src/base-styles/_nx-btn.scss
rpokorny: > Why this change?

So for one thing, it turns out the UA stylesheet (in Chrome at least) had this set to `border-box` by default for `<button>`s.  I think I changed it to make the the width calculations here and in the mixin slightly simpler. However it also made the min-width calculation in the pagination button more complicated, and I just realized I have that calculation wrong (forgot borders).  So I've now changed this including removing this `box-sizing`. 3568b477

glenhunter:
Safari is unhappy with the buttons in the dropdowns. I checked Master and it's fine.

Screen Shot 2021-04-16 at 12 22 32 PM

glenhunter:
Safari also disrespects the 42px icon only button:

Screen Shot 2021-04-16 at 2 21 58 PM

rpokorny:
> Safari is unhappy...
> Safari also disrespects...

Safari placated here: 9549551d

glenhunter (approved):

PR #363, opened 2021-04-16
Page Title Additions RSC-468

glenhunter:
https://issues.sonatype.org/browse/RSC-468

This started as a task to add a sub-title, but grew to fix a variety of spacing issues and other design requests.

I cleaned up the examples a little bit, and added a visual test for the "everything" example.


gallery/src/pageConfig.ts
```diff
@@ -173,15 +173,15 @@ const pageConfig: PageConfig = {
     'nx-card': NxCardPage,
     'nx-code': NxCodePage,
     'nx-counter': NxCounterPage,
+    'nx-fieldset': NxFieldsetStylePage,
+    'nx-form-group': NxFormGroupStylePage,
+    'nx-form-select': NxFormSelectPage,
```
glenhunter, line 178: These were out of alphabetical order so I moved them.

lib/src/base-styles/_nx-page.scss
```diff
@@ -29,19 +31,30 @@
...
+  margin: 0;
+  padding-left: $nx-spacing-md;
+}
+
+.nx-page-title__description {
+  @include container-vertical;
+  grid-area: description;
+  margin-top: $nx-spacing-xs;
+
+  & + .nx-page-title__tags {
```
glenhunter, line 46: The design requires that the first block below the page title have `8px` of top spacing, and the next one should have `16px`. The description always precedes the tags so give both `8px` in case they are alone, and give the tags `16px` only when they follow the description.

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -7,28 +7,32 @@
...
+      </NxP>
+      <NxP>
+        In addition <code className="nx-code">.nx-page-title</code> can have buttons inline with the title.
+      </NxP>
+      <NxP>
         Note: <code className="nx-code">.nx-page-title</code> replaces
         <code className="nx-code">.nx-tile--top-tile</code> and <code className="nx-code">.nx-tile--title-only</code>.
-      </p>
+      </NxP>
       <table className="nx-table nx-table--gallery-props">
```
rpokorny, line 36: Not from this PR, but the second row in this table is missing a cell

glenhunter: > Not from this PR, but the second row in this table is missing a cell

That whole row should have been removed. @dturner32 requested that we remove the icon from page headers. Also the class was never defined in the SCSS.

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -48,13 +52,6 @@ const NxPageTitlePage = () =>
...
             </td>
           </tr>
-          <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-page-title__page-icon</code></td>
-            <td className="nx-cell">Element</td>
-            <td className="nx-cell">
-              Class for an icon that appears to the left of the page title.
-            </td>
-          </tr>
           <tr className="nx-table-row">
```
rpokorny, line 55: Needs a row for the subtitle class

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -69,37 +66,54 @@ const NxPageTitlePage = () =>
...
       A simple example of an <code className="nx-code">nx-page-title</code>.
     </GalleryExampleTile>
 
-    <GalleryExampleTile title="NX Page Title with Actions Example"
+    <GalleryExampleTile title="Page title with actions"
                         id="nx-page-title-actions-example"
                         defaultCheckeredBackground={true}
                         htmlExample={nxPageTitleActionsCode}
                         codeExamples={nxPageTitleActionsCode}>
       An example of <code className="nx-code">nx-page-title</code> with actions in the form of a drop-down.
```
rpokorny, line 90: It isn't a dropdown

lib/src/base-styles/_nx-page.scss
```diff
@@ -11,8 +11,10 @@
   display: grid;
   grid-template-columns: auto 1fr auto;
   grid-template-areas:
-    "title tags actions"
-    "description description description";
+    "title sub-title actions"
+    "description description description"
+    "tags tags tags";
+  column-gap: $nx-spacing-md;
```
rpokorny, line 17: You've got this and also left margin on the nx-btn-bar, adding up to a total of 40px (should be 24)
image

lib/src/base-styles/_nx-page.scss
```diff
@@ -11,8 +11,10 @@
   display: grid;
   grid-template-columns: auto 1fr auto;
```
rpokorny, line 12: A really long main title can completely squeeze out the subtitle.  Do we want to guard against that in any way (like a min-width on the subtitle)?
image

gallery/src/styles/NxPageTitle/NxPageTitleSubtitleExample.html
```diff
@@ -0,0 +1,24 @@
...
+    This program and the accompanying materials are made available under
+    the terms of the Eclipse Public License 2.0 which accompanies this
+    distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+
+-->
+<div class="nx-page-title">
+  <h1 class="nx-h1">
+    Page Title
+  </h1>
+  <h2 class="nx-h2 nx-page-title__sub-title">
```
rpokorny, line 13: Technically I feel like this and the `h1` should be wrapped together in an `hgroup`.  Otherwise everything from here down to the next header will be considered a section under the `h2`, outline wise.

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -7,28 +7,32 @@
...
-        have an icon and descriptive text.
-      </p>
-      <p className="nx-p">
-        In addition <code className="nx-code">.nx-page-title</code> can have tags or actions inline with the title.
-      </p>
-      <p className="nx-p">
+        have a sub-title, descriptive text, and space for tags or other indicators.
+      </NxP>
+      <NxP>
+        In addition <code className="nx-code">.nx-page-title</code> can have buttons inline with the title.
```
rpokorny, line 30: Do we want to document more clearly how to do the buttons (e.g. adding a row to the classes table for `.nx-btn-bar`) or is documenting by example enough?

rpokorny:
In this mockup, what is the "Distributed" thing?  It kind of seems like another tag, but it's styled a little bit differently and is not in the tag section.
image

rpokorny:
You have the grid configured for the description and tag areas to go all the way across.  None of the mockups are clear on whether or not that is intended.

lib/src/base-styles/_nx-page.scss
```diff
@@ -11,8 +11,10 @@
...
   grid-template-areas:
-    "title tags actions"
-    "description description description";
+    "title sub-title actions"
+    "description description description"
+    "tags tags tags";
+  column-gap: $nx-spacing-md;
   margin: 0 0 $nx-spacing-l 0;
```
rpokorny, line 18: The individual tags have 8px of bottom margin which adds to this

lib/src/base-styles/_nx-page.scss
```diff
@@ -29,19 +31,30 @@
   }
 }
 
+.nx-page-title__sub-title {
+  @include nx-truncate-ellipsis;
+  border-left: 1px solid $nx-indigo-100;
```
rpokorny, line 36:
```suggestion
  border-left: $nx-border-default;
```

rpokorny:
What is the intended vertical alignment on the action button?  It seems a little high in the mockup, I'm not sure what it's aligned to there.
image

lib/src/base-styles/_nx-page.scss
glenhunter: @dturner32 ^^

glenhunter:
> https://issues.sonatype.org/browse/RSC-468
> 
> This started as a task to add a sub-title, but grew to fix a variety of spacing issues and other design requests.



> You have the grid configured for the description and tag areas to go all the way across. None of the mockups are clear on whether or not that is intended.

@dturner32 ^^

glenhunter:
> What is the intended vertical alignment on the action button? It seems a little high in the mockup, I'm not sure what it's aligned to there.
> image

@dturner32 ^^

glenhunter:
> In this mockup, what is the "Distributed" thing? It kind of seems like another tag, but it's styled a little bit differently and is not in the tag section.
> image

@dturner32 ^^

lib/src/base-styles/_nx-page.scss
dturner32: > A really long main title can completely squeeze out the subtitle. Do we want to guard against that in any way (like a min-width on the subtitle)?
> image

I'll have to mock something up as an example, but yes I think the subtitle should have a min width in a case like this one

dturner32:
> In this mockup, what is the "Distributed" thing? It kind of seems like another tag, but it's styled a little bit differently and is not in the tag section.
> image

That's from an old ALP mock up, before the tags got updated, and It just slipped my mind to fix it for this mock up. It is a tag and should be styled accordingly and in line with the other tags

dturner32:
> > What is the intended vertical alignment on the action button? It seems a little high in the mockup, I'm not sure what it's aligned to there.
> > image
> 
> @dturner32 ^^

its aligned with the top of the page title

dturner32:
> https://issues.sonatype.org/browse/RSC-468

That was intended

rpokorny:
> > What is the intended vertical alignment on the action button? It seems a little high in the mockup, I'm not sure what it's aligned to there.
> 
> its aligned with the top of the page title

I question whether it should be.  It's adjacent to two other elements that are center-aligned with each other, and in other places (`nx-tile`) we effectively center align the title and the action buttons to each other

dturner32:
> > > What is the intended vertical alignment on the action button? It seems a little high in the mockup, I'm not sure what it's aligned to there.
> > 
> > 
> > its aligned with the top of the page title
> 
> I question whether it should be. It's adjacent to two other elements that are center-aligned with each other, and in other places (`nx-tile`) we effectively center align the title and the action buttons to each other

let's center it for consistency then if its centered elsewhere

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
glenhunter: > It isn't a dropdown

Removed reference to dropdown here: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/cfe9ef4c8fe3058d8ed0cef6ad2ab43643a303fb

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
glenhunter: > Needs a row for the subtitle class

Fixed and added convenience class: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/8fd0ec709b78832a8333fd0668801d4e1ff617be

gallery/src/styles/NxPageTitle/NxPageTitleSubtitleExample.html
glenhunter: > Technically I feel like this and the `h1` should be wrapped together in an `hgroup`. Otherwise everything from here down to the next header will be considered a section under the `h2`, outline wise.

Good point: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/78ca66b4343fccc6421e96044299e5583a7f2486

glenhunter:
> > > > What is the intended vertical alignment on the action button? It seems a little high in the mockup, I'm not sure what it's aligned to there.
> > > 
> > > 
> > > its aligned with the top of the page title
> > 
> > 
> > I question whether it should be. It's adjacent to two other elements that are center-aligned with each other, and in other places (`nx-tile`) we effectively center align the title and the action buttons to each other
> 
> let's center it for consistency then if its centered elsewhere

Ok, that's how it is implemented now.

lib/src/base-styles/_nx-page.scss
glenhunter: > The individual tags have 8px of bottom margin which adds to this

Removed. https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/efd0b5bd7ecab9f0a1acd4d3bbdd6f895fffed5b

lib/src/base-styles/_nx-page.scss
dturner32: > A really long main title can completely squeeze out the subtitle. Do we want to guard against that in any way (like a min-width on the subtitle)?
> image

the min width for the subtitle is 150px 
Screen Shot 2021-04-21 at 3 48 15 PM

dturner32:
> You have the grid configured for the description and tag areas to go all the way across. None of the mockups are clear on whether or not that is intended.

there is no set max width for this

lib/src/base-styles/_nx-page.scss
glenhunter: > You've got this and also left margin on the nx-btn-bar, adding up to a total of 40px (should be 24)
> image

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/efd0b5bd7ecab9f0a1acd4d3bbdd6f895fffed5b

lib/src/base-styles/_nx-page.scss
glenhunter: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/6133fe086af3b375fd557cd5741d286c92489469

lib/src/base-styles/_nx-page.scss
rpokorny: That solution creates problems when there are enough tags to wrap.
image

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -49,10 +53,22 @@ const NxPageTitlePage = () =>
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-page-title__page-icon</code></td>
+            <td className="nx-cell"><code className="nx-code">.nx-page-title__headings</code></td>
+            <td className="nx-cell"><code className="nx-code"></code></td>
```
rpokorny, line 57: The empty `.nx-code` doesn't look right.  Either the whole cell should be empty, or ideally this should have a convenience component

lib/src/base-styles/_nx-page.scss
```diff
@@ -23,25 +25,41 @@
     white-space: nowrap;
   }
 
-  .nx-btn-bar {
+  .nx-btn-bar, .nx-dropdown {
```
rpokorny, line 28: I believe the correct way was to have the dropdown inside of a btn-bar

lib/src/base-styles/_nx-page.scss
```diff
@@ -23,25 +25,41 @@
...
+  }
+}
+
+.nx-page-title__tags {
+  @include container-horizontal;
+
+  grid-area: tags;
+  margin-top: $nx-spacing-xs;
+
+  .nx-policy-violation-indicator, .nx-tag {
```
rpokorny, line 62: I worry about how specific this is.  We might have other things in the future that could fit here

lib/src/base-styles/_nx-page.scss
glenhunter: > I believe the correct way was to have the dropdown inside of a btn-bar

I had it that way and then I second guessed myself. Reverted here: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/32969860b6a0d8b1e41d5cd37d0a8c7ab2af55d0

lib/src/base-styles/_nx-page.scss
glenhunter: > I worry about how specific this is. We might have other things in the future that could fit here

Adjusted here: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/8e28369200a8ae39884be11e3644cc402453e8c0

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
glenhunter: > The empty `.nx-code` doesn't look right. Either the whole cell should be empty, or ideally this should have a convenience component

That was a mistake: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/6ceb255ac0af3789c1fc381f299d11a089419a36

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
glenhunter: > Do we want to document more clearly how to do the buttons (e.g. adding a row to the classes table for `.nx-btn-bar`) or is documenting by example enough?

I think that's worth documenting: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/ca7fdcb455dcc36e625599b5665502e0cf3152c5

glenhunter:
NOTE: while working on this ticket I noticed that `NxPolicyViolationIndicator` and `NxTag` specify their lateral spacing inconsistently. I changed `NxPolicyViolationIndicator` to be consistent with `NxTag` here: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/7cf051caa00ff6ef126b474eb5706481d658e1a1

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -49,10 +53,31 @@ const NxPageTitlePage = () =>
...
+            <td className="nx-cell"><code className="nx-code">NxPageTitle.Subtitle</code></td>
+            <td className="nx-cell">Element</td>
+            <td className="nx-cell">
+              If there is sub-title text it should be wrapped in a containing
+              <code className="nx-code">&lt;H2&gt;</code> with this class.
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell"><code className="nx-code">.nx-btn-bar</code></td>
+            <td className="nx-cell"><code className="nx-code">NxPageTitle.Subtitle</code></td>
```
rpokorny, line 76:
```suggestion
            <td className="nx-cell"><code className="nx-code">NxBtnBar</code></td>
```

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -49,10 +53,31 @@ const NxPageTitlePage = () =>
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-page-title__page-icon</code></td>
+            <td className="nx-cell"><code className="nx-code">.nx-page-title__headings</code></td>
+            <td className="nx-cell"></td>
             <td className="nx-cell">Element</td>
             <td className="nx-cell">
-              Class for an icon that appears to the left of the page title.
+              If there is a sub-title then the <NxCode>&lt;h1&gt; &amp; &lt;h2&gt;</NxCode> should both be wrapped
```
rpokorny, line 60:
```suggestion
              If there is a sub-title then the <NxCode>&lt;h1&gt;</NxCode> &amp; <NxCode>&lt;h2&gt;</NxCode> should both be wrapped
```

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
```diff
@@ -49,10 +53,31 @@ const NxPageTitlePage = () =>
             </td>
           </tr>
           <tr className="nx-table-row">
-            <td className="nx-cell"><code className="nx-code">.nx-page-title__page-icon</code></td>
+            <td className="nx-cell"><code className="nx-code">.nx-page-title__headings</code></td>
+            <td className="nx-cell"></td>
```
rpokorny, line 57: So no convenience component for this then?

lib/src/base-styles/_nx-page.scss
```diff
@@ -25,23 +27,42 @@
...
+.nx-page-title__tags {
+  @include container-horizontal;
+
+  display: flex;
+  flex-wrap: wrap;
+  grid-area: tags;
+  row-gap: $nx-spacing-xs;
+  margin-top: $nx-spacing-xs;
+
+  > :nth-child(1n) {
```
rpokorny, line 65: This seems like a weird way of selecting every child.  Is this just to increase specificity? Would `> *, > :first-child, > :last-child` work?

rpokorny:
> NOTE: while working on this ticket I noticed that NxPolicyViolationIndicator and NxTag specify their lateral spacing inconsistently. I changed NxPolicyViolationIndicator to be consistent with NxTag here: 7cf051c

Did you clear that with @dturner32 ? I don't see why they'd necessarily be the same.

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <div className="nx-page-title__description">
+        <p className="nx-p">This is a page description.</p>
+        <p className="nx-p">
+          jeans sign papier-mache assassin San Francisco rifle physical 3D-printed denim tanto courier concrete dolphin
+          rebar free-market. tank-traps papier-mache dead free-market tanto drone concrete dolphin sunglasses weathered
+          dead jeans office vehicle nodal point. motion film meta- monofilament knife vinyl post- bridge jeans city
+          Tokyo alcohol marketing girl vehicle.
+        </p>
+      </div>
+      <div className="nx-page-title__tags">
```
rpokorny, line 48: This would be a good spot to demonstrate a mix of `NxTag`s and `NxPolicyViolationIndicator`s

lib/src/base-styles/_nx-page.scss
glenhunter: > This seems like a weird way of selecting every child. Is this just to increase specificity? Would `> *, > :first-child, > :last-child` work?

Yes it's specificity. `> *` did not work. It's overridden by the specific `.class`'s of the components

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
glenhunter: > This would be a good spot to demonstrate a mix of `NxTag`s and `NxPolicyViolationIndicator`s

I debated that but I didn't think that we'd ever mix them in one line/container, and I'm not of the opinion that we should demonstrate that. IMO if a page title has both tags and indicators they should be in separate containers / on separate lines.

rpokorny: > I debated that but I didn't think that we'd ever mix them in one line/container, and I'm not of the opinion that we should demonstrate that. IMO if a page title has both tags and indicators they should be in separate containers / on separate lines.

The mockup shows it:
image

lib/src/base-styles/_nx-page.scss
rpokorny: > Yes it's specificity. > * did not work. It's overridden by the specific .class's of the components

What about what I suggested above?

lib/src/base-styles/_nx-page.scss
rpokorny, line 17: It appears that when the subtitle is omitted, the column gaps on either side of that grid cell double up: 
image

glenhunter: > It appears that when the subtitle is omitted, the column gaps on either side of that grid cell double up:
> image

I noticed that but I cannot find a solution. I didn't think that it was worth worrying about.

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
glenhunter: @dturner32 is this a design direction that we can/will mix tags, indicators, etc on a line?

glenhunter:
> > NOTE: while working on this ticket I noticed that NxPolicyViolationIndicator and NxTag specify their lateral spacing inconsistently. I changed NxPolicyViolationIndicator to be consistent with NxTag here: [7cf051c](https://github.com/sonatype/sonatype-react-shared-components/commit/7cf051caa00ff6ef126b474eb5706481d658e1a1)
> 
> Did you clear that with @dturner32 ? I don't see why they'd necessarily be the same.

Why wouldn't they be consistent? Why would we need clearance from design about how spacing is applied?

lib/src/components/SimpleComponents.tsx
```diff
@@ -63,6 +63,7 @@ export const NxList = Object.assign(withClass('ul', 'nx-list'), {
 });
 
 export const NxPageTitle = Object.assign(withClass('div', 'nx-page-title'), {
+  Subtitle: withClass('h2', 'nx-page-title__sub-title'),
```
rpokorny, line 67: There are unit tests for these

rpokorny:
> Why would we need clearance from design about how spacing is applied?

Because they're the source of truth and decision making on layout

lib/src/base-styles/_nx-page.scss
glenhunter: > > Yes it's specificity. > * did not work. It's overridden by the specific .class's of the components
> 
> What about what I suggested above?

It works. Why exactly it is better than what I have?

glenhunter:
> > Why would we need clearance from design about how spacing is applied?
> 
> Because they're the source of truth and decision making on layout
No left spacing is indicated in Zeplin when they are below a Page Title.

Screen Shot 2021-04-26 at 9 34 47 AM

lib/src/base-styles/_nx-page.scss
rpokorny: > I noticed that but I cannot find a solution. I didn't think that it was worth worrying about.

Howe about using margins instead of gap?  Luckily this is the only row of the grid that has multiple items should it should be pretty straightforward

glenhunter:
> > Why would we need clearance from design about how spacing is applied?
> 
> Because they're the source of truth and decision making on layout

Yes, design supplies guidance on the spacing between objects, it's up to us whether that spacing is applied to the left/right/both of the object to get the spacing required. I think that we (you and I) should be consistent about where we apply that spacing.

gallery/src/styles/NxPageTitle/NxPageTitlePage.tsx
glenhunter: > So no convenience component for this then?

https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/320049f434ad835a62447b37f287478041e74fe5

lib/src/base-styles/_nx-page.scss
glenhunter: > It appears that when the subtitle is omitted, the column gaps on either side of that grid cell double up:
> image

Margin works, and I can't remember why I switched that out for `column-gap`: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/b4660408e77ea0f60e58dc9f88587b385780dbd3

lib/src/base-styles/_nx-page.scss
glenhunter: > > > Yes it's specificity. > * did not work. It's overridden by the specific .class's of the components
> > 
> > 
> > What about what I suggested above?
> 
> It works. Why exactly it is better than what I have?

I accidentally merged it: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/b4660408e77ea0f60e58dc9f88587b385780dbd3

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
dturner32: > @dturner32 is this a design direction that we can/will mix tags, indicators, etc on a line?

yes, that is the intention

lib/src/base-styles/_nx-page.scss
rpokorny: > Why exactly it is better than what I have?

I think it's better in two ways
1. The `nth-child` expression was confusing.  I looked at it, said to myself, "wait, what children does that apply to, isn't it all of them?" and then had to go into the browser to confirm, and then ponder why it was written that way.
2. What I suggested is more clear about what other specificities it has to beat (other `:first-child` and `:last-child` selectors)

dturner32:
> > > NOTE: while working on this ticket I noticed that NxPolicyViolationIndicator and NxTag specify their lateral spacing inconsistently. I changed NxPolicyViolationIndicator to be consistent with NxTag here: [7cf051c](https://github.com/sonatype/sonatype-react-shared-components/commit/7cf051caa00ff6ef126b474eb5706481d658e1a1)
> > 
> > 
> > Did you clear that with @dturner32 ? I don't see why they'd necessarily be the same.
> 
> Why wouldn't they be consistent? Why would we need clearance from design about how spacing is applied?

I agree with Glen about maintaining consistency here

rpokorny:
> Yes, design supplies guidance on the spacing between objects, it's up to us whether that spacing is applied to the left/right/both of the object to get the spacing required. I think that we (you and I) should be consistent about where we apply that spacing.

I understand now.  You just moved the spacing to the other side.  I was thinking you'd changed the amount.

lib/src/components/SimpleComponents.tsx
glenhunter: > There are unit tests for these

I did not know that: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/b5db6af974b1b5654cc5577d5b65d86f02a5cf50

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
glenhunter: > > @dturner32 is this a design direction that we can/will mix tags, indicators, etc on a line?
> 
> yes, that is the intention

Mixed example: https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/2f4ef4d8b695bc63eb54ad3908347a7878512aa4

rpokorny: > Mixed example: 2f4ef4d

The mockups specify 24px between a violation indicator and tags.  Which begs some questions - does that apply only at the switchover from one type to the other? What about in the opposite direction, going from tags to violation indicators?  Or is that ordering just unsupported?

lib/src/components/__tests__/SimpleComponents.test.tsx
```diff
@@ -296,6 +296,18 @@ describe('NxPageTitle', function() {
...
 });
 
+describe('NxPageTitle.Headings', function() {
+  it('makes a <hgroup> tag with an nx-page-title__headings class', function() {
+    expect(shallow(<NxPageTitle.Headings/>)).toMatchSelector('div.nx-page-title__headings');
+  });
+});
+
+describe('NxPageTitle.Subtitle', function() {
+  it('makes a <h2> tag with nx-h3 and nx-page-title__sub-title classes', function() {
```
rpokorny, line 306:
```suggestion
  it('makes a <h2> tag with nx-h2 and nx-page-title__sub-title classes', function() {
```

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
glenhunter: @derrek?

Also can other indicators appear here? `NxThreatCounter` for example? Or the mysterious "Distributed" label that's still in the mockups? 

What happens if `NxTag`s are mixed with `NxPolicyViolationIndicator`s?

dturner32: > > Mixed example: [2f4ef4d](https://github.com/sonatype/sonatype-react-shared-components/commit/2f4ef4d8b695bc63eb54ad3908347a7878512aa4)
> 
> The mockups specify 24px between a violation indicator and a tag. Which begs some questions - does that apply only at the switchover from one type to the other? What about in the opposite direction, going from tags to violation indicators? Or is that ordering just unsupported?

the 24px remains the standard spacing between the indicator groupings and tag groupings and only applies when going from indicators to tags, otherwise the spacing between tags remains the same (8px) and the spacing between indicators remains the same (16px). Indicators will always show up first in terms of order. If there are threat counts, those will show up second, and then tags third.

dturner32: > @derrek?
> 
> Also can other indicators appear here? `NxThreatCounter` for example? Or the mysterious "Distributed" label that's still in the mockups?
> 
> What happens if `NxTag`s are mixed with `NxPolicyViolationIndicator`s?

Other indicators can also appear (i.e. threat counts). The order would be: policy indicators, threat counts, tags. Indicators and tags will remain grouped and never be mixed together, in other words, tags will never show up in the same grouping as indicators (i.e. indicator, tag, indicator)

glenhunter: > > @derrek?
> > Also can other indicators appear here? `NxThreatCounter` for example? Or the mysterious "Distributed" label that's still in the mockups?
> > What happens if `NxTag`s are mixed with `NxPolicyViolationIndicator`s?
> 
> Other indicators can also appear (i.e. threat counts). The order would be: policy indicators, threat counts, tags. Indicators and tags will remain grouped and never be mixed together, in other words, tags will never show up in the same grouping as indicators (i.e. indicator, tag, indicator)

https://github.com/sonatype/sonatype-react-shared-components/pull/363/commits/5f080cb9f8ac77ddbb116d5e0bfa32745b4994d7

I resisted creating wrapping containers for `NxPolicyViolationIndicator`s and `Nxtag`s (`NxThreatCounter`s are wrapped in a `dl`), to keep the complexity down.

rpokorny (approved):

PR #369, opened 2021-04-22
NxTable improved child inspection - RSC-530

rpokorny:
https://issues.sonatype.org/browse/RSC-530

This PR improves the way that the NxTable components inspect their children in order to determine column count and emptiness.  The previous approach inspected the child virtual DOM.  This had a major limitation in that custom components placed in between the NxTable layers would break the inspection.  For instance, if a consuming app wanted to wrap up its use of `<NxTableHead>` within a given table in a custom component of their own, this would prevent `NxTable` from figuring out the column count.

This PR changes the approach to inspect the native DOM elements after rendering rather than inspecting the virtual DOM.  This ensures that regardless of react layers, the end result is what is checked.

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -72,6 +72,7 @@
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@material-ui/core": "^4.11.2",
     "@react-hook/resize-observer": "^1.1.0",
+    "@rooks/use-mutation-observer": "^4.11.0",
     "@sonatype/react-commonmark": "^3.0.1",
     "@types/classnames": "^2.2.11",
     "@types/prop-types": "^15.7.3",
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "5.5.0",
+  "version": "5.5.1",
   "description": "Sonatype shared UI components and utilities written in React",
   "main": "index.js",
   "repository": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -92,5 +92,8 @@
   "sideEffects": [
     "index.js",
     "*.scss"
-  ]
+  ],
+  "resolutions": {
+    "jest-environment-jsdom": "26.6.2"
+  }
 }
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "5.4.1",
+  "version": "5.4.2",
   "description": "Sonatype shared UI components and utilities written in React",
   "main": "index.js",
   "repository": {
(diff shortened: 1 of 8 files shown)


lib/src/components/NxTable/NxTable.tsx
```diff
@@ -4,30 +4,34 @@
...
+      tableRef = useRef<HTMLTableElement>(null),
+      [columnCount, setColumnCount] = useState(1),
+      updateColumnCount = useCallback(function updateColumnCount() {
+        if (tableRef.current) {
+          setColumnCount(tableRef.current.querySelectorAll('thead > tr:first-child > th').length);
+        }
+      }, []);
+
+  useEffect(updateColumnCount, []);
+  useMutationObserver(tableRef, updateColumnCount, mutationObserverConfig);
```
rpokorny, line 29: This sets up a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) that will trigger the column count update logic any time a DOM node is added or removed within the table.

I had initially gone with a simpler approach: `useEffect` with the `children` as its dependency, so that the logic re-runs every time the `children` changes.  I realized however that this was problematic in two ways:
1. it would re-run basically any time anything happened at the `NxTable` level, as the child nodes, even if semantically identical, would typically be new instances
2. it would fail to catch changes that result strictly from stateful logic within a child component.  The `MutationObserver` guarantees that not matter how the child DOM gets changed, we'll notice

lib/package.json
```diff
@@ -91,5 +92,8 @@
   "sideEffects": [
     "index.js",
     "*.scss"
-  ]
+  ],
+  "resolutions": {
+    "jest-environment-jsdom": "26.6.2"
```
rpokorny, line 97: `jest-environment-enzyme` still depends on an older version of this that pulls in a `jsdom` version that doesn't have `MutationObserver`.  So we need to override this to use a more current `jsdom`.

Blacktiger (approved):

lib/src/components/NxTable/NxTable.tsx
```diff
@@ -17,18 +17,25 @@ import { ColumnCountContext } from './contexts';
...
   const {className, children, ...attrs} = props,
-      thead = only(children, NxTableHead),
-      tbody = only(children, NxTableBody),
-      trow = thead && only(thead.props.children, NxTableRow),
-      columns = trow ? React.Children.count(trow.props.children) : 0;
+      tableRef = useRef<HTMLTableElement>(null),
+      [columnCount, setColumnCount] = useState(1),
+      updateColumnCount = useCallback(function updateColumnCount() {
+        if (tableRef.current) {
+          setColumnCount(tableRef.current.querySelectorAll('thead > tr:first-child > th').length);
```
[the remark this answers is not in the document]
rpokorny: > It occurs to me that this could be incorrect if the first header row contains cells that have a colspan value set. One option would be to scan all the table rows and use the max number of cells in any particular row for the column count.

I'm not really inclined to scan the whole table for performance reasons.  We could maybe look for and add up `colspan` attributes in the first row.

PR #372, opened 2021-04-26
Global Sidebar Page Header - RSC-487

rpokorny:
https://issues.sonatype.org/browse/RSC-487

Followup tickets:
https://issues.sonatype.org/browse/RSC-565
https://issues.sonatype.org/browse/RSC-566


gallery/visualtests/nxPageLocalHeader.js
```diff
@@ -0,0 +1,27 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+describe('NxPageLocalHeader', function() {
+  it('looks right with full content', async function() {
+    await browser.url('#/NxPageLocalHeaderFullExample');
+    await browser.eyesSnapshot(null);
+  });
+
+  it('looks right without action bar', async function() {
+    await browser.url('#/NxPageLocalHeaderNoActionsExample');
```
raveneyex, line 14: `async` and `await`? Fancy lads!

lib/src/components/__tests__/SimpleComponents.test.tsx
```diff
@@ -337,3 +338,15 @@ describe('NxTableContainer.Footer', function() {
     expect(shallow(<NxTableContainer.Footer/>)).toMatchSelector('div.nx-table-container__footer');
   });
 });
+
+describe('NxPageLocalHeader', function() {
+  it('makes an <aside> with an nx-page-local-header class', function() {
+    expect(shallow(<NxPageLocalHeader/>)).toMatchSelector('aside.nx-page-local-header');
```
BigAB, line 344: Would you consider `header`, `section` or even `div` (if it is just for layout)  instead of an `aside` element?

> The HTML \<aside\> element represents a portion of a document whose content is only indirectly related to the document's main content. [[mdn]](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)

> The \<aside\> element is used to identify content that is related to the primary content of the webpage, but does not constitute the primary content of the page. Author information, related links, related content, and advertisements are examples of content that may be found in an aside element. [[html.com]](https://html.com/tags/aside/)

Our primary use would be global Nav options, not generally related to the main content.

BigAB (approved):
LGTM

lib/src/components/__tests__/SimpleComponents.test.tsx
BigAB: That could've actually references https://github.com/sonatype/sonatype-react-shared-components/pull/372/files#diff-ff772c2424b5566111a42aca9953a7e851d8d341d9fea979605a3e706f904c56R79 instead of line 344, but I didn't recognize the change in `SimpleComponents.tsx` for what it was

rpokorny: > Would you consider header, section or even div (if it is just for layout) instead of an aside element?

Could you clarify in what way this header doesn't seem like a good fit for `aside`?  It contains a back button to another page and the sort of global navigations/buttons that exist in the top-right corner of the IQ UI.  To me those seem like a fit for the purpose of `aside`.

> Our primary use would be global Nav options

Right, the global sidebar is also an `aside`.

BigAB: The `aside` element is supposed to be used for supplementary content related to, if tangentially, the main content.

Our uses of the aside are not related to the main content, nor are they really content.

We have what probably should be a `section` in the case of the global side-nav with nav element for the list of "main nav links" in that, and possibly `header` or `div` element for the "header" section and a nav for the "utility nav links" menus.  The back button / breadcrumbs could also be in `nav` elements in the `header` or not.

The "side nav" and "main header" appear on every page so they are not really related to the main content, so aside seems inappropriate. The html.com link above 

That all being said, semantic elements are great and all, but sometimes the minutia of it, and especially an not-well-defined always kinda-iffy element like `aside`, is not worth debating about, so I wouldn't consider it a blocker, more of a request, or even request for discussion.

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderEmptyExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+  NxPageMain
+} from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink } from '@fortawesome/free-solid-svg-icons';
+
+const sidebarLogoPath = require('../../assets/logo-plaid-villain-text.png');
+
+export default function NxPageLocalHeaderEmptyExample() {
+  return (
+    <>
+      <aside className="nx-page-local-header"></aside>
```
glenhunter, line 21: Why "local"? Since this is meant to be used with the "global" sidebar I'd have expected a classname that would have connected the two in a clearer way.

rpokorny: > Why "local"? Since this is meant to be used with the "global" sidebar I'd have expected a classname that would have connected the two in a clearer way.

It was a difficult choice.  The problem is not only how to link it to the global sidebar, but how to differentiate it from the old page header.  Really, in the old designs the page header was global while the sidebar was local to the page.  In the new designs the situation is reversed.  So it really seemed wrong to call this header the "global" one when it is in pretty much every way less global than the old header.  It also contains some page-specific content in the form of the back button.  I'm open to suggestions but I really don't think anything mentioning "global" is right for this.

lib/src/components/__tests__/SimpleComponents.test.tsx
glenhunter: I think that this is an appropriate use of `<aside>`. 

W3C says: "Example of contents: pull quotes, **sidebars, advertising**, **groups of nav elements**, etc. Other content that is considered separate from the main content of the page." Emphasis mine.

MDN describes it likewise, "The HTML <aside> element represents a portion of a document whose content is only **indirectly related to the document's main content**. Asides are frequently presented as sidebars or call-out boxes." again emphasis mine.

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderFullExample.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { faArrowLeft, faArrowRight, faLink, faEdit } from '@fortawesome/free-solid-svg-icons';
+
+const sidebarLogoPath = require('../../assets/logo-plaid-villain-text.png');
+
+export default function NxPageLocalHeaderFullExample() {
+  return (
+    <>
+      <aside className="nx-page-local-header">
+        <NxBackButton href="#/pages/nx-page-local-header" targetPageTitle="Documentation" />
+        <div className="nx-page-local-header__actions">
```
glenhunter, line 27: Not `btn-bar`?

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderEmptyExample.tsx
glenhunter: The more I think about the more I dislike "local" here. I don't think that there's any intent that this header would be used with anything but `NxGlobalSidebar` so I think that they should be tied intrinsically together.

lib/src/components/__tests__/SimpleComponents.test.tsx
rpokorny: Right I still think this is a correct use of `aside` after re-reading the specs and docs posted in this thread.  If ads are relevant enough to be an aside, then surely this header is.  And at the same time I don't believe this header is part of the "main content" of the page.

> The "side nav" and "main header" appear on every page so they are not really related to the main content, so aside seems inappropriate.

I think it is actually extremely common for `aside`s to be things that appear on every page.

> nor are they really content.

What's your definition of content?  Everything rendered on the page is content I think.

> We have what probably should be a section in the case of the global side-nav with nav element for the list of "main nav links" in that

FWIW while the global sidebar itself is an `aside`, the nav links within are a `nav`.  

> and possibly header or div element for the "header" section 

From WHATWG on `header` elements

> The header element represents a group of introductory or navigational aids.

> A header element is intended to usually contain the section's heading ..., but this is not required.

Hmm so I'm now convinced that it would also be valid as a `header` due to the mention of "navigational aids".  @BigAB would that be your preference? @glenhunter what do you think?

> and a nav for the "utility nav links" menus

I hesitate to insist on that element being a `<nav>` because I'm not confident that everything in it will be navigation.  For instance in IQ that includes the notifications menu.

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderFullExample.tsx
rpokorny: > Not .btn-bar?

I guess I wasn't confident that it would only contain buttons, at also it has different styling from a usual `.nx-btn-bar` due to the left border and surrounding spacing.

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderEmptyExample.tsx
glenhunter: > I'm open to suggestions but I really don't think anything mentioning "global" is right for this.

I'll think about it. In retrospect "Global" was a shitty name choice. Sometimes I think that we should have just numbered the damn things. Sidebar1 goes with Header1, Sidebar2 goes with Header2, etc.

glenhunter: > less global than the old header

Is this true though? The buttons to the right are always present aren't they?

lib/src/components/__tests__/SimpleComponents.test.tsx
glenhunter: I wouldn't have any issues with it being a `header`, that _might_ make more sense than `aside` but it might be a choice based on layout rather than content.

I do not think that it should be `nav`.

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderEmptyExample.tsx
rpokorny: > The more I think about the more I dislike "local" here. I don't think that there's any intent that this header would be used with anything but NxGlobalSidebar so I think that they should be tied intrinsically together.

Renamed: 866f7b7

lib/src/components/__tests__/SimpleComponents.test.tsx
rpokorny: > I wouldn't have any issues with it being a header, that might make more sense than aside but it might be a choice based on layout rather than content.

I think I'm going to stick with `aside`

gallery/src/styles/NxPageLocalHeader/NxPageLocalHeaderEmptyExample.tsx
raveneyex: >  I think that we should have just numbered the damn things. Sidebar1 goes with Header1, Sidebar2 goes with Header2, etc.

+1!

lib/src/components/__tests__/SimpleComponents.test.tsx
BigAB: I am totally fine with sticking to `aside` but just for furthering the conversation and to address the sidebar thing, please have a look at the **Incorrect use of \<aside\>** section in this article: http://html5doctor.com/understanding-aside/

And actually the whole article as well.

raveneyex: > I am totally fine with sticking to `aside` but just for furthering the conversation and to address the sidebar thing, please have a look at the **Incorrect use of <aside>** section in this article: http://html5doctor.com/understanding-aside/
> 
> And actually the whole article as well.

From the [updated article](http://html5doctor.com/aside-revisited/):

> It did not appear to be an appropriate element for holding secondary content related to the site as a whole, typically known as a “sidebar”. From the comments here and elsewhere, you can see this definition was not accepted by developers. The spec writers listened, and aside is now acceptable for secondary content when not nested within an article element.

BigAB: Interesting but I don't think that's relevant to what we are talking about.

raveneyex: > Interesting but I don't think that's relevant to what we are talking about.

What do you mean it isn't relevant? we're discussing wether or not this kind of secondary content is valid as an `<aside>` element. Your article just pointed out it is.

BigAB: The correction and the thing you highlighted is specifically about using an **`aside`** element **within an article element**. Which was part of the original article and then removed.

raveneyex: > The correction and the thing you highlighted is specifically about using an **`aside`** element **within an article element**. Which was part of the original article and then removed.

"aside is now acceptable for secondary content when **not** nested within an article element."
AFAIK we're not using `<article>` — which makes that point relevant to our current discussion.

BigAB: > Your article just pointed out it is.

If you read the **Conclusion** part of the updated article it has:

> For content that is not the primary focus of an article (or page) **but is still related to the article (or page)**, aside is what you need, regardless of its visual design.

Which is where my mind stays, if the aside is not related to the main content, and it is just a global nav and menus on every page, is it the proper element?

rpokorny: > Which is where my mind stays, if the aside is not related to the main content, and it is just a global nav and menus on every page, is it the proper element?

The more clear quote from the updated article:

> When used within an article element, the contents should be specifically related to that article (e.g., a glossary). When used outside of an article element, the contents should be related to the site (e.g., a blogroll, groups of additional navigation, and even advertising if that content is related to the page).

Further, see the third example here for a more canonical source: https://html.spec.whatwg.org/multipage/sections.html#the-aside-element

raveneyex (approved):

lib/src/components/__tests__/SimpleComponents.test.tsx
BigAB: I see those all as supporting my point.

**From the "more clear quote":**
> if that content is related to the page

**From the whatwg spec:**
> consists of content that is tangentially related to the content

> this aside contains two sections that are tangentially related
  to the page, namely, links to other blogs, and links to blog posts
  from this blog
  
The H1 heading of that page being `<h1>My wonderful blog</h1>` and the content being a list of blog posts.


#### Content
As for my definition of **content**, it would be any images, videos, tables of information or prose that relays information on a topic.

I would consider navs, buttons, and inputs as **controls**, not content but rather a way to interact with the data or site as a whole. There are exceptions of course, for example a list of related links could be nav, but it is also content in the sense that it is showing titles of other articles on the content topic.

I think the unfortunate use of "side-bar" in the spec and therefore related blog articles, is the main cause of any confusion, because they don't mean what we traditionally in web think of as a sidebar but more:

>side·bar /ˈsīdˌbär/
> a short article in a newspaper or magazine, typically boxed, placed alongside a main article, and containing additional or explanatory material.

This has been a very good discussion, but please don't hold anything up because of it. Use `aside`, there will be literally no harm done by using it, but I do think that asides are meant for content with a relation

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -16,27 +16,35 @@
...
   min-width: var(--page-width-min);
+  height: 100%;
+  width: 100%;
+
+  // NOTE: Most of this template is only used when using nx-page-content--full-width (or no nx-page-content at all).
+  // In traditional layouts that use plain nx-page-content, that element manages everything below the second
+  // row of this template.
+  grid-template:
+    'old-header     old-header  old-header' auto
+    'global-sidebar   notices      notices' auto
```
glenhunter, line 33: So I see that we're defining the position of "notices" here as above the header but I cannot find any mockups that reflect this. @dturner32

rpokorny: > So I see that we're defining the position of "notices" here as above the header but I cannot find any mockups that reflect this. @dturner32

I actually asked this morning in slack if my assumptions here are correct.  That conversation seems to have stalled out however

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -16,27 +16,35 @@
...
 .nx-page-content {
   @include container-horizontal;
 
+  justify-self: center;
   box-sizing: border-box;
   display: flex;
-  flex-grow: 1;
+
+  // See the comment above on nx-page's grid template
+  grid-area: 3 / 1 / -1 / -1;
```
glenhunter, line 47: I think that this could use further explanation.

glenhunter: Why doesn't this start at row 2?

lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -48,32 +56,41 @@
...
   box-sizing: border-box;
-  flex: none;
 
-  // don't go beyond the max that it would have in a normal .nx-page-content, even when in an
-  // .nx-page-content--full-width
+  // sizing properties for when in .nx-page-sidebar
+  flex: 0 0 var(--sidebar-width);
+
+  // sizing properties for when in .nx-page-side--full-width or directly within .nx-page.
+  // Unfortunately we have to use vw units here because percent is taken relative to the grid cell
```
glenhunter, line 74: Why "unfortunately"?

rpokorny: > Why "unfortunately"?

Because it ties this layout to the assumption that the parent grid's width is equal to the viewport width

lib/src/base-styles/_nx-page-layout.scss
rpokorny: > Why doesn't this start at row 2?

This way the system notices are supported in the old layout as well

rpokorny: > I think that this could use further explanation.

c66781d5

glenhunter: > > Why doesn't this start at row 2?
> 
> This way the system notices are supported in the old layout as well

This puts us in a weird place though, we're allocating space for a notice that RSC doesn't support, for which we have no designs, and for which we don't even have a ticket in the backlog.

rpokorny: > This puts us in a weird place though, we're allocating space for a notice that RSC doesn't support, for which we have no designs, and for which we don't even have a ticket in the backlog.

One of the followup PRs that I have linked in the description is to document the existence of this slot.  For the time being it would be up to products what they put in the slot, so there isn't really a need for mockups at that level.  However, @dturner32 has also just created RSC-570 which will make RSC components (or at least guidance on usage of existing components) in this slot.  I assume there will be mockups for that ticket.

The reason I felt the need to do this now is because I didn't want downstream projects that need a places for system notices in the short term to have to completely rewrite our carefully-crafted grid template in order to fit them in.

glenhunter: > > This puts us in a weird place though, we're allocating space for a notice that RSC doesn't support, for which we have no designs, and for which we don't even have a ticket in the backlog.
> 
> One of the followup PRs that I have linked in the description is to document the existence of this slot. For the time being it would be up to products what they put in the slot, so there isn't really a need for mockups at that level. However, @dturner32 has also just created RSC-570 which will make RSC components (or at least guidance on usage of existing components) in this slot. I assume there will be mockups for that ticket.
> 
> The reason I felt the need to do this now is because I didn't want downstream projects that need a places for system notices in the short term to have to completely rewrite our carefully-crafted grid template in order to fit them in.

The existence of the tickets makes me happy

glenhunter:
The file name of the visual test is still "local"

glenhunter (approved):
Aside from renaming the visual test it all looks good.

rpokorny:
> The file name of the visual test is still "local"

570b8484

PR #373, opened 2021-04-27
nx-grid restyle - RSC-557

rpokorny:
https://issues.sonatype.org/browse/RSC-557


glenhunter:
The mockups show 16px above the first row.

glenhunter:
There's also this business about scrolling tile content areas:

Screen Shot 2021-04-29 at 11 21 54 AM

We should probably demonstrate that since it appears that the line does not scroll so would be an addition to our tile styling

rpokorny:
> The mockups show 16px above the first row.

Which it has, when in a tile as shown in the mockup.  I went back and forth with @dturner32 a few times trying to get the mockups clearer about what they're trying to show, but the rules are 24px spacing above and below grid rows and horizontal keylines, but subject to our usual rules about being the first or last child in a container with padding.

glenhunter:
> > The mockups show 16px above the first row.
> 
> Which it has, when in a tile as shown in the mockup. I went back and forth with @dturner32 a few times trying to get the mockups clearer about what they're trying to show, but the rules are 24px spacing above and below grid rows and horizontal keylines, but subject to our usual rules about being the first or last child in a container with padding.

The Sketch files show `16px` between the tile header and the start of the first grid row. The tile header itself has no bottom spacing, but `nx-tile-content` has `16px` of top padding and `nx-grid-row` has `24px` of top margin. That's going to combine to create `40px` of space between the tile header and the grid.

rpokorny:
> The Sketch files show 16px between the tile header and the start of the first grid row. The tile header itself has no bottom spacing, but nx-tile-content has 16px of top padding and nx-grid-row has 24px of top margin. That's going to combine to create 40px of space between the tile header and the grid.

It doesn't.  Observe the example I added on the `nx-tile` page.  `nx-tile-content` uses the `container-vertical` mixin which is why it works as expected.

glenhunter:
> > The Sketch files show 16px between the tile header and the start of the first grid row. The tile header itself has no bottom spacing, but nx-tile-content has 16px of top padding and nx-grid-row has 24px of top margin. That's going to combine to create 40px of space between the tile header and the grid.
> 
> It doesn't. Observe the example I added on the `nx-tile` page. `nx-tile-content` uses the `container-vertical` mixin which is why it works as expected.

Got it

rpokorny:
> Tiles don't have any scrolling in general. The grid cells themselves would be the scrollables here. I can add an example

8885e5bb

glenhunter (approved):

PR #374, opened 2021-04-28
Remove page title icons RSC-541

glenhunter:
https://issues.sonatype.org/browse/RSC-541

Per directions from design remove the icons from page headers


rpokorny:
So this class already isn't present in our styles right?  If it was, I'd want that removed in this PR too, though that would be a breaking change.  `.nx-page-title__page-icon` is used within IQ.

glenhunter:
> So this class already isn't present in our styles right? If it was, I'd want that removed in this PR too, though that would be a breaking change. `.nx-page-title__page-icon` is used within IQ.

Here's the funny thing. I went to do just that as part of this task and while the class is used, it is never actually defined. There is nothing to remove and I'm not sure that it constitutes a breaking change if we do not remove a class that never existed.

rpokorny:
> I'm not sure that it constitutes a breaking change if we do not remove a class that never existed.

Right if that's the case I agree

rpokorny (approved):

PR #376, opened 2021-04-28
NxSubmitMask a11y - RSC-422

rpokorny:
https://issues.sonatype.org/browse/RSC-422

Adding the `status` role causes the submit mask to be a politely-announced live region, having the desired effect that its contents are announced as they change.  Previously, the mask was not announced at all.


glenhunter (approved):
I'm not sure what you mean by the region being announced. In Voice Over the message is read e.g. "Loading"), which is what I expected.

rpokorny:
> I'm not sure what you mean by the region being announced. In Voice Over the message is read e.g. "Loading"), which is what I expected.

Right, and then it should say Success when the mask gets to that phase.  None of that happened before, in ChromeVox at least.

PR #377, opened 2021-04-28
Fix toolip error message in NxTreeView - RSC-546

rpokorny:
https://issues.sonatype.org/browse/RSC-546


glenhunter (approved):

glenhunter:
Just an observation, Material (I assume) also adds an empty class to the empty `div`. 

Screen Shot 2021-04-29 at 10 38 25 AM

glenhunter:
Should we have an example of this in the disabled section?

rpokorny:
> Should we have an example of this in the disabled section?

The disabled `NxTreeView*Select` examples have it, is that adequate?

PR #379, opened 2021-04-29
Improvement Day: bundle size

rpokorny:
Mess with webpack configuration to decrease bundle size from 16.95MB to 11.29MB, mainly by allowing it to delete unused dependency code from the bundle.


glenhunter (approved):
Big savings!

PR #383, opened 2021-05-04
CSS Variables - RSC-549

rpokorny:
https://issues.sonatype.org/browse/RSC-549

This PR includes updates to the higher-level color variables and the non-color variables, but does not include the updates to the swatch variables.  Those are in another PR. These two PRs will be merged to a feature branch and released together, along with RSC-554


lib/src/base-styles/_nx-globals.scss
```diff
@@ -42,16 +42,6 @@ $nx-scrollable-default-height: 400px;
   float: right;
 }
 
-.nx-error {
```
rpokorny, line 45: No longer used anywhere

lib/src/base-styles/_nx-colors.scss
```diff
@@ -0,0 +1,43 @@
...
+  --nx-color-border-subsection: #{$nx-grey-200};
+  --nx-color-border-form-element: #{$nx-grey-1100};
+  --nx-color-border-form-element-pristine: #{$nx-indigo-100};
+  --nx-color-border-disabled: #{$nx-grey-500};
+  --nx-color-border-focus: #{$nx-teal-300};
+
+  --nx-color-box-shadow-focus: #{transparentize($nx-teal-300, 0.7)};
+
+  --nx-color-background-component: #{$nx-white};
+  --nx-color-background-site: #{$nx-grey-50};
```
glenhunter, line 18: I think that these should be `nx-color-comonent-background` and `nx-color-site-background`

rpokorny: > I think that these should be nx-color-comonent-background and nx-color-site-background

That would be inconsistent with the naming scheme we agreed on

lib/src/base-styles/_nx-colors.scss
```diff
@@ -0,0 +1,43 @@
...
+  --nx-color-background-selected: #{$nx-blue-200};
+  --nx-color-background-active: #{$nx-blue-200};
+  --nx-color-background-hover: #{$nx-grey-100};
+  --nx-color-background-disabled: #{$nx-grey-200};
```
glenhunter, line 22: These are too vague IMO, they could be the background of anything. Maybe something like `nx-color-interactive-selected` etc?

lib/src/base-styles/_nx-colors.scss
```diff
@@ -0,0 +1,43 @@
...
+ */
+:root {
+  --nx-color-border-default: #{$nx-indigo-100};
+  --nx-color-border-subsection: #{$nx-grey-200};
+  --nx-color-border-form-element: #{$nx-grey-1100};
+  --nx-color-border-form-element-pristine: #{$nx-indigo-100};
+  --nx-color-border-disabled: #{$nx-grey-500};
+  --nx-color-border-focus: #{$nx-teal-300};
+
+  --nx-color-box-shadow-focus: #{transparentize($nx-teal-300, 0.7)};
```
glenhunter, line 15: I'd group this with the other interactive states below and name it similarly.

lib/src/base-styles/_nx-variables.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  --nx-z-index-submit-mask: 10;
+  --nx-z-index-modal-backdrop: 1050;
+
+  --nx-border-radius: 6px;
+
+  --nx-border-default: 1px solid var(--nx-color-border-default);
+  --nx-border-subsection: 1px solid var(--nx-color-border-subsection);
+
+  --nx-box-shadow-focus: 0 0 3px 1px var(--nx-focus-box-shadow-color);
```
glenhunter, line 45: `--nx-focus-box-shadow-color` is like a jumbled version of  `--nx-color-box-shadow-focus` below.

glenhunter:
Should we have a deprecated section for the spacing variables (at least)? Otherwise this is going to be a very big breaking change for a number of projects.

rpokorny:
> Should we have a deprecated section for the spacing variables (at least)? Otherwise this is going to be a very big breaking change for a number of projects.

I don't think so.  A breaking change is a breaking change, I'd rather rip that band-aid off.  I do intend to help with at least one of the upgrades too.

lib/src/base-styles/_nx-colors.scss
glenhunter: > > I think that these should be nx-color-comonent-background and nx-color-site-background
> 
> That would be inconsistent with the naming scheme we agreed on

I disagree. Looking at the other variables the naming is `nx-color-uniqueThing-specificPartOfThing`. The background is not the unique thing, the site or component is.

For example: `--nx-color-border-default`, `--nx-color-link-hover`or `--nx-color-threat-moderate`.

Looking at it another way if the site had a border we wanted to encapsulate in a variable we'd want:
`--nx-color-site-background`
`--nx-color-site-border`

lib/src/base-styles/_nx-colors.scss
```diff
@@ -0,0 +1,43 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+:root {
+  --nx-color-border-default: #{$nx-indigo-100};
+  --nx-color-border-subsection: #{$nx-grey-200};
+  --nx-color-border-form-element: #{$nx-grey-1100};
+  --nx-color-border-form-element-pristine: #{$nx-indigo-100};
+  --nx-color-border-disabled: #{$nx-grey-500};
+  --nx-color-border-focus: #{$nx-teal-300};
```
glenhunter, line 13: I think that these two variables should be grouped with the other interactive states below.

I'm wondering if we should name them something like `--nx-color-element-border-disabled` etc.

lib/src/base-styles/_nx-colors.scss
glenhunter: Per my comment here: https://github.com/sonatype/sonatype-react-shared-components/pull/383/files#r633547053 and also on the focus states, maybe:

`--nx-color-element-background-disabled` etc.

lib/src/base-styles/_nx-colors.scss
rpokorny: > I disagree. Looking at the other variables the naming is nx-color-uniqueThing-specificPartOfThing. The background is not the unique thing, the site or component is.

Actually it's `nx-color-thingBeingColored-state`.  Looking now I see that `thingBeingColored` is sometimes a property (e.g. background), sometimes a "thing" (e.g. link), and sometimes both (`border-form-element`, `background-site`.  I guess really, the full structure that I arrived at (without realizing it) is
```
nx-color-[property-][thing-][state]
```
where _at least one_ of `property` and `thing` are required.

> Looking at it another way if the site had a border we wanted to encapsulate in a variable we'd want:
> `--nx-color-site-background`
> `--nx-color-site-border`
Consistent with what we've got for other variables, I'd want to keep those as `--nx-color-background-site` and `--nx-color-border-site`.

lib/src/base-styles/_nx-colors.scss
rpokorny: > I'd group this with the other interactive states below and name it similarly.
>
> --nx-color-element-focus-glow?

Why? That's less consistent than what I have now.  Especially renaming it to "glow" rather than an actual CSS property name

lib/src/base-styles/_nx-variables.scss
rpokorny: > --nx-focus-box-shadow-color is a jumbled version of --nx-color-box-shadow-focus below and I cannot find its definition, are they supposed to be the same?

Yes.  I had actually noticed and fixed that on the swatch branch.  Cherry-picked over to this branch in e3d74e12

lib/src/base-styles/_nx-colors.scss
rpokorny: > I think that these two variables should be grouped with the other interactive states below.
>
> I'm wondering if we should name them something like --nx-color-element-border-disabled etc.

I see nothing wrong with the way I've got them grouped.  All of the borders are together

lib/src/base-styles/_nx-colors.scss
glenhunter: > > I'd group this with the other interactive states below and name it similarly.
> > --nx-color-element-focus-glow?
> 
> Why? That's less consistent than what I have now. Especially renaming it to "glow" rather than an actual CSS property name

I'm not wedded to "glow" but the idea is to name it according to the effect achieved, not how the effect was achieved.

lib/src/base-styles/_nx-colors.scss
glenhunter: > > I think that these two variables should be grouped with the other interactive states below.
> > I'm wondering if we should name them something like --nx-color-element-border-disabled etc.
> 
> I see nothing wrong with the way I've got them grouped. All of the borders are together

Right I get that, but that's not the only way to group things. I'm suggesting that we group all of the interactive states together which I makes more sense in terms of how they're going to be used. For example the disabled border is almost always used with the disabled background, grouping them together makes it easier to see all of the interactive states in one glance rather than looking multiple places within the file.

lib/src/base-styles/_nx-colors.scss
glenhunter: > Consistent with what we've got for other variables

I suspect that most of the places where you see this consistency I think that those variables should be renamed.

I think that the number of variables that should be in the form of `nx-color-[property-][state]` is low and should only be considered for things like `nx-border-default` which are intended to be used in many places and are therefore very generic. I think that every other border variable in that list should be in the form of:

`--nx-color-subsection-border`
`--nx-color-form-element-border`
`--nx-color-form-element-border-pristine`

This matches the `text`, `link`, `threat` & `validation` variables below.

rpokorny: > I think that the number of variables that should be in the form of nx-color-[property-][state] is low and should only be considered for things like nx-border-default which are intended to be used in many places and are therefore very generic

There's some sort of disconnect in the way we're reading things.  By my count, 21 of the 28 color variables have states - not a small amount.

I feel like what you may be advocating for is reordering the `[property-][thing-]` in the naming scheme to `[thing-][property-]`, and then grouping by `thing` instead of by `property`, and making the `thing` of `interactive` explicit (it currently is implicit).  Doing that results in something like this:

```
:root {
  --nx-color-border: #{$nx-indigo-100};

  --nx-color-subsection-border: #{$nx-grey-200};

  --nx-color-component-background: #{$nx-white};

  --nx-color-site-background: #{$nx-grey-50};

  --nx-color-form-element-border: #{$nx-grey-1100};
  --nx-color-form-element-border-pristine: #{$nx-indigo-100};

  --nx-color-interactive-background-selected: #{$nx-blue-200};
  --nx-color-interactive-background-active: #{$nx-blue-200};
  --nx-color-interactive-background-hover: #{$nx-grey-100};
  --nx-color-interactive-background-disabled: #{$nx-grey-200};
  --nx-color-interactive-border-disabled: #{$nx-grey-500};
  --nx-color-interactive-border-focus: #{$nx-teal-300};
  --nx-color-interactive-box-shadow-focus: #{transparentize($nx-teal-300, 0.7)};

  --nx-color-text-dark: #{$nx-grey-1100};
  --nx-color-text: #{$nx-grey-900};
  --nx-color-text-disabled: #{$nx-grey-500};

  --nx-color-link: #{$nx-blue-600};
  --nx-color-link-active: #{$nx-blue-400};
  --nx-color-link-hover: #{$nx-blue-800};
  --nx-color-link-disabled: #{$nx-grey-500};

  --nx-color-threat-unspecified: #{$nx-indigo-100};
  --nx-color-threat-none: #{$nx-blue-300};
  --nx-color-threat-low: #{$nx-blue-600};
  --nx-color-threat-moderate: #{$nx-yellow-700};
  --nx-color-threat-severe: #{$nx-orange-700};
  --nx-color-threat-critical: #{$nx-red-800};

  --nx-color-validation-valid: #{$nx-green-700};
  --nx-color-validation-invalid: #{$nx-red-800};

}

```

This results in more variables that aren't grouped at all, and makes the `interactive` variables longer, but otherwise I'm be ok with it.

lib/src/base-styles/_nx-colors.scss
rpokorny: > I'm not wedded to "glow" but the idea is to name it according to the effect achieved, not how the effect was achieved.

How about just "shadow" instead of "box-shadow"? Since it can also be used as a drop-shadow

glenhunter: > > I'm not wedded to "glow" but the idea is to name it according to the effect achieved, not how the effect was achieved.
> 
> How about just "shadow" instead of "box-shadow"? Since it can also be used as a drop-shadow

Sure, that makes sense.

rpokorny: > Sure, that makes sense.

2cddcd6d

lib/src/base-styles/_nx-colors.scss
glenhunter: > > I think that the number of variables that should be in the form of nx-color-[property-][state] is low and should only be considered for things like nx-border-default which are intended to be used in many places and are therefore very generic
> 
> There's some sort of disconnect in the way we're reading things. By my count, 21 of the 28 color variables have states - not a small amount.
> 
> I feel like what you may be advocating for is reordering the `[property-][thing-]` in the naming scheme to `[thing-][property-]`, and then grouping by `thing` instead of by `property`, and making the `thing` of `interactive` explicit (it currently is implicit). Doing that results in something like this:
> 
> This results in more variables that aren't grouped at all, and makes the `interactive` variables longer, but otherwise I'm ok with it.

Yes, this is exactly what I was suggesting. I agree about the `interactive` variables and I'm open to changing the name, but aside from the length `interactive` works.

rpokorny: > Yes, this is exactly what I was suggesting. I agree about the interactive variables and I'm open to changing the name, but aside from the length interactive works.

7a3d0824 and 286afe5e

lib/src/base-styles/_nx-colors.scss
rpokorny: > Per my comment here: https://github.com/sonatype/sonatype-react-shared-components/pull/383/files#r633547053 and also on the focus states, maybe:
>
> --nx-color-element-background-disabled etc.

Addressed in other thread: https://github.com/sonatype/sonatype-react-shared-components/pull/383#discussion_r635581410 7a3d082

lib/src/base-styles/_nx-colors.scss
rpokorny: > Right I get that, but that's not the only way to group things. I'm suggesting that we group all of the interactive states together which I makes more sense in terms of how they're going to be used. For example the disabled border is almost always used with the disabled background, grouping them together makes it easier to see all of the interactive states in one glance rather than looking multiple places within the file.

Addressed in other thread: https://github.com/sonatype/sonatype-react-shared-components/pull/383#discussion_r635581410 7a3d082

glenhunter (approved):
Thanks for making those naming changes. I find the variables easier to understand now.

PR #387, opened 2021-05-07
Global sidebar footer RSC-562

glenhunter:
https://issues.sonatype.org/browse/RSC-562

- added a class for content between the nav and footer because it has styling
- overrode the text colour variables locally (within the sidebar) so that the default colour is white
- all of the footer content is optional

Sketch: https://www.sketch.com/s/854743a8-74f5-4529-be90-fd703d583732/a/v8EkrRM


gallery/src/components/NxGlobalSidebar/NxGlobalSidebarExample.tsx
```diff
@@ -37,14 +48,24 @@ export default function NxGlobalSidebarExample() {
...
           </NxH3>
           <NxP>
             neural decay saturation point assault camera neon concrete engine
           </NxP>
         </section>
+        <NxGlobalSidebarFooter showSupport={true}
+                               supportText="Support for RSC"
+                               supportLink="https://github.com/sonatype/sonatype-react-shared-components"
+                               releaseText="React Shared Components"
+                               releaseNumber={packageJson.version}
```
rpokorny, line 65: I'd rather just hardcode something here for the example

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+              <NxTable.Row>
+                <NxTable.Cell>Prop</NxTable.Cell>
+                <NxTable.Cell>Type</NxTable.Cell>
+                <NxTable.Cell>Required</NxTable.Cell>
+                <NxTable.Cell>Details</NxTable.Cell>
+              </NxTable.Row>
+            </NxTable.Head>
+            <NxTable.Body>
+              <NxTable.Row>
+                <NxTable.Cell>showSupport</NxTable.Cell>
```
rpokorny, line 179: I don't think we should have this property.  Just base it off the `supportLink`

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+                <NxTable.Cell>Boolean</NxTable.Cell>
+                <NxTable.Cell>Yes</NxTable.Cell>
+                <NxTable.Cell>
+                  If true the support link is shown. At minimum <NxCode>supportLink</NxCode> below should be
+                  provided.
+                </NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>supportIcon</NxTable.Cell>
```
rpokorny, line 188: Do we have a requirement for this to be customizable?

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>
+                  Alternate text that should appear in the support link, by default the text reads "Help and Support".
+                </NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>supportLink</NxTable.Cell>
+                <NxTable.Cell>string</NxTable.Cell>
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>URL</NxTable.Cell>
```
rpokorny, line 208: Needs a better description

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>releaseText</NxTable.Cell>
```
rpokorny, line 211: Should this be referred to as something with a bit more meaning? Like `productName`?

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+                <NxTable.Cell>releaseNumber</NxTable.Cell>
+                <NxTable.Cell>String</NxTable.Cell>
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>A release number.</NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>productText</NxTable.Cell>
+                <NxTable.Cell>String</NxTable.Cell>
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>Text string to denote product family or other custom text.</NxTable.Cell>
```
rpokorny, line 226: It's not clear to me from this description what this would typically be

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>showSonatype</NxTable.Cell>
```
rpokorny, line 229: `showCreatedBy`?

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,89 @@ export default function NxGlobalSidebarPage() {
...
+                <NxTable.Cell>Boolean</NxTable.Cell>
+                <NxTable.Cell>Yes</NxTable.Cell>
+                <NxTable.Cell>
+                  If true the "Created by Sonatype" tagline is displayed unless alternate text is provided below.
+                </NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>sonatypeText</NxTable.Cell>
```
rpokorny, line 237: What's the use case for this customization?  If arbitrary additional content is needed we should probably define a place for that elsewhere

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -149,3 +154,52 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
...
   }
 }
+
+.nx-global-sidebar__other-content {
+  @include container-vertical;
+
+  padding-bottom: $nx-spacing-md;
+
+  + .nx-global-sidebar__footer {
+    border-top: 1px solid $nx-global-sidebar-separator-color;
```
rpokorny, line 164: Is there a reason this isn't a bottom border on the `.nx-global-sidebar__other_content`?

rpokorny:
Since this PR adds not only the footer but the middle section, I think questions about how scrolling of the global sidebar should behave need to be answered here.  I think there are a number of important related questions:
* is there a minimum supported page height?
* is there a max number of supported links in the nav section?
* are there rules on the amount of content that can be crammed into the various string props on this footer?

Depending on the answers to those questions, it may be that we need to worry about scrolling even in the absence of the middle section of the sidebar.  If that's not the case, we still surely need to worry about it when when that section is present.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -149,3 +154,52 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
...
+  font-size: $nx-font-size-xs;
+  margin-bottom: $nx-spacing-xl;
+}
+
+.nx-global-sidebar__product {
+  font-size: $nx-font-size-xs;
+  margin-bottom: $nx-spacing-xs;
+}
+
+.nx-global-sidebar__sonatype {
```
rpokorny, line 203: I definitely would prefer this have "created-by" or something instead of "sonatype" in the class name.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -149,3 +154,52 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
...
+  margin-bottom: $nx-spacing-xxl;
+
+  .nx-text-link {
+    align-items: center;
+    display: flex;
+  }
+}
+
+.nx-global-sidebar__support-text {
+  font-size: $nx-font-size-s;
```
rpokorny, line 190: Having the icon and the text at different font sizes causes alignment problems.  Also I'm not seeing a mockup for this ticket so I can check what sizes they should be.

lib/src/components/NxGlobalSidebar/types.ts
```diff
@@ -53,3 +53,31 @@ export const nxGlobalSidebarNavigationLinkPropTypes: ValidationMap<NxGlobalSideb
   text: PropTypes.string.isRequired,
   href: PropTypes.string.isRequired
 };
+
+export interface NxGlobalSidebarFooterProps {
+  showSupport: boolean;
+  supportIcon?: IconDefinition;
```
rpokorny, line 59: No `| null` on this one?

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > > Is there a reason this isn't a bottom border on the .nx-global-sidebar__other_content?
> 
> I guess the `other-content` doesn't necessarily expand to where this border would be

The other issue is that the border is only present when there is `other-content` followed by a footer.

lib/src/components/NxGlobalSidebar/types.ts
glenhunter: > No `| null` on this one?

Now there is: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/0e6dd7b2dbf27a77c724a8c498aee6a862550936

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > I definitely would prefer this have "created-by" or something instead of "sonatype" in the class name.

I waffled back and forth between the two: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/0e6dd7b2dbf27a77c724a8c498aee6a862550936

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > What's the use case for this customization? If arbitrary additional content is needed we should probably define a place for that elsewhere

I was just covering my bases, but we could make the wording of this one mandatory.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > `showCreatedBy`?

Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/aafb1c00e72ed08119b37e91426f5bf06e5a6d61

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > Should this be referred to as something with a bit more meaning? Like `productName`?

I avoided `productName` both because it's too similar to the text below (which I have since renamed `productName`) but also because it could take the form of `Release 1.1.1` or `Version 1.1.1`, etc

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > I don't think we should have this property. Just base it off the `supportLink`

Ok, done here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/f6c2dc51636eb3dd205c3ef3bcb6073a9fb19c25

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > Needs a better description

Improved: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/f6c2dc51636eb3dd205c3ef3bcb6073a9fb19c25

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarExample.tsx
glenhunter: > I'd rather just hardcode something here for the example

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/b974a4ebcab15cdc3edfe587c17fd6c0e546c2fe

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > It's not clear to me from this description what this would typically be

This better? https://github.com/sonatype/sonatype-react-shared-components/commit/f6c2dc51636eb3dd205c3ef3bcb6073a9fb19c25

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
rpokorny: > Ok, done here: f6c2dc5

I don't think that's the right commit.  Looking around more broadly though it looks like the prop has been removed but the documentation for it is still there.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
rpokorny: > We do not have a requirement that says that it is required, or customizable. I'm anticipating a need.

That seems premature to me.  A lot of the customizability you've allowed for in this component does.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
rpokorny: > but we could make the wording of this one mandatory.

That would make sense to me

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -163,6 +161,92 @@ export default function NxGlobalSidebarPage() {
...
+                <NxTable.Cell>Text that appears before the release number.</NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>releaseNumber</NxTable.Cell>
+                <NxTable.Cell>String</NxTable.Cell>
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>A release number.</NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>productName</NxTable.Cell>
```
rpokorny, line 223: I disagree pretty strongly with this being `productName`.  "Powered by Insight" is not a product name.  Even "Insight" is not a product name.  I'd maybe call this `productTagline`.

lib/src/components/NxGlobalSidebar/NxGlobalSidebarFooter.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <footer className={classes} {...otherProps}>
+      { supportLink ?
+        <div className="nx-global-sidebar__support">
+          <NxTextLink href={supportLink || undefined}>
+            <NxFontAwesomeIcon icon={supportIcon || faQuestionCircle} />
+            <span className="nx-global-sidebar__support-text nx-global-sidebar__expanded-content">
+              {supportText || 'Help and Support'}
+            </span>
+          </NxTextLink>
+        </div> : null
```
rpokorny, line 42:
```suggestion
      { supportLink &&
        <div className="nx-global-sidebar__support">
          <NxTextLink href={supportLink || undefined}>
            <NxFontAwesomeIcon icon={supportIcon || faQuestionCircle} />
            <span className="nx-global-sidebar__support-text nx-global-sidebar__expanded-content">
              {supportText || 'Help and Support'}
            </span>
          </NxTextLink>
        </div>
```
and similarly on the other conditions below...

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > > Ok, done here: [f6c2dc5](https://github.com/sonatype/sonatype-react-shared-components/commit/f6c2dc51636eb3dd205c3ef3bcb6073a9fb19c25)
> 
> I don't think that's the right commit. Looking around more broadly though it looks like the prop has been removed but the documentation for it is still there.

Ah right, removed here and I also highlighted that without `supportLink` the `supportText` will not appear. https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/cfe692dc56f8fc65d0451f439156802c00a22d08

glenhunter:
> Since this PR adds not only the footer but the middle section, I think questions about how scrolling of the global sidebar should behave need to be answered here. I think there are a number of important related questions:
> 
> * is there a minimum supported page height?
> * is there a max number of supported links in the nav section?
> * are there rules on the amount of content that can be crammed into the various string props on this footer?
> 
> Depending on the answers to those questions, it may be that we need to worry about scrolling even in the absence of the middle section of the sidebar. If that's not the case, we still surely need to worry about it when when that section is present.

Per conversation at wall walk I have created a scrolling example here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/c6abe5e014f5c00bfcf4105d2c6b307d42537e2d

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > I disagree pretty strongly with this being `productName`. "Powered by Insight" is not a product name. Even "Insight" is not a product name. I'd maybe call this `productTagline`.

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/86a16180d56cbe08ca38d4b6e41d15ea25adafc9

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > > but we could make the wording of this one mandatory.
> 
> That would make sense to me

Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/9e9973513a85f2a33fbe3404ba77ebb2c5b3481c

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > > We do not have a requirement that says that it is required, or customizable. I'm anticipating a need.
> 
> That seems premature to me. A lot of the customizability you've allowed for in this component does.

@dturner32 care to weigh in?

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
dturner32: the mockups under RSC-plaid in sketch cloud for the side nav are the most recently updated and should be referred to for this

dturner32:
> > Since this PR adds not only the footer but the middle section, I think questions about how scrolling of the global sidebar should behave need to be answered here. I think there are a number of important related questions:
> > 
> > * is there a minimum supported page height?
> > * is there a max number of supported links in the nav section?
> > * are there rules on the amount of content that can be crammed into the various string props on this footer?
> > 
> > Depending on the answers to those questions, it may be that we need to worry about scrolling even in the absence of the middle section of the sidebar. If that's not the case, we still surely need to worry about it when when that section is present.
> 
> Per conversation at wall walk I have created a scrolling example here: [c6abe5e](https://github.com/sonatype/sonatype-react-shared-components/commit/c6abe5e014f5c00bfcf4105d2c6b307d42537e2d)

• the minimum supported height for the side nav is determined by the height of the viewport. If 768px is the minimum viewport height, then the entire side nav would adjust to that meaning that the middle section would adjust to fit in between the main nav content section and the footer

• The number of nav items that appear in the middle section should be left up to the team implementing it, and they would need to be mindful of how much content appears in that section. As for the main nav, that section is capped at 10 items

• the content in the footer should never exceed more than the help and support text, release number and additional 'powered by...' text. Any other content outside of those three main bits would be restricted.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
dturner32: is this in reference to specific content in the footer?

glenhunter: > is this in reference to specific content in the footer?

Yes, there is an icon next to the support text. The question is whether that's the only icon that can be used here in which case we'll hard code it or whether it is customizable.

dturner32: the support icon should not be custom and should always be the question mark

glenhunter: > the support icon should not be custom and should always be the question mark

Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/76bc776d69f2e7850cce375d90144cbe1b9ca8bb

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > Having the icon and the text at different font sizes causes alignment problems. Also I'm not seeing a mockup for this ticket so I can check what sizes they should be.

Current mockup still shows 14px for the text and 16px for the icon: https://www.sketch.com/s/854743a8-74f5-4529-be90-fd703d583732/a/v8EkrRM#Inspector

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
rpokorny: > Done here: 76bc776

Docs remain

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
rpokorny: > but also because it could take the form of Release 1.1.1 or Version 1.1.1, etc

I don't follow, wouldn't that be the `releaseNumber` not the `releaseText`?

lib/src/components/NxGlobalSidebar/NxGlobalSidebarFooter.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <NxTextLink href={supportLink || undefined}>
+            <NxFontAwesomeIcon icon={faQuestionCircle} />
+            <span className="nx-global-sidebar__support-text nx-global-sidebar__expanded-content">
+              {supportText || 'Help and Support'}
+            </span>
+          </NxTextLink>
+        </div>
+      }
+
+      { releaseText || releaseNumber &&
```
rpokorny, line 43: When `releaseText` is present it gets rendered as a plain text node rather than rendering the div structure below. This can be seen in the first example in the gallery.


That this wasn't caught by a unit test brings my attention to the fact that this PR is missing unit tests!

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
rpokorny: > Current mockup still shows 14px for the text and 16px for the icon: https://www.sketch.com/s/854743a8-74f5-4529-be90-fd703d583732/a/v8EkrRM#Inspector

Ok.  The alignment happens to more or less match the mockup too (and looks good) so this is fine.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -115,6 +114,9 @@ export default function NxGlobalSidebarPage() {
             <NxCode>NxGlobalSidebarNavigation</NxCode> is a container for navigation links.
             It accepts all standard <NxCode>&lt;div&gt;</NxCode> HTML attributes.
           </NxP>
+          <NxP>
+            Note that it is strongly suggested that the number of navigation links used be kept to 10 or less.
```
rpokorny, line 118: I think this should be stressed more strongly.  I'd make it a warning alert and use "must" language, not "strongly suggested" language.  I'd also add the reason why - that it's the most that can fit along with the footer at minimum page height.

glenhunter: > I think this should be stressed more strongly. I'd make it a warning alert and use "must" language, not "strongly suggested" language. I'd also add the reason why - that it's the most that can fit along with the footer at minimum page height.

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/1075ed3f27240a4c9f7f650619e8ef3be39fa421

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > > Done here: [76bc776](https://github.com/sonatype/sonatype-react-shared-components/commit/76bc776d69f2e7850cce375d90144cbe1b9ca8bb)
> 
> Docs remain

Docs removed: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/25c3048b360be327c90a2513c0db203d452a97e6

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > > but also because it could take the form of Release 1.1.1 or Version 1.1.1, etc
> 
> I don't follow, wouldn't that be the `releaseNumber` not the `releaseText`?

In `Version 1.1.1` "version" is the `releaseText` and "1.1.1" is the `releaseNumber`.

rpokorny: > In Version 1.1.1 "version" is the releaseText and "1.1.1" is the releaseNumber.

In that case, is "it could take the form of Release 1.1.1 or Version 1.1.1, etc" actually something we want to support? Or do we want to insist/enforce that it's always "<product name> x.y.z"?  The mockups just say "Product Release Number" on that line.

rpokorny: > In that case, is "it could take the form of Release 1.1.1 or Version 1.1.1, etc" actually something we want to support? Or do we want to insist/enforce that it's always "<product name> x.y.z"? The mockups just say "Product Release Number" on that line.

A related question: the mockup shows this section as "Release 123" split across two lines when the sidebar is collapsed.  What is the intended relationship between that text and the text when it's expanded?

glenhunter: > > In Version 1.1.1 "version" is the releaseText and "1.1.1" is the releaseNumber.
> 
> In that case, is "it could take the form of Release 1.1.1 or Version 1.1.1, etc" actually something we want to support? Or do we want to insist/enforce that it's always "<product name> x.y.z"? The mockups just say "Product Release Number" on that line.

There's actually one closed mockup that shows "Release 123" just to confuse matters. I am taking the position that we want to provide for "Text" +  "#" where both are optional and that the "Text" portion is _not_ visible in the collapsed state - which I agree does not match the mockups, but we can't know how long that text string is so it's either don't show it or truncate it. @dturner32 what is your guidance here?

lib/src/components/NxGlobalSidebar/NxGlobalSidebarFooter.tsx
glenhunter: > When `releaseText` is present it gets rendered as a plain text node rather than rendering the div structure below. This can be seen in the first example in the gallery.
> 
> That this wasn't caught by a unit test brings my attention to the fact that this PR is missing unit tests!

Fixed the text here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/1fe5571bb3a04815a954540f315880cc109e4e3a

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
dturner32: in the collapsed view, I say don't show the text in agreement with not knowing how long the text string could potentially be

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -25,6 +25,11 @@ $nx-global-sidebar-focus-border: 1px solid $nx-focus-border-color;
 $nx-global-sidebar-separator-color: $nx-grey-900;
 
 .nx-global-sidebar {
+  // Override default dark grey & black text colours for white in the sidebar
+  $nx-heading-color: $nx-white;
+  $nx-text-color: $nx-white;
+  $nx-link-color: $nx-white;
```
rpokorny, line 31: None of these do anything.  This overrides these values in a lexical sense not in a CSS cascade sense.  In other words, only usages of `$nx-heading-color`, `$nx-text-color`, and `$nx-link-color` that are present within this scss block would be affected by these declarations – and there aren't any.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import NxTextLink from '../../NxTextLink/NxTextLink';
+
+describe('NxGlobalSidebarFooter', function() {
+  const minimalProps: Props = {
+        supportText: 'Support for RSC',
+        supportLink: '#supporturl',
+        releaseText: 'React Shared Components',
+        releaseNumber: '3.1.4',
+        productTagLine: 'Powered by PLAID VILLAIN',
+        showCreatedBy: true
```
rpokorny, line 21: `minimalProps` is intended to be the minimum that you can pass to the component without it crashing or failing to compile.  Many of these are optional props which shouldn't be present here.  For each optional prop there should be tests both with and without that prop in use to verify functionality in both cases.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  it('renders the Support div with a link with the passed href and text', function() {
+    const support = getShallowComponent().find('div.nx-global-sidebar__support'),
+        supportLink = getShallowComponent().find('div.nx-global-sidebar__support').find(NxTextLink);
+
+    expect(support).toContainMatchingElement('span.nx-global-sidebar__support-text');
+    expect(support.find('span.nx-global-sidebar__support-text')).toHaveText('Support for RSC');
+    expect(supportLink).toHaveProp('href', '#supporturl');
+  });
+
+  it('renders the Release div with a text and version number', function() {
```
rpokorny, line 38: This tests when `releaseText` and `releaseNumber` are both present.  You also need 3 more tests for when either, and both, aren't present.  I believe you will find that things do not work correctly when both are absent, due to your usage of square brackets instead of parentheses in the conditional expression around the rendering of `.nx-global-sidebar__release`

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        releaseNumber: '3.1.4',
+        productTagLine: 'Powered by PLAID VILLAIN',
+        showCreatedBy: true
+      },
+      getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxGlobalSidebarFooter, minimalProps);
+
+  it('renders an <footer> with nx-global-sidebar__footer class', function() {
+    expect(getShallowComponent()).toMatchSelector('footer.nx-global-sidebar__footer');
+  });
+
```
rpokorny, line 28: Needs tests for custom class names and attributes.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -46,6 +51,7 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
   display: flex;
   height: $nx-global-sidebar-logo-height;
   margin-bottom: $nx-spacing-xl;
+  padding: 0 $nx-spacing-l;
```
rpokorny, line 54: Could/should be margin?

rpokorny, line 54: We probably want a variable for this side spacing, similar to how we've got `$nx-modal-padding` in `NxModal.scss`

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -149,3 +155,54 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
     visibility: hidden;
   }
 }
+
+.nx-global-sidebar__other-content {
+  @include container-vertical;
+
+  flex-grow: 1;
+  padding: $nx-spacing-md $nx-spacing-l;
+  max-height: 100%;
```
rpokorny, line 164:
```suggestion
  max-height: unset;
```

lib/src/components/NxGlobalSidebar/types.ts
```diff
@@ -53,3 +53,25 @@ export const nxGlobalSidebarNavigationLinkPropTypes: ValidationMap<NxGlobalSideb
...
   href: PropTypes.string.isRequired
 };
+
+export interface NxGlobalSidebarFooterProps {
+  supportText?: string | null;
+  supportLink?: string | null;
+  releaseText?: string | null;
+  releaseNumber?: string | null;
+  productTagLine?: string | null;
+  showCreatedBy: boolean;
```
rpokorny, line 63: I feel like this boolean should probably be optional as well, defaulting one way or the other depending on what we think is more common (`true` I would think)

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > Could/should be margin?

It could be but we use padding for l/r spacing for every other container in the sidebar so I was keeping it consistent. I can't see a reason it matters.

glenhunter: > We probably want a variable for this side spacing, similar to how we've got `$nx-modal-padding` in `NxModal.scss`

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/b92f7b4328c1463abfe3627d566376ae75ecc837

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > None of these do anything. This overrides these values in a lexical sense not in a CSS cascade sense. In other words, only usages of `$nx-heading-color`, `$nx-text-color`, and `$nx-link-color` that are present within this scss block would be affected by these declarations – and there aren't any.

Oh yeah those should have been removed a while ago: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/215b76ae1d8d651406416bab2f53e5f43329836f

lib/src/components/NxGlobalSidebar/types.ts
glenhunter: > I feel like this boolean should probably be optional as well, defaulting one way or the other depending on what we think is more common (`true` I would think)

Ok: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/deb6e5507c9d5586545ae5632577305aced2563f

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > This tests when `releaseText` and `releaseNumber` are both present. You also need 3 more tests for when either, and both, aren't present. I believe you will find that things do not work correctly when both are absent, due to your usage of square brackets instead of parentheses in the conditional expression around the rendering of `.nx-global-sidebar__release`

At least part of this is no longer true bc we've dropped `releaseName` altogether: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/a3582fe31ea98556bd1bdbc6dcace8e73f18cc2d

rpokorny: > At least part of this is no longer true bc we've dropped releaseName altogether: a3582fe

Is the most typical expected usage that it would just be something like "Release 1.2.3" or is it expected to be something like "React Shared Components 1.2.3" or "React Shared Components Release 1.2.3" as you had before?  If it's expected to be one of the latter I think the prop would be better named `releaseText`, and our examples should reflect that

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
rpokorny: > It could be but we use padding for l/r spacing for every other container in the sidebar so I was keeping it consistent. I can't see a reason it matters.

I guess it comes down to whether we'd rather be consistent with the other children within this component, or consistent with the way we do things more broadly, only deviating where really necessary.  I don't feel strongly about it.

lib/src/components/NxGlobalSidebar/NxGlobalSidebarFooter.tsx
```diff
@@ -0,0 +1,65 @@
...
+import classnames from 'classnames';
+
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import NxTextLink from '../NxTextLink/NxTextLink';
+
+import { NxGlobalSidebarFooterProps as Props, nxGlobalSidebarFooterPropTypes } from './types';
+import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
+export { NxGlobalSidebarFooterProps } from './types';
+
+const NxGlobalSidebarFooter: FunctionComponent<Props> = function NxNavigationSidebarLinks(props) {
```
rpokorny, line 18:
```suggestion
const NxGlobalSidebarFooter: FunctionComponent<Props> = function NxGlobalSidebarFooter(props) {
```

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > > At least part of this is no longer true bc we've dropped releaseName altogether: [a3582fe](https://github.com/sonatype/sonatype-react-shared-components/commit/a3582fe31ea98556bd1bdbc6dcace8e73f18cc2d)
> 
> Is the most typical expected usage that it would just be something like "Release 1.2.3" or is it expected to be something like "React Shared Components 1.2.3" or "React Shared Components Release 1.2.3" as you had before? If it's expected to be one of the latter I think the prop would be better named `releaseText`, and our examples should reflect that

Sure: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/b0313b5bed7fc1d67f877619fb6cf42bc4a34a69

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > Needs tests for custom class names and attributes.

Here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/dcb088dbde236bda5ab9f146b439c8b4beb1d501

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > `minimalProps` is intended to be the minimum that you can pass to the component without it crashing or failing to compile. Many of these are optional props which shouldn't be present here. For each optional prop there should be tests both with and without that prop in use to verify functionality in both cases.

Ah, fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/dcb088dbde236bda5ab9f146b439c8b4beb1d501

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > > It could be but we use padding for l/r spacing for every other container in the sidebar so I was keeping it consistent. I can't see a reason it matters.
> 
> I guess it comes down to whether we'd rather be consistent with the other children within this component, or consistent with the way we do things more broadly, only deviating where really necessary. I don't feel strongly about it.

I am content with consistency within the component, but likely nobody but you and I will notice that consistency (or lack thereof).

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    expect(getShallowComponent({ className: 'test-class' }))
+        .toMatchSelector('footer.nx-global-sidebar__footer.test-class');
+  });
+
+  it('renders the Support div with a link with the passed href and text and does not render when empty', function() {
+    const supportEmpty = getShallowComponent().find('div.nx-global-sidebar__support'),
+        support = getShallowComponent({ supportText: 'Support for RSC', supportLink: '#supporturl' })
+            .find('div.nx-global-sidebar__support'),
+        supportLink = getShallowComponent({ supportText: 'Support for RSC', supportLink: '#supporturl' })
+            .find('div.nx-global-sidebar__support').find(NxTextLink);
```
rpokorny, line 28:
```suggestion
        supportLink = support.find(NxTextLink);
```

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
rpokorny: > Sure: b0313b5

> and our examples should reflect that

What I meant here is that at least some of the examples should go back to saying something like "React Shared Components 1.2.3"

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
rpokorny: > Here: dcb088d

Still need a test for arbitrary attributes.  For example https://github.com/sonatype/sonatype-react-shared-components/blob/master/lib/src/components/NxColorPicker/__tests__/NxColorPicker.test.tsx#L30

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > > Sure: [b0313b5](https://github.com/sonatype/sonatype-react-shared-components/commit/b0313b5bed7fc1d67f877619fb6cf42bc4a34a69)
> 
> > and our examples should reflect that
> 
> What I meant here is that at least some of the examples should go back to saying something like "React Shared Components 1.2.3"

Some did, but I think that they all do now: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/4916e577a5ea2d0331ef672b9cc87fe76505d37c

Make sure you look at the page for `NxGlobalSidebarFooter`.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -185,14 +188,21 @@ export default function NxGlobalSidebarPage() {
...
               </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell><NxCode>nx-global-sidebar__other-content</NxCode></NxTable.Cell>
+                <NxTable.Cell>element</NxTable.Cell>
+                <NxTable.Cell>
+                  A simple container for content that will appears below the navigation and above the footer.
+                </NxTable.Cell>
+              </NxTable.Row>
             </NxTable.Body>
           </NxTable>
```
rpokorny, line 199: I think we need to document on this page that there is a footer component and that it is required.

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterMinimalExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+  NxGlobalSidebarNavigationLink,
+  NxGlobalSidebarFooter,
+  useToggle,
+  NxPageMain }
+  from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink, faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons';
+
+const logoImg = require('../../assets/images/logo-plaid-villain-text.png');
+
+export default function NxGlobalSidebarFooterExample() {
```
rpokorny, line 20: I don't see this example used

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterMinimalExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <>
+      <NxGlobalSidebar isOpen={sidebarOpen}
+                       toggleOpenIcon={faArrowLeft}
+                       toggleCloseIcon={faArrowRight}
+                       onToggleClick={onToggleCollapse}
+                       logoImg={logoImg}
+                       logoAltText="RSC Plaid Villain"
+                       logoLink="#">
+        <NxGlobalSidebarNavigation>
```
rpokorny, line 32: IMO all of the footer examples should use identical content outside of the footer, to draw attention to the intended differences between the examples

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+import React from 'react';
+import CodeExample from '../../CodeExample';
+import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
+import { NxP, NxCode, NxTable, NxTile, NxH2, NxH3 } from '@sonatype/react-shared-components';
+
+const NxGlobalSidebarFooterExample = require('./NxGlobalSidebarFooterExample.tsx?raw'),
+    NxGlobalSidebarFooterScrollingExample = require('./NxGlobalSidebarFooterScrollingExample.tsx?raw'),
+    NxGlobalSidebarFooterMinimalExample = require('./NxGlobalSidebarFooterMinimalExample.tsx?raw');
+
+export default function NxGlobalSidebarPage() {
```
rpokorny, line 16:
```suggestion
export default function NxGlobalSidebarFooterPage() {
```

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+                  support text to appear.
+                </NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>releaseText</NxTable.Cell>
+                <NxTable.Cell>String</NxTable.Cell>
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>
+                  Text that indicates what version of the software is currently being used. You may use just a version
+                  number or a mix of text and a number, e.g. "Release 3.1.4".
```
rpokorny, line 62: This is where I think we should mention that this would most typically be the product/application name followed by the version.

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+                </NxTable.Cell>
+              </NxTable.Row>
+              <NxTable.Row>
+                <NxTable.Cell>productTagLine</NxTable.Cell>
+                <NxTable.Cell>String</NxTable.Cell>
+                <NxTable.Cell>No</NxTable.Cell>
+                <NxTable.Cell>
+                  This text is meant to indicate when a product is part of a suite or family of products for
```
rpokorny, line 70:
```suggestion
                  This text is meant to indicate when a product is part of a suite or family of products, for
```

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterMinimalExample.tsx
rpokorny: > I don't see this example used

Nvm my branch was out of date

gallery/src/Application.tsx
```diff
@@ -112,6 +123,21 @@ function Application() {
...
+        <Route exact path="/NxGlobalSidebarFooterScrollingExample">
+          <SectionScrollingWrapper>
+            <NxGlobalSidebarFooterScrollingExample />
+          </SectionScrollingWrapper>
+        </Route>
+        <Route exact path="/NxGlobalSidebarFooterMinimalExample">
+          <SectionScrollingWrapper>
+            <NxGlobalSidebarFooterMinimalExample />
+          </SectionScrollingWrapper>
+        </Route>
```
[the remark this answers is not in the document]
rpokorny: > I see three footer examples here but only two on the footer doc page

Nvm my branch was out of date.

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterScrollingExample.tsx
```diff
@@ -0,0 +1,116 @@
...
+  NxTreeView,
+  NxCheckbox,
+  NxRadio,
+  NxTreeViewChild }
+  from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink, faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons';
+
+const logoImg = require('../../assets/images/logo-plaid-villain-text.png');
+
+export default function NxGlobalSidebarFooterScrollingExample() {
```
rpokorny, line 26: Are there meaningful differences _in the footer_ between this example and the basic one?  If not this feels redundant/misnamed.  We already demonstrate the middle section scrolling on the sidebar page

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -135,6 +137,12 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
     display: none;
   }
 
+  // used for the footer - hidden content still takes up space when sidebar is collapsed
+  .nx-global-sidebar__hide-when-collapsed {
```
rpokorny, line 141: Is this intended to be used by custom content in downstream apps? If not I'm not sure it should be it's own class rather than just these properties applied to the right elements in the right circumstances.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > > Here: [dcb088d](https://github.com/sonatype/sonatype-react-shared-components/commit/dcb088dbde236bda5ab9f146b439c8b4beb1d501)
> 
> Still need a test for arbitrary attributes. For example https://github.com/sonatype/sonatype-react-shared-components/blob/master/lib/src/components/NxColorPicker/__tests__/NxColorPicker.test.tsx#L30

Here it is: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/fe8b4bdc987e12295a8a53ce6f74632864160d17

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -149,3 +157,58 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
     visibility: hidden;
   }
 }
+
+.nx-global-sidebar__other-content {
```
rpokorny, line 161: I think this class only work correctly when used in conjunction with `nx-scrollable`.  Is that intended or should we just build the scrolling CSS into this class itself.  If use with `.nx-scrollable` is intended, it should be documented.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -149,3 +157,58 @@ $nx-global-sidebar-separator-color: $nx-grey-900;
...
+
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  margin-top: auto;
+  padding: $nx-spacing-xxl $nx-global-sidebar-left-right-spacing 0 $nx-global-sidebar-left-right-spacing;
+  text-align: center;
+
+  &:empty {
+    padding-top: 0;
```
rpokorny, line 185: We'll need a fully empty example to run applitools on.  Maybe don't link to it from the documentation page though since it isn't an endorsed way of using the component

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterMinimalExample.tsx
glenhunter: > IMO all of the footer examples should use identical content outside of the footer, to draw attention to the intended differences between the examples

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/74b60f8e85c31534b0a13344944cb02698498850

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
glenhunter: > I think we need to document on this page that there is a footer component and that it is required.

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/cb864450f5887f39b15466869fae4bc868842500

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
glenhunter: > This is where I think we should mention that this would most typically be the product/application name followed by the version.

I'm not sure that's true though, I think we should leave it open. It could be the product name, it could be something like "version", "release", or even "v".

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterScrollingExample.tsx
glenhunter: > Are there meaningful differences _in the footer_ between this example and the basic one? If not this feels redundant/misnamed. We already demonstrate the middle section scrolling on the sidebar page

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/ce9878ac59ecb6215142c3d38f17850a334d360d

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > Is this intended to be used by custom content in downstream apps? If not I'm not sure it should be it's own class rather than just these properties applied to the right elements in the right circumstances.

My intent was that it would be available to us, I did not document it.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > I think this class only work correctly when used in conjunction with `nx-scrollable`. Is that intended or should we just build the scrolling CSS into this class itself. If use with `.nx-scrollable` is intended, it should be documented.

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/e677742ec7626551ce3a9c7fdde7534188919e7d

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > We'll need a fully empty example to run applitools on. Maybe don't link to it from the documentation page though since it isn't an endorsed way of using the component

https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/8029b2495340ea0811ab01a00a0ebe452bd1d92a

I also captured the minimal example, and left a note that the full examples (plus open/closed states) are captured by the `NxGlobalSidebar`

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
rpokorny: > Some did, but I think that they all do now: 4916e57
>
> Make sure you look at the page for NxGlobalSidebarFooter.

This didn't address what I said.  It is my understanding that, most typically, the text for this property will be the product name followed by the version number.  In your examples you just have "Release" followed by the version number.  I am suggesting that we make them more realistic by replace "Release" with a product name ("React Shared Components" or maybe something made up).

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
rpokorny: > I'm not sure that's true though, I think we should leave it open. It could be the product name, it could be something like "version", "release", or even "v".

That's why I said "most typically"

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
rpokorny: > My intent was that it would be available to us, I did not document it.

But in that case why does it exist? Why not just style the existing classes as needed?

glenhunter: > > My intent was that it would be available to us, I did not document it.
> 
> But in that case why does it exist? Why not just style the existing classes as needed?

One way I add multiple identical attributes to multiple classes, this way I create one class with those attributes. Then if for some reason we need this behaviour for something else we have it.

We have utility JS we chose not to expose, why not have utility CSS we chose not to expose?

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
glenhunter: > > I'm not sure that's true though, I think we should leave it open. It could be the product name, it could be something like "version", "release", or even "v".
> 
> That's why I said "most typically"

I'm not even sure that "most typically" is true, we won't know until we see what the designers design. I have expanded the documentation though and used the "most typically" language: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/fc8acd2c5de712af6bb1e38e480c7f64f901a793

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
glenhunter: > > Some did, but I think that they all do now: [4916e57](https://github.com/sonatype/sonatype-react-shared-components/commit/4916e577a5ea2d0331ef672b9cc87fe76505d37c)
> > Make sure you look at the page for NxGlobalSidebarFooter.
> 
> This didn't address what I said. It is my understanding that, most typically, the text for this property will be the product name followed by the version number. In your examples you just have "Release" followed by the version number. I am suggesting that we make them more realistic by replacing "Release" with a product name ("React Shared Components" or maybe something made up).

I changed one of the examples here: https://github.com/sonatype/sonatype-react-shared-components/pull/387/commits/95b8f3c13b4f7dbd8d20a6033540c1f70e74ef95

But I'm not sure where you're getting the understanding that "most typically, the text for this property will be the product name followed by the version number" since the mockups just say "Product Release Name" and I can't recall any discussion on the subject (outside of this PR).

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarScrollingExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+  NxTreeViewChild }
+  from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink, faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons';
+
+const logoImg = require('../../assets/images/logo-plaid-villain-text.png');
+
+export default function NxGlobalSidebarExample() {
+  const [sidebarOpen, onToggleCollapse] = useToggle(true);
+  const [is1Open, onToggle1Collapse] = useToggle(false),
+      [is2Open, onToggle2Collapse] = useToggle(false);
```
rpokorny, line 29: Minor, for the sake of consistency since this is an example
```suggestion
  const [sidebarOpen, onToggleCollapse] = useToggle(true),
      [is1Open, onToggle1Collapse] = useToggle(false),
      [is2Open, onToggle2Collapse] = useToggle(false);
```

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarScrollingExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+              Francisco post- j-pop systemic monofilament free-market A.I. Legba receding bomb beef noodles boy
+              papier-mache.
+            </NxP>
+          </div>
+        </section>
+        <NxGlobalSidebarFooter supportText="Support for RSC"
+                               supportLink="https://github.com/sonatype/sonatype-react-shared-components"
+                               releaseText="3.1.4"
+                               productTagLine="Powered by PLAID VILLAIN">
+        </NxGlobalSidebarFooter>
```
rpokorny, line 108:
```suggestion
        <NxGlobalSidebarFooter supportText="Support for RSC"
                               supportLink="https://github.com/sonatype/sonatype-react-shared-components"
                               releaseText="3.1.4"
                               productTagLine="Powered by PLAID VILLAIN" />
```

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarExample.tsx
```diff
@@ -33,18 +41,17 @@ export default function NxGlobalSidebarExample() {
...
-          </NxH3>
-          <NxP>
-            neural decay saturation point assault camera neon concrete engine
-          </NxP>
-        </section>
+        <NxGlobalSidebarFooter supportText="Support for RSC"
+                               supportLink="https://github.com/sonatype/sonatype-react-shared-components"
+                               releaseText="Release 3.1.4"
+                               productTagLine="Powered by PLAID VILLAIN">
+        </NxGlobalSidebarFooter>
```
rpokorny, line 54:
```suggestion
                               productTagLine="Powered by PLAID VILLAIN" />
```

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterExample.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <NxGlobalSidebarNavigationLink icon={faLink}
+                                         text="NxGlobalSidebarFooter"
+                                         href="#/pages/NxGlobalSidebarFooter"/>
+          <NxGlobalSidebarNavigationLink icon={faExternalLinkSquareAlt} text="NxLoadError" href="#/pages/NxLoadError"/>
+        </NxGlobalSidebarNavigation>
+        <NxGlobalSidebarFooter supportText="Support for RSC"
+                               supportLink="https://github.com/sonatype/sonatype-react-shared-components"
+                               releaseText="React Shared Components 3.1.4"
+                               productTagLine="Powered by PLAID VILLAIN">
+        </NxGlobalSidebarFooter>
```
rpokorny, line 42:
```suggestion
                               productTagLine="Powered by PLAID VILLAIN" />
```

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterEmptyExample.tsx
```diff
@@ -0,0 +1,43 @@
...
+  NxGlobalSidebarNavigationLink,
+  NxGlobalSidebarFooter,
+  useToggle,
+  NxPageMain }
+  from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink, faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons';
+
+const logoImg = require('../../assets/images/logo-plaid-villain-text.png');
+
+export default function NxGlobalSidebarFooterExample() {
```
rpokorny, line 20:
```suggestion
export default function NxGlobalSidebarFooterEmptyExample() {
```

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarScrollingExample.tsx
```diff
@@ -0,0 +1,113 @@
...
+  NxTreeView,
+  NxCheckbox,
+  NxRadio,
+  NxTreeViewChild }
+  from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink, faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons';
+
+const logoImg = require('../../assets/images/logo-plaid-villain-text.png');
+
+export default function NxGlobalSidebarExample() {
```
rpokorny, line 26:
```suggestion
export default function NxGlobalSidebarScrollingExample() {
```

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterMinimalExample.tsx
```diff
@@ -0,0 +1,43 @@
...
+  NxGlobalSidebarNavigationLink,
+  NxGlobalSidebarFooter,
+  useToggle,
+  NxPageMain }
+  from '@sonatype/react-shared-components';
+import { faArrowLeft, faArrowRight, faLink, faExternalLinkSquareAlt } from '@fortawesome/free-solid-svg-icons';
+
+const logoImg = require('../../assets/images/logo-plaid-villain-text.png');
+
+export default function NxGlobalSidebarFooterExample() {
```
rpokorny, line 20:
```suggestion
export default function NxGlobalSidebarFooterMinimalExample() {
```

gallery/src/components/NxGlobalSidebarFooter/NxGlobalSidebarFooterPage.tsx
```diff
@@ -0,0 +1,135 @@
...
+            </a>
+          </NxP>
+          <CodeExample content={NxGlobalSidebarFooterExample} />
+        </NxTile.Content>
+      </NxTile>
+      <NxTile>
+        <NxTile.Header>
+          <NxTile.HeaderTitle>
+            <NxH2>Minimal Global Sidebar Example</NxH2>
+            <NxP>In it's most minimal layout the footer shows only the "Created by Sonatype" text.</NxP>
```
rpokorny, line 121:
```suggestion
            <NxP>In its most minimal layout the footer shows only the "Created by Sonatype" text.</NxP>
```

gallery/visualtests/nxGlobalSidebarFooter.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+describe('NxGlobalFooter', function() {
```
rpokorny, line 7:
```suggestion
describe('NxGlobalSidebarFooter', function() {
```

rpokorny (approved):
+1 aside from the various code suggestions I just left

PR #388, opened 2021-05-11
nx-read-only grid - RSC-545

rpokorny:
https://issues.sonatype.org/browse/RSC-545

### DIFF
--- a/gallery/src/styles/NxReadOnly/NxReadOnlyPage.tsx
+++ b/gallery/src/styles/NxReadOnly/NxReadOnlyPage.tsx
@@ -7,7 +7,7 @@
 import React from 'react';
 
 import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
-import { NxCode } from '@sonatype/react-shared-components';
+import { NxCode, NxP, NxTextLink } from '@sonatype/react-shared-components';
 
 import './NxReadOnlyGridExample.scss';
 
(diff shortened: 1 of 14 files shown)


glenhunter:
Why do we need both the grid example on `nx-read-only` and the Layout Example page? It's confusingly redundant.

gallery/src/styles/NxReadOnly/NxReadOnlyPage.tsx
```diff
@@ -58,14 +76,37 @@ const NxReadOnlyPage = () => {
...
       </GalleryExampleTile>
+
+      <GalleryExampleTile title="Grid Example"
+                          id="nx-read-only-grid-example"
+                          codeExamples={gridCode}
+                          htmlExample={nxReadOnlyGridExampleCode}>
+        An <NxCode>nx-read-only</NxCode> arranged in a grid. The exact layout of the grid is custom to this example.
+        In typical usage, the layout of the grid would be usage-specific requiring the use
+        of <NxCode>grid-template</NxCode> and/or <NxCode>grid-area</NxCode> CSS properties.
+      </GalleryExampleTile>
```
glenhunter, line 109: I think that it's worth calling out the use of `nx-read-only__item` here as well. 

e.g. "Note the use of `nx-read-only__item` wrapping each set of `dt/dd`s."

glenhunter:
If we keep both examples I think that we should call out the existence of the layout page on both the `nx-grid` and `nx-read-only` pages.

rpokorny:
> Why do we need both the grid example on nx-read-only and the Layout Example page? It's confusingly redundant. I understand why you have it, I think that it could just be incorporated into the nx-read-only page.

It's similar to why we have examples on the `NxFormGroup` page and then also the "Form Layout Styles" page.  One to show the component in isolation and another to show how several components can be used together to implement a larger common pattern.

rpokorny:
> If we keep both examples I think that we should call out the existence of the layout page on both the nx-grid and nx-read-only pages.

I added a link to it on the `nx-read-only` page in 55e444e5.  I don't feel like it makes as much sense to link to it from the `nx-grid` page as well.

gallery/src/styles/NxReadOnly/NxReadOnlyPage.tsx
rpokorny: > I think that it's worth calling out the use of nx-read-only__item here as well.

97c9a682

glenhunter (approved):

PR #391, opened 2021-05-17
nx-table cell padding values RSC-555

glenhunter:
https://issues.sonatype.org/browse/RSC-555

Not just the filter rows, verify all table cell padding values against the mockups and adjust accordingly.

### DIFF
--- a/lib/src/base-styles/_nx-tables.scss
+++ b/lib/src/base-styles/_nx-tables.scss
@@ -212,14 +212,6 @@
       @include nx-small-icon-btn;
     }
   }
-
-  &:first-child {
-    padding-left: $nx-spacing-md;
-  }
-
-  &:last-child {
-    padding-right: $nx-spacing-md;
-  }
 }
 
 .nx-cell--label {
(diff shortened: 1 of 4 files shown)


lib/src/base-styles/_nx-tables.scss
```diff
@@ -212,6 +212,14 @@
       @include nx-small-icon-btn;
     }
   }
+
+  &:first-child {
+    padding-left: $nx-spacing-md;
+  }
+
+  &:last-child {
+    padding-right: $nx-spacing-md;
```
rpokorny, line 221: Both of these are being overridden by more specific selectors which also happen to be right

glenhunter: > Both of these are being overridden by more specific selectors which also happen to be right

Ah so they are: https://github.com/sonatype/sonatype-react-shared-components/pull/391/commits/61930a1161ab4e54949f41c387fe6cf897802c59

rpokorny (approved):

PR #392, opened 2021-05-19
Tile Form Error Spacing - RSC-579

rpokorny:
https://issues.sonatype.org/browse/RSC-579

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "6.2.3",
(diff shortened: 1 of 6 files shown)


glenhunter:
At 1366px there's no l/r spacing between the alert and the tile - below 1366px the alert breaks out of the tile.

Screen Shot 2021-05-19 at 2 53 30 PM

rpokorny:
> At 1366px there's no l/r spacing between the alert and the tile - below 1366px the alert breaks out of the tile.

1f92a22b

glenhunter (approved):

PR #395, opened 2021-05-19
Adjacent Viewport-Sized Scrollables - RSC-553

rpokorny:
https://issues.sonatype.org/browse/RSC-553


glenhunter (approved):
So many scrollbars!

PR #400, opened 2021-05-25
Fix NxCloseButton deprecation warning - RSC-584

rpokorny:
https://issues.sonatype.org/browse/RSC-584

### DIFF
--- a/lib/src/util/classUtil.ts
+++ b/lib/src/util/classUtil.ts
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import { curryN } from "ramda";
+
+/**
+ * @return whether the space-separated className string (fullClassName) includes the specified class
+ */
+export const includesClass = curryN(2,
+  (cls: string, fullClassName?: string) => !!(fullClassName && fullClassName.match(`\\b${cls}\\b`))
+);
+
+export const includesDisabledClass = includesClass('disabled');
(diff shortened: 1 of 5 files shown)


raveneyex (approved):

lib/src/util/classUtil.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// given a class name, create a regex that will match HTML class attrs that have that class name
+const classMatchRegex = (cls: string) => new RegExp(String.raw`(\s|^)${cls}(\s|$)`);
+
+/**
+ * @return whether the space-separated className string (fullClassName) includes the specified class
+ */
+export const includesClass = curryN(2,
+    (cls: string, fullClassName?: string) => !!(fullClassName && fullClassName.match(classMatchRegex(cls)))
+) as ((cls: string) => (className?: string) => boolean) & ((cls: string, className?: string) => boolean);
```
raveneyex, line 17: These functional-programming functions are nearly impossible to read in TypeScript.

rpokorny: > These functional-programming functions are nearly impossible to read in TypeScript.

This cast is necessary due to a bug in the typings for `curryN` - it loses the optionalness of `fullClassName`.

Blacktiger: Why not just do something like `fullClassName.split(/\s/).some(className => className === cls)`?

rpokorny: > Why not just do something like fullClassName?.split(/\s/).some(className => className === cls)?

That doesn't address the type signature issue, but if you're asking separately why I used a regex and not split, I figured it might be slightly more performant due to not having to process any more of the string than necessary.  Of course, class strings are never particularly long so it hardly matters.

Blacktiger: I was also thinking that maybe it would get rid of the need for curryN but I guess not.

PR #403, opened 2021-06-02
Fix palette loading logic - RSC-594

rpokorny:
https://issues.sonatype.org/browse/RSC-594


glenhunter (approved):
Can't comment on the exact solution, but can confirm that the problem seems to be fixed and that's good enough for me.

PR #405, opened 2021-06-02
nx-counters in NxTreeview titles RSC-569 

glenhunter:
https://issues.sonatype.org/browse/RSC-569

Added a max-width to the counters and provided a couple of expanded examples to make sure that it works and demonstrate text in the counter.


rpokorny:
The gap between the title and the counter is supposed to be 16px, but it's 24.

glenhunter:
> The gap between the title and the counter is supposed to be 16px, but it's 24.

The mockups showed both but @dturner32 has confirmed that it should be `24px` and will fix the mockups.

rpokorny (approved):

PR #408, opened 2021-06-03
NxTreeview twisty focus styles RSC-580, RSC-597 

glenhunter:
https://issues.sonatype.org/browse/RSC-580
https://issues.sonatype.org/browse/RSC-597

Adjusted the CSS to allow the focus styles as specified by design.

In the course of doing so I identified a bug (RSC-597) where the twisty did not move when there was a tooltip. Since I was already in the NxTreeview SCSS I fixed it as well.

I also noticed that the indentation of the tree view children did not match the mockups so I adjusted that in the SCSS as well.


lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -18,16 +18,18 @@ $twisty-proportion-difference: 2px;
 */
 .nx-tree-view {
   box-sizing: border-box;
-  margin: var(--nx-spacing-4x) 0;
+  margin: var(--nx-spacing-4x) -8px;
```
rpokorny, line 21: When the tree is expanded, the `<path>` appears to stick out 1px.  See this screenshot where I have the content area of the parent (the `.gallery-example-live`) colored red.
image

rpokorny, line 21: Should this use a spacing var?

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -18,16 +18,18 @@ $twisty-proportion-difference: 2px;
...
 
   // Design specifies 300px wide not including the caret and its padding (which are themselves 30px)
   max-width: 330px;
```
rpokorny, line 24: Does this need adjusting to account for the extra space on the left?

lib/src/components/NxTreeView/NxTreeView.tsx
```diff
@@ -50,7 +50,7 @@ const NxTreeView: FunctionComponent<Props> =
           triggerTooltipProps ? (
             // div necessary to avoid error message when tooltip is on disabled button
             <NxTooltip { ...triggerTooltipProps } >
-              <div>
+              <div className="nx-tree-view__tooltip">
```
rpokorny, line 53: I'd rather a better name for this – it isn't the tooltip itself.  Maybe `nx-tree-view__tooltip-wrapper`?

lib/src/components/NxTreeView/NxTreeView.scss
glenhunter: > Should this use a spacing var?

I didn't think that we could use a spacing var as a negative number.

lib/src/components/NxTreeView/NxTreeView.tsx
glenhunter: > I'd rather a better name for this – it isn't the tooltip itself. Maybe `nx-tree-view__tooltip-wrapper`?

https://github.com/sonatype/sonatype-react-shared-components/pull/408/commits/9b3afaf8deb50a3720d690413af3f00a0ddf7e3c

lib/src/components/NxTreeView/NxTreeView.scss
glenhunter: > > Should this use a spacing var?
> 
> I didn't think that we could use a spacing var as a negative number.

Feels hacky but uses the variable: https://github.com/sonatype/sonatype-react-shared-components/pull/408/commits/9b3afaf8deb50a3720d690413af3f00a0ddf7e3c

glenhunter: > When the tree is expanded, the `<path>` appears to stick out 1px. See this screenshot where I have the content area of the parent (the `.gallery-example-live`) colored red.
> image

I think this solves it: https://github.com/sonatype/sonatype-react-shared-components/pull/408/commits/dc9bb12c1b104f0cad2126b54adf0abfe65074a9

I can't really see it on my monitor, the pixels are so small

lib/src/components/NxTreeView/NxTreeView.scss
glenhunter: > Does this need adjusting to account for the extra space on the left?

Direction from @dturner32 is a max-width of 360px including the focus states. https://github.com/sonatype/sonatype-react-shared-components/pull/408/commits/09e528f7a9a6fdd52c0b3fe7b1b1dda4662f2a86 and where I realized I shouldn't have included the negative margin: https://github.com/sonatype/sonatype-react-shared-components/pull/408/commits/1fda0c78374ce8ed751afe67c48f0e309dd02750

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -18,16 +18,19 @@ $twisty-proportion-difference: 2px;
 */
 .nx-tree-view {
   box-sizing: border-box;
-  margin: var(--nx-spacing-4x) 0;
+  // left gets 7px which keeps the caret from overhanging the container
+  margin: var(--nx-spacing-4x) calc(0px - var(--nx-spacing-2x)) var(--nx-spacing-4x) calc(1px - var(--nx-spacing-2x));
```
rpokorny, line 22: I think this fully reflects what's going on here:
```suggestion
  margin: var(--nx-spacing-4x) calc(0px - var(--nx-spacing-2x)) var(--nx-spacing-4x) calc(#{$twisty-proportion-difference / 2} - var(--nx-spacing-2x));
```

lib/src/components/NxTreeView/NxTreeView.scss
rpokorny: > Direction from @dturner32 is a max-width of 360px including the focus states. 09e528f and where I realized I shouldn't have included the negative margin: 1fda0c7

If I'm not mistaken, the intent is effectively that the `.nx-tree-view__text` be 300px wide.  It is currently coming out to 323.75px wide.

rpokorny: > If I'm not mistaken, the intent is effectively that the .nx-tree-view__text be 300px wide. It is currently coming out to 323.75px wide.

Oh nvm I think it's the `.nx-tree-view__child` content box that's supposed to be 300px wide. Which it is.

rpokorny: > Oh nvm I think it's the .nx-tree-view__child content box that's supposed to be 300px wide. Which it is.

Something's still wrong though.  The mockups show the children (when clickable) as being 350px wide but in the impl they are currently 360px wide.
image

glenhunter: > > Oh nvm I think it's the .nx-tree-view__child content box that's supposed to be 300px wide. Which it is.
> 
> Something's still wrong though. The mockups show the children (when clickable) as being 350px wide but in the impl they are currently 360px wide.

That looks like @dturner32 either hasn't updated the mockups or that he's just showing an example in a constrained container.

dturner32: > > > Oh nvm I think it's the .nx-tree-view__child content box that's supposed to be 300px wide. Which it is.
> > 
> > 
> > Something's still wrong though. The mockups show the children (when clickable) as being 350px wide but in the impl they are currently 360px wide.
> 
> That looks like @dturner32 either hasn't updated the mockups or that he's just showing an example in a constrained container.
> 
> Seems like it's both. I have no problem with the mockups showing constrained examples but we should be stating the max width somewhere.

the mocks have just been updated to reflect this

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -18,16 +18,19 @@ $twisty-proportion-difference: 2px;
 */
 .nx-tree-view {
   box-sizing: border-box;
-  margin: var(--nx-spacing-4x) 0;
+  // left gets 7px which keeps the caret from overhanging the container
+  margin: var(--nx-spacing-4x) calc(0px - var(--nx-spacing-2x)) var(--nx-spacing-4x) calc(#{$twisty-proportion-difference / 2} - var(--nx-spacing-2x));
```
rpokorny, line 22: Line too long.  Not sure what you think would be the best way to split it up

rpokorny:
The mockups are inconsistent (or trying to show something we don't understand).  In some, they show the hover highlight on child items as being aligned with the left edge of the caret while in others it shows them as being farther left (aligned with the new focus style I guess)
image
image

dturner32:
> The mockups are inconsistent (or trying to show something we don't understand). In some, they show the hover highlight on child items as being aligned with the left edge of the caret while in others it shows them as being farther left (aligned with the new focus style I guess)
> image
> image

correct, aligned with the new focus styles based on the updated style for the treeview title.

lib/src/components/NxTreeView/NxTreeView.scss
rpokorny: > I think this fully reflects what's going on here:

Checking again, this isn't right.  If you look at the other usages of `$twisty-proportion-difference`, you'll see that it's already intended to be the value on one side of the twisty, not the sum of both.  So we shouldn't be dividing by 2 here.  

At the same time, it makes me wonder why using that variable here is necessary at all – the twisty already gets its own extra padding using this variable when it is expanded, which should take care of this situation.  I'm thinking maybe the variable needs to be 3, not 2.  Or maybe 2.5.  The true difference between the height at width of the `path` is 5.53px.

glenhunter: > > I think this fully reflects what's going on here:
> 
> Checking again, this isn't right. If you look at the other usages of `$twisty-proportion-difference`, you'll see that it's already intended to be the value on one side of the twisty, not the sum of both. So we shouldn't be dividing by 2 here.
> 
> At the same time, it makes me wonder why using that variable here is necessary at all – the twisty already gets its own extra padding using this variable when it is expanded, which should take care of this situation. I'm thinking maybe the variable needs to be 3, not 2. Or maybe 2.5. The true difference between the height and width of the `path` is 5.53px.

So the upshot is that we'll leave this as-is?

lib/src/components/NxTreeView/NxTreeView.scss
glenhunter: > Line too long. Not sure what you think would be the best way to split it up

This line is only too long if we take your suggestion to use the `$twisty-proportion-difference` variable which it sounds like you've talked yourself out of above. Frankly I think that `1px` is much clearer.

lib/src/components/NxTreeView/NxTreeView.scss
rpokorny: > So the upshot is that we'll leave this as-is?

No, the upshot is that we shouldn't have had to do anything special about the twisty on this line as the styles it already had were supposed to be handling the problem we're having here.  They aren't, because the value of that variable is wrong.

glenhunter: > > So the upshot is that we'll leave this as-is?
> 
> No, the upshot is that we shouldn't have had to do anything special about the twisty on this line as the styles it already had were supposed to be handling the problem we're having here. They aren't, because the value of that variable is wrong.

Based on the `5.53px` value you mentioned above I set the variable to `2.75px`. Zoomed in very close on my monitor it looks perfect. https://github.com/sonatype/sonatype-react-shared-components/pull/408/commits/3e3535b08a1d9720a57ee7078f83f0a14de37d4d

glenhunter:
> > The mockups are inconsistent (or trying to show something we don't understand). In some, they show the hover highlight on child items as being aligned with the left edge of the caret while in others it shows them as being farther left (aligned with the new focus style I guess)
> 
> correct, aligned with the new focus styles based on the updated style for the treeview title.

I think that this matches what is present in the gallery.

rpokorny:
> > correct, aligned with the new focus styles based on the updated style for the treeview title.
>
> I think that this matches what is present in the gallery.

I don't follow what either of you are saying here.  The mockups are inconsistent with themselves.

glenhunter:
> > > correct, aligned with the new focus styles based on the updated style for the treeview title.
> > 
> > 
> > I think that this matches what is present in the gallery.
> 
> I don't follow what either of you are saying here. The mockups are inconsistent with themselves.

They're all consistent expect this one:

Screen Shot 2021-06-08 at 9 58 43 AM

Which I assume is a mistake since all the others including the other focused example show `16px`.

@dturner32 ^^

rpokorny:
> They're all consistent expect this one:

That isn't true.  Look at all the of the ones farther to the right.

dturner32:
> > > > correct, aligned with the new focus styles based on the updated style for the treeview title.
> > > 
> > > 
> > > I think that this matches what is present in the gallery.
> > 
> > 
> > I don't follow what either of you are saying here. The mockups are inconsistent with themselves.
> 
> They're all consistent expect this one:
> 
> Screen Shot 2021-06-08 at 9 58 43 AM
> 
> Which I assume is a mistake since all the others including the other focused example show `16px`.
> 
> @dturner32 ^^

this was one I missed in the mockup. I just updated it

glenhunter:
> > > > > correct, aligned with the new focus styles based on the updated style for the treeview title.
> > > > 
> > > > 
> > > > I think that this matches what is present in the gallery.
> > > 
> > > 
> > > I don't follow what either of you are saying here. The mockups are inconsistent with themselves.
> > 
> > 
> > They're all consistent expect this one:

> > Which I assume is a mistake since all the others including the other focused example show `16px`.
> > @dturner32 ^^
> 
> this was one I missed in the mockup. I just updated it

Mockups have been updated after discussion in Slack.

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -11,23 +11,26 @@
...
-$twisty-proportion-difference: 2px;
+$twisty-proportion-difference: 2.75px;
 
 /*
   #Tree view
 */
 .nx-tree-view {
   box-sizing: border-box;
-  margin: var(--nx-spacing-4x) 0;
+  // left gets 7px which keeps the caret from overhanging the container
```
rpokorny, line 21:
```suggestion
```

lib/src/components/NxTreeView/NxTreeView.scss
```diff
@@ -58,32 +61,27 @@ $twisty-proportion-difference: 2px;
...
   box-sizing: border-box;
   color: var(--nx-color-text-dark);
   cursor: pointer;
   display: flex;
   font: inherit;
   font-weight: 600;
   text-align: left;
-  padding: 0;
+  // subtract 1 from all sides to make room for focus border
+  padding: calc(var(--nx-spacing-base) - 1px) calc(var(--nx-spacing-2x) - 1px );
```
rpokorny, line 74:
```suggestion
  padding: calc(var(--nx-spacing-base) - 1px) calc(var(--nx-spacing-2x) - 1px);
```

rpokorny (approved):

PR #410, opened 2021-06-08
toggle control positioning RSC-585 

glenhunter:
https://issues.sonatype.org/browse/RSC-585

I had many positioning problems when I moved the control to the left side, chiefly that the margin shenanigans required to hide the input in Safari caused the label's box to start part way across the control and the first object below (in this case the paragraph showing the toggle results) to be similarly indented.

To solve these problems I used CSS Grid instead of Flexbox. I created two template areas, one for the text and one for the toggle. I then assign both the input and the toggle control to the same area and indented the input so that it disappeared under the control.

I also removed the gapless example and CSS as it's no longer applicable.

All of the screenshots below are from Safari.

Screen Shot 2021-06-08 at 1 41 10 PM
Screen Shot 2021-06-08 at 1 40 54 PM
Screen Shot 2021-06-08 at 1 40 48 PM


gallery/src/components/NxToggle/NxTogglePage.tsx
```diff
@@ -10,10 +10,8 @@ import { NxTextLink } from '@sonatype/react-shared-components';
...
 
 import NxToggleExample from './NxToggleExample';
-import NxToggleGaplessExample from './NxToggleGaplessExample';
 
-const exampleCode = require('./NxToggleExample?raw'),
-    exampleGaplessCode = require('./NxToggleGaplessExample?raw');
+const exampleCode = require('./NxToggleExample?raw');
 
 const NxTogglePage = () =>
   <>
```
rpokorny, line 17: Text a few lines down needs to be updated: "Child VDOM will be used as a label preceeding the toggle control."

lib/src/components/NxToggle/NxToggle.scss
```diff
@@ -7,12 +7,17 @@
...
+$nx-toggle-input-hide-hack-spacing: var(--nx-spacing-2x);
+
+$nx-toggle-control-width: 44px;
 
 .nx-toggle {
-  display: flex;
+  column-gap: 16px;
+  display: grid;
+  grid-template-areas: "toggle content";
+  grid-template-columns: $nx-toggle-control-width 1fr;
```
rpokorny, line 20: I think the first column can just be `auto`

lib/src/components/NxToggle/NxToggle.scss
```diff
@@ -7,12 +7,17 @@
 @import '../../scss-shared/nx-text-helpers';
 
 // In safari, the width: 0 and -moz-appearance settings fail to hide the <input>.
-// So instead, we use margin on the checkbox and corresponding negative margin on the toggle to hide
+// So instead, we use negative margin on the checkbox and offsetting positive margin on the toggle to hide
 // the <input> under the toggle
-$nx-toggle-input-hide-hack-spacing: 5px;
+$nx-toggle-input-hide-hack-spacing: var(--nx-spacing-2x);
```
rpokorny, line 12: Unused

lib/src/components/NxToggle/NxToggle.scss
```diff
@@ -7,12 +7,17 @@
 @import '../../scss-shared/nx-text-helpers';
 
 // In safari, the width: 0 and -moz-appearance settings fail to hide the <input>.
-// So instead, we use margin on the checkbox and corresponding negative margin on the toggle to hide
+// So instead, we use negative margin on the checkbox and offsetting positive margin on the toggle to hide
```
rpokorny, line 10: I don't think this is up to date

lib/src/components/NxToggle/NxToggle.scss
glenhunter: > I think the first column can just be `auto`

It can but:
1. why does it matter?
2. Why be _less_ precise?

lib/src/components/NxToggle/NxToggle.scss
glenhunter: > I don't think this is up to date

Rewritten and moved here: https://github.com/sonatype/sonatype-react-shared-components/pull/410/commits/ca98f32a4ab06e790c0adeb13a3c8f83bc849e5b

lib/src/components/NxToggle/NxToggle.scss
glenhunter: > Unused

Removed: https://github.com/sonatype/sonatype-react-shared-components/pull/410/commits/ca98f32a4ab06e790c0adeb13a3c8f83bc849e5b

gallery/src/components/NxToggle/NxTogglePage.tsx
glenhunter: > Text a few lines down needs to be updated: "Child VDOM will be used as a label preceeding the toggle control."

https://github.com/sonatype/sonatype-react-shared-components/pull/410/commits/584e1116291272c6f86f598fb18d95a4f4b2797c

lib/src/components/NxToggle/NxToggle.scss
rpokorny: > 1. why does it matter?

It minimizes use of magic numbers, and in this case removes the need for the `$nx-toggle-control-width` variable.

> 2. Why be less precise?

IMO it's more precise to say "make it the width of its contents" than to say "make it 44px wide... which btw should be the width of its contents".

glenhunter: > > 1. why does it matter?
> 
> It minimizes use of magic numbers, and in this case removes the need for the `$nx-toggle-control-width` variable.
> 

Well it's not like the magic number is going away, it's still required.

> > 1. Why be less precise?
> 
> IMO it's more precise to say "make it the width of its contents" than to say "make it 44px wide... which btw should be the width of its contents".

Meh, I think it's a stretch to say that `auto` is _more_ precise than `44px`.

It also doesn't matter to me as long as it works: https://github.com/sonatype/sonatype-react-shared-components/pull/410/commits/04eced974483834ca09e87ee06c09e45aa69196b

rpokorny (approved):

PR #412, opened 2021-06-09
Left align alerts RSC-586

glenhunter:
https://issues.sonatype.org/browse/RSC-586

The ticket made it seem that all alerts were being left aligned but that is not the case. Some alerts are left aligned and some remain centred. Looking at the designs I decided that it was easiest to make left aligned the default and centre aligned the exception. 

I added an example of an alert inside a form layout for testing purposes.


lib/src/base-styles/_nx-alert.scss
```diff
@@ -17,7 +17,7 @@
   border-radius: var(--nx-border-radius);
   box-sizing: border-box;
   display: flex;
-  margin: 16px auto 24px auto;
+  margin: 16px 0 24px 0;
```
rpokorny, line 20: Shouldn't the right margin be auto?

glenhunter: > Shouldn't the right margin be auto?

Why? 

I understand the difference in terms of what the browser does with the space, but why does it matter in terms of the display, or the correctness of the code?

rpokorny: > Why?
>
> I understand the difference in terms of what the browser does with the space, but why does it matter in terms of the display, or the correctness of the code?

I've changed my mind after thinking more about how margin works in block layout.  I think what you have is fine.  But for good measure, here's what I was about to say before second-guessing it:

> Currently you're asking the impossible of the browser: that an element within a container of unknown width simultaneously have a specific width and also no space on either side.  Unless the container happens to be just the right size, the browser has to pick one of those constraints to break.  Which one it picks _is_ presumably well defined and aligns with what we want, but I'd rather not ask for impossibilities in the first place.

rpokorny (approved):

PR #414, opened 2021-06-14
Fix alignment of NxTextInput and NxFilterInput when adjacent - RSC-599

rpokorny:
https://issues.sonatype.org/browse/RSC-599


lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -30,15 +30,10 @@
...
-  .nx-text-input__box {
-    padding-bottom: 0;
-  }
-
-  .nx-text-input__input {
-    padding: 0 0 8px 0;
-  }
-
-  .nx-icon--filter-icons {
-    margin-bottom: 8px;
```
rpokorny, line 42: These deleted styles were basically just doing the same thing the nx-text-input styles do but in a different way.  Deleting them made no difference and made it simpler to understand and solve the alignment problem.

glenhunter:
Unfortunately Safari is not playing along: 
Screen Shot 2021-06-17 at 12 18 19 PM

and Chrome seems to be off by just a pixel: 
Screen Shot 2021-06-17 at 12 18 10 PM

rpokorny:
I can reproduce that it's wrong in safari but I just double checked that chrome is perfect for me

rpokorny:
Fix for safari: bf7c786e

glenhunter:
Right now there is no default spacing between the form fields, I remember that was mentioned in WW this week but I can't recall the outcome and it isn't recorded anywhere. @dturner32 ?

rpokorny:
> Right now there is no default spacing between the form fields, I remember that was mentioned in WW this week but I can't recall the outcome and it isn't recorded anywhere. @dturner32 ?

I don't think we came to much of a conclusion on it since the one place that currently needs this has spacing needs that wouldn't be what we'd want more generally.

glenhunter:
> > Right now there is no default spacing between the form fields, I remember that was mentioned in WW this week but I can't recall the outcome and it isn't recorded anywhere. @dturner32 ?
> 
> I don't think we came to much of a conclusion on it since the one place that currently needs this has spacing needs that wouldn't be what we'd want more generally.

I thought we said something like just `4px`

rpokorny:
> I thought we said something like just 4px

I know we talked _about_ that, but in my recollection we didn't agree to go with it.

dturner32:
> Right now there is no default spacing between the form fields, I remember that was mentioned in WW this week but I can't recall the outcome and it isn't recorded anywhere. @dturner32 ?



> > I thought we said something like just 4px
> 
> I know we talked _about_ that, but in my recollection we didn't agree to go with it.

if I recall, we landed on 16px being the text input and the filter input

rpokorny:
> if I recall, we landed on 16px being the text input and the filter input

99a778a4

glenhunter (approved):
Safari issues are resolved, looks good.

PR #415, opened 2021-06-14
nx-form accessibility RSC-427

glenhunter:
https://issues.sonatype.org/browse/RSC-427

Per the description in the ticket in order to have the `<form>` tag recognized as part of the landmark structure it must have one of `title`, `aria-label`, or `aria-labelled-by` defined. While `title` works in Voice Over, it is not an accepted attribute of the `<form>` element (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). Note that while `name` is an attribute of `<form>` it doesn't have any effect where VO is concerned.

I added examples of `aria-label` to the three example pages and wrote a paragraph explaining the addition.

### DIFF
--- a/gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
+++ b/gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
@@ -7,6 +7,7 @@
 import React from 'react';
 
 import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
+import { NxTextLink } from '@sonatype/react-shared-components';
 
 import NxFormLayoutExample from './NxFormLayoutExample';
 import NxFormHorizontalLayoutExample from './NxFormHorizontalLayoutExample';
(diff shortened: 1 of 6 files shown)


rpokorny:
> While title works in Voice Over, it is not an accepted attribute of the `<form>` element

image
`title` is a global attribute, valid on all HTML elements

rpokorny:
I think we should make `title` a "soft-required" prop on `NxForm`.  That is, document it as strongly recommend and only optional for backwards compatibility.

rpokorny:
> I think we should make title a "soft-required" prop on NxForm. That is, document it as strongly recommend and only optional for backwards compatibility.

Though in light of the option to use `aria-describedby`, maybe we should just leave it as optional and stick with the documentation you've written.

glenhunter:
> > While title works in Voice Over, it is not an accepted attribute of the `<form>` element
> 
> `title` is a global attribute, valid on all HTML elements

Damn. How did I not know that?

glenhunter:
> > I think we should make title a "soft-required" prop on NxForm. That is, document it as strongly recommend and only optional for backwards compatibility.
> 
> Though in light of the option to use `aria-describedby`, maybe we should just leave it as optional and stick with the documentation you've written.

I think that sticking with the `aria` attributes is good enough. I feel like mixing in `title` could just lead to confusion.

gallery/src/styles/NxFormLayout/NxFormLayoutPage.tsx
```diff
@@ -128,6 +129,19 @@ const NxFormLayoutPage = () =>
           </tr>
         </tbody>
       </table>
+      <section className="nx-tile__subsection">
+        <h2 className="nx-h2">&lt;form&gt; Accessibility</h2>
+        <p className="nx-p">
+          Larger forms should be identified using either the <code className="nx-code">aria-label</code>
+          {' '}attribute to provide a descriptive title, or the <code className="nx-code">aria-labelled-by</code>
```
rpokorny, line 136:
```suggestion
          {' '}attribute to provide a descriptive title, or the <code className="nx-code">aria-labelledby</code>
```

rpokorny (approved):
+1 except for an important typo

PR #418, opened 2021-06-18
Add none count to NxThreatCounter - CLM-18720

mealingr:
Related story: https://issues.sonatype.org/browse/CLM-18720
Build: https://jenkins.ci.sonatype.dev/job/uxui/job/sonatype-react-shared-components/job/add_none_count_to_nx_threat_counter/lastBuild

image


mealingr:
> 4 out of 337 visual tests detected diffs

Not sure what visual tests are failing (no access), maybe because I've changed the gallery page? Hopefully if these changes are good the diffs can be accepted 🤞

rpokorny:
Thanks for contributing.

Our understanding of the initial requirements though was that the None count would be optional.  Then, as more discussion with @ChurchRyan (and you I assume) happened, it was decided that really all of the counts should be optional.  Therefore our plan for when we got to this was going to be to make it so that each count prop is optional, and the corresponding part of the UI only shows up if its actually a number (incl. 0).  If a downstream dev doesn't want a given count to show up in a given situation, they'd pass null or undefined (or just not specify anything).

Additionally I think design work is needed to confirm what the overall layout is supposed to look like with various counts present or missing.

Also our specific ticket for this is RSC-608.

rpokorny:
> Hopefully if these changes are good the diffs can be accepted crossed_fingers

Some of the diffs are a little problematic unfortunately:
image

mealingr:
> Additionally I think design work is needed to confirm what the overall layout is supposed to look like with various counts present or missing.
> 
> Our specific ticket for this is RSC-608.

Thanks for the explanation, I see where we want to go now, I'm happy to try and update this to match the expectations (or I can close for now if that's preferred)

rpokorny:
> Thanks for the explanation, I see where we want to go now, I'm happy to try and update this to match the expectations (or I can close for now if that's preferred)

I'd say talk to @dturner32 first about what it should do when some counters aren't displayed.  if the answer is non-trivial or not yet known I'd say you'll need to hold off on it

mealingr:
> I'd say talk to @dturner32 first about what it should do when some counters aren't displayed. if the answer is non-trivial or not yet known I'd say you'll need to hold off on it

I know this is somewhat liable to change given https://issues.sonatype.org/browse/RSC-608 isn't through design yet, but Derrek mentioned that the current thinking at least is to have all counts optional and to just hide the counts we don't want.

I've attempted this in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/010bf9d9de295e3f735e928377ab65bc15cc5081 the row/column layouts seem to work but maybe the grid layout only makes sense with 4 for a nice 2x2

Still triggering applitools too but hopefully things look closer to being acceptable
image

rpokorny:
> I've attempted this in 010bf9d the row/column layouts seem ok but maybe the grid layout only makes sense with 4 for a nice 2x2

I'd say the grid is behaving sanely at any rate.  We can either get @dturner32 's ok of its current behavior or just add documentation saying to only use that one with a total of 4 counts showing.  Or if @dturner32 has other ideas of what it should do...

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
```diff
@@ -22,7 +22,10 @@ const NxThreatCounterPage = () =>
     <GalleryDescriptionTile>
       <p className="nx-p">
         Component for a series of small counters which display a number representing a count of threats/violations,
-        and a short pre-set text string which displays the severity (critical, severe, moderate, and low).
+        and a short pre-set text string which displays the severity (critical, severe, moderate, low, and none).
+      </p>
+      <p className="nx-p">
+        Each count is optional. If no counts are provided, then nothing will be rendered.
```
rpokorny, line 28: I'd rephrase this a little bit, as using this component to render nothing at all wouldn't really be a useful/supported case.  How about this?
```suggestion
        Each count is optional. If no value is provided for a given count, then the indicator for that severity level will not be rendered.
```

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
```diff
@@ -41,26 +44,42 @@ const NxThreatCounterPage = () =>
...
+            <td className="nx-cell">
+              A numerical value, the number of low threats. If no value is provided, then the count will be hidden.
+            </td>
+          </tr>
+          <tr className="nx-table-row">
+            <td className="nx-cell">noneCount</td>
+            <td className="nx-cell">number</td>
+            <td className="nx-cell">No</td>
+            <td className="nx-cell">
+              A numerical value, the number of none threats. If no value is provided, then the count will be hidden.
```
rpokorny, line 81:
```suggestion
              A numerical value, the number of counted items posing no threat. If no value is provided, then the count will be hidden.
```

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
```diff
@@ -57,7 +57,8 @@ function NxPageTitleEverythingExample() {
         <NxThreatCounter criticalCount={45}
                          severeCount={21114}
                          moderateCount={12}
-                         lowCount={45}/>
+                         lowCount={45}
+                         noneCount={21}/>
```
rpokorny, line 61: While we do certainly want these examples that have every count shown, I'm wondering if we also want examples of each layout fewer counts shown.  Keep in mind that these examples double as our visual test items, so sometimes it makes sense to show every possibility even if it might be considered excessive for the sake of human documentation.

lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "7.1.2",
+  "version": "7.1.3",
```
rpokorny, line 3: This would be a minor version

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -49,10 +49,14 @@
...
 
+.nx-threat-counter--none {
+  border-color: var(--nx-color-threat-none);
+}
+
 .nx-threat-counter-container {
   display: inline-grid;
   gap: var(--nx-spacing-4x);
-  grid-template-columns: auto auto auto auto;
+  grid-template-columns: auto auto auto auto auto;
```
rpokorny, line 59: At this point it's probably long enough to warrant using `repeat`
```suggestion
  grid-template-columns: repeat(5, auto);
```

lib/src/components/NxThreatCounter/NxThreatCounter.tsx
```diff
@@ -13,30 +13,47 @@ export { Props } from './types';
...
   const layoutClasses = classnames('nx-threat-counter-container', className, {
     [`nx-threat-counter-container--${layout}`]: layout
   });
 
+  if (typeof criticalCount !== 'number' &&
+      typeof severeCount !== 'number' &&
+      typeof moderateCount !== 'number' &&
+      typeof lowCount !== 'number' &&
+      typeof noneCount !== 'number') {
+    return null;
```
rpokorny, line 27: I'd log a console warning here as it's not really a valid use of the component

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
```diff
@@ -7,47 +7,99 @@
...
+      const props = getCounts(value);
+      const countName = name + 'Count';
+      props[countName] = 0;
+      const container = getShallowComponent(props).find('.nx-threat-counter-container');
+      expect(container).toExist();
+      const counts = container.find('div');
+      expect(counts.length).toBe(1);
+      const count = counts.at(0);
+      expect(count.find('.nx-threat-counter__count')).toHaveText('0');
+      expect(count.find('.nx-threat-counter__text')).toHaveText(capitalize(name));
```
rpokorny, line 59: This shouldn't be working, the text isn't capitalized at the DOM level but at the styling level

rpokorny: > This shouldn't be working, the text isn't capitalized at the DOM level but at the styling level

Oh nvm I was thinking fully uppercased not capitalized.  At any rate, this is a bit too moch logic for a test IMO.  I'd rather it all be spelled out in a more hardcoded way

lib/package.json
mealingr: > This would be a minor version

Minor version bumped instead in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/ad865014ca44474766838735c2cd9ac8beead1bf

lib/src/components/NxThreatCounter/NxThreatCounter.tsx
mealingr: > I'd log a console warning here as it's not really a valid use of the component

Added in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/ec3631afe405fb01d0c9406fc4d0e68496ddb5da

lib/package.json
rpokorny: > Minor version bumped instead in ad86501

Should be 7.2.0.  Please use `yarn version --no-git-tag-version --minor`or similar  in the `lib` directory for version bumps

rpokorny:
> We can either get @dturner32 's ok of its current behavior

He has ok'd it

lib/package.json
mealingr: > Should be 7.2.0. Please use `yarn version --no-git-tag-version --minor`or similar in the `lib` directory for version bumps

Fixed in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/7e78c6d3b9bc49b11756ec499cf15244ff796b0f

lib/src/components/NxThreatCounter/types.ts
```diff
@@ -8,19 +8,21 @@ import { HTMLAttributes, ValidationMap } from 'react';
...
-  lowCount: number;
-  layout?: string | null;
-  className?: string | null;
+  criticalCount?: number | null | undefined;
+  severeCount?: number | null | undefined;
+  moderateCount?: number | null | undefined;
+  lowCount?: number | null | undefined;
+  noneCount?: number | null | undefined;
+  layout?: string | null | undefined;
+  className?: string | null | undefined;
```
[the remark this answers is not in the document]
mealingr: > The `?` adds `undefined` to the union automatically; you don't need to make it explicit.

Simplified in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/982f36a97c78c738a80127ef6a1e62839f7bceac

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
mealingr: > While we do certainly want these examples that have every count shown, I'm wondering if we also want examples of each layout fewer counts shown. Keep in mind that these examples double as our visual test items, so sometimes it makes sense to show every possibility even if it might be considered excessive for the sake of human documentation.

I've added some more examples to the existing pages https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/eff6a8606c1b78944ee9d99b15de9dabff6510fa https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/e5941cc6215ee25c84289e53ead4335f4b13b760 https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/c7a4d1230c210154b7502dc9699f8e7bdd329f0c) what do you think of these?
image
image
image

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
mealingr: > Oh nvm I was thinking fully uppercased not capitalized. At any rate, this is a bit too much logic for a test IMO. I'd rather it all be spelled out in a more hardcoded way

Attempted to make it slightly more explicit https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/bbbc191c18b7ba8b720c3b782cff9e474889c15f or I can break it into separate tests if you prefer `it('renders only critical if the others are undefined') ...`

rpokorny: > Attempted to make it slightly more explicit bbbc191 or I can break it into separate tests if you prefer it('renders only critical if the others are undefined') ...

I'd have a test like `it('renders an .nx-threat-counter--critical when criticalCount is specified'` and then similarly for the other four levels.  And then one with all of them together.

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
rpokorny: > I've added some more examples to the existing pages eff6a86 e5941cc c7a4d12 1915f11 what do you think of these?

Hmm unfortunately I think having them displayed like that makes it difficult for the user to tell what they're looking at, e.g. that in the first one that is 5 separate components laid out one after another vertically.  I think for each layout type, two distinct examples are warranted: one with all the counters and one with less than all of them.

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
mealingr: > I'd have a test like `it('renders an .nx-threat-counter--critical when criticalCount is specified'` and then similarly for the other four levels. And then one with all of them together.

Done in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/c296dfb05ada74f6812ddba6a75396b9945f6d76

gallery/src/styles/NxPageTitle/NxPageTitleEverythingExample.tsx
mealingr: > Hmm unfortunately I think having them displayed like that makes it difficult for the user to tell what they're looking at, e.g. that in the first one that is 5 separate components laid out one after another vertically. I think for each layout type, two distinct examples are warranted: one with all the counters and one with less than all of them.

Done in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/c296dfb05ada74f6812ddba6a75396b9945f6d76 (each layout with an example of all counts and just 2 counts)

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
```diff
@@ -13,41 +13,149 @@ describe('NxThreatCounter', function() {
         criticalCount: 66,
```
rpokorny, line 13: Now that these are all optional, I'd say the `minimalProps` should be an empty object.  This would simplify many of the tests below.

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
```diff
@@ -13,41 +13,149 @@ describe('NxThreatCounter', function() {
...
+
+  it('renders all five indicators', function() {
     expect(getShallowComponent().find('.nx-threat-counter--critical')).toExist();
     expect(getShallowComponent().find('.nx-threat-counter--severe')).toExist();
     expect(getShallowComponent().find('.nx-threat-counter--moderate')).toExist();
     expect(getShallowComponent().find('.nx-threat-counter--low')).toExist();
+    expect(getShallowComponent().find('.nx-threat-counter--none')).toExist();
   });
 
   it('renders all of the values set in minimalProps', function() {
```
rpokorny, line 135: This test and the one above it can be merged

rpokorny:
Visual tests of the non-full counts are needed.  You can find the visual tests in `gallery/visualtests/nxVulnerabilityIndicator.js` (whoops, looks like we should rename that file)

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
```diff
@@ -92,7 +119,14 @@ const NxThreatCounterPage = () =>
                         id="nx-threat-counter-row"
                         liveExample={NxThreatCounterExample}
                         codeExamples={NxThreatCounterCode}>
-      <code className="nx-code">nx-threat-counter</code>s in the default layout.
+      <code className="nx-code">nx-threat-counter</code>s in the default (row) layout.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="NxThreatCounter Small Default Example"
+                        id="nx-threat-counter-row"
```
rpokorny, line 126: Duplicate id here and on the other new ones.  Aside from that generally being invalid, you'll want to use the ids here as you add to the visual tests.

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
mealingr: > Now that these are all optional, I'd say the `minimalProps` should be an empty object. This would simplify many of the tests below.

Done in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/8cce8f0e7e32b56afa23137b9a7b1a8cb948a2bd

lib/src/components/NxThreatCounter/__tests__/NxThreatCounter.test.tsx
mealingr: > This test and the one above it can be merged

Merged in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/8cce8f0e7e32b56afa23137b9a7b1a8cb948a2bd

gallery/src/components/NxThreatCounter/NxThreatCounterPage.tsx
mealingr: > Duplicate id here and on the other new ones. Aside from that generally being invalid, you'll want to use the ids here as you add to the visual tests.

Thanks I missed that, changed the ids and added more visual tests in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/8cce8f0e7e32b56afa23137b9a7b1a8cb948a2bd

mealingr:
> Visual tests of the non-full counts are needed. You can find the visual tests in `gallery/visualtests/nxVulnerabilityIndicator.js` (whoops, looks like we should rename that file)

Renamed in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/21dc09fd6424ab10169006be4e06f7d10cb8b4a9 to `nxThreatCounter.js` if that works (couldn't see if i needed to change it's name anywhere else)

rpokorny (approved):
Thanks again for contributing to RSC!

rpokorny:
> couldn't see if i needed to change it's name anywhere else

Nope, you don't

kakumara (approved):
+1

mealingr:
> 4 out of 340 visual tests detected diffs

Would you mind checking these for me @rpokorny ? Or if you prefer I'm happy to check myself if I can get access (hoping this time they're acceptable)

rpokorny:
> Would you mind checking these for me @rpokorny ? Or if you prefer I'm happy to check myself if I can get access (hoping this time they're acceptable crossed_fingers )

:thinking: not sure about this one:
image

rpokorny:
> :thinking: not sure about this one:

The thing is, a row-based NxThreatCounter is implemented to render in a single line.  It used to fit on a single line along with potentially multiple `NxPolicyViolationIndicator`s in front of it.  Now it seems that it doesn't.  Granted, if there were 3 `NxPolicyViolationIndicator` rather than two, it probably wouldn't have fit before.  So while it used to break a little in an extreme edge case, now it breaks a little in a slightly less extreme edge case.  I'd like input from @dturner32 on whether this is ok

dturner32:
> > 🤔 not sure about this one:
> 
> The thing is, a row-based NxThreatCounter is implemented to render in a single line. It used to fit on a single line along with potentially multiple `NxPolicyViolationIndicator`s in front of it. Now it seems that it doesn't. Granted, if there were 3 `NxPolicyViolationIndicator` rather than two, it probably wouldn't have fit before. So while it used to break a little in an extreme edge case, now it breaks a little in a slightly less extreme edge case. I'd like input from @dturner32 on whether this is ok

yeah, this one seems a little weird to me that it's breaking after only showing two indicators on one line. I know the example here is an extreme edge case, but my thinking is that two or more indicators should still be able to all fit on one line and fill what ever space they are allowed to fill before wrapping to a new line.

rpokorny:
> yeah, this one seems a little weird to me that it's breaking after only showing two indicators on one line. I know the example here is an extreme edge case, but my thinking is that two or more indicators should still be able to all fit on one line and fill what ever space they are allowed to fill before wrapping to a new line.

Is this an accurate way of putting it: within the page header tags area, the `NxThreatCounter` shouldn't lay out in a "row" style but rather in an "inline" style?

lib/src/base-styles/_nx-page.scss
```diff
@@ -71,10 +71,14 @@
...
-  
+
   .nx-threat-counter-container {
     margin-top: 0;
-  
+    display: contents;
+    .nx-threat-counter {
+      margin: 0 var(--nx-spacing-4x) var(--nx-spacing-2x) 0;
+    }
+
```
mealingr, line 81: How about this for threat counters in the page title tags area to get them to wrap? 
image

rpokorny: > How about this for threat counters in the page title tags area to get them to wrap?

That's getting there but there are more style adjustments needed.  The indivdual counters should have their bottom margin removed.  The margin between the last counter and the first tag is too large.  @dturner32 please clarify what it's supposed to be.  Additionally there are a few margin styles on the `.nx-threat-counter-container` here (a top margin and a left margin) which have no effect due to `display: contents` and which should be removed

mealingr: > That's getting there but there are more style adjustments needed. The indivdual counters should have their bottom margin removed. The margin between the last counter and the first tag is too large. @dturner32 please clarify what it's supposed to be. Additionally there are a few margin styles on the `.nx-threat-counter-container` here (a top margin and a left margin) which have no effect due to `display: contents` and which should be removed

Should be done in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/358fd340bc90acad3ef36c05159efa3af04461e4 which
- Removed the bottom margin on the counters
- Removed the right margin on the last counter
- Decreased the left margin on the tag following the counter container to make it the same distance as between counters

image

rpokorny: See the relevant mockup: 
image

I believe the intent was that there be 24px of space between each _kind_ of indicator.  So you took too much out of the gap between the threat indicators and the tags.

mealingr: > I believe the intent was that there be 24px of space between each _kind_ of indicator. So you took too much out of the gap between the threat indicators and the tags.

Added the missing space back in https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/0b23e70fe31be91d7e5e360a7c2b43043a24644b

rpokorny: That's on one side, but judging from master there's a similar matter on the other side: 
image

mealingr: > That's on one side, but judging from master there's a similar matter on the other side:

I think this fixes it https://github.com/sonatype/sonatype-react-shared-components/pull/418/commits/6b2972f9e5040d72a30027ad8b8e7ac58add94f8

rpokorny (approved):

PR #420, opened 2021-06-21
nx-form-select disabled example code - RSC-606

rpokorny:
https://issues.sonatype.org/browse/RSC-606

In addition to displaying the _correct_ code, this PR removes the use of the disabled class (which doesn't really work adequately here, in isolation) in favor of the attribute.


glenhunter:
I'm a bit surprised that the disabled text doesn't use `--nx-text-color-disabled`.

rpokorny:
> I'm a bit surprised that the disabled text doesn't use --nx-text-color-disabled.

I'd guess it's supposed to.  I'm also looking at the color of the triangle.  I don't think it can be controlled by a css variable because it's embedded in that inline svg.  So it's out of date currently but also it doesn't change when disabled.  I'll see what I can do with it

rpokorny:
> I'm also looking at the color of the triangle. I don't think it can be controlled by a css variable because it's embedded in that inline svg. So it's out of date currently but also it doesn't change when disabled. I'll see what I can do with it

Here's the best I could do for it. 7603dd0a

glenhunter (approved):

PR #421, opened 2021-06-22
Clarify restrictions on NxDropdown children - RSC-536

rpokorny:
https://issues.sonatype.org/browse/RSC-536

### DIFF
--- a/gallery/src/components/NxDropdown/NxDropdownPage.tsx
+++ b/gallery/src/components/NxDropdown/NxDropdownPage.tsx
@@ -215,7 +215,7 @@ const NxDropdownPage = () =>
               <NxTableCell>
                 Sometimes it is desirable to add a icon-only button to a menu row which performs an action distinct
                 from clicking on the row itself - for instance, a trashcan button which deletes the row. For this
-                scenario, the icon button, a sibling of the main row button, must include this class and{' '}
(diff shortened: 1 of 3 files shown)


gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -198,10 +213,9 @@ const NxDropdownPage = () =>
                 An additional icon-only button which can appear on the right side of a menu row
               </NxTableCell>
               <NxTableCell>
-                Sometimes it is desireable to add a icon-only button to a menu row which performs an action distinct
+                Sometimes it is desirable to add a icon-only button to a menu row which performs an action distinct
                 from clicking on the row itself - for instance, a trashcan button which deletes the row. For this
-                scenario, the icon button, a sibling of the main row button, must include this class and
-                the <code className="nx-code">nx-dropdown-button-container</code> and{' '}
+                scenario, the icon button, a sibling of the main row button, must include this class and{' '}
```
glenhunter, line 218: Probably worth mentioning that the icon button must appear _before_ the link or the display breaks (because of the right float).

rpokorny: > Probably worth mentioning that the icon button must appear before the link or the display breaks (because of the right float).

c2f67692

glenhunter (approved):

PR #424, opened 2021-06-29
Centre collapsed buttons RSC-503

glenhunter:
https://issues.sonatype.org/browse/RSC-603

For the main nav buttons I added a single pixel of padding to the left and that centres the icons in the collapsed state.

The support icon had right margin on the icon that I had to remove when the sidebar is collapsed.

### DIFF
--- a/lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
+++ b/lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
@@ -95,6 +95,8 @@ $nx-global-sidebar-separator-color: var(--nx-swatch-indigo-40);
   padding: var(--nx-spacing-base)
            $nx-global-sidebar-left-right-spacing
            var(--nx-spacing-base)
+           // We need to add an extra pixel to the left to properly centre
+           // the icons when the sidebar is collapsed
            calc(#{$nx-global-sidebar-left-right-spacing} + 1px);
 
   .nx-icon {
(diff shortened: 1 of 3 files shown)


rpokorny:
So the decision was that the icons should be 1px farther right both when it's collapsed and when it's expanded?

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -92,7 +92,10 @@ $nx-global-sidebar-separator-color: var(--nx-swatch-indigo-40);
 
   border: 1px solid transparent;
   display: block;
-  padding: var(--nx-spacing-base) $nx-global-sidebar-left-right-spacing;
+  padding: var(--nx-spacing-base)
+           $nx-global-sidebar-left-right-spacing
+           var(--nx-spacing-base)
+           calc(#{$nx-global-sidebar-left-right-spacing} + 1px);
```
rpokorny, line 98: Could use a code comment

rpokorny (approved):

glenhunter:
> So the decision was that the icons should be 1px farther right both when it's collapsed and when it's expanded?

I think that it's important that the distance be the same open and closed, the 1px jump was noticeable to me when I tried it.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
glenhunter: > Could use a code comment

https://github.com/sonatype/sonatype-react-shared-components/pull/424/commits/570d8aef9a7c6c6c8ebd0c4f104c24668002aeb5

PR #425, opened 2021-06-29
focused radio checkbox RSC-600

avnvbhatta:
https://issues.sonatype.org/browse/RSC-600

- Added focus styles on checked radio buttons and checkboxes.
- Modified visual tests to align with new design spec.

Achieved by using an absolutely positioned `::after` pseudoelement that is visible on focus.

### DIFF
--- a/gallery/visualtests/nxCheckbox.js
+++ b/gallery/visualtests/nxCheckbox.js
@@ -19,7 +19,7 @@ describe('NxCheckbox', function() {
     it('has a light grey border and white background by default', simpleTest(selector));
     it('has a black border when hovered', hoverTest(selector));
 
-    it('has a blue background and white checkmark and a light blue outer border when clicked', async function() {
+    it('has a blue background and white checkmark when clicked', async function() {
       const targetElement = await browser.$(selector);
 
       await targetElement.scrollIntoView({ block: 'center' });
--- a/gallery/visualtests/nxCheckbox.js
+++ b/gallery/visualtests/nxCheckbox.js
@@ -20,13 +20,14 @@ describe('NxCheckbox', function() {
     it('has a black border when hovered', hoverTest(selector));
 
     it('has a blue background and white checkmark when clicked', async function() {
-      const targetElement = await browser.$(selector);
+      const blurSelector = `${selector} input`,
+          [targetElement, blurElement] = await Promise.all([browser.$(selector), browser.$(blurSelector)]);
 
       await targetElement.scrollIntoView({ block: 'center' });
       await targetElement.click();
       await browser.execute(function(el) {
         el.blur();
-      }, targetElement);
+      }, blurElement);
 
       try {
         await browser.eyesRegionSnapshot(null, Target.region(targetElement));
--- a/gallery/visualtests/nxCheckbox.js
+++ b/gallery/visualtests/nxCheckbox.js
@@ -24,6 +24,9 @@ describe('NxCheckbox', function() {
 
       await targetElement.scrollIntoView({ block: 'center' });
       await targetElement.click();
+      await browser.execute(function(el) {
+        el.blur();
+      }, targetElement);
 
       try {
         await browser.eyesRegionSnapshot(null, Target.region(targetElement));
--- a/gallery/visualtests/nxCheckbox.js
+++ b/gallery/visualtests/nxCheckbox.js
@@ -19,7 +19,7 @@ describe('NxCheckbox', function() {
     it('has a light grey border and white background by default', simpleTest(selector));
     it('has a black border when hovered', hoverTest(selector));
 
-    it('has a blue background and white checkmark when clicked', async function() {
+    it('has a blue background and white checkmark and a light blue outer border when clicked', async function() {
       const targetElement = await browser.$(selector);
 
       await targetElement.scrollIntoView({ block: 'center' });
@@ -34,7 +34,8 @@ describe('NxCheckbox', function() {
       }
     });
 
-    it('has a blue background, white checkmark, and glow when clicked and focused', async function() {
+    it(`has a blue background, white checkmark, a light blue outer border,
+      and glow when clicked and focused`, async function() {
       const focusSelector = `${selector} input`,
           [targetElement, focusElement] = await Promise.all([browser.$(selector), browser.$(focusSelector)]);
 
@@ -54,7 +55,8 @@ describe('NxCheckbox', function() {
       }
     });
 
-    it('has a blue background and white checkmark when clicked, focused, and hovered', async function() {
+    it(`has a blue background and white checkmark with a light blue outer border
+      when clicked, focused, and hovered`, async function() {
       const focusSelector = `${selector} input`,
           [targetElement, focusElement] = await Promise.all([browser.$(selector), browser.$(focusSelector)]);
 
@@ -74,8 +76,8 @@ describe('NxCheckbox', function() {
       }
     });
 
-    it('has a light blue border and glow when focused', focusTest(selector));
-    it('has a dark border when focused and hovered', focusAndHoverTest(selector));
+    it('has a light blue outer border and glow when focused', focusTest(selector));
+    it('has a light blue outer border and a dark border when focused and hovered', focusAndHoverTest(selector));
   });
 
   describe('Attribute-Disabled NxCheckbox', function() {
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -22,10 +22,6 @@
   max-width: max-content;
   position: relative;
 
-  &::after {
-    display: none;
-  }
-
   &:hover {
     cursor: pointer;
 
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -23,14 +23,7 @@
   position: relative;
 
   &::after {
-    border: 1px solid var(--nx-color-interactive-border-focus);
-    box-shadow: var(--nx-box-shadow-focus);
-    content: '';
     display: none;
-    height: 22px;
-    position:absolute;
-    left: -4px;
-    width: 22px;
   }
 
   &:hover {
@@ -44,7 +37,11 @@
 
   &:focus-within {
     &::after {
+      border: 1px solid var(--nx-color-interactive-border-focus);
+      box-shadow: var(--nx-box-shadow-focus);
+      content: '';
       display: block;
+      position: absolute;
     }
 
     .nx-radio-checkbox__control {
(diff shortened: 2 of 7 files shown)


lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -20,6 +20,18 @@
...
+  &::after {
+    border: 1px solid var(--nx-color-interactive-border-focus);
+    filter: drop-shadow(var(--nx-drop-shadow-focus));
+    content: '';
+    width: 22px;
+    height: 22px;
+    position:absolute;
+    left: -4px;
+    display: none;
+  }
```
glenhunter, line 34: This is minor, and a personal thing (Ross doesn't share this view), but I prefer to see CSS attributes in alphabetical order.

Exceptions are that min/max/height, min/max/width are grouped, and top/right/bottom/left are grouped after `position`.

glenhunter:
I like your solution, it seems though that while the mockups show the glow outside the focus outline that isn't so in the implementation.

avnvbhatta:
> I like your solution, it seems though that while the mockups show the glow outside the focus outline that isn't so in the implementation.

It seems like decreasing the blur on `--nx-drop-shadow-focus` from 3px to 1px and increasing the alpha on the HSLA values for `--nx-color-interactive-shadow-focus` from 0.3 to 1:

image

Another option is that we could use `box-shadow`:

image

Are there any better alternatives?

lib/src/base-styles/_nx-radio-checkbox.scss
avnvbhatta: > This is minor, and a personal thing (Ross doesn't share this view), but I prefer to see CSS attributes in alphabetical order.
> 
> Exceptions are that min/max/height, min/max/width are grouped, and top/right/bottom/left are grouped after `position`.

Noted. Fixed.
b711591bb80b4e5c3cae2dc00d0587dd74f80419

glenhunter:
> > I like your solution, it seems though that while the mockups show the glow outside the focus outline that isn't so in the implementation.
> 
> It seems like decreasing the blur on `--nx-drop-shadow-focus` from 3px to 1px and increasing the alpha on the HSLA values for `--nx-color-interactive-shadow-focus` from 0.3 to 1:
> 
> Another option is that we could use `box-shadow`:
> 
> Are there any better alternatives?

I think that we use `box-shadow` most places.

avnvbhatta:
> > > I like your solution, it seems though that while the mockups show the glow outside the focus outline that isn't so in the implementation.
> > 
> > 
> > It seems like decreasing the blur on `--nx-drop-shadow-focus` from 3px to 1px and increasing the alpha on the HSLA values for `--nx-color-interactive-shadow-focus` from 0.3 to 1:
> > Another option is that we could use `box-shadow`:
> > Are there any better alternatives?
> 
> I think that we use `box-shadow` most places.

Switched it to use `box-shadow`.
ea808ce4706983c532ea6f531673da20be691fbe

gallery/visualtests/nxCheckbox.js
```diff
@@ -19,7 +19,7 @@ describe('NxCheckbox', function() {
     it('has a light grey border and white background by default', simpleTest(selector));
     it('has a black border when hovered', hoverTest(selector));
 
-    it('has a blue background and white checkmark when clicked', async function() {
+    it('has a blue background and white checkmark and a light blue outer border when clicked', async function() {
```
rpokorny, line 22: The test below is for when it is clicked and focused, so this was supposed to be the test for when it is clicked and _not_ focused. Maybe it was failing to be that - all the more indication that the old focus style was too hard to see.  But my point is that I don't think this test should be expecting the focus style, and it should be modified so that the element is not in fact focused (or hovered) when the screenshot is taken

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -31,9 +43,11 @@
   }
 
   &:focus-within {
+    &::after {
+      display: block;
```
rpokorny, line 47: Why not define the whole thing here rather than having most of it higher up with `display: none`?

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -20,6 +20,18 @@
...
   width: calc(min(#{$max-width}, 100%));
   max-width: max-content;
+  position: relative;
+
+  &::after {
+    border: 1px solid var(--nx-color-interactive-border-focus);
+    box-shadow: var(--nx-box-shadow-focus);
+    content: '';
+    display: none;
+    height: 22px;
```
rpokorny, line 30: I worry a little bit about the fact that the width and height here are assuming the dimesions of the control, even though those dimensions are defined in the individual `NxCheckbox.scss` and `NxRadio.scss` files rather than here.  it might make sense to move the `height`, `width`, and maybe `left` to those files

gallery/visualtests/nxCheckbox.js
avnvbhatta: > The test below is for when it is clicked and focused, so this was supposed to be the test for when it is clicked and _not_ focused. Maybe it was failing to be that - all the more indication that the old focus style was too hard to see. But my point is that I don't think this test should be expecting the focus style, and it should be modified so that the element is not in fact focused (or hovered) when the screenshot is taken

I accidentally added the "focused" test when the checkbox/radio is clicked and _not_ focused. This has been fixed.
982993dfc75068be14ca66b90b7028342c5923e4

lib/src/base-styles/_nx-radio-checkbox.scss
avnvbhatta: > Why not define the whole thing here rather than having most of it higher up with `display: none`?

I agree. Defining it here makes it clearer.
aa7d9db5def38dd8cfb54ae2af7e45f49d3f4fba

lib/src/base-styles/_nx-radio-checkbox.scss
avnvbhatta: > I worry a little bit about the fact that the width and height here are assuming the dimesions of the control, even though those dimensions are defined in the individual `NxCheckbox.scss` and `NxRadio.scss` files rather than here. It might make sense to move the `height`, `width`, and maybe `left` to those files

Separating them out makes sense. I initially had them grouped because the `width`, `height` and `left` were the same for both the radio and the checkboxe. I have implemented the change.
aa7d9db5def38dd8cfb54ae2af7e45f49d3f4fba

gallery/visualtests/nxCheckbox.js
rpokorny: > I accidentally added the "focused" test when the checkbox/radio is clicked and not focused. This has been fixed.
982993d

Does the focus border show up in the applitools baseline for this test? I know the test doesn't explicitly focus the input, but generally things get focus when you click on them, which is why I'm thinking this test, as currently written, would actually pick up the focus state too.

lib/src/base-styles/_nx-radio-checkbox.scss
rpokorny: > I initially had them grouped because the width, height and left were the same for both the radio and the checkboxe.

Right you just have to consider whether they're the same due to some fundamental link or just coincidentally.  Since the dimensions of the checkbox box and radio circle are defined in their respective files and not the shared file, that would suggest that they're only coincidentally the same.

gallery/visualtests/nxCheckbox.js
avnvbhatta: > > I accidentally added the "focused" test when the checkbox/radio is clicked and not focused. This has been fixed.
> > [982993d](https://github.com/sonatype/sonatype-react-shared-components/commit/982993dfc75068be14ca66b90b7028342c5923e4)
> 
> Does the focus border show up in the applitools baseline for this test? I know the test doesn't explicitly focus the input, but generally things get focus when you click on them, which is why I'm thinking this test, as currently written, would actually pick up the focus state too.

The focus border previously (while clicked) was so subtle that it's incredibly difficult to see if there's even a focus border present. It's really difficult to tell whether the baseline picks up the focus state or not, both for the radios and checkboxes.

Checkbox clicked baseline:
image

Radio clicked baseline:
image

rpokorny: > The focus border previously (while clicked) was so subtle that it's incredibly difficult to see if there's even a focus border present. It's really difficult to tell whether the baseline picks up the focus state or not, both for the radios and checkboxes.

In the past, sure.  But that doesn't matter now thanks to this branch.  Are the new focus borders visible in the screenshot for this test, in this branch?

rpokorny: > Yes, the focus borders are visible for this test (Default NxCheckbox has a blue background and white checkmark when clicked)

Which makes it effectively identical to the next test below, which isn't what we want.  So this test needs to be modified to blur the element after clicking on it

avnvbhatta: > > Yes, the focus borders are visible for this test (Default NxCheckbox has a blue background and white checkmark when clicked)
> 
> Which makes it effectively identical to the next test below, which isn't what we want. So this test needs to be modified to blur the element after clicking on it

Added blur after click to both checkboxes and radiobuttons.
7a75d5873a9038d9f66c7cd22b1bfae945ae03b1

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -20,6 +20,11 @@
   // the end of the text
   width: calc(min(#{$max-width}, 100%));
   max-width: max-content;
+  position: relative;
+
+  &::after {
```
rpokorny, line 25: Is this block actually necessary? I thought `::after` didn't show up until `content` was set on it.

avnvbhatta: > Is this block actually necessary? I thought `::after` didn't show up until `content` was set on it.

You're right. `::after` does not show up unless `content` is set. So that block is unnecessary.
RSC-600-focused-radio-checkbox

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -31,9 +32,15 @@
...
+      box-shadow: var(--nx-box-shadow-focus);
+      content: '';
+      display: block;
+      position: absolute;
+    }
+
     .nx-radio-checkbox__control {
-      border-color: var(--nx-color-interactive-border-focus);
-      stroke: var(--nx-color-interactive-border-focus);
       filter: drop-shadow(var(--nx-drop-shadow-focus));
```
rpokorny, line 44: Shouldn't this be removed? It creates an additional glow around the radio/checkbox itself.

avnvbhatta: > Shouldn't this be removed? It creates an additional glow around the radio/checkbox itself.

It has been removed.
37849b5e287821b39691cdefdf4c1a6165f13bfb

rpokorny (approved):

PR #426, opened 2021-07-01
NxModal Close and Focus refactor - RSC-510

rpokorny:
https://issues.sonatype.org/browse/RSC-510

This deviates a bit from what the ticket was originally about.  I am against making `onClose` optional as it would be too easy for future modals to forget about it, missing that usability feature.  Instead I took another pass at making `NxModal` work in conjunction with other elements which can be closed with the ESC key.  Additionally this caused me to take a closer look at the focus behavior around the opening and closing of modals, which is a closely related issue.  The approach I came up with to make everything work as well as possible is as follows:

In browsers which have more-or-less full support for `<dialog>` elements (only Chromium-based browsers currently), the native `cancel` event on the modal is used as the signal to close the modal.  This leaves it up to the browser what standard ways of canceling the modal are made available to the user, but allowing them to press ESC is the typical one cited in the HTML standards.  In practice, what this means is that if an ESC keydown event is not cancelled while a modal is up, then after that even finishes bubbling a `cancel` event is fired on the modal itself.  Note that if we still reacted to the `keydown` instead of listening for `cancel`, that would (and *does*, on master) cause problems when multiple modals are present in that the inner modal would be removed programmatically in the `onClose` handler and the browser would then `cancel` the _outer_ modal, which can be seen by the fact that the outer modal's `open` attribute gets removed.  This puts things into an inconsistent state - the browser considers the outer modal to be closed, but our styles and JavaScript still have it presented and interactive.  In practical terms, what this meant was that when you ESC out of a inner modal, the focus trapping on the outer modal no longer works.  Listening for `cancel` instead as is done in this PR fixes that.

The `onClose` prop has been renamed to `onCancel` to more closely align with the meanings of "close" and "cancel" in relation to `<dialog>` elements.  `onClose` has been kept as a deprecated alias for `onCancel`.

In browsers which do not fully support `<dialog>`, the keydown listening is still done similarly to before.  However, to avoid complications with other elements, the listening is no longer done on the `document` but rather on the dialog backdrop.  A keydown which closes the modal here gets `stopPropagation` called on it so that it does not fire any other keydown listeners attached to the document.  Note that the modal focus-trapping does not work in these browsers, so we didn't have to worry about breaking it.  Also note however that with these changes ESC will not close a modal if the user has accidentally tabbed out of the modal.

`NxDropdown` and `NxSegmentedButton` have also been adjusted in this PR to play more nicely with modals and other ESC-closed elements.  The existing behavior for these elements has been that ESC can close them if they are focused when it is pressed.  This is still the case, but they have been adjusted so that the keydown which closes them gets canceled (that is, `preventDefault` is called on it) so that it does not close any other surrounding elements, such as a modal containing the dropdown.

In addition to all that, there were several changes to explicit focus management which needed to be made to prevent focus from defaulting back to the `<body>` and thus escaping modals:
* When a modal opens, it remembers which element had focus before and it resets focus back to that element when it closes.  According to the spec, this is supposed to be built-in behavior for `<dialog>`s but I didn't observe it even in browsers which otherwise support that element.
* When a modal is opened on browsers which don't support `<dialog>`, the `<dialog>` element itself is explicitly focused.  Without this, focus remained on the background page.  In browsers which do support `<dialog>`, the built-in behavior already was focusing the first focusable element within the modal.  That behavior remains.
* When a dropdown is closed and focus is on one of the dropdown menu items, focus gets moved to the dropdown toggle button.

### DIFF
--- a/gallery/src/components/NxModal/NxModalEscExample.tsx
+++ b/gallery/src/components/NxModal/NxModalEscExample.tsx
@@ -17,12 +17,15 @@ export default function NxModalStackedExample() {
       modal1CancelHandler = () => setShowModal(false),
       modal2CancelHandler = () => setShowModal2(false);
 
+  function customComponentKeydownListener(evt: KeyboardEvent) {
+    if (evt.key === 'Escape' || evt.key === 'Esc') {
+      setShowCustomComponent(false);
+    }
+  }
+
   useEffect(() => {
-    document.addEventListener('keydown', (evt) => {
-      if (evt.key === 'Escape' || evt.key === 'Esc') {
-        setShowCustomComponent(false);
-      }
-    });
+    document.addEventListener('keydown', customComponentKeydownListener);
+    return () => { document.removeEventListener('keydown', customComponentKeydownListener); };
   }, []);
 
   return (
--- a/gallery/src/components/NxModal/NxModalEscExample.tsx
+++ b/gallery/src/components/NxModal/NxModalEscExample.tsx
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, {useState, useEffect} from 'react';
+
+import {NxModal, NxFontAwesomeIcon, NxButton, NxDropdown, NxP, useToggle} from '@sonatype/react-shared-components';
+import {faAngry} from '@fortawesome/free-solid-svg-icons';
+
+export default function NxModalStackedExample() {
+  const [showModal, setShowModal] = useState(false),
+      [showModal2, setShowModal2] = useState(false),
+      [showDropdown, toggleDropdown] = useToggle(false),
+      [showCustomComponent, setShowCustomComponent] = useState(false),
+      modal1CancelHandler = () => setShowModal(false),
+      modal2CancelHandler = () => setShowModal2(false);
+
+  useEffect(() => {
+    document.addEventListener('keydown', (evt) => {
+      if (evt.key === 'Escape' || evt.key === 'Esc') {
+        setShowCustomComponent(false);
+      }
+    });
+  }, []);
+
+  return (
+    <>
+      <NxButton onClick={() => setShowCustomComponent(true)}>Open Custom Expandable Panel</NxButton>
+      {showCustomComponent &&
+        <div className="gallery-custom-expandable">
+          <NxP>Press ESC anywhere on the page to close me!</NxP>
+          <NxButton onClick={() => setShowModal(true)}>Open Modal 1</NxButton>
+        </div>
+      }
+      {showModal &&
+        <NxModal id="nx-modal-stacked-example" onCancel={modal1CancelHandler}>
+          <header className="nx-modal-header">
+            <h2 className="nx-h2">
+              <NxFontAwesomeIcon icon={faAngry} />
+              <span>NxModal with a dropdown that opens another modal</span>
+            </h2>
+          </header>
+          <div className="nx-modal-content">
+            <NxDropdown label="Click to Expand"
+                        isOpen={showDropdown}
+                        onToggleCollapse={toggleDropdown}
+                        onCloseClick={evt => evt.preventDefault()}>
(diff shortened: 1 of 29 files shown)


rpokorny:
I've tested these changes locally in IQ, both with IQ's current code and with adjustments to bring IQ more in line with how we want NxModal to be used.

gallery/src/components/NxModal/NxModalEscExample.tsx
```diff
@@ -0,0 +1,88 @@
...
+      modal1CancelHandler = () => setShowModal(false),
+      modal2CancelHandler = () => setShowModal2(false);
+
+  useEffect(() => {
+    document.addEventListener('keydown', (evt) => {
+      if (evt.key === 'Escape' || evt.key === 'Esc') {
+        setShowCustomComponent(false);
+      }
+    });
+  }, []);
```
avnvbhatta, line 26: Should we remove the `eventListener` on unmount?

rpokorny: > Should we remove the eventListener on unmount?

Yes, thanks for the reminder. cb5dcebd

lib/src/components/NxModal/NxModal.tsx
```diff
@@ -27,60 +27,98 @@ const NxModal: FunctionComponent<Props> = ({className, onClose, variant, role, .
...
+     *
+     * Note: not supported in safari, which is why the conditional is here. Once safari gets
+     * it together, we should be able to take advantage of the "top layer" functionality built into
+     * the browser around modals to simplify the NxModal styling around z-index handling
+     */
+    if (hasNativeModalSupport) {
+      el.showModal();
+    }
+    else {
+      // without native support we don't trap focus in the modal, but we can at least start it off there
```
glenhunter, line 79: Is that "we don't" or "we can't"?

rpokorny: > Is that "we don't" or "we can't"?

Don't. It's something that can attempted in JavaScript, more or less successfully, but I always found it hacky and error prone and now I feel comfortable just blaming browsers for not supporting `<dialog>` if anyone points out that we don't do it.

marushkevych (commented):
+1 for the overall approach

rpokorny:
> +1 for the overall approach

@marushkevych Is that different from a full +1?

marushkevych:
> Is that different from a full +1?

yes, I did not do the full code review

lib/src/components/NxModal/NxModal.tsx
```diff
@@ -27,60 +27,98 @@ const NxModal: FunctionComponent<Props> = ({className, onClose, variant, role, .
...
+    if (dialog && hasNativeModalSupport) {
+      /* eslint-disable @typescript-eslint/no-non-null-assertion */
+      dialog.addEventListener('cancel', onCancel!);
+      return () => { dialog.removeEventListener('cancel', onCancel!); };
+      /* eslint-enable @typescript-eslint/no-non-null-assertion */
     }
-  }, []);
+    else {
+      return undefined;
+    }
```
avnvbhatta, line 103: Is this necessary?

rpokorny: > Is this necessary?

Only to make typescript happy

gallery/src/components/NxModal/NxModalEscExample.tsx
```diff
@@ -0,0 +1,91 @@
...
+          </header>
+          <div className="nx-modal-content">
+            <NxDropdown label="Click to Expand"
+                        isOpen={showDropdown}
+                        onToggleCollapse={toggleDropdown}
+                        onCloseClick={evt => evt.preventDefault()}>
+              <button onClick={() => setShowModal2(true)} className="nx-dropdown-button">
+                Open Modal 2
+              </button>
+            </NxDropdown>
```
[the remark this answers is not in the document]
rpokorny: > A little unrelated, but on the subject, when the first modal is open, pressing Esc closes the dropdown menu, but clicking outside the dropdown doesn't (works on NxDropdown example in the gallery).

That's actually deliberate.  This example is meant to emulate aspects of real situations in IQ.  In one of those situations, there is a dropdown that contains buttons that open modals. That dropdown is specially not supposed to close when the buttons are clicked.

> When the dropdown is clicked and the menu is open, closing and re-opening the first modal seems to have the dropdown menu open.

2ad9857e

avnvbhatta (approved):

PR #428, opened 2021-07-02
CSS Variable docs - RSC-592

rpokorny:
https://issues.sonatype.org/browse/RSC-592

Question for the team: I felt that the Mixins and Helpers section was the best place to put this in the nav, and the `styles` directory the best place in the file structure.  I feel like these variables are "helpers" more than they are anything else in our nav.  This is however inconsistent with where the Color Palette docs are located, both in the nav and the file structure, where they are considered "guidelines".  I strongly suspect that where one of these is located, so should be the other one.  So either the Color Palette docs should be moved to Mixins and Helpers (and the `styles` dir) or this new page should be moved to "Guidelines".  Which do you prefer?


gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -0,0 +1,334 @@
...
+        To avoid naming conflicts with other custom properties which may be present, all RSC custom properties
+        begin with <NxCode>--nx-</NxCode>. Following that prefix, the variables follow one of several consistent
+        naming schemes, depending on what type of value the variable holds.
+      </NxP>
+    </GalleryDescriptionTile>
+    <GalleryTile title="Spacing Properties">
+      <NxP>
+        RSC has a set of spacing properties which are intended to be used for margin and padding values. They are
+        not intended to be used for sizing of heights and widths. These properties are all simple multiples of
+        a base spacing value, and are named based on what multiple they are. The base value itself is available
```
glenhunter, line 61:
```suggestion
        a base spacing value (<NxCode>4px</NxCode>), and are named based on their multiple. The base value itself is available
```

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -0,0 +1,334 @@
...
+        using the <NxCode>--foo: blue</NxCode> syntax, while "variables" refers to the <em>usage</em> of custom
+        properties within <NxCode>var()</NxCode> expressions to substitute the named value into another CSS
+        property definition. They are defined as properties, but used as variables.
+      </NxP>
+      <NxP>
+        It is worth noting that as the spec points out, there are conceivably other uses of custom properties beside
+        CSS variables. For instance, arbitrary text could be included as the property value and then extracted via
+        JavaScript using the CSSOM APIs. In an extreme example from the spec, the value of a custom property could
+        even be a snippet of JavaScript code which a script extracts and evals!{' '}
+        <small className="gallery-small-parenthetical">(What a horrible idea)</small>
```
glenhunter, line 329:
```suggestion
        <small className="gallery-small-parenthetical">(What a horrible idea, please don't do this.)</small>
```

glenhunter:
I don't have any issue with this page moving to the "Guidelines" section. It's more documentation than a 'mixin' or 'helper'.  

"Guidelines" may have outlived its usefulness as a name and something more representative could be found for that section.

glenhunter (approved):
Minor tweaks suggested. Really nice documentation work.

gallery/src/styles/CssVariables/CssVariablesPage.tsx
rpokorny: I'll do you one better. 8560f931

gallery/src/styles/CssVariables/CssVariablesPage.tsx
rpokorny: 2c16fb75

rpokorny:
> It's more documentation than a 'mixin' or 'helper'.

I'm not sure I agree.  Aside from the Color Palette page (which I am suggesting moving), everything under Guidelines consists pretty much entirely of long explanatory text and examples, not really enumerations of available technical entities (like class names or react component props).  This new page has a bit of both but I think more than anything else it is an enumeration of the available variables (which I'd say are technical entities every bit as much as class names are).  This is why it feels to me like it fits in more with the other sections of RSC than with the Guidelines.  I'm feeling the same reasoning with regards to the Color Palette page.

rpokorny:
> I'm not sure I agree. Aside from the Color Palette page (which I am suggesting moving), everything under Guidelines consists pretty much entirely of long explanatory text and examples, not really enumerations of available technical entities (like class names or react component props). This new page has a bit of both but I think more than anything else it is an enumeration of the available variables (which I'd say are technical entities every bit as much as class names are). This is why it feels to me like it fits in more with the other sections of RSC than with the Guidelines. I'm feeling the same reasoning with regards to the Color Palette page.

Moved the Color Palette page in 7ac7401

PR #429, opened 2021-07-06
Fix tooltip blips - RSC-605

rpokorny:
https://issues.sonatype.org/browse/RSC-605

The issue only appeared to affect `<button>` elements.  The underlying problem appears to be that the `<button>` element was being removed and re-added to the DOM every time the tooltip appeared or disappeared.  Since it was removed before the tooltip disappearance animation could complete, the tooltip would move to the corner of the screen.  I also wonder if this replacement behavior could cause other subtle bugs.

Strangely, I found that this removal only happens when the tooltip is a plain string.  When the tooltip is JSX, it works fine.  So my solution is to make sure it's always JSX.


glenhunter (approved):
Checked in Safari, FF, and Chrome, no blipping.

PR #430, opened 2021-07-06
Improve NxToggle focus style RSC-602

avnvbhatta:
https://issues.sonatype.org/browse/RSC-602

- Added focus styles on `NxToggle` that shows a light outer blue border when focused.
- Modified visual tests to align with new design spec.

Achieved by using an absolutely positioned `::after` pseudoelement that is visible on focus.

### DIFF
--- a/gallery/visualtests/nxToggle.js
+++ b/gallery/visualtests/nxToggle.js
@@ -21,11 +21,15 @@ describe('NxToggle', function() {
     it('has a black border when hovered', hoverTest(selector));
 
     it('has a blue background and white indicator when clicked', async function() {
-      const targetElement = await browser.$(selector);
+      const blurSelector = `${selector} input`,
+          [targetElement, blurElement] = await Promise.all([browser.$(selector), browser.$(blurSelector)]);
 
       await targetElement.scrollIntoView({ block: 'center' });
       await targetElement.click();
       await targetElement.moveTo({ xOffset: -10, yOffset: -10 });
+      await browser.execute(function(el) {
+        el.blur();
+      }, blurElement);
 
       try {
         await browser.eyesRegionSnapshot(null, Target.region(targetElement));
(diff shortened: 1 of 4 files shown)


gallery/visualtests/nxToggle.js
```diff
@@ -56,7 +61,8 @@ describe('NxToggle', function() {
       }
     });
 
-    it('dark border, blue background and white indicator when clicked, focused, and hovered', async function() {
+    it(`dark border, blue background and white indicator with light outer blue border
```
rpokorny, line 64: The hover border is black on these actually

lib/src/components/NxToggle/NxToggle.scss
```diff
@@ -31,12 +32,22 @@
...
+        border: 1px solid var(--nx-color-interactive-border-focus);
+        border-radius: 16px;
+        box-shadow: var(--nx-box-shadow-focus);
+        content: '';
+        display: block;
+        height: 30px;
+        position:absolute;
+        left: -4px;
+        top: -4px;
+        width: 50px;
```
rpokorny, line 49: Should we turn the `height` and `width` of the control into local scss variables and derive the focus border sizes from it?

avnvbhatta: I think it's a good idea, but I think would it make sense to do it under a different JIRA ticket and use variables for focus borders of radios and checkboxes as well?

rpokorny: > I think it's a good idea, but I think would it make sense to do it under a different JIRA ticket and use variables for focus borders of radios and checkboxes as well?

Sure.  Put in a tech debt ticket and link it here.

avnvbhatta: > > I think it's a good idea, but I think would it make sense to do it under a different JIRA ticket and use variables for focus borders of radios and checkboxes as well?
> 
> Sure. Put in a tech debt ticket and link it here.

Tech Debt ticket: https://issues.sonatype.org/browse/RSC-620

rpokorny (approved):

PR #431, opened 2021-07-06
Hide radio and checkbox inputs RSC-613

avnvbhatta:
https://issues.sonatype.org/browse/RSC-613

Fixed by adding `-webkit-appearance: none;` that hides the input element on webkit based browsers.

### DIFF
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -109,6 +109,7 @@ svg.nx-radio-checkbox__control {
   // in FF, the checkboxes don't respect the width: 0 and we need this in order to hide them
   // Fix discovered via the title of https://bugzilla.mozilla.org/show_bug.cgi?id=605985
   -moz-appearance: none;
+  -webkit-appearance: none;
 
   &:focus {
     // less-nice IE11 focus style because IE11 doesn't support :focus-within
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -104,7 +104,6 @@ svg.nx-radio-checkbox__control {
 
 .nx-radio-checkbox__input {
   margin: 0;
-  width: 0;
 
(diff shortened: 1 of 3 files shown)


lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -109,6 +109,7 @@ svg.nx-radio-checkbox__control {
   // in FF, the checkboxes don't respect the width: 0 and we need this in order to hide them
   // Fix discovered via the title of https://bugzilla.mozilla.org/show_bug.cgi?id=605985
   -moz-appearance: none;
+  -webkit-appearance: none;
```
rpokorny, line 112: Does just `appearance` work? Also with this do we still need the `width: 0`?

avnvbhatta: Just having `appearance: none;` doesn't seem to work. And no, it looks like we do not need the `width: 0;` anymore.
19fb2c9f28cc17fde8b1ac0f68650d8f6a026e3a

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -104,11 +104,11 @@ svg.nx-radio-checkbox__control {
 
 .nx-radio-checkbox__input {
   margin: 0;
-  width: 0;
 
   // in FF, the checkboxes don't respect the width: 0 and we need this in order to hide them
   // Fix discovered via the title of https://bugzilla.mozilla.org/show_bug.cgi?id=605985
```
rpokorny, line 109: I think at this point this comment can just be removed.  We are consistently using `appearance` to hide the native input so this comment about firefox specifics isn't important.

rpokorny (approved):

lib/src/base-styles/_nx-radio-checkbox.scss
avnvbhatta: > I think at this point this comment can just be removed. We are consistently using `appearance` to hide the native input so this comment about firefox specifics isn't important.

Removed.
f7ab44bcecb80e17b5fa8b7859baf0de80fad44f

PR #432, opened 2021-07-07
max-width of 800px for wrapping text RSC-590

glenhunter:
https://issues.sonatype.org/browse/RSC-590

I created a variable `--nx-width-text-wrapping` and applied it to `nx-p`, as well as `nx-blockquote`, `nx-pre`, `nx-read-only` and `nx-list`s.

I added it to form labels and legends 9and sub-labels) and added an wrapping example to the form layout page.

I also added it to the basic elements in `NxVulnerabilityDetails` despite the fact that those elements are currently constrained by their parent.


rpokorny:
> I created a variable --nx-width-text-wrapping and applied it to nx-p, as well as nx-blockquote, nx-pre, nx-read-only and nx-lists.
>
> I added it to form labels and legends 9and sub-labels) and added an wrapping example to the form layout page.

What about things like tile subtitles and page-title descriptions?

glenhunter:
> > I created a variable --nx-width-text-wrapping and applied it to nx-p, as well as nx-blockquote, nx-pre, nx-read-only and nx-lists.
> > I added it to form labels and legends 9and sub-labels) and added an wrapping example to the form layout page.
> 
> What about things like tile subtitles and page-title descriptions?

Here's the tile subtitles: https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/1103ec9688286afdd63b368ace079a609b8cc98c

The page title description in the example is an `nx-p` and so inherits the max width and IMO a paragraph is the correct element for that usage.

rpokorny:
> The page title description in the example is an nx-p and so inherits the max width and IMO a paragraph is the correct element for that usage.

I'm surprised to see those `<p>` tags in those examples – we typically don't demand that people use extra paragraph tags where there's already a surrounding container.  For instance, why do we have that there but not in the tile subtitle? I think it would be safer to go ahead and put the max-width on the page-title description as well.

rpokorny:
Subheader in accordion?
image

rpokorny:
Adjusting the `max-width` of `nx-alert` was also part of this ticket

lib/src/base-styles/_nx-read-only.scss
```diff
@@ -23,8 +23,10 @@
 .nx-read-only__label {
   @include bold();
   margin: var(--nx-spacing-8x) 0 var(--nx-spacing-4x) 0;
+  max-width: var(--nx-width-text-wrapping);
 }
 
 .nx-read-only__data {
   margin: 0 0 var(--nx-spacing-2x) 0;
+  max-width: var(--nx-width-text-wrapping);
```
rpokorny, line 31: I suspect we don't want this set for `.nx-read-only__data` that are children of `.nx-read-only--grid`.  The grid should be in charge of the layout there.

lib/src/base-styles/_nx-lists.scss
```diff
@@ -70,12 +70,15 @@
   grid-area: list-text;
   margin: 0;
   padding-right: var(--nx-spacing-6x);
+  max-width: var(--nx-width-text-wrapping);
```
rpokorny, line 73: Centering of empty text is broken.
image

lib/src/base-styles/_nx-lists.scss
```diff
@@ -70,12 +70,15 @@
...
+
 
   &.nx-truncate-ellipsis {
     // Something about the grid layout gets confused due to overflow: hidden on this element
     // and adds vertical space to the second row of the grid.
     // Experimentally, setting an explicit height here fixes that.
     height: calc(var(--nx-line-height) * 1em);
+    max-width: unset;
   }
 }
```
rpokorny, line 84: Should we add the max-width to `.nx-list__description` as well?

gallery/src/components/NxOverflowTooltip/NxOverflowTooltipExample.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+#nx-overflow-tooltip-dynamic-example {
+  .nx-p {
+    max-width: unset;
```
rpokorny, line 9: Or should this be on `.nx-truncate-ellipsis` in general?

rpokorny:
Should we build in handling of long unbroken strings or let them be?

lib/src/base-styles/_nx-forms.scss
```diff
@@ -39,6 +39,7 @@
 .nx-label, .nx-legend {
   display: block;
   padding: 0;
+  max-width: var(--nx-width-text-wrapping);
```
rpokorny, line 42: I don't think we've ever considered whether these are supposed to wrap or truncate in the first place. We've always assumed they should/would be short enough for it not to matter.  Which should they do? Or if we support neither and just demand they be short, this property isn't really needed here.

gallery/src/components/NxOverflowTooltip/NxOverflowTooltipExample.scss
glenhunter: > Or should this be on `.nx-truncate-ellipsis` in general?

Wouldn't that cause issues where there's a `max-width` set that we want to keep? That seems like an overly broad and potentially dangerous change.

Note also that in this case I'm just tweaking an example so that a visual test runs correctly.

glenhunter:
> Should we build in handling of long unbroken strings or let them be?

I'm of the opinion that we should let them be. since if they are there and long enough there's probably (hopefully) a good reason

lib/src/base-styles/_nx-forms.scss
glenhunter: > I don't think we've ever considered whether these are supposed to wrap or truncate in the first place. We've always assumed they should/would be short enough for it not to matter. Which should they do? Or if we support neither and just demand they be short, this property isn't really needed here.

IMO they should be short and this is _should_ be unnecessary, but I think that it's safer to add it here. Call it CYA. Should there be an absurdly long label I think that wrapping is more appropriate than truncation.

lib/src/base-styles/_nx-lists.scss
glenhunter: > Centering of empty text is broken.

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/f270a3be8f1a7edfe74f870805a9f4f9f2d72aae

lib/src/base-styles/_nx-lists.scss
glenhunter: > Should we add the max-width to `.nx-list__description` as well?

I think so: https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/c7ceeebcde0087f3015b66cb27125a666f3cefdb

rpokorny: > I debated adding it to the terms as well (see note re CYA below) but in the end I did not.

On way or another terms should match form labels since that's the whole intent with `nx-read-only`

glenhunter:
> Subheader in accordion?

Per Derrek's Slack message applied here to all headings: https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/971552b5264ee278d2b4ec8b9117f9d534fe295e

glenhunter:
> Adjusting the `max-width` of `nx-alert` was also part of this ticket

https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/029edce6555e98483bbaf3cfd8a92644bc72b484

lib/src/base-styles/_nx-lists.scss
glenhunter: > > I debated adding it to the terms as well (see note re CYA below) but in the end I did not.
> 
> One way or another terms should match form labels since that's the whole intent with `nx-read-only`

The terms have `width: 250px` on them so I think they're good.

lib/src/base-styles/_nx-read-only.scss
glenhunter: > I suspect we don't want this set for `.nx-read-only__data` that are children of `.nx-read-only--grid`. The grid should be in charge of the layout there.

https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/471e9c58d35433af6175f3ce8ef936f8c7149b7a

glenhunter:
> > The page title description in the example is an nx-p and so inherits the max width and IMO a paragraph is the correct element for that usage.
> 
> I'm surprised to see those `<p>` tags in those examples – we typically don't demand that people use extra paragraph tags where there's already a surrounding container. For instance, why do we have that there but not in the tile subtitle? I think it would be safer to go ahead and put the max-width on the page-title description as well.

I was worried that a `max-width` on the container could come back and bite us later but I think we'll probably be ok. Done here: https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/1534e2ffbaddae6ee94b724e37879b8c2f1faa35

lib/src/base-styles/_nx-lists.scss
rpokorny: > The terms have width: 250px on them so I think they're good.

Ah right I was thinking of `nx-read-only` when I wrote that.

rpokorny:
> Per Derrek's Slack message applied here to all headings: 971552b and b6f0481

Re that second commit, which slack message? I don't think I agree with that change; I'm generally against max-widths on elements that don't wrap, and we have many headers that don't

rpokorny:
> > Per Derrek's Slack message applied here to all headings: 971552b and b6f0481
>
> Re that second commit, which slack message? I don't think I agree with that change; I'm generally against max-widths on elements that don't wrap, and we have many headers that don't

For instance, if you look at the `nx-tile` example with action buttons, and remove one of the action buttons, you get this:
image

I don't think there's any reason for the header to cut off preemptively there.

glenhunter:
> > Per Derrek's Slack message applied here to all headings: [971552b](https://github.com/sonatype/sonatype-react-shared-components/commit/971552b5264ee278d2b4ec8b9117f9d534fe295e) and [b6f0481](https://github.com/sonatype/sonatype-react-shared-components/commit/b6f0481cb5e1af65bf9d0e0184f4aa372ab525a7)
> 
> Re that second commit, which slack message? I don't think I agree with that change; I'm generally against max-widths on elements that don't wrap, and we have many headers that don't

https://sonatype.slack.com/archives/CJUF9G7T2/p1625850868291700

glenhunter:
> > > Per Derrek's Slack message applied here to all headings: [971552b](https://github.com/sonatype/sonatype-react-shared-components/commit/971552b5264ee278d2b4ec8b9117f9d534fe295e) and [b6f0481](https://github.com/sonatype/sonatype-react-shared-components/commit/b6f0481cb5e1af65bf9d0e0184f4aa372ab525a7)
> > 
> > 
> > Re that second commit, which slack message? I don't think I agree with that change; I'm generally against max-widths on elements that don't wrap, and we have many headers that don't
> 
> For instance, if you look at the `nx-tile` example with action buttons, and remove one of the action buttons, you get this:
> image
> 
> I don't think there's any reason for the header to cut off preemptively there.

There is, now all headers have a max-width, so I need to sort that out

rpokorny:
> There is, now all headers have a max-width, so I need to sort that out

I still think that's a bad idea, as I've argued many times before

glenhunter:
> > There is, now all headers have a max-width, so I need to sort that out
> 
> I still think that's a bad idea, as I've argued many times before

@dturner32 ^^

glenhunter:
> > There is, now all headers have a max-width, so I need to sort that out
> 
> I still think that's a bad idea, as I've argued many times before

There were three places that was an issue: https://github.com/sonatype/sonatype-react-shared-components/pull/432/commits/b3a17febf6b97bc8113a1fd4a901c4ffeae8ed22

rpokorny (approved):

PR #433, opened 2021-07-07
Add SCSS variables for focus borders RSC-620

avnvbhatta:
https://issues.sonatype.org/browse/RSC-620

- The focus border style's `width`, `height` and absolute positioning values (`left`, `top`) were previously hardcoded. 
- New implementation makes use of dynamic calculations to correctly position the focus borders on `NxToggle`, `NxRadio` and `NxCheckbox`.
- All calculations are based on a `1px` border width of the focus border.
- Introduced `mixins` to reuse calculation logic.

### DIFF
--- a/lib/src/components/NxCheckbox/NxCheckbox.scss
+++ b/lib/src/components/NxCheckbox/NxCheckbox.scss
@@ -4,7 +4,7 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-@import '../../scss-shared/_nx-focus-border-helpers';
+@import '../../scss-shared/nx-focus-border-helpers';
 
 $nx-checkbox__box-height: 16px;
 $nx-checkbox__box-width: 16px;
--- a/lib/src/components/NxCheckbox/NxCheckbox.scss
+++ b/lib/src/components/NxCheckbox/NxCheckbox.scss
@@ -8,7 +8,6 @@
 
 $nx-checkbox__box-height: 16px;
 $nx-checkbox__box-width: 16px;
-$nx-checkbox-focus-border-offset: 6px;
 
 .nx-checkbox__box {
   border-radius: 2px;
@@ -37,7 +36,7 @@ $nx-checkbox-focus-border-offset: 6px;
 
 .nx-checkbox:focus-within {
   &::after{
-    @include nx-radio-checkbox-focus-border($nx-checkbox__box-height, $nx-checkbox__box-width, $nx-checkbox-focus-border-offset);
+    @include nx-radio-checkbox-focus-border($nx-checkbox__box-height, $nx-checkbox__box-width);
     border-radius: 4px;
   }
 }
--- a/lib/src/components/NxCheckbox/NxCheckbox.scss
+++ b/lib/src/components/NxCheckbox/NxCheckbox.scss
@@ -6,15 +6,15 @@
  */
 @import '../../scss-shared/nx-focus-border-helpers';
 
-$nx-checkbox__box-height: 16px;
-$nx-checkbox__box-width: 16px;
+$nx-checkbox-height: 16px;
+$nx-checkbox-width: 16px;
 
 .nx-checkbox__box {
   border-radius: 2px;
   box-sizing: border-box;
-  height: $nx-checkbox__box-height;
+  height: $nx-checkbox-height;
   position: relative;
-  width: $nx-checkbox__box-width;
+  width: $nx-checkbox-width;
 
   .fa-check {
     color: var(--nx-swatch-white);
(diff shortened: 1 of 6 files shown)


lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -4,13 +4,18 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
+@import '../../scss-shared/_nx-focus-border-helpers';
```
rpokorny, line 7: Scss has goofy rules and shortcuts about its import paths. Generally the `_` and file extension aren't intended to be included, though it works fine if the are.
```suggestion
@import '../../scss-shared/nx-focus-border-helpers';
```

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -4,13 +4,18 @@
...
+
+$nx-checkbox__box-height: 16px;
+$nx-checkbox__box-width: 16px;
```
rpokorny, line 10: Using BEM style for variable names isn't something we've done so far.  @glenhunter do you think that's something we want to introduce or should these be changed?

lib/src/scss-shared/_nx-focus-border-helpers.scss
```diff
@@ -0,0 +1,23 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+// For dynamically calculating focus border styles
+// based on height, width and focus border offset
+// Calculations based on a 1px border
+
+// NxRadio and NxCheckbox require the left property for correct positioning of focus border
+@mixin nx-radio-checkbox-focus-border($height, $width, $offset) {
```
rpokorny, line 13: It's not super clear to me what the `$offset` represents.  It seems to be the total extra space between the control and the focus border on each axis? That seems a little confusing.  Additionally we are using the same value everywhere for it so far so I'd be inclined not to provide a parameter for it at all.

lib/src/components/NxCheckbox/NxCheckbox.scss
glenhunter: > Using BEM style for variable names isn't something we've done so far. @glenhunter do you think that's something we want to introduce or should these be changed?

IIRC we decided a long time ago that variables would not use BEM. If we were going to change that it should have happened in the shift to CSS variables. So I say that they should be changed

lib/src/components/NxCheckbox/NxCheckbox.scss
avnvbhatta: > Scss has goofy rules and shortcuts about its import paths. Generally the `_` and file extension aren't intended to be included, though it works fine if the are.

Fixed. 
c67fd99ba9304eb1436725fae06965fb906c349c

lib/src/scss-shared/_nx-focus-border-helpers.scss
avnvbhatta: > It's not super clear to me what the `$offset` represents. It seems to be the total extra space between the control and the focus border on each axis? That seems a little confusing. Additionally we are using the same value everywhere for it so far so I'd be inclined not to provide a parameter for it at all.

You are correct. It is the total space between the border and the focus border in each axis. I initially put $offset as a variable to provide support for custom spacing, but looking back, I doubt that there will ever be a case when we would need custom spacing for something like this. So this has been reverted to have fixed spacing and the parameter has been removed.
fa154a89ff2754976fbd5640f59e3fdb15d88671

lib/src/components/NxCheckbox/NxCheckbox.scss
avnvbhatta: > > Using BEM style for variable names isn't something we've done so far. @glenhunter do you think that's something we want to introduce or should these be changed?
> 
> IIRC we decided a long time ago that variables would not use BEM. If we were going to change that it should have happened in the shift to CSS variables. So I say that they should be changed

I understand now that for something as trivial as this, we do not require BEM. 
5b2f6050d4210671e331277e019a6cc41f19475f

lib/src/scss-shared/_nx-focus-border-helpers.scss
```diff
@@ -0,0 +1,23 @@
...
+@mixin nx-radio-checkbox-focus-border($height, $width) {
+    height: calc(#{$height} + 6px);
+    left: -4px;
+    width: calc(#{$width} + 6px);
+}
+
+// NxToggle requires an additional top property for correct positioning of focus border
+@mixin nx-toggle-focus-border($height, $width) {
+    @include nx-radio-checkbox-focus-border($height, $width);
+    top: -4px;
```
rpokorny, line 22: Since this is specific to the one case I'm not sure it needs to be a second mixin, it could just be directly in the nx-toggle styles

rpokorny (approved):

lib/src/scss-shared/_nx-focus-border-helpers.scss
avnvbhatta: 13087749db49d32aec38c7cef7040a9bfa4b8ed2

PR #434, opened 2021-07-09
NxFormSelect pristine state - RSC-500

rpokorny:
https://issues.sonatype.org/browse/RSC-500


glenhunter:
I think that we should replace the HTML form select with this component on the form layout page

rpokorny:
> I think that we should replace the HTML form select with this component on the form layout page

5b1c0081

glenhunter:
On the `NxCheckbox` and `NxRadio` pages we demonstrate text changing in accordance with the radios or checkboxes that are selected, should we do that here as well?

gallery/src/components/NxFormSelect/NxFormSelectPage.tsx
```diff
@@ -0,0 +1,122 @@
...
+            </NxTable.Row>
+          </NxTable.Head>
+          <NxTable.Body>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>isPristine</NxCode></NxTable.Cell>
+              <NxTable.Cell>boolean</NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell>false</NxTable.Cell>
+              <NxTable.Cell>
+                Should be set to true when the user has not yet adjusted the value of this form field.
```
glenhunter, line 46: Why isn't the default of this "true"?

rpokorny: > Why isn't the default of this "true"?

Why should it be?  I'd rather have a field that always looks non-pristine than one that always looks pristine.  I'd make this required but I don't want to break backwards compatibility (with the former convenience component) right now.

glenhunter: > > Why isn't the default of this "true"?
> 
> Why should it be? I'd rather have a field that always looks non-pristine than one that always looks pristine. I'd make this required but I don't want to break backwards compatibility (with the former convenience component) right now.

I was thinking that the default state would be pristine because in almost all cases the select would be in a pristine state when the form is initially loaded. I'm ok with not making it the default to avoid breaking the backwards compatibility.

rpokorny:
> On the NxCheckbox and NxRadio pages we demonstrate text changing in accordance with the radios or checkboxes that are selected, should we do that here as well?

813cf081

glenhunter (approved):

PR #435, opened 2021-07-09
Add button type attr where needed - RSC-626

rpokorny:
https://issues.sonatype.org/browse/RSC-626

The ticket only mentions `NxTreeView` (and its derivatives) but really there are a number of places where we have buttons that should not trigger form submissions if they are included within a form.


lib/src/components/NxSegmentedButton/NxSegmentedButton.tsx
```diff
@@ -46,13 +47,15 @@ const NxSegmentedButton = forwardRef<HTMLDivElement, Props>(
 
       return (
         <div ref={ref} className={classes} onKeyDown={onKeyDown} { ...attrs }>
-          <NxButton variant={variant}
+          <NxButton type={type || undefined}
```
glenhunter, line 50: Rather than `undefined` would we be safer to make the default `button`?

rpokorny: > Rather than undefined would we be safer to make the default button?

I decided against that because we try to make our button components behave similar to HTML `button`.  `NxButton` deliberately stays consistent with `<button>` in regards to the `type` attr, and I think `NxSegmentedButton` should too.

glenhunter (approved):

PR #437, opened 2021-07-09
Fix additional visual artifacts on focus RSC-627

avnvbhatta:
https://issues.sonatype.org/browse/RSC-627

- Fix for hiding the additional visual artifacts on focus in Firefox and Safari.
- Added `outline: 0` on the input element.
- Adding this property should not impact anything as the input element itself already has a `width: 0` and `appearance: none` property.


lib/src/components/NxToggle/NxToggle.scss
```diff
@@ -71,6 +71,7 @@ $nx-toggle-width: 44px;
   margin: 0 0 0 var(--nx-spacing-3x);
   width: 0;
   -moz-appearance: none;
+  outline: 0;
```
rpokorny, line 74: Does `none` work? That feels cleaner to me

rpokorny (approved):

lib/src/components/NxToggle/NxToggle.scss
avnvbhatta: > Does `none` work? That feels cleaner to me

Changed. 
a617179ea71e9e80b0be56c9a4b0a0c3a88d6fa7

PR #440, opened 2021-07-14
NxModal a11y label RSC-615

glenhunter:
https://issues.sonatype.org/browse/RSC-615

I could not see any way that we could automatically link the modal header to the `aria-labelledy` attribute so I left it as a manual step. I documented both `aria-labelledby` and `aria-label` making it clear that `aria-labelledby` is preferred.

I modified the examples of `NxModal` to use `aria-labelledby` except one which demonstrates `aria-label`.

I did not make the use of either mandatory because that would be a breaking change and I didn't think that it was worth making this a breaking change right now. Then next time we make a major release we can revisit this decision.

### DIFF
--- a/gallery/src/components/NxModal/NxModalEscExample.tsx
+++ b/gallery/src/components/NxModal/NxModalEscExample.tsx
@@ -72,9 +72,9 @@ export default function NxModalStackedExample() {
         </NxModal>
       }
       {showModal2 &&
-        <NxModal id="nx-modal-esc-example-modal2" onCancel={modal2CancelHandler}>
+        <NxModal id="nx-modal-esc-example-modal2" onCancel={modal2CancelHandler} aria-labelledby="modal-esc-header2">
           <header className="nx-modal-header">
-            <h2 className="nx-h2">
+            <h2 className="nx-h2" id="modal-esc-header2">
               <NxFontAwesomeIcon icon={faAngry} />
               <span>NxModal stacked example</span>
             </h2>
--- a/gallery/src/components/NxModal/NxModalEscExample.tsx
+++ b/gallery/src/components/NxModal/NxModalEscExample.tsx
@@ -44,9 +44,9 @@ export default function NxModalStackedExample() {
         </div>
       }
       {showModal &&
-        <NxModal id="nx-modal-esc-example-modal" onCancel={modal1CancelHandler}>
+        <NxModal id="nx-modal-esc-example-modal" onCancel={modal1CancelHandler} aria-labelledby="modal-esc-header">
           <header className="nx-modal-header">
-            <h2 className="nx-h2">
+            <h2 className="nx-h2" id="modal-esc-header">
               <NxFontAwesomeIcon icon={faAngry} />
               <span>NxModal with a dropdown that opens another modal</span>
             </h2>
--- a/gallery/src/components/NxModal/NxModalPage.tsx
+++ b/gallery/src/components/NxModal/NxModalPage.tsx
@@ -178,13 +178,23 @@ export default function NxModalPage() {
             </tr>
           </tbody>
         </table>
+        <NxInfoAlert>
+          Note: Placing content into the modal which exceeds its horizontal bounds is not supported. The resulting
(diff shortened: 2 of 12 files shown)


rpokorny:
These aria attributes are being applied to the wrong element and (hopefully for that reason) are not having any effect in ChromVox

gallery/src/components/NxModal/NxModalPage.tsx
```diff
@@ -182,6 +182,44 @@ export default function NxModalPage() {
           Note: Placing content into the modal which exceeds its horizontal bounds is not supported. The resulting
           layout is unspecified.
         </NxInfoAlert>
+        <h3>Accessibility</h3>
+        <p className="nx-p">
+          Elements with the <code className="nx-code">dialog</code> role are supposed to have a label specified by the
```
rpokorny, line 187: It probably isn't obvious to the reader that `NxModal` has the `dialog` role by default.

gallery/src/components/NxModal/NxModalStackedExample.tsx
```diff
@@ -19,9 +19,9 @@ export default function NxModalStackedExample() {
     <>
       <NxButton onClick={() => setShowModal(true)}>Open modal with stacked example</NxButton>
       {showModal &&
-        <NxModal id="nx-modal-stacked-example" onCancel={modal1CloseHandler}>
+        <NxModal id="nx-modal-stacked-example" onCancel={modal1CloseHandler} aria-labelledby="modal-stacked-header">
```
rpokorny, line 22: Missing label in the other modal

gallery/src/components/NxModal/NxModalEscExample.tsx
```diff
@@ -44,9 +44,9 @@ export default function NxModalStackedExample() {
         </div>
       }
       {showModal &&
-        <NxModal id="nx-modal-esc-example-modal" onCancel={modal1CancelHandler}>
+        <NxModal id="nx-modal-esc-example-modal" onCancel={modal1CancelHandler} aria-labelledby="modal-esc-header">
           <header className="nx-modal-header">
-            <h2 className="nx-h2">
+            <h2 className="nx-h2" id="modal-esc-header">
```
rpokorny, line 49: MIssing label in the other modal

glenhunter:
> These aria attributes are being applied to the wrong element and (hopefully for that reason) are not having any effect in ChromVox

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/440/commits/6096e086870fbd40c76307f1ccca4f1253915635

gallery/src/components/NxModal/NxModalPage.tsx
glenhunter: > It probably isn't obvious to the reader that `NxModal` has the `dialog` role by default.

Ah, now it should be: https://github.com/sonatype/sonatype-react-shared-components/pull/440/commits/f25b9117f3094c3bbaf8f2e5c6269dfe176fbb6e

gallery/src/components/NxModal/NxModalStackedExample.tsx
glenhunter: > Missing label in the other modal

Added: https://github.com/sonatype/sonatype-react-shared-components/pull/440/commits/90f65c1b76c9c3e995743f184c788f721810fefa

gallery/src/components/NxModal/NxModalEscExample.tsx
glenhunter: > MIssing label in the other modal

Added: https://github.com/sonatype/sonatype-react-shared-components/pull/440/commits/90f65c1b76c9c3e995743f184c788f721810fefa

rpokorny:
> Fixed here: 6096e08

ChromeVox now reads them but it seems to work better with `aria-label` than `aria-labelledby`.  With the latter, it reads the label out _after_ it reads the automatically focused button, which is a little bit confusing.

rpokorny (approved):

PR #442, opened 2021-07-16
Create React-vis examples RSC-568

avnvbhatta:
https://issues.sonatype.org/browse/RSC-568

- Installed `react-vis` package for generating charts for data visualization
- Created examples for 3 types of charts as discussed (Line chart,  Bar chart and Donut chart)
- Created a 4th chart demonstrating how certain graphs can be stacked
- Added a new item at the end on the sidebar treeview called 'Data Visualization'
- First item under 'Data Visualization' contains basic API documentation
- Rest of the items are individual examples for charts mentioned above


lib/package.json
```diff
@@ -81,7 +82,8 @@
     "classnames": "^2.2.6",
     "fuse.js": "^6.4.6",
     "prop-types": "^15.7.2",
-    "ramda": "^0.27.1"
+    "ramda": "^0.27.1",
+    "react-vis": "^1.11.7"
```
rpokorny, line 86: Do we want this to be a lib dependency? It doesn't seem like it really needs to be

lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "7.4.4",
+  "version": "7.5.0",
```
rpokorny, line 3: Why is this a minor bump and not a patch?

gallery/src/data-visualization/ReactVis.scss
```diff
@@ -0,0 +1,8 @@
+@import "~react-vis/dist/style";
+
+.nx-donut-hint{
```
rpokorny, line 3:
```suggestion
.nx-donut-hint {
```

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { GalleryDescriptionTile, GalleryExampleTile } from '../gallery-components/GalleryTiles';
+import ReactVisBarGraphExample from './examples/ReactVisBarGraphExample';
+import { NxTable, NxTableBody, NxTableCell, NxTableHead, NxTableRow } from '@sonatype/react-shared-components';
+
+const reactVisBarGraphExampleCode = require('./examples/ReactVisBarGraphExample?raw');
+
+const ReactVisBarGraphPage = () =>
+  <>
+    <GalleryDescriptionTile >
+      <p className="nx-p">
```
rpokorny, line 17: Don't forget we have the convenience components which make a lot of these simpler

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+        React-vis is a React-based data visualization library. It is highly customizable, easy-to-use, and very
+        React-friendly. The library is composed of React components that are used to render common data visualization
+        charts like line/area/bar charts, scatterplots, pie and donut charts etc.
+      </p>
+      <h3 className="nx-h2">React-vis Basics</h3>
+      <h3 className="nx-h3">XYPlot</h3>
+      <p className="nx-p">
+        Every series chart that React-vis generates (excludes RadialChart, Sankey, Sunburst and Treemap)
+        is inside a parent component called <code className="nx-code">XYPlot</code>
+        that requires two props: <code className="nx-code">height</code> and <code className="nx-code">width</code>.
```
rpokorny, line 25: These are actually required? That's disappointing, I would have expected it to be possible to set them in CSS.

lib/package.json
avnvbhatta: > Do we want this to be a lib dependency? It doesn't seem like it really needs to be

I thought it would be convenient for users to just be able to use the library without having to install it themselves. Should I just remove it?

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      </p>
+      <p className="nx-p">
+        <code className="nx-code">{'<XYPlot height={300} width = {300} />'}</code>
```
rpokorny, line 28:
```suggestion
        <code className="nx-code">{'<XYPlot height={300} width={300} />'}</code>
```

Though really I'd probably just remove this, its already part of each real example.

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      <p className="nx-p">
+        React-vis is a React-based data visualization library. It is highly customizable, easy-to-use, and very
+        React-friendly. The library is composed of React components that are used to render common data visualization
+        charts like line/area/bar charts, scatterplots, pie and donut charts etc.
+      </p>
+      <h3 className="nx-h2">React-vis Basics</h3>
+      <h3 className="nx-h3">XYPlot</h3>
+      <p className="nx-p">
+        Every series chart that React-vis generates (excludes RadialChart, Sankey, Sunburst and Treemap)
+        is inside a parent component called <code className="nx-code">XYPlot</code>
```
rpokorny, line 24: Missing space after `XYPlot`.

Rather than add `{' '}` as you have in many other places, it is often easier to choose your line break such that the JSX element is not at the start or end of a line.

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+        that requires two props: <code className="nx-code">height</code> and <code className="nx-code">width</code>.
+      </p>
+      <p className="nx-p">
+        <code className="nx-code">{'<XYPlot height={300} width = {300} />'}</code>
+      </p>
+      <h3 className="nx-h3">Ordering</h3>
+      <p className="nx-p">
+        Similar to traditional React, order matters as components are drawn in order. For example, placing
```
rpokorny, line 32: Missing space after "placing".

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+        <code className="nx-code">{'<VerticalGridLines />'}</code> before
```
rpokorny, line 33: MIssing space after "before"

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+        <code className="nx-code">{'<LineSeries data={data} />'}</code> will draw the chart on top of the gridlines.
+      </p>
+      <h3 className="nx-h3">Colors</h3>
+      <p className="nx-p">
+        Colors are assigned to charts by passing in the <code className="nx-code">color</code> or the
```
rpokorny, line 38: Does this have to be done with props is is CSS an option?  If CSS can be used, I'd prefer to steer our users towards that primarily.

lib/package.json
avnvbhatta: > Why is this a minor bump and not a patch?

I needed some clarification on this as well. As most of my prior commits were bug fixes, I went with patches. Since this was more of an addition, I thought it would be more of a minor bump.

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      <p className="nx-p">
+        <code className="nx-code">{'<RadialChart />'}</code> uses the <code className="nx-code">color</code>
+        {' '} property on the <code className="nx-code">data</code> object to determine colors of the bars.
+      </p>
+      <p className="nx-p">
+        RSC's <code className="nx-code">{'selectableColors'}</code> array provides a seamlessly way of
+        assigning colors based on RSC color guidelines. Usage is quite simple. Simply import the
+        <code className="nx-code">{'selectableColors'}</code> array and pass it on the
+        <code className="nx-code">stroke</code> or the <code className="nx-code">color</code> prop. An example
+        of using the color purple (<code className="nx-code">selectableColors[0]</code>) is shown below:
```
rpokorny, line 58: This doesn't work.  `selectableColors[0]` does in fact currently evaluate to "purple", but passing "purple" as the `stroke` causes the line color to be the CSS value "purple" (i.e. the browser's built-in purple #800080) and not the RSC purple selectable color (either `hsl(280, 100%, 80%)` or `hsl(280, 100%, 95%)` depending on whether you meant the light or dark variant).

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      </p>
+      <p className="nx-p">
+        <code className="nx-code">{'<LineSeries data={chartData} stroke={selectableColors[0]}/>'}</code>
+      </p>
+      <NxWarningAlert>
+        When assigning colors to charts, it is critical to choose colors from RSC's
```
rpokorny, line 64: I think saying it's critical is going a bit too far.  I think it makes sense to recommend the use of the selectable colors by default, and to steer consumers away from just using the default react-vis colors, but it's entirely possible that other colors will make sense in certain situations, especially when colors with actual meanings are involved (for instance, the threat level colors).

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      </p>
+      <p className="nx-p">
+        RSC's <code className="nx-code">{'selectableColors'}</code> array provides a seamlessly way of
+        assigning colors based on RSC color guidelines. Usage is quite simple. Simply import the
+        <code className="nx-code">{'selectableColors'}</code> array and pass it on the
+        <code className="nx-code">stroke</code> or the <code className="nx-code">color</code> prop. An example
+        of using the color purple (<code className="nx-code">selectableColors[0]</code>) is shown below:
+      </p>
+      <p className="nx-p">
+        <code className="nx-code">{'<LineSeries data={chartData} stroke={selectableColors[0]}/>'}</code>
```
rpokorny, line 61: Does `LineSeries` represent a single line or potentially multiple? If multiple, how do you get different lines to be different colors?

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+        When assigning colors to charts, it is critical to choose colors from RSC's
+        {' '} <code className="nx-code">{'selectableColors'}</code> array.
+        For more details on colors, please check the Color Palettes and Selectable Colors section.
+      </NxWarningAlert>
+      <section className="nx-tile-subsection">
+        <header className="nx-tile-subsection__header">
+          <h3 className="nx-h3">Basic API Reference</h3>
+        </header>
+        <p className="nx-p">
+          Note: The following table only describles the props used for examples on this page.
```
rpokorny, line 73: There are no examples on this page

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+          For the full official API documentation, please click {' '}
+          <a href="https://uber.github.io/react-vis/documentation/welcome-to-react-vis" target="_blank">here.</a>
```
rpokorny, line 75:
```suggestion
          <NxTextLink href="https://uber.github.io/react-vis/documentation/welcome-to-react-vis" external>here.</NxTextLink>
```

rpokorny: Links where the link text somewhat describes the resource (as opposed to just saying "here") are preferred.  Maybe rephrase as "Please see the official API documentation for more details" where "official API documentation" is the link.

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <NxTableRow>
+              <NxTableCell>Name</NxTableCell>
+              <NxTableCell>Type</NxTableCell>
+              <NxTableCell>Required</NxTableCell>
+              <NxTableCell>Description</NxTableCell>
+            </NxTableRow>
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell>animation</NxTableCell>
```
rpokorny, line 88: I'd be in favor of removing `animation` from the examples and this table.  It seems like more of an edge case than a core capability that we need to demonstrate IMO.

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      <section className="nx-tile-subsection">
+        <header className="nx-tile-subsection__header">
+          <h3 className="nx-h3">Basic API Reference</h3>
+        </header>
+        <p className="nx-p">
+          Note: The following table only describles the props used for examples on this page.
+          For the full official API documentation, please click {' '}
+          <a href="https://uber.github.io/react-vis/documentation/welcome-to-react-vis" target="_blank">here.</a>
+        </p>
+        <NxTable>
```
rpokorny, line 77: It's a bit confusing having this single table that appears to combine docs for props on a number of different components.

gallery/src/data-visualization/ReactVisPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+                object: expects an object formatting like
+                <code className="nx-code">{'{damping: NUMBER, stiffness: NUMBER}'}</code>
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>data</NxTableCell>
+              <NxTableCell>object array</NxTableCell>
+              <NxTableCell>Yes</NxTableCell>
+              <NxTableCell>
+                Each react-vis char components require a   property, through which
```
rpokorny, line 104:
```suggestion
                Each react-vis chart components require a   property, through which
```

gallery/src/data-visualization/examples/ReactVisLineGraphExample.tsx
```diff
@@ -0,0 +1,71 @@
...
+    {x: 3, y: 9},
+    {x: 4, y: 1},
+    {x: 5, y: 7},
+    {x: 6, y: 6},
+    {x: 7, y: 3},
+    {x: 8, y: 2},
+    {x: 9, y: 0}
+  ];
+
+  const [isLoading, setIsLoading] = useState(true);
```
rpokorny, line 28: I think this example is more complicated than it needs to be.  I'd remove the loading and the button that generates new data.

gallery/src/data-visualization/examples/ReactVisLineGraphExample.tsx
```diff
@@ -0,0 +1,71 @@
...
+        isLoading ? <NxLoadingSpinner /> :
+        <>
+          {
+            chartData &&
+            <XYPlot width={400} height={300} animation onMouseLeave={() => setHintValue(null)}>
+              <XAxis/>
+              <YAxis/>
+              <HorizontalGridLines />
+              <VerticalGridLines />
+              <LineSeries data={chartData} onNearestXY={v => setHintValue(v)} stroke={selectableColors[1]}/>
```
rpokorny, line 62: The hint doesn't work the way I would have expected.  I would expect it to show the point for whatever x-value I am closest to.  But if I'm far away from some of the points, it can skip them which is confusing to the user:

https://user-images.githubusercontent.com/3630091/126197049-c17a7d47-4b45-49e1-99af-822e804e61d4.mp4

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+const ReactVisLineGraphPage = () =>
+  <>
+    <GalleryDescriptionTile >
+      <p className="nx-p">
+        Line graphs are part of  <code className="nx-code">{'<LineSeries />'}</code> in React-vis.
+      </p>
+      <p className="nx-p">
+        The following documentation only contains basics on how to get started with creating basic line graphs. <br />
+        For the full, official API documentation, please click {' '}
+        <a href="https://uber.github.io/react-vis/documentation/series-reference/line-series" target="_blank">here.</a>
```
rpokorny, line 23: Use `NxTextLink`

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </NxTable>
+      </section>
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Line Graph Example"
+                        id="react-vis-line-graph-example"
+                        liveExample={ReactVisLineGraphExample}
+                        codeExamples={reactVisLineGraphExampleCode}>
+      The following example demonstrates how to generate a simple line graph using
+      {' '} <code className="nx-code">{'<LineSeries />'}</code> <br />
```
rpokorny, line 94: What's with the `<br/>`?

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+            </NxTableRow>
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell>barWidth</NxTableCell>
+              <NxTableCell>number/string/object</NxTableCell>
+              <NxTableCell>No</NxTableCell>
+              <NxTableCell>
+                The percentage for which each bar fills the designated bucket. 1.0 means that the bar fills the whole
+                bucket (no padding between bars), while a smaller percentage means more whitespace between the bars.
```
rpokorny, line 88: Needs to explain the format for passing an `object` here.

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+              <NxTableCell>
+                Assigns the specified color to the series.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>stroke</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>no</NxTableCell>
+              <NxTableCell>
+                In LineSeries, stroke is used instead of color to specify the color of the lines.
```
rpokorny, line 104: But this page isn't about `LineSeries`

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+                For example, <code className="nx-code">{'tickFormat={val => `$${val}`}'}</code> adds a $ before
+                the tick label.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>title</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>No</NxTableCell>
+              <NxTableCell>
+                Shows the title for the axis.
```
rpokorny, line 122: Which axis

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+                The percentage for which each bar fills the designated bucket. 1.0 means that the bar fills the whole
+                bucket (no padding between bars), while a smaller percentage means more whitespace between the bars.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>color</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>No</NxTableCell>
+              <NxTableCell>
+                Assigns the specified color to the series.
```
rpokorny, line 96: The semantics of this value are not what one would expect and should definitely be documented: "By default the color is interpreted as number to be scaled to a color range" (from the react-vis docs).

It's not super clear to me what that actually means and how it works.  We need to investigate more.  For instance I don't fully understand how the color in the example is being derived.

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+        </NxTable>
+      </section>
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Bar Graph Example"
+                        id="react-vis-bar-graph-example"
+                        liveExample={ReactVisBarGraphExample}
+                        codeExamples={reactVisBarGraphExampleCode}>
+      The following example demonstrates how to generate a simple vertical bar graph using
+      {' '} <code className="nx-code">{'<VerticalBarSeries />'}</code> <br />
```
rpokorny, line 135: Same question as before about the `<br/>`

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  <>
+    <GalleryDescriptionTile >
+      <p className="nx-p">
+        Donut charts are a type of <code className="nx-code">{'<RadialChart />'}</code> in React-vis. Both pie charts
+        and donut charts are supported by React-vis.
+      </p>
+      <p className="nx-p">
+        The following documentation only contains basics on how to get started with creating basic donut charts. <br />
+        For the full, official API documentation, please click {' '}
+        <a href="https://uber.github.io/react-vis/documentation/other-charts/radial-chart" target="_blank">here.</a>
```
rpokorny, line 24: `NxTextLink`

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <NxTableCell>Description</NxTableCell>
+            </NxTableRow>
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell>angle</NxTableCell>
+              <NxTableCell>number</NxTableCell>
+              <NxTableCell>yes</NxTableCell>
+              <NxTableCell>
+                This determines the angular size of each wedge.
```
rpokorny, line 52: What unit is this in?

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>color</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>yes</NxTableCell>
```
rpokorny, line 58: This doesn't appear to be required

rpokorny, line 58: Bears more explaining per the react-vis docs

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <NxTableCell>Description</NxTableCell>
+            </NxTableRow>
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell>colorType</NxTableCell>
+              <NxTableCell>string</NxTableCell>
+              <NxTableCell>No</NxTableCell>
+              <NxTableCell>
+                Override to let React-vis know that the colors provided are literal color values
```
rpokorny, line 101: Needs elaboration

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+                How far the labels should be from the center of the chart as a function of the radius of the chart.
+                If not specified, the default value of 1.1 is used (slightly outside of the chart).
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>onValueMouseOut</NxTableCell>
+              <NxTableCell>function</NxTableCell>
+              <NxTableCell>no</NxTableCell>
+              <NxTableCell>
+                This handler is triggered either when the user's mouse leaves a mark. The handler passes two arguments,
```
rpokorny, line 127: either?

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+                This handler is triggered either when the user's mouse enters a mark. The handler passes two arguments,
+                the corresponding datapoint and the actual event.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>padAngle</NxTableCell>
+              <NxTableCell>number/function</NxTableCell>
+              <NxTableCell>no</NxTableCell>
+              <NxTableCell>
+                The padding to be applied between arcs.
```
rpokorny, line 145: If it's a function, does it take arguments?

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>radius</NxTableCell>
+              <NxTableCell>number</NxTableCell>
+              <NxTableCell>no</NxTableCell>
+              <NxTableCell>
+                The distance between the origin and the outside of the arc. This values is scaled linearly by default
```
rpokorny, line 153: How else would it scale?

gallery/src/data-visualization/examples/ReactVisDonutChartExample.tsx
```diff
@@ -0,0 +1,81 @@
...
+      setChartData(data);
+      setIsLoading(false);
+    }, 1000);
+  }, []);
+
+  // Simulate new data load on button click
+  const updateData = () => {
+    const tempData: RadialChartPoint[] = [];
+    let randomNum = randomNumberGenerator(1, 10, 0);
+    tempData.push({angle: randomNum, color: selectableColors[3], label: 'Severe', subLabel: String(randomNum)});
```
rpokorny, line 36: These sublabels are concerningly hideous
image

lib/package.json
rpokorny: > I needed some clarification on this as well. As most of my prior commits were bug fixes, I went with patches. Since this was more of an addition, I thought it would be more of a minor bump.

Keep in mind the version are primarily versions of the library, not the gallery.  As such any PRs which only change the gallery are automatically patches.

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > It's a bit confusing having this single table that appears to combine docs for props on a number of different components.

I initially had a single page that contained both the docs and the examples, but @glenhunter and I talked about it and he suggested that we break it into a "main page" that has descriptions of the common props and then break out each example graph into its own page and have the prop descriptions for props that are _only_ used in that specific example.

Would it be better if each graph example page had descriptions of _all_ of the props used in that example? And perhaps remove the documentation table from the main page entirely?

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
avnvbhatta: > What's with the `<br/>`?

All `GalleryExampleTile` children are `<p>` tags. I wanted some text to appear on a new line, so I used `<br />`

gallery/src/data-visualization/ReactVis.scss
avnvbhatta: > 

Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/commits/2e4db1dfde848ef990f90db8c18ece20a6268f2b#diff-409f3ff074a6dcfa3fc680b9ee01d070e4d8c608476721faa63801ef1b9dcf83R1-R6

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
avnvbhatta: > Don't forget we have the convenience components which make a lot of these simpler

I have changed the code to use convenience components where required.
2e4db1dfde848ef990f90db8c18ece20a6268f2b

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > These are actually required? That's disappointing, I would have expected it to be possible to set them in CSS.

Looks like they are required. The react-vis docs also suggest that the `height` and `width` must be specified.

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > Though really I'd probably just remove this, its already part of each real example.

I agree. This has been removed.

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > Missing space after `XYPlot`.
> 
> Rather than add `{' '}` as you have in many other places, it is often easier to choose your line break such that the JSX element is not at the start or end of a line.

I have tried to space out line breaks as much as possible to avoid using `{' '}`

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > Missing space after "placing".

Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/commits/2e4db1dfde848ef990f90db8c18ece20a6268f2b#diff-022b661d769d97fab4aefe7ceba79b9f9821ecda32291b98754a4eb02a9efe02R27-R30

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > MIssing space after "before"

Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/commits/2e4db1dfde848ef990f90db8c18ece20a6268f2b#diff-022b661d769d97fab4aefe7ceba79b9f9821ecda32291b98754a4eb02a9efe02R27-R30

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > Does this have to be done with props is is CSS an option? If CSS can be used, I'd prefer to steer our users towards that primarily.

I think it can be done with CSS, but would require to pass in a prop for `className`.
Check 65093a218837d323181c3b2788135f6ca536421a

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > This doesn't work. `selectableColors[0]` does in fact currently evaluate to "purple", but passing "purple" as the `stroke` causes the line color to be the CSS value "purple" (i.e. the browser's built-in purple `#800080`) and not the RSC purple selectable color (either `hsl(280, 100%, 80%)` or `hsl(280, 100%, 95%)` depending on whether you meant the light or dark variant).

Changed to use `selectableColorClasses` instead.
65093a218837d323181c3b2788135f6ca536421a

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > I think saying it's critical is going a bit too far. I think it makes sense to recommend the use of the selectable colors by default, and to steer consumers away from just using the default react-vis colors, but it's entirely possible that other colors will make sense in certain situations, especially when colors with actual meanings are involved (for instance, the threat level colors).

Changed to _recommended_ instead of critical. I agree. Critical sounds a bit harsh.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/commits/65093a218837d323181c3b2788135f6ca536421a#diff-022b661d769d97fab4aefe7ceba79b9f9821ecda32291b98754a4eb02a9efe02R74-R75

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > Does `LineSeries` represent a single line or potentially multiple? If multiple, how do you get different lines to be different colors?

A single `<LineSeries />` under an` <XYPlot />` renders a single line. To have multiple lines, you could have several `<LineSeries />` children under an `<XYPlot />`.

The Stacked Graph example shows how multiple lines can have different colors. You would simply just pass in a different `selectableColorClasses` element in the `className` prop.
Check
https://github.com/sonatype/sonatype-react-shared-components/pull/442/commits/65093a218837d323181c3b2788135f6ca536421a#diff-94e6d9001a55378902b0b0d8de05036607ed84ff6ff80227a771ba9bfc323fe9R46-R59

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > There are no examples on this page

This was meant to be removed.

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > Links where the link text somewhat describes the resource (as opposed to just saying "here") are preferred. Maybe rephrase as "Please see the official API documentation for more details" where "official API documentation" is the link.

This has been fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-022b661d769d97fab4aefe7ceba79b9f9821ecda32291b98754a4eb02a9efe02R85-R90

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > I'd be in favor of removing `animation` from the examples and this table. It seems like more of an edge case than a core capability that we need to demonstrate IMO.

`animation` has been removed from all charts.

gallery/src/data-visualization/ReactVisPage.tsx
avnvbhatta: > 

Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-022b661d769d97fab4aefe7ceba79b9f9821ecda32291b98754a4eb02a9efe02R104-R106

gallery/src/data-visualization/examples/ReactVisLineGraphExample.tsx
avnvbhatta: > I think this example is more complicated than it needs to be. I'd remove the loading and the button that generates new data.

All examples are now simple with a static data set. The loading and update data buttons have been removed.
b2164bcc3751943a5e2340545f7c1c3c1a3a2133

gallery/src/data-visualization/examples/ReactVisLineGraphExample.tsx
avnvbhatta: > The hint doesn't work the way I would have expected. I would expect it to show the point for whatever x-value I am closest to. But if I'm far away from some of the points, it can skip them which is confusing to the user:
> 
>  2021-07-19.12-48-48.mp4

I was mistakenly using `onNearestXY` where I should've been using `onNearestX`. The hover now works as intended.

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
avnvbhatta: > Use `NxTextLink`

Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-d56acd07c265273046a9edbf323c951b328af933c0b40e157632773f54d08452R23-R28

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+              <NxTableCell>Description</NxTableCell>
+            </NxTableRow>
+          </NxTableHead>
+          <NxTableBody>
+            <NxTableRow>
+              <NxTableCell>x</NxTableCell>
+              <NxTableCell>VerticalBarSeries: string/number <br /> HorizontalBarSeries: number</NxTableCell>
+              <NxTableCell>yes</NxTableCell>
+              <NxTableCell>
+                The x position in coordinates of the box to be used.
```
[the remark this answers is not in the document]
avnvbhatta: > Needs to explain what the `string` datatype is for

https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-01f9e2b8feee08c87c5551bc7fcfd37908269fed4267f1a52de73e7c54d8ceb0R51-R68

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
avnvbhatta: > Needs to explain the format for passing an `object` here.

That was actually a typo. Should only be passing number for `barWidth` 
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-01f9e2b8feee08c87c5551bc7fcfd37908269fed4267f1a52de73e7c54d8ceb0R84-R88

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
avnvbhatta: > But this page isn't about `LineSeries`

Should have been removed from this page, so has been removed.

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+              <NxTableCell>
+                In LineSeries, stroke is used instead of color to specify the color of the lines.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>tickFormat</NxTableCell>
+              <NxTableCell>function</NxTableCell>
+              <NxTableCell>No</NxTableCell>
+              <NxTableCell>
+                Format function for the tick label. Typically used to transform ticks to show addtional information.
```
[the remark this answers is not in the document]
avnvbhatta: > What is the "tick"?

https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-01f9e2b8feee08c87c5551bc7fcfd37908269fed4267f1a52de73e7c54d8ceb0R104-R109

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
avnvbhatta: > Which axis

https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-01f9e2b8feee08c87c5551bc7fcfd37908269fed4267f1a52de73e7c54d8ceb0R112-R118

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
avnvbhatta: > What unit is this in?

Not sure. It doesn't say in the official documentation either.

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
avnvbhatta: > Same question as before about the `<br/>`

Has been removed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-01f9e2b8feee08c87c5551bc7fcfd37908269fed4267f1a52de73e7c54d8ceb0R125-R132

gallery/src/data-visualization/examples/ReactVisBarGraphExample.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export default function ReactVisBarGraphExample() {
+  const data: VerticalBarSeriesPoint[] = [
+    {x: 'Q1', y: 2.8},
+    {x: 'Q2', y: 3.2},
+    {x: 'Q3', y: 2.5},
+    {x: 'Q4', y: 3.5}
+  ];
+
+  const [isLoading, setIsLoading] = useState(true);
```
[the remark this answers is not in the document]
avnvbhatta: > As with the other example, I think this one would be better simplified.

Simplified.
b2164bcc3751943a5e2340545f7c1c3c1a3a2133

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
avnvbhatta: > `NxTextLink`

Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-d3cb9f23dd6b85016a9b3a4521568efea78292d6a2254345b4ce92c0edba9c54R23-R28

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
avnvbhatta: > This doesn't appear to be required

It's not required per the library, but I figured I would mark this as required so that users would be forced to use RSC colors. If `color` is not specified, react-vis creates charts with its default colors.

avnvbhatta: > Bears more explaining per the react-vis docs

What do you think should be added to this? The official documentation doesn't really provide a whole lot of details on this either. I merely added this data reference part just to give users a brief overview of what is required to generate donut charts with specific colors.

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
avnvbhatta: > Needs elaboration

I have added a bit of detail to this part. Again, not a whole lot of details on the official docs either. Basically, we need to override react-vis assigning custom colors to the wedges by providing `colorType="literal"`

https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-d3cb9f23dd6b85016a9b3a4521568efea78292d6a2254345b4ce92c0edba9c54R98-R107

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+                Override to let React-vis know that the colors provided are literal color values
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>innerRadius</NxTableCell>
+              <NxTableCell>number (in pixels)</NxTableCell>
+              <NxTableCell>no</NxTableCell>
+              <NxTableCell>
+                If radius is not set on the data then this can be used to set the innerRadius for all of the rows.
+                This can be useful for building donut charts.
```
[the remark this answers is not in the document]
avnvbhatta: > Needs to specify the default

Need some design guidance on this. Perhaps @dturner32 can weigh in on this.

Currently, the radius (radius for the entire circle) is 140 and the inner radius (the radius/size of the wedge) is 100. 

image
*Ignore the labels. Yes, they're horrible.*

Do we want the wedge to be bigger/smaller or do you think this is ideal?

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
avnvbhatta: > If it's a function, does it take arguments?

Not entirely sure. Official docs do not provide details.
image

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
avnvbhatta: > How else would it scale?

Again, no details on it.
image

gallery/src/data-visualization/examples/ReactVisDonutChartExample.tsx
avnvbhatta: > These sublabels are concerningly hideous
> image

I agree, and from the looks of it, there's no straight-forward way to style these sublabels either. And there's no way to hide just the sublabels. It's either hide the labels and sublabels entirely or show both the labels and the sublabels.

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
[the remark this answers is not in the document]
rpokorny: > Need some design guidance on this. Perhaps @dturner32 can weigh in on this.

Oh, I just meant you should explain what happens when this prop isn't provided (I'm assuming it defaults to 0 e.g. a pie chart not a donut chart)

gallery/src/data-visualization/examples/ReactVisDonutChartExample.tsx
rpokorny: > I agree, and from the looks of it, there's no straight-forward way to style these sublabels either. And there's no way to hide just the sublabels. It's either hide the labels and sublabels entirely or show both the labels and the sublabels.

We need to figure out how to work with them, they're not usable the way they're shown here.

lib/package.json
rpokorny: > Should I just remove it?

I can see arguments either way.  I'll ask around in the JavaScript slack channel to see what would make more sense to people.

rpokorny: > I can see arguments either way. I'll ask around in the JavaScript slack channel to see what would make more sense to people.

Actually on second thought I've decided against it.  IMO any codebase which references a given module in one of its `import` statements should always declare a direct dependency on that module.  So following that rule, downstream projects which use react-vis should always have their own explicit dependency on it, so there isn't really anything gained by having the RSC lib also declare a dependency on it.

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
rpokorny: > Fixed.
https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-d56acd07c265273046a9edbf323c951b328af933c0b40e157632773f54d08452R23-R28

Needs the `external` prop

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
rpokorny: > https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-01f9e2b8feee08c87c5551bc7fcfd37908269fed4267f1a52de73e7c54d8ceb0R112-R118

So this table is actually a combination of props for different components? I think that needs to be clearer.  Either a column needs to be added saying which component each prop goes on, or if you want to be significantly more thorough, you could create an entire tile subsection for each relevant component, with each having its own props table.

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
rpokorny: > https://github.com/sonatype/sonatype-react-shared-components/pull/442/files#diff-d3cb9f23dd6b85016a9b3a4521568efea78292d6a2254345b4ce92c0edba9c54R23-R28

Needs `external`

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
rpokorny: > Not sure. It doesn't say in the official documentation either.

Then we need to figure it out and document it.

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
rpokorny: > It's not required per the library, but I figured I would mark this as required so that users would be forced to use RSC colors. If color is not specified, react-vis creates charts with its default colors.

The problem with that is that there are a number of different ways to set the color, and this isn't necessarily the preferred one.  Which leads to something else I've been meaning to mention, I think we need to do significantly more to document and demonstrate how custom classnames would be added and used in order to color the various chart components - as that would be my preferred way of doing so.

rpokorny: > What do you think should be added to this? The official documentation doesn't really provide a whole lot of details on this either. I merely added this data reference part just to give users a brief overview of what is required to generate donut charts with specific colors.

Here's what the react-vis docs say:

> The color of a box in the series. By default the color is interpreted as number to be scaled to a color range. This can be over-ridden by providing the prop colorType="literal" to the series itself. This property can also be defined on the series level.

So there's some non-trivial complexity here that we need to first understand ourselves and then figure out how to communicate to our users.  The docs you have currently would leave people assuming that they can plug any CSS color value in here and the donut segment would just be that color, and that doesn't seem to be the case.

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
rpokorny: > I have added a bit of detail to this part. Again, not a whole lot of details on the official docs either. Basically, we need to override react-vis assigning custom colors to the wedges by providing colorType="literal"

What happens if `colorType` is not `literal`?

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
rpokorny: > Not entirely sure. Official docs do not provide details.

I think we should either figure it out or just not document the `function` case

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
rpokorny: > Again, no details on it.

I'd be in favor of dropping that note from our docs then.  The default is what one would expect, so we can leave it to the react-vis docs to explain whatever other possibilities there are.

gallery/src/data-visualization/ReactVis.scss
```diff
@@ -0,0 +1,33 @@
+@import "~react-vis/dist/style";
```
rpokorny, line 1: I assume the styles in this import are required when using these charts? We need to document that.

rpokorny:
Here's the overall explanation of how the colors work: https://github.com/uber/react-vis/blob/master/docs/colors.md

gallery/src/data-visualization/ReactVis.scss
```diff
@@ -0,0 +1,33 @@
...
+
+.nx-donut-hint {
```
rpokorny, line 3: Are these the standard styles for how we want these to appear? If so this class should be in the lib, not the gallery

gallery/src/data-visualization/ReactVis.scss
```diff
@@ -0,0 +1,33 @@
...
+  background-color: black;
+  border-radius: 5px;
+  color: white;
+  font-size: 12px;
+  padding: 8px;
+}
+
+.nx-graph-shape-fill-dark {
```
rpokorny, line 11: After looking at https://github.com/uber/react-vis/blob/master/docs/colors.md, it looks like coloring via CSS isn't really supported.  I think what we need to do is create a `colorRange` based on the RSC selectable colors (or maybe two, one for the light selectable colors and one for the dark) and document how to use that.

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,200 @@
...
+  <>
+    <GalleryDescriptionTile >
+      <NxP>
+        Bar graphs are part of <NxCode>{'<BarSeries />'}</NxCode> in React-vis. Both vertical and
+        horizontal bar graphs are supported by React-vis.
+      </NxP>
+      <NxP>
+        The following documentation only contains basics on how to get started with creating basic bar graphs.
+        For more details, please see the
+        <NxTextLink href="https://uber.github.io/react-vis/documentation/series-reference/bar-series">
```
rpokorny, line 26:
```suggestion
        <NxTextLink external href="https://uber.github.io/react-vis/documentation/series-reference/bar-series">
```

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,200 @@
...
+              <NxTableCell>yes</NxTableCell>
+              <NxTableCell>
+                Each react-vis chart components require a property, through which
+                we pass an array of objects. These properties correspond to various visual characteristics of the
+                corresponding marks. For example, x and y, which are required for most series types, affect the position
+                of each mark.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>height</NxTableCell>
```
rpokorny, line 113: Another place where it's confusing to figure out which component is being referenced.

gallery/src/data-visualization/ReactVisBarGraphPage.tsx
```diff
@@ -0,0 +1,200 @@
...
+                This handler fires when the user moves their mouse or touches point somewhere on the plot.
+                The handler fires a function that takes two arguments: the datapoint with the x value closest to the
+                cursor or touch point, and a second object containing: the <NxCode>innerX</NxCode> value (x
+                coordinates of the cursor relative to the left of the plot), <NxCode>index</NxCode> (position
+                of this datapoint in the dataset, where 0 is the first datapoint, 1 is the second, etc)
+                plus the actual event as <NxCode>event</NxCode>.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>tickFormat</NxTableCell>
```
rpokorny, line 143: We should demonstrate this prop

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+  <>
+    <GalleryDescriptionTile >
+      <NxP>
+        Donut charts are a type of <NxCode>{'<RadialChart />'}</NxCode> in React-vis. Both pie charts
+        and donut charts are supported by React-vis.
+      </NxP>
+      <NxP>
+        The following documentation only contains basics on how to get started with creating basic donut charts.
+        For more details, please see the
+        <NxTextLink href="https://uber.github.io/react-vis/documentation/other-charts/radial-chart">
```
rpokorny, line 25:
```suggestion
        <NxTextLink externalhref="https://uber.github.io/react-vis/documentation/other-charts/radial-chart">
```

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+          {' official API documentation.'}
```
rpokorny, line 26: Why is this quoted?

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>data</NxTableCell>
+              <NxTableCell>object array</NxTableCell>
+              <NxTableCell>yes</NxTableCell>
+              <NxTableCell>
+                Each react-vis chart components require a property, through which
+                we pass an array of objects. These properties correspond to various visual characteristics of the
+                corresponding marks. For example, x and y, which are required for most series types, affect the position
+                of each mark.
```
rpokorny, line 114: Talking about x and y on this page doesn't really make sense

gallery/src/data-visualization/ReactVisDonutChartPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+                This handler is triggered when the user's mouse enters a mark. The handler passes two arguments,
+                the corresponding datapoint and the actual event.
+              </NxTableCell>
+            </NxTableRow>
+            <NxTableRow>
+              <NxTableCell>padAngle</NxTableCell>
+              <NxTableCell>number/function</NxTableCell>
+              <NxTableCell>no</NxTableCell>
+              <NxTableCell>
+                The padding to be applied between arcs.
```
rpokorny, line 175: What units is this in?

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { GalleryDescriptionTile, GalleryExampleTile } from '../gallery-components/GalleryTiles';
+import ReactVisLineGraphExample from './examples/ReactVisLineGraphExample';
+
+const reactVisLineGraphExampleCode = require('./examples/ReactVisLineGraphExample?raw');
+
+const ReactVisLineGraphPage = () =>
+  <>
+    <GalleryDescriptionTile >
+      <NxP>
+        Line graphs are part of  <NxCode>{'<LineSeries />'}</NxCode> in React-vis.
```
rpokorny, line 19:
```suggestion
        Line graphs are part of <NxCode>{'<LineSeries />'}</NxCode> in React-vis.
```

gallery/src/data-visualization/ReactVisLineGraphPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+      </NxP>
+      <NxP>
+        The following documentation only contains basics on how to get started with creating basic line graphs.
+        For more details, please see the
+        <NxTextLink href="https://uber.github.io/react-vis/documentation/series-reference/line-series">
```
rpokorny, line 24:
```suggestion
        <NxTextLink external href="https://uber.github.io/react-vis/documentation/series-reference/line-series">
```

gallery/src/data-visualization/ReactVisStackedGraphPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+  <>
+    <GalleryDescriptionTile >
+      <NxP>
+        React-vis also supports stacked graphs where two or more graphs can be stacked on top of each other.
+      </NxP>
+      <section className="nx-tile-subsection">
+        <header className="nx-tile-subsection__header">
+          <h3 className="nx-h3">Basic API Reference</h3>
+        </header>
+        <NxTable>
```
rpokorny, line 25: This page doesn't actually introduce any components that aren't already documented elsewhere, so I think this props table is redundant

PR #443, opened 2021-07-20
Fix cutoff focus styles in modal body - RSC-631

rpokorny:
https://issues.sonatype.org/browse/RSC-631


glenhunter (approved):
Looks good to me. My only suggestion would be to add NxRadio button examples to one of the two form examples just for the sake of visual testing, but NxChcekbox alone is probably good enough.

PR #444, opened 2021-07-20
Consolidate border variables RSC-614

glenhunter:
https://issues.sonatype.org/browse/RSC-614

I decided to use ` --nx-color-border-secondary` as the variable color name and `--nx-border-secondary` as the border variable name.

I replaced the instances of subsection borders that I felt were not semantically correct.

### DIFF
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -181,6 +181,9 @@ const CssVariablesPage = () => {
           <PropertyDocItem propertyVar="--nx-border-default">
             Standard border styles for higher-level elements.
           </PropertyDocItem>
+          <PropertyDocItem propertyVar="--nx-border-secondary">
+            Lighter border styles generally used inside components, or after a default border.
+          </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-border-subsection">
             Standard border styles for lower-level elements such as tile subsections.
           </PropertyDocItem>
@@ -210,6 +213,9 @@ const CssVariablesPage = () => {
           <NxTable.Body>
             <ColorDocRow colorVar="--nx-color-border">
               Color for standard higher-level borders
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-border-secondary">
+              Color for lighter borders
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-subsection-border">
               Color for standard lower-level borders such as those of tile subsections
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -182,7 +182,12 @@ const CssVariablesPage = () => {
             Standard border styles for higher-level elements.
           </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-border-secondary">
-            Secondary border styles generally used inside components, or after a default border.
+            Secondary border styles generally used inside components or after a default border.
+          </PropertyDocItem>
+          <PropertyDocItem propertyVar="--nx-border-subsection">
+            <NxWarningAlert>
+              Deprecated. Standard border styles for lower-level elements such as tile subsections.
+            </NxWarningAlert>
           </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-box-shadow-focus">
             The standard box-shadow to apply to focused elements in addition to a border or outline.
@@ -213,6 +218,11 @@ const CssVariablesPage = () => {
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-border-secondary">
               Color for secondary borders
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-border-subsection">
+              <NxWarningAlert>
+                Deprecated. Standard border styles for lower-level elements such as tile subsections.
+              </NxWarningAlert>
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-component-background">
               The background color of most RSC components including tiles and form fields. Note that many RSC
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -214,9 +214,6 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-border-secondary">
               Color for lighter borders
             </ColorDocRow>
-            <ColorDocRow colorVar="--nx-color-subsection-border">
-              Color for standard lower-level borders such as those of tile subsections
-            </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-component-background">
               The background color of most RSC components including tiles and form fields. Note that many RSC
               components simply have transparent backgrounds. For those that have an actual color however, it is
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -221,7 +221,7 @@ const CssVariablesPage = () => {
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-border-subsection">
               <NxWarningAlert>
-                Deprecated. Standard border styles for lower-level elements such as tile subsections.
+                Deprecated. Standard border color for lower-level elements such as tile subsections.
               </NxWarningAlert>
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-component-background">
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -212,7 +212,7 @@ const CssVariablesPage = () => {
               Color for standard higher-level borders
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-border-secondary">
-              Color for lighter borders
+              Color for secondary borders
             </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-component-background">
               The background color of most RSC components including tiles and form fields. Note that many RSC
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -182,10 +182,7 @@ const CssVariablesPage = () => {
             Standard border styles for higher-level elements.
           </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-border-secondary">
-            Lighter border styles generally used inside components, or after a default border.
-          </PropertyDocItem>
-          <PropertyDocItem propertyVar="--nx-border-subsection">
-            Standard border styles for lower-level elements such as tile subsections.
+            Secondary border styles generally used inside components, or after a default border.
           </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-box-shadow-focus">
             The standard box-shadow to apply to focused elements in addition to a border or outline.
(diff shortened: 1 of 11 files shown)


lib/src/base-styles/_nx-variables.scss
```diff
@@ -42,6 +42,7 @@
   --nx-border-radius: 6px;
 
   --nx-border-default: 1px solid var(--nx-color-border);
+  --nx-border-secondary: 1px solid var(--nx-color-border-secondary);
   --nx-border-subsection: 1px solid var(--nx-color-subsection-border);
```
avnvbhatta, line 46: Looks like we're replacing `--nx-border-subsection` with` --nx-border-secondary.` Do we even need `--nx-border-subsection` anymore?

rpokorny: > Do we even need --nx-border-subsection anymore?

For backwards compatibility, yes.  But it should be defined as just an alias to `--nx-border-secondary` and should be documented as deprecated.  Same with `--nx-color-subsection-border`.

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -181,6 +181,9 @@ const CssVariablesPage = () => {
           <PropertyDocItem propertyVar="--nx-border-default">
             Standard border styles for higher-level elements.
           </PropertyDocItem>
+          <PropertyDocItem propertyVar="--nx-border-secondary">
+            Lighter border styles generally used inside components, or after a default border.
```
rpokorny, line 185: Do we want to make it explicit in the documentation like this that they are lighter?  I don't think there's anything inherently lighter about them that's just the styling that was chosen.

lib/src/base-styles/_nx-colors.scss
```diff
@@ -6,7 +6,7 @@
  */
 :root {
   --nx-color-border: var(--nx-swatch-indigo-90);
-
+  --nx-color-border-secondary: var(--nx-swatch-indigo-95);
```
rpokorny, line 9: Hmm should it be this or `--nx-color-secondary-border`? Is "secondary" more of an "element" or a "circumstance" (to use the terminology from the CSS variable docs page)? I especially ask in light of there being `--nx-color-form-element-border`

glenhunter: > Hmm should it be this or `--nx-color-secondary-border`? Is "secondary" more of an "element" or a "circumstance" (to use the terminology from the CSS variable docs page)? I especially ask in light of there being `--nx-color-form-element-border`

IMO the form variables are incorrect. The variables names should be general to the left and more specific to the right.  The most general is that it's a color, next is that it's a border, next what kind of border, etc

rpokorny: > The most general is that it's a color, next is that it's a border, next what kind of border, etc

I don't really agree.  Consider that some types of elements have variables for both border and background.  `color -> border|background -> element` isn't any more correctly sorted by than `color -> element -> border|background`.  They're both equally valid in terms of "general to the left and more specific to the right".  The latter is what we agreed on and wrote down though when we created the variable naming scheme, so it's what we should stick to.

glenhunter: > > The most general is that it's a color, next is that it's a border, next what kind of border, etc
> 
> I don't really agree. Consider that some types of elements have variables for both border and background. `color -> border|background -> element` isn't any more correctly sorted by than `color -> element -> border|background`. They're both equally valid in terms of "general to the left and more specific to the right". The latter is what we agreed on and wrote down though when we created the variable naming scheme, so it's what we should stick to.

I think that you're correct in terms of borders/backgrounds that are a part of elements. Forms, lists, tables, etc. But in this case what you are suggesting is `color -> variation -> element`, because `secondary` isn't an element it's a variation of the element.

gallery/src/styles/CssVariables/CssVariablesPage.tsx
glenhunter: > Do we want to make it explicit in the documentation like this that they are lighter? I don't think there's anything inherently lighter about them that's just the styling that was chosen.

Good point, fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/444/commits/82dc27415b4429cf0d7336cfd3ea0a0e507da291

lib/src/base-styles/_nx-variables.scss
glenhunter: > > Do we even need --nx-border-subsection anymore?
> 
> For backwards compatibility, yes. But it should be defined as just an alias to `--nx-border-secondary` and should be documented as deprecated. Same with `--nx-color-subsection-border`.

Subsection borders are deprecated here: https://github.com/sonatype/sonatype-react-shared-components/pull/444/commits/0832a751def52ca8ab1bf66b5ddf5bcfd85df38d and the doc entry was removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/444/commits/82dc27415b4429cf0d7336cfd3ea0a0e507da291

lib/src/base-styles/_nx-colors.scss
rpokorny: > because secondary isn't an element it's a variation of the element.

Well that's what I was getting at when I said 

> Is "secondary" more of an "element" or a "circumstance" (to use the terminology from the CSS variable docs page)?

My thinking is that there aren't single elements that would have the primary border in some circumstances and a secondary border in other circumstances.  Rather, the elements that would get a secondary border are a completely different set than those that would get a primary border.  So to me secondary seems like an element as far as our naming scheme goes.  After all, the main purpose of this variable is to replace `--nx-color-subsection-border`, and `subsection` was definitely an element, so shouldn't this be too?

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -181,8 +181,8 @@ const CssVariablesPage = () => {
           <PropertyDocItem propertyVar="--nx-border-default">
             Standard border styles for higher-level elements.
           </PropertyDocItem>
-          <PropertyDocItem propertyVar="--nx-border-subsection">
```
rpokorny, line 184: Usually when we deprecate a variable we don't completely remove its docs, we rewrite them to explain that its deprecated, often using a warning alert.  See `NxModal` `onClose`

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -211,8 +211,8 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-border">
               Color for standard higher-level borders
             </ColorDocRow>
-            <ColorDocRow colorVar="--nx-color-subsection-border">
-              Color for standard lower-level borders such as those of tile subsections
+            <ColorDocRow colorVar="--nx-color-border-secondary">
+              Color for lighter borders
```
rpokorny, line 215: Another place where it says "lighter" and probably shouldn't

lib/src/base-styles/_nx-colors.scss
```diff
@@ -42,4 +41,7 @@
 
   --nx-color-validation-valid: var(--nx-swatch-green-30);
   --nx-color-validation-invalid: var(--nx-swatch-red-45);
+
+  // Deprecated variables
+  --nx-color-subsection-border: var(--nx-color-border);
```
rpokorny, line 46: Shouldn't this be an alias for the secondary color not the primary one? They were both the same before this PR, but now that they are different I'd assume this should follow the secondary.

lib/src/base-styles/_nx-grid.scss
```diff
@@ -39,7 +39,7 @@
 
 .nx-grid-h-keyline {
   border: none;
-  border-top: var(--nx-border-subsection);
+  border-top: var(--nx-border-default);
```
rpokorny, line 42: I'll have to assume that changing these from secondary to primary was stipulated by @dturner32 , I have no direct way of knowing whether they're correct or not.

lib/src/base-styles/_nx-colors.scss
glenhunter: > > because secondary isn't an element it's a variation of the element.
> 
> Well that's what I was getting at when I said
> 
> > Is "secondary" more of an "element" or a "circumstance" (to use the terminology from the CSS variable docs page)?
> 
> My thinking is that there aren't single elements that would have the primary border in some circumstances and a secondary border in other circumstances. Rather, the elements that would get a secondary border are a completely different set than those that would get a primary border. So to me secondary seems like an element as far as our naming scheme goes. After all, the main purpose of this variable is to replace `--nx-color-subsection-border`, and `subsection` was definitely an element, so shouldn't this be too?

We were also trying to get rid of the `subsection` border in general because it was being used by things that weren't subsections.

> My thinking is that there aren't single elements that would have the primary border in some circumstances and a secondary border in other circumstances

One example of such a thing is the list item borders where the first list item has a default border but all subsequent list items have a secondary border. Tables are similar where `thead` rows tend to be primary and `tbody` rows tend to be secondary.

When we talk about borders that are named specifically for the element they are attached to (form element, etc) it makes sense to name them `nx-color-form-element-border` because they will be in a set of other `nx-color-form-element-*` items (presumably) which makes them easy to group and easy to find. There isn't a set of `nx-color-secondary-*` elements, nor do I think that there should be. To me it makes more sense that the "basic" elements like `nx-color-border` and `nx-color-text` be named such that the variations are last, which if you look at `_nx-colors.scss` the majority are:

```
  --nx-color-interactive-background-active: var(--nx-swatch-blue-95);
  etc
  
  --nx-color-text-dark: var(--nx-swatch-grey-05);
  --nx-color-text: var(--nx-swatch-grey-10);
  etc
  
  --nx-color-link: var(--nx-swatch-blue-40);
  --nx-color-link-active: var(--nx-swatch-blue-60);
  etc

  --nx-color-threat-unspecified: var(--nx-swatch-indigo-95);
  --nx-color-threat-none: var(--nx-swatch-blue-70);
  etc
```

lib/src/base-styles/_nx-colors.scss
glenhunter: > Shouldn't this be an alias for the secondary color not the primary one? They were both the same before this PR, but now that they are different I'd assume this should follow the secondary.

The intended use for this was specifically tile subsections and according to @dturner32 they will be the default color moving forward

lib/src/base-styles/_nx-grid.scss
glenhunter: > I'll have to assume that changing these from secondary to primary was stipulated by @dturner32 , I have no direct way of knowing whether they're correct or not.

Yes, we chatted via Slack

gallery/src/styles/CssVariables/CssVariablesPage.tsx
glenhunter: > Another place where it says "lighter" and probably shouldn't

Thanks: https://github.com/sonatype/sonatype-react-shared-components/pull/444/commits/44327d26832612d170fd268994dbc3a5d004ac1c

gallery/src/styles/CssVariables/CssVariablesPage.tsx
glenhunter: > Usually when we deprecate something we don't completely remove its docs, we rewrite them to explain that its deprecated, often using a warning alert. See `NxModal` `onClose`

Right: https://github.com/sonatype/sonatype-react-shared-components/pull/444/commits/b4683e8f1bf6bd06abe8563034dc2552e93f5f08

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -211,8 +216,13 @@ const CssVariablesPage = () => {
...
               Color for standard higher-level borders
             </ColorDocRow>
-            <ColorDocRow colorVar="--nx-color-subsection-border">
-              Color for standard lower-level borders such as those of tile subsections
+            <ColorDocRow colorVar="--nx-color-border-secondary">
+              Color for secondary borders
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-border-subsection">
+              <NxWarningAlert>
+                Deprecated. Standard border styles for lower-level elements such as tile subsections.
```
rpokorny, line 224:
```suggestion
                Deprecated. Standard border color for lower-level elements such as tile subsections.
```

rpokorny (approved):

PR #445, opened 2021-07-26
Transfer List - RSC-547

rpokorny:
https://issues.sonatype.org/browse/RSC-547

### DIFF
--- a/lib/src/components/NxTransferList/NxTransferList.scss
+++ b/lib/src/components/NxTransferList/NxTransferList.scss
@@ -17,6 +17,7 @@ $nx-transfer-list-visible-item-count: 10;
   display: grid;
   grid-template-columns: 1fr 1fr;
   gap: var(--nx-spacing-6x);
+  margin-top: var(--nx-spacing-2x);
   width: 720px;
   min-width: 720px;
 
@@ -25,7 +26,7 @@ $nx-transfer-list-visible-item-count: 10;
 .nx-transfer-list__half-header {
   font-size: var(--nx-font-size);
   font-weight: bold;
-  margin-bottom: var(--nx-spacing-2x);
+  margin: 0 0 var(--nx-spacing-2x) 0;
 }
 
 .nx-transfer-list__control-box {
--- a/lib/src/components/NxTransferList/NxTransferList.scss
+++ b/lib/src/components/NxTransferList/NxTransferList.scss
@@ -118,6 +118,7 @@ $nx-transfer-list-visible-item-count: 10;
   appearance: none;
   -webkit-appearance: none;
   -moz-appearance: none;
+  margin: 0;
 }
 
 .nx-transfer-list__item-text {
--- a/lib/src/components/NxTransferList/NxTransferList.scss
+++ b/lib/src/components/NxTransferList/NxTransferList.scss
@@ -8,7 +8,7 @@
 @import '../../scss-shared/nx-text-helpers';
 
 // the padding on the sides of each transfer list control box
-$nx-transfer-list-control-box-side-padding: var(--nx-spacing-6x);
+$nx-transfer-list-control-box-side-spacing: var(--nx-spacing-6x);
 $nx-transfer-list-item-vertical-padding: var(--nx-spacing-2x);
 $nx-transfer-list-item-vertical-margin: var(--nx-spacing-2x);
 $nx-transfer-list-visible-item-count: 10;
@@ -50,7 +50,7 @@ $nx-transfer-list-visible-item-count: 10;
 }
 
 .nx-transfer-list__filter {
-  margin: var(--nx-spacing-6x) $nx-transfer-list-control-box-side-padding 0 $nx-transfer-list-control-box-side-padding;
+  margin: var(--nx-spacing-6x) $nx-transfer-list-control-box-side-spacing 0 $nx-transfer-list-control-box-side-spacing;
   width: auto;
 }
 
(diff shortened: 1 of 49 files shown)


glenhunter:
@dturner32 I think that the filters would look better full width on the default (width constrained) version

rpokorny:
> @dturner32 I think that the filters would look better full width on the default (width constrained) version

That's my mistake, they are full width in the mockups

lib/src/components/NxTransferList/NxTransferList.scss
```diff
@@ -0,0 +1,138 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-container-helpers';
+@import '../../scss-shared/nx-text-helpers';
+
+// the padding on the sides of each transfer list control box
+$nx-transfer-list-control-box-side-padding: var(--nx-spacing-6x);
```
glenhunter, line 11: It always bugs me when we set up variables named like this that we then use interchangeably as margin & padding. Why not `$nx-transfer-list-control-box-side-spacing`?

rpokorny:
> > I think that the filters would look better full width on the default (width constrained) version
>
> That's my mistake, they are full width in the mockups

A little trickier to fix than I expected too.  000ba09b

lib/src/components/NxTransferList/NxTransferList.scss
rpokorny: > It always bugs me when we set up variables named like this that we then use interchangeably as margin & padding. Why not $nx-transfer-list-control-box-side-spacing?

18b08922

rpokorny:
Tagging other teams that are known to have an interest in this component

lib/src/components/NxTransferList/NxTransferList.scss
```diff
@@ -0,0 +1,142 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-container-helpers';
+@import '../../scss-shared/nx-text-helpers';
+
+// the padding on the sides of each transfer list control box
+$nx-transfer-list-control-box-side-spacing: var(--nx-spacing-6x);
+$nx-transfer-list-item-vertical-padding: var(--nx-spacing-2x);
+$nx-transfer-list-item-vertical-margin: var(--nx-spacing-2x);
+$nx-transfer-list-visible-item-count: 10;
```
glenhunter, line 14: Would this better as a CSS variable? In case a consuming project needs/wants to make the columns taller for some reason?

rpokorny: > Would this better as a CSS variable? In case a consuming project needs/wants to make the columns taller for some reason?

That's only useful if they want to make them taller but still define their height in terms of a number of visible items.  If they just want to make it some number of pixels tall they'd use a normal CSS override style.  So I think it's too much of an edge case to worry about given that no one's asking for it.  Keep in mind if we did that we'd have to go through the trouble of documenting it, worrying about backwards compatibility for it, etc.

Blacktiger (approved):
I think this should work fine for the NXRM use-cases. I think we'll probably use the NxStatefulTransferList mostly.

glenhunter:
The `:focus` styles on the move all buttons aren't great

Screen Shot 2021-07-30 at 3 11 21 PM

rpokorny:
> The :focus styles on the move all buttons aren't great

Like all of our other links (and things styled like links) we rely on the browser-default focus styling here.  I believe that was at some point a conscious decision

glenhunter:
> > The :focus styles on the move all buttons aren't great
> 
> Like all of our other links (and things styled like links) we rely on the browser-default focus styling here. I believe that was at some point a conscious decision

Yeah I confirmed this with NxTextLink, I think we should revisit this decision - but not specifically here: https://issues.sonatype.org/browse/RSC-648

glenhunter (approved):

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -0,0 +1,170 @@
...
+    onChangeProp(newSelectedItems);
+  }
+
+  return (
+    <div className={classnames('nx-transfer-list', classNameProp)} { ...attrs }>
+      <NxFieldset className="nx-transfer-list__half" label={availableItemsLabel || 'Available Items'}>
+        <div className="nx-transfer-list__control-box">
+          <NxFilterInput className="nx-transfer-list__filter"
+                         value={availableItemsFilter}
+                         onChange={onAvailableItemsFilterChange} />
```
avnvbhatta, line 130: Should we add a placeholder? I think the mockups had a "Filter" placeholder.

rpokorny: > Should we add a placeholder? I think the mockups had a "Filter" placeholder.

Good question.  Are you suggesting just hardcoding it to "Filter" or having it be configurable?  I think the former might be adequate

rpokorny: > I think the former might be adequate

b6bebb7

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -0,0 +1,170 @@
...
+          <div className="nx-transfer-list__footer">
+            {availableItemsCountFormatter(availableCount)}
+          </div>
+        </div>
+      </NxFieldset>
+      <NxFieldset className="nx-transfer-list__half" label={selectedItemsLabel || 'Transferred Items'}>
+        <div className="nx-transfer-list__control-box">
+          <NxFilterInput className="nx-transfer-list__filter"
+                         value={selectedItemsFilter}
+                         onChange={onSelectedItemsFilterChange} />
```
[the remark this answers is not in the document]
rpokorny: > Similar to the available column's missing placeholder.

b6bebb7

lib/src/components/NxTransferList/NxTransferList.tsx
rpokorny: >  Are you suggesting just hardcoding it to "Filter" or having it be configurable? I think the former might be adequate

@alexalyons126 has confirmed that hardcoding "Filter" as the placeholder is adequate

PR #447, opened 2021-07-28
Deprecate lime swatch - RSC-645

rpokorny:
https://issues.sonatype.org/browse/RSC-645

### DIFF
--- a/gallery/src/styles/ColorPalette/ColorPalettePage.tsx
+++ b/gallery/src/styles/ColorPalette/ColorPalettePage.tsx
@@ -8,6 +8,7 @@ import React from 'react';
 
 import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
 import ColorPaletteExample from './ColorPaletteExample';
+import { NxWarningAlert } from '@sonatype/react-shared-components';
 
 const ColorPalettePage = () =>
   <>
@@ -22,6 +23,9 @@ const ColorPalettePage = () =>
         CSS custom properties declared within the RSC base styles. The swatches below show all of the colors
         available in RSC along with their hex value and corresponding RSC variable.
       </p>
+      <NxWarningAlert>
+        The lime swatch is deprecated and is currently an alias for the green swatch.
+      </NxWarningAlert>
       <section className="nx-tile-subsection">
         <header className="nx-tile-subsection__header">
           <h3 className="nx-h3">Using Colors</h3>
--- a/gallery/src/styles/ColorPalette/ColorPalettePage.tsx
+++ b/gallery/src/styles/ColorPalette/ColorPalettePage.tsx
@@ -24,7 +24,7 @@ const ColorPalettePage = () =>
         available in RSC along with their hex value and corresponding RSC variable.
       </p>
       <NxWarningAlert>
-        The lime swatch is deprecated and is currently an alias for the green swatch.
+        The lime palette is deprecated and is currently an alias for the green palette.
       </NxWarningAlert>
       <section className="nx-tile-subsection">
         <header className="nx-tile-subsection__header">
--- a/lib/src/base-styles/_nx-color-swatches.scss
+++ b/lib/src/base-styles/_nx-color-swatches.scss
@@ -71,13 +71,14 @@
   --nx-swatch-pink-95: hsl(var(--nx-swatch-pink-hs), 95%);
 
   // Green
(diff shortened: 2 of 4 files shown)


gallery/src/styles/ColorPalette/ColorPalettePage.tsx
```diff
@@ -22,6 +23,9 @@ const ColorPalettePage = () =>
         CSS custom properties declared within the RSC base styles. The swatches below show all of the colors
         available in RSC along with their hex value and corresponding RSC variable.
       </p>
+      <NxWarningAlert>
+        The lime swatch is deprecated and is currently an alias for the green swatch.
```
glenhunter, line 27:
```suggestion
        The lime swatches are deprecated and are currently an alias for the green swatches.
```

lib/src/base-styles/_nx-color-swatches.scss
```diff
@@ -87,23 +88,22 @@
...
-  --nx-swatch-lime-45: hsl(var(--nx-swatch-lime-hs), 45%);
-  --nx-swatch-lime-50: hsl(var(--nx-swatch-lime-hs), 50%);
-  --nx-swatch-lime-60: hsl(var(--nx-swatch-lime-hs), 60%);
-  --nx-swatch-lime-70: hsl(var(--nx-swatch-lime-hs), 70%);
-  --nx-swatch-lime-75: hsl(var(--nx-swatch-lime-hs), 75%);
-  --nx-swatch-lime-80: hsl(var(--nx-swatch-lime-hs), 80%);
-  --nx-swatch-lime-85: hsl(var(--nx-swatch-lime-hs), 85%);
-  --nx-swatch-lime-90: hsl(var(--nx-swatch-lime-hs), 90%);
-  --nx-swatch-lime-95: hsl(var(--nx-swatch-lime-hs), 95%);
+  // Lime - deprecated
```
glenhunter, line 91: I realize that this is automated but it's hard to see the value in displaying these swatches, not to mention that the title still appears as just lime so the unwary might not realize that they are deprecated. Perhaps we should move all of these swatches elsewhere so they don't appear on this page.

gallery/src/styles/ColorPalette/ColorPalettePage.tsx
rpokorny: The semantics I've settled on, and which are reflected to varying degrees in the docs I've written in a variety of places, are that a single color value is a "sample", a group of all lightness variations of a single hue/saturation is a "swatch", and all the swatches together are a "palette".

lib/src/base-styles/_nx-color-swatches.scss
rpokorny: > Perhaps we should move all of these swatches elsewhere so they don't appear on this page.

Which source file they are in has no bearing on whether the automated logic will pick them up unfortunately.  If they are in the `gallery.css` output from the build, it will find them

gallery/src/styles/ColorPalette/ColorPalettePage.tsx
glenhunter: > a group of all lightness variations of a single hue/saturation is a "swatch", and all the swatches together are a "palette"

What's the difference between these two things?

rpokorny: The definitions I've been using in my head:

Sample: 
image

Swatch:
image

Palette:
image

rpokorny: If we want to say "swatch" means a single color (what I've been calling "sample"), then we need to decide what to call the thing I'm currently calling a "swatch".

glenhunter: A swatch is by definition a single colour (technically a single sample since the same term is used for fabrics).

Based off what you have above: sample => swatch, swatch => palette, palette => palettes, because a palette can be all the blues and also all the colours.

rpokorny: > Based off what you have above: sample => swatch, swatch => palette, palette => palettes, because a palette can be all the blues and also all the colours.

b0238cb7.  I've created RSC-646 to make sure the rest of the docs follow this meaning of the terms as well.

glenhunter (approved):

PR #453, opened 2021-07-30
Improve tooltip screen reading - RSC-423

rpokorny:
https://issues.sonatype.org/browse/RSC-423

Before this PR, tabbing to an element with a tooltip would cause the screenreader to read only the tooltip text and not the actual text content of the element.  This was due to some extra code that was added to NxTooltip to improve the way table sort headers were read.  However. we have found that it actually works better without that, reading both the text content and the tooltip - in Voiceover.  ChromeVox unfortunately doesn't read the tooltip.

Note that this is how it is with tooltips that appear on focus/hover.  Tooltips that are already open when the screenreader gets to them behave in the opposite manner - ChromeVox reads them but VoiceOver does not.

I addition to resetting that behavior, I have added more explicit aria labels to certain elements (sortable table headers and form submit buttons) to ensure that they get read in a useful manner

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "7.8.2",
+  "version": "7.8.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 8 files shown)


glenhunter:
Is this known/intentional? 
Screen Shot 2021-08-18 at 9 20 07 AM

glenhunter (approved):
Works very nicely in VO

PR #456, opened 2021-08-04
.nx-spinner spacing in tiles RSC-650

glenhunter:
https://issues.sonatype.org/browse/RSC-650

Because an `NxForm` can be in an error or loading state and both states have the same spacing issue it made sense to use the same styles for both.

### DIFF
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -23,6 +23,10 @@ $nx-tile-header-height-actions: 43px; // Height with sub-title and/or right acti
   border-radius: var(--nx-border-radius);
   margin: var(--nx-spacing-2x) 0 var(--nx-spacing-4x) 0;
   padding: var(--nx-tile-spacing);
+
+  > .nx-footer, > .nx-form > .nx-footer {
+    margin-top: var(--nx-spacing-8x);
+  }
 }
 
 // Tile headers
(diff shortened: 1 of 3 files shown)


rpokorny:
When we fixed this in `NxModal` we took a more maintainable approach by giving the modal padding rather than having to specify margin on various specific children.  The downside being that negative margin had to be used on those elements which are supposed to be flush with the edge of the modal.

rpokorny:
Also see RSC-651 which I discovered while checking the modal implementation.

glenhunter:
> Also see RSC-651 which I discovered while checking the modal implementation.

In my tests the spinner was left aligned inside of the tile

glenhunter:
> When we fixed this in `NxModal` we took a more maintainable approach by giving the modal padding rather than having to specify margin on various specific children. The downside being that negative margin had to be used on those elements which are supposed to be flush with the edge of the modal.

A bit more work, but cleaner default code in the end: https://github.com/sonatype/sonatype-react-shared-components/pull/456/commits/73abd1d970cd91fd6a7dd57cc9ec3d8b84012ab0

lib/src/base-styles/_nx-tile.scss
```diff
@@ -16,25 +16,13 @@ $nx-tile-header-height-actions: 43px; // Height with sub-title and/or right acti
 .nx-tile {
   @include container-vertical;
 
-  // the "padding" around the inside edge of the nx-tile, which must be implemented using padding/margin
-  // on children since nx-tile-content--accordion-container is supposed to go all the way to the edges
-  --nx-tile-padding: var(--nx-spacing-6x);
+  // the "padding" around the inside edge of the nx-tile
+  --nx-tile-spacing: var(--nx-spacing-6x);
```
rpokorny, line 20: Does this still need to be a variable?

rpokorny (approved):

PR #458, opened 2021-08-05
Use non-deprecated sass library - RSC-636

rpokorny:
https://issues.sonatype.org/browse/RSC-636
https://issues.sonatype.org/browse/RSC-635

Downstream projects also have to upgrade due to the new syntax, so this is a breaking change.


glenhunter (approved):
Going to take a while to get used to using `math`

PR #462, opened 2021-08-17
Update wide form field width - RSC-664

rpokorny:
https://issues.sonatype.org/browse/RSC-664

This comes from the Kaliedo v2 sketch library. (the tartan sketch libraries are deprecated)

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "7.8.1",
+  "version": "7.8.2",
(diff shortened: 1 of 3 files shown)


glenhunter (approved):
I think that we need to make a note that this change comes from Kaleido v2 and not Tartan v1 or v2, both in this PR and in the JIRA ticket.

rpokorny:
> I think that we need to make a note that this change comes from Kaleido v2 and not Tartan v1 or v2, both in this PR and in the JIRA ticket.

Done

PR #464, opened 2021-08-19
Fix Numeric Sort Column Header Alignment - RSC-663

rpokorny:
https://issues.sonatype.org/browse/RSC-663


glenhunter (approved):
I appreciate the improved example data.

PR #467, opened 2021-08-24
NxSearchDropdown - RSC-658

rpokorny:
https://issues.sonatype.org/browse/RSC-658


rpokorny:
Added in 9c43572 and 227aa04d.  I'll create a separate ticket for documenting (and testing) the new prop on `NxFilterIcon` as I don't want even more semi-unrelated work in this PR.

rpokorny:
> I'll create a separate ticket for documenting (and testing) the new prop on NxFilterIcon as I don't want even more semi-unrelated work in this PR.

RSC-676

glenhunter:
Something odd is going on with the text of the error message
Screen Shot 2021-08-30 at 3 33 07 PM

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
```diff
@@ -0,0 +1,41 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-container-helpers';
+
+.nx-search-dropdown {
+  display: inline-block;
+  position: relative;
+
+  &__menu {
```
glenhunter, line 13: Up until this point we have not used this syntax and I don't remember if there was a reason why we didn't. I'm curious as to why you decided to use it here.

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
```diff
@@ -0,0 +1,41 @@
...
+    // only show when the component is focused
```
glenhunter, line 14: This isn't true, it only shows when it's focused and the user has typed something

lib/src/components/NxDropdownMenu/types.tsx
```diff
@@ -4,9 +4,16 @@
...
   onClosing: () => void;
-  children?: ReactNode | null;
+  className?: string | null;
 }
+
+export const propTypes: PropTypes.ValidationMap<Props> = {
+  onClosing: PropTypes.func.isRequired,
+  className: PropTypes.string,
+  children: PropTypes.node
+};
```
avnvbhatta, line 19: Is there a reason why we're omitting `className` and then redefining the type?

gallery/visualtests/nxSearchDropdown.js
```diff
@@ -0,0 +1,174 @@
...
+    await input.setValue('1');
+    await dropdownButton.waitForDisplayed();
+
+    const { x, y } = await component.getLocation();
+    const region = new Region(parseInt(x, 10), parseInt(y, 10), 300, 376);
+
+    await browser.eyesRegionSnapshot(null, Target.region(region));
+  });
+
+  it('looks right with a truncated results', async function() {
```
avnvbhatta, line 60:
```suggestion
  it('looks right with truncated results', async function() {
```

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+    const retryButton = component.find('button.nx-load-error__retry').getDOMNode() as HTMLElement;
+    retryButton.focus();
+    expect(onSearch).not.toHaveBeenCalled();
+
+    filterInput.focus();
+    expect(onSearch).not.toHaveBeenCalled();
+  });
+
+  it('does not call doSeach if focus moves into the component from an outside window while there is an error',
```
avnvbhatta, line 296:
```suggestion
  it('does not call doSearch if focus moves into the component from an outside window while there is an error',
```

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+        expect(onSearch).not.toHaveBeenCalled();
+
+        (document.activeElement! as HTMLElement).blur();
+        expect(onSearch).not.toHaveBeenCalled();
+
+        filterInput.focus();
+        expect(onSearch).not.toHaveBeenCalled();
+      }
+  );
+});
```
avnvbhatta, line 321: Should we add the test where if there's an error querying the data, and if the user navigates away from the component and comes back to it, the search should be retried automatically (as mentioned [here](https://github.com/sonatype/sonatype-react-shared-components/pull/467/files#diff-22194081f00f47900c6cba95001a93b3add1e569d6189026fe59d27a50f7ffc9R65-R70)). 

There is a test for calling `onSearch` with the current trimmed searchText if focus enters the component from elsewhere, but we don't have an explicit case when there's an error in querying and focus enters from elsewhere.

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -9,13 +9,14 @@ import classnames from 'classnames';
 import { faPlusCircle, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
 import { toLower, filter, includes, groupBy, partial, identity, prop, pipe } from 'ramda';
 
-import { Props, TransferListItemProps, DataItem, FilterFn } from './types';
+import DataItem from '../../util/DataItem';
+import { Props, TransferListItemProps, FilterFn } from './types';
 import NxFilterInput from '../NxFilterInput/NxFilterInput';
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
```
avnvbhatta, line 15: Not sure if `NxTransferList` changes are supposed to be here.

gallery/src/components/NxSearchDropdown/NxSearchDropdownPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+              <NxTable.Cell><NxCode>onSearch</NxCode></NxTable.Cell>
+              <NxTable.Cell>Function (string =&gt; void)</NxTable.Cell>
+              <NxTable.Cell>true</NxTable.Cell>
+              <NxTable.Cell></NxTable.Cell>
+              <NxTable.Cell>
+                Callback that fires when a new search query should be performed. The <em>trimmed</em> value of the
+                filter text is passed as an argument. This callback is executed whenever the user changes
+                the trimmed value of the filter input, whenever the error alert's Retry button is clicked, and whenever
+                the component gains focus while in an error state (which is intended to automatically trigger a retry
+                attempt. The handler passed in for this prop should, in addition to performing the search, update
```
avnvbhatta, line 74: Missing closing bracket.
```suggestion
                filter text is passed as an argument. This callback is executed whenever the user changes
                the trimmed value of the filter input, whenever the error alert's Retry button is clicked, and whenever
                the component gains focus while in an error state (which is intended to automatically trigger a retry
                attempt). The handler passed in for this prop should, in addition to performing the search, update
```

gallery/src/components/NxSearchDropdown/NxSearchDropdownPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+              <NxTable.Cell>HTML <NxCode>&lt;div&gt;</NxCode> Attributes</NxTable.Cell>
+              <NxTable.Cell>
+                <NxTextLink external href="https://developer.mozilla.org/en/docs/Web/HTML/Element/div">
+                  Div Attributes
+                </NxTextLink>
+              </NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell></NxTable.Cell>
+              <NxTable.Cell>
+                <NxCode>NxSearchDropdown</NxCode> supports any html attribute that's normally supported by{' '}
```
avnvbhatta, line 156:
```suggestion
                <NxCode>NxSearchDropdown</NxCode> supports any HTML attribute that's normally supported by{' '}
```

gallery/src/components/NxStatefulSearchDropdown/NxStatefulSearchDropdownPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+          </NxTable.Head>
+          <NxTable.Body>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>defaultSearchText</NxCode></NxTable.Cell>
+              <NxTable.Cell>string</NxTable.Cell>
+              <NxTable.Cell>false</NxTable.Cell>
+              <NxTable.Cell>(empty string)</NxTable.Cell>
+              <NxTable.Cell>
+                The initial value for the search text. Note that if this is provided, the <NxCode>matches</NxCode> and
+                <NxCode>loading</NxCode> props should be set to match. That is, if non-empty search text is set by
```
avnvbhatta, line 51:
```suggestion
                {' '}<NxCode>loading</NxCode> props should be set to match. That is, if non-empty search text is set by
```

rpokorny:
> Huh, ok, I figured out that the search text is the last part of the message, but also if you keep typing the error message keeps growing in real time with each key stroke. Is that intended?

That's all just a result of how the example is set up, and no necessarily reflective of what you'd do with it IRL

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
rpokorny: > This isn't true, it only shows when it's focused and the user has typed something

What you said is a subset of what I said.  But only the "when focused" part of that is implemented in this spot.

lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  return (
+    <div ref={ref} className={className} onFocus={handleComponentFocus} { ...attrs }>
+      <NxFilterInput ref={filterRef}
+                     className={filterClassName}
+                     value={searchText}
+                     onChange={handleFilterChange}
+                     disabled={disabled || undefined}
+                     searchIcon />
```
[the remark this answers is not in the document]
rpokorny: @alexalyons126 is the placeholder on this component intended to be hardcoded to the word "Search" or should it be configurable?

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -18,18 +18,16 @@ export { Props } from './types';
...
-      const classes = classnames('nx-filter-input', props.className),
+      const { className: classNameProp, searchIcon, ...otherProps } = props,
+          className = classnames('nx-filter-input', classNameProp),
 
           // just in case these props get passed in, avoid passing them to NxTextInput as they would cause
           // malfunction
-          otherProps = omit(['validatable', 'validationErrors', 'type'], props),
-          filterIcon = <NxFontAwesomeIcon icon={faFilter} className="nx-icon--filter-icons" />;
+          cleanedProps = omit(['validatable', 'validationErrors', 'type'], otherProps),
+          filterIcon = searchIcon ? faSearch : faFilter,
```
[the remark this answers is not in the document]
rpokorny: > I'm curious: in what case would these props get passed in? Is it in the context where it's wrapped within an NxForm?

Is this referring to the `searchIcon` prop or the props being omitted on the line above that?

lib/src/components/NxDropdownMenu/types.tsx
rpokorny: > Is there a reason why we're omitting className and then redefining the type?

Hmm, when I made these changes I got some sort of type error without that redefinition.  But I don't remember what it was, and I can't reproduce it now.  So I've tentatively undone that. 2bb95d2f

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
rpokorny: > Should we add the test where if there's an error querying the data, and if the user navigates away from the component and comes back to it, the search should be retried automatically (as mentioned here).
>
> There is a test for calling onSearch with the current trimmed searchText if focus enters the component from elsewhere, but we don't have an explicit case when there's an error in querying and focus enters from elsewhere.

Is that not what https://github.com/sonatype/sonatype-react-shared-components/pull/467/files#diff-8faf831d9f7a4427b8d4674137223bd7eb914e2b245eddc1308562fa0ce5a3f9R250 is?  That behavior is only supposed to happen when there's an error so if you're saying we have a test that asserts that it happens when there _isn't_ an error, that'd be a problem.

lib/src/components/NxTransferList/NxTransferList.tsx
rpokorny: > Not sure if NxTransferList changes are supposed to be here.

Yes. This PR moves `DataItem` from `NxTransferList`'s types to a general utility file

lib/src/components/NxFilterInput/NxFilterInput.tsx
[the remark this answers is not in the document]
avnvbhatta: > > I'm curious: in what case would these props get passed in? Is it in the context where it's wrapped within an NxForm?
> 
> Is this referring to the `searchIcon` prop or the props being omitted on the line above that?

Oh I'm sorry, I meant the props being omitted.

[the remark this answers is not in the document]
rpokorny: > Oh I'm sorry, I meant the props being omitted.

There isn't any specific case where they _would_ be passed in, but it would be possible in a non-typescript project for someone to try to pass them in.  And if they did so I believe that would break this component if not for this check.

lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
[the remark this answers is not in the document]
rpokorny: > Missing placeholder.

Placeholder hardcoded per discussion in WW today. d7f26c9a

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
avnvbhatta: > > > I'm curious as to why you decided to use it here.
> > 
> > 
> > It aligned with my feelings of laziness that day. Up to you if you like it and want us to try to use it more going forward, or don't like it and want me to rip it out.
> 
> I'm not opposed to using this syntax going forward, so longs well all agree. @avnvbhatta @entroform ?

Works for me.

avnvbhatta (approved):
🚀

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
entroform: +1 for me as well!

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@import '../../scss-shared/nx-container-helpers';
```
entroform, line 7: Just wondering if we should be using `@use` instead of `@import` since we moved to dart-sass?

glenhunter (approved):

entroform (approved):

PR #474, opened 2021-09-01
Move textarea validation icon gutter from the right to the bottom RSC-683

avnvbhatta:
https://issues.sonatype.org/browse/RSC-683

- Used `flex-direction: column` to move icon gutter to bottom.
- Disabled horizontal resize by setting both `max-width` and `min-width` to be 100%.
- Added top margin of validation icon to be `var(--nx-spacing-2x)`.

### DIFF
--- a/lib/src/base-styles/_nx-text-input.scss
+++ b/lib/src/base-styles/_nx-text-input.scss
@@ -122,7 +122,7 @@
     flex-direction: column;
 
     &::before {
-      content: unset;
+      content: none;
     }
 
     svg {
--- a/lib/src/base-styles/_nx-text-input.scss
+++ b/lib/src/base-styles/_nx-text-input.scss
@@ -125,7 +125,7 @@
       content: none;
     }
 
-    svg {
+    .nx-icon {
       align-self: flex-end;
       margin: var(--nx-spacing-2x) 0 0 0;
     }
--- a/lib/src/base-styles/_nx-text-input.scss
+++ b/lib/src/base-styles/_nx-text-input.scss
@@ -117,7 +117,7 @@
 
 .nx-text-input--textarea {
   .nx-text-input__box {
-    align-items: unset;
+    align-items: initial;
     display: flex;
     flex-direction: column;
 
--- a/lib/src/base-styles/_nx-text-input.scss
+++ b/lib/src/base-styles/_nx-text-input.scss
@@ -133,11 +133,9 @@
 }
 
 textarea.nx-text-input__input {
-  //To disable horizontal resize of textarea
-  max-width: 100%;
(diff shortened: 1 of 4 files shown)


lib/src/base-styles/_nx-text-input.scss
```diff
@@ -117,11 +117,25 @@
 
 .nx-text-input--textarea {
   .nx-text-input__box {
-    align-items: flex-end;
+    align-items: unset;
```
rpokorny, line 120: I'd feel more comfortable with `initial` here as unset feels more undefined in terms of what it would do
```suggestion
    align-items: initial;
```

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -117,11 +117,25 @@
...
+    display: flex;
+    flex-direction: column;
+
+    &::before {
+      content: unset;
```
rpokorny, line 125:
```suggestion
      content: none;
```

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -117,11 +117,25 @@
...
+      align-self: flex-end;
+      margin: var(--nx-spacing-2x) 0 0 0;
+    }
   }
 }
 
 textarea.nx-text-input__input {
+  //To disable horizontal resize of textarea
+  max-width: 100%;
+  min-width: 100%;
```
rpokorny, line 138: Apparently there is a CSS `resize` property which I had somehow never heard of before now, which is a better way to accomplish this. https://developer.mozilla.org/en-US/docs/Web/CSS/resize

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -117,11 +117,25 @@
...
-    align-items: flex-end;
+    align-items: unset;
+    display: flex;
+    flex-direction: column;
+
+    &::before {
+      content: unset;
+    }
+
+    svg {
```
rpokorny, line 128: We prefer to select by class
```suggestion
    .nx-icon {
```

lib/src/base-styles/_nx-text-input.scss
avnvbhatta: > Apparently there is a CSS `resize` property which I had somehow never heard of before now, which is a better way to accomplish this. https://developer.mozilla.org/en-US/docs/Web/CSS/resize

I initially used `resize`, but didn't see it being used anywhere so went with an alternative approach.

changed to use `resize` here: c5741da20d55c5520c03381a3cfd9b5a4f966bfd

rpokorny (approved):

PR #476, opened 2021-09-07
NxDivider - RSC-680

avnvbhatta:
https://issues.sonatype.org/browse/RSC-680


rpokorny:
This is very strange and may just be my monitor, but for some reason the hr's are alternating between being slightly thicker and thinner.  Can you see it in this screenshot? I've added a few extra dividers to prove out that it's happening in an alternating manner.
Screenshot_20210908_142729

lib/src/components/NxDivider/NxDivider.scss
```diff
@@ -0,0 +1,13 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ .nx-divider {
+  background-color: var(--nx-swatch-indigo-90);
+  border-radius: 1px;
+  border-style: none;
+  margin: 24px 0;
+  min-height: 1px;
+ }
```
rpokorny, line 13:
```suggestion
.nx-divider {
  background-color: var(--nx-swatch-indigo-90);
  border-radius: 1px;
  border-style: none;
  margin: 24px 0;
  min-height: 1px;
}
```

rpokorny:
Given that this is really just styles, it should probably be documented as a style (with mention of its convenience component) rather than documenting it primarily as a component.  This is how we prefer to document things that don't have any logic to encapsulate in order to facilitate their use outside of react projects.

avnvbhatta:
> This is very strange and may just be my monitor, but for some reason the hr's are alternating between being slightly thicker and thinner. Can you see it in this screenshot? I've added a few extra dividers to prove out that it's happening in an alternating manner.
> Screenshot_20210908_142729


Yes, it's clear that it's alternating in your screenshot. Maybe that might just be your monitor? It looks fine to me on Chrome, FF and Safari.
image

avnvbhatta:
> Given that this is really just styles, it should probably be documented as a style (with mention of its convenience component) rather than documenting it primarily as a component. This is how we prefer to document things that don't have any logic to encapsulate in order to facilitate their use outside of react projects.

This has been implemented. I have added `nx-divider` under styles and changed `NxDivider` to be under `SimpleComponents.tsx` as a convenience component rather than a standalone component.
Check bcdcb7bb385487554967379afcc1ad75cde17ad5

lib/src/components/NxDivider/NxDivider.scss
avnvbhatta: > 

Implemented [here.](https://github.com/sonatype/sonatype-react-shared-components/commit/bcdcb7bb385487554967379afcc1ad75cde17ad5#diff-8ebd6ae106060ccab3a733488cccc3045602704ea05aefa02e7cbcac65a5fb0bR7-R13)

glenhunter:
> > Given that this is really just styles, it should probably be documented as a style (with mention of its convenience component) rather than documenting it primarily as a component. This is how we prefer to document things that don't have any logic to encapsulate in order to facilitate their use outside of react projects.
> 
> This has been implemented. I have added `nx-divider` under styles and changed `NxDivider` to be under `SimpleComponents.tsx` as a convenience component rather than a standalone component.
> Check [bcdcb7b](https://github.com/sonatype/sonatype-react-shared-components/commit/bcdcb7bb385487554967379afcc1ad75cde17ad5)

It seems that the components files are still present though

glenhunter:
I think that there is value in creating `.nx-divider__vertical` for instances where we want a vertical border where one is not normally present or we want to override a lighter thinner vertical border. We could just pick left or right.

lib/src/base-styles/_nx-divider.scss
```diff
@@ -0,0 +1,13 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-divider {
+  background-color: var(--nx-swatch-indigo-90);
+  border-radius: 1px;
+  border-style: none;
+  margin: 24px 0;
+  min-height: 1px;
```
glenhunter, line 12: why not height?

gallery/src/styles/NxDivider/NxDividerHorizontalExample.tsx
```diff
@@ -0,0 +1,35 @@
...
+        of letters, as opposed to using 'Content here, content here', making it look like readable English. Many
+        desktop publishing packages and web page editors now use Lorem Ipsum as their default model text, and a
+        search for 'lorem ipsum' will uncover many web sites still in their infancy. Various versions have evolved
+        over the years, sometimes by accident, sometimes on purpose (injected humour and the like).
+      </NxP>
+      <NxDivider />
+      <NxButtonBar>
+        <NxButton variant="primary">Enter The Matrix</NxButton>
+        <NxButton variant="primary" disabled>But Is It Real?</NxButton>
+      </NxButtonBar>
```
rpokorny, line 32: I think we should avoid this part of the example, since buttons in this sort of position will more or less always be part of a footer, which has its own border and which does not involve `NxDivider`

gallery/src/styles/NxDivider/NxDividerHorizontalExample.tsx
```diff
@@ -0,0 +1,35 @@
...
+export default function NxDividerHorizontalExample() {
+  return (
+    <>
+      <NxP>This is a simple paragraph</NxP>
+      <NxDivider/>
+      <NxP>
+        This example simply demonstrates how to use the <NxCode>NxDivider</NxCode> to render
+        simple horizontal dividers.
+      </NxP>
+      <NxDivider />
```
rpokorny, line 19: We demo styles using plain HTML elements and not convenience components.  In fact for especially simple setups such as this one, we typically write the demo in actual HTML rather than in JavaScript with react.  See most of the tile demos for example.

lib/src/base-styles/_nx-divider.scss
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-divider {
+  background-color: var(--nx-swatch-indigo-90);
+  border-radius: 1px;
```
glenhunter, line 9: I can't see any difference toggling this on and off, what's its purpose?

gallery/src/styles/NxDivider/NxDividerPage.tsx
```diff
@@ -0,0 +1,87 @@
...
+      A simple example of <NxCode>NxDivider</NxCode>.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Simple Example Within NxForm"
+                        liveExample={NxDividerHorizontalFormExample}
+                        codeExamples={NxDividerHorizontalFormExampleCode}>
+      An example of <NxCode>NxDivider</NxCode> used within an <NxCode>NxForm</NxCode>.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Simple Vertical Example"
```
rpokorny, line 76: I think this example is unnecessary given the mention that you have at the top of the page about using `nx-grid` for vertical dividers.  `nx-grid` already has its own docs and examples

glenhunter:
For all of the examples I see `.nx-divider`'s rules twice in the inspector
Screen Shot 2021-09-09 at 3 31 45 PM

gallery/src/styles/NxDivider/NxDividerPage.tsx
```diff
@@ -0,0 +1,87 @@
...
+      </NxTile.Subsection>
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Simple Example"
+                        liveExample={NxDividerHorizontalExample}
+                        codeExamples={NxDividerHorizontalExampleCode}>
+      A simple example of <NxCode>NxDivider</NxCode>.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Simple Example Within NxForm"
```
rpokorny, line 70: Personally I also think this example is unnecessary – there isn't anything special about the usage of `.nx-divider` in a form vs other contexts that is being demonstrated here.

lib/src/base-styles/_nx-divider.scss
rpokorny, line 9: I'm not sure a border radius here is part of the spec.  The divider in sketch seems squared-off no matter how far I zoom in.  Not to mention, wouldn't this need to be 0.5?

lib/src/base-styles/_nx-divider.scss
rpokorny, line 12: Is there a reason this isn't just `height`?

lib/src/base-styles/_nx-grid.scss
```diff
@@ -17,7 +17,7 @@
 .nx-grid-col {
   @include nx-container-helpers.container-vertical;
 
-  border-left: var(--nx-border-default);
+  border-left: var(--nx-border-2x);
```
rpokorny, line 20: The vertical lines actually _do_ have a border radius to them in sketch, which is problematic as we can't do that on a border.

lib/src/components/NxDivider/NxDivider.scss
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ .nx-divider {
```
rpokorny, line 7: This file needs to be deleted

lib/src/components/NxDivider/NxDivider.tsx
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import withClass from '../../util/withClass';
+import './NxDivider.scss';
+
+export const NxDivider = withClass('hr', 'nx-divider');
```
rpokorny, line 10: Needs to be deleted

lib/src/components/NxDivider/__tests__/NxDivider.test.tsx
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import { NxDivider } from '../NxDivider';
+import { getShallowComponent } from '../../../__testutils__/enzymeUtils';
+
+describe('NxDivider', function() {
```
rpokorny, line 10: Can be simplified and moved to the SimpleComponents tests

rpokorny: > Can be simplified and moved to the SimpleComponents tests

Oh I see it's already there too.  This needs to be deleted as well then.

lib/src/index.ts
```diff
@@ -237,3 +237,5 @@ export { default as NxTransferList, Props as NxTransferListProps, DataItem as Nx
 
 export { default as NxStatefulTransferList, Props as NxStatefulTransferListProps }
   from './components/NxTransferList/stateful/NxStatefulTransferList';
+
+export { NxDivider } from './components/NxDivider/NxDivider';
```
rpokorny, line 241: I'm surprised it isn't causing compilation errors to have both this and the copy is SimpleCompnonents defined and exported.  Shouldn't that cause a naming collision?

rpokorny:
> I think that there is value in creating .nx-divider__vertical for instances where we want a vertical border where one is not normally present or we want to override a lighter thinner vertical border. We could just pick left or right.

That still comes back to the question of what does that mean/how does it work though.  You can't have a border between adjacent elements without having the element be adjacent in the first place, which isn't the way things lay out be default.  So that means having a container to define that adjacency, and then that container might as well define the vertical divider too.  And we already have this, it's exactly what `nx-grid` is.

How would an `.nx-divider__vertical` work in absence of something like `.nx-grid`?

glenhunter:
> > I think that there is value in creating .nx-divider__vertical for instances where we want a vertical border where one is not normally present or we want to override a lighter thinner vertical border. We could just pick left or right.
> 
> That still comes back to the question of what does that mean/how does it work though. You can't have a border between adjacent elements without having the element be adjacent in the first place, which isn't the way things lay out by default. So that means having a container to define that adjacency, and then that container might as well define the vertical divider too. And we already have this, it's exactly what `nx-grid` is.
> 
> How would an `.nx-divider__vertical` work in absence of something like `.nx-grid`?

It wouldn't. It would just be a class that could be added. I see it being used in places (like `nx-table`) where there's already a border but design wants a more prominent border than the default.

I've seen table mockups from design with this specific use-case so I'm just suggesting that we create the style here. We can pick whether it's on the left or right side and it would be applied accordingly.

rpokorny:
> I am also fine with leaving this out completely if nobody else sees the point. The problem is easy enough to solve with custom classes.

It's not that I don't see the point it's that I don't even see what this could _be_ in practice.  If I were asked to implement this I wouldn't know what I was supposed to be implementing.  If you or @avnvbhatta want to implement it so I can see what you're talking about I'd be happy to look at it.

glenhunter:
> > I am also fine with leaving this out completely if nobody else sees the point. The problem is easy enough to solve with custom classes.
> 
> It's not that I don't see the point it's that I don't even see what this could _be_ in practice. If I were asked to implement this I wouldn't know what I was supposed to be implementing. If you or @avnvbhatta want to implement it so I can see what you're talking about I'd be happy to look at it.

`<td class="nx-cell nx-divider__vertical">` would display a darker two pixel border on the left (or right) edge instead of the standard lighter `1px` border. I looked through the gallery and it's hard to see any other place where such a style might be used excepting (maybe) some of the page headers and maybe some of the wider versions of `nx-card`

avnvbhatta:
> > > Given that this is really just styles, it should probably be documented as a style (with mention of its convenience component) rather than documenting it primarily as a component. This is how we prefer to document things that don't have any logic to encapsulate in order to facilitate their use outside of react projects.
> > 
> > 
> > This has been implemented. I have added `nx-divider` under styles and changed `NxDivider` to be under `SimpleComponents.tsx` as a convenience component rather than a standalone component.
> > Check [bcdcb7b](https://github.com/sonatype/sonatype-react-shared-components/commit/bcdcb7bb385487554967379afcc1ad75cde17ad5)
> 
> It seems that the component's files are still present though

I seem to have copied the previous implementation of cutting it.
The files have been removed.
e1f79dfe5abbe6861ef08f7b224747b2b200e2dd

gallery/src/styles/NxDivider/NxDividerHorizontalExample.tsx
avnvbhatta: > I think we should avoid this part of the example, since buttons in this sort of position will more or less always be part of a footer, which has its own border and which does not involve `NxDivider`

I can see why this example might be confusing. It has been removed.
04aaa88dd382180096cd6250cb831f5b251214b3

gallery/src/styles/NxDivider/NxDividerHorizontalExample.tsx
avnvbhatta: > We demo styles using plain HTML elements and not convenience components. In fact for especially simple setups such as this one, we typically write the demo in actual HTML rather than in JavaScript with react. See most of the tile demos for example.

The example has been changed to in HTML rather than JS/React.
eb82df227306a3b7f7e0b0b35f58a3a7c488ef33

gallery/src/styles/NxDivider/NxDividerPage.tsx
avnvbhatta: > I think this example is unnecessary given the mention that you have at the top of the page about using `nx-grid` for vertical dividers. `nx-grid` already has its own docs and examples

This example has been removed.
eb82df227306a3b7f7e0b0b35f58a3a7c488ef33

avnvbhatta:
> For all of the examples I see `.nx-divider`'s rules twice in the inspector
> Screen Shot 2021-09-09 at 3 31 45 PM

Hmm interesting. That doesn't seem to be the case for me.
image

gallery/src/styles/NxDivider/NxDividerPage.tsx
avnvbhatta: > Personally I also think this example is unnecessary – there isn't anything special about the usage of `.nx-divider` in a form vs other contexts that is being demonstrated here.

The example has been removed.
eb82df227306a3b7f7e0b0b35f58a3a7c488ef33

lib/src/base-styles/_nx-divider.scss
avnvbhatta: > I'm not sure a border radius here is part of the spec. The divider in sketch seems squared-off no matter how far I zoom in. Not to mention, wouldn't this need to be 0.5?

Removed here: 1c8d806169e9f3fcf770a7d8dc42b9e9ab95436c
The sketch file shows that there was a radius of 1, so I initially had it as 1. But for something with a `height` of `1px`, like you said, it's really difficult to see the `border-radius` anyway.

lib/src/base-styles/_nx-divider.scss
avnvbhatta: > Is there a reason this isn't just `height`?

This was a 'to-be-removed' remains of me messing around with both horizontal and vertical borders.
Check 1c8d806169e9f3fcf770a7d8dc42b9e9ab95436c

lib/src/base-styles/_nx-grid.scss
avnvbhatta: > The vertical lines actually _do_ have a border radius to them in sketch, which is problematic as we can't do that on a border.

You're right. Not sure how we would want to go about this. 
Thoughts about this @alexalyons126, @glenhunter, @entroform  ?

lib/src/components/NxDivider/NxDivider.scss
avnvbhatta: > This file needs to be deleted

Check 1c8d806169e9f3fcf770a7d8dc42b9e9ab95436c

lib/src/components/NxDivider/NxDivider.tsx
avnvbhatta: > Needs to be deleted

Check e1f79dfe5abbe6861ef08f7b224747b2b200e2dd

lib/src/components/NxDivider/__tests__/NxDivider.test.tsx
avnvbhatta: > > Can be simplified and moved to the SimpleComponents tests
> 
> Oh I see it's already there too. This needs to be deleted as well then.

Check e1f79dfe5abbe6861ef08f7b224747b2b200e2dd

lib/src/index.ts
avnvbhatta: > I'm surprised it isn't causing compilation errors to have both this and the copy is SimpleCompnonents defined and exported. Shouldn't that cause a naming collision? Regardless, this should be deleted.

Not sure why I didn't get any compilation errors. This should've been caught easily.
Removed. Check e1f79dfe5abbe6861ef08f7b224747b2b200e2dd

rpokorny:
> Hmm interesting. That doesn't seem to be the case for me.

It was happening due to the duplicated implementations, so it would be gone now that that's cleared up.

avnvbhatta:
> > > I am also fine with leaving this out completely if nobody else sees the point. The problem is easy enough to solve with custom classes.
> > 
> > 
> > It's not that I don't see the point it's that I don't even see what this could _be_ in practice. If I were asked to implement this I wouldn't know what I was supposed to be implementing. If you or @avnvbhatta want to implement it so I can see what you're talking about I'd be happy to look at it.
> 
> `<td class="nx-cell nx-divider__vertical">` would display a darker two pixel border on the left (or right) edge instead of the standard lighter `1px` border. I looked through the gallery and it's hard to see any other place where such a style might be used excepting (maybe) some of the page headers and maybe some of the wider versions of `nx-card`.
> 
> Note that I see this as only changing border attributes, no spacing would change.

So I've added the class here 71c46947a3c9292db9e20be041065593d820590d
And going along with Glen's example, it works:
image

For a default, 2px border on the right (or left), it's fine as it stands, but if users want specific spacing values or want to alter the divider's position, it'd be quite hacky (or even impossible). So if this is the default and "most likely to be used" use case then it works, if not, maybe we should consider alternatives, or even not use this approach at all.

rpokorny:
> So I've added the class here 71c4694
And going along with Glen's example, it works:

I guess for this approach I'd ask it it makes sense as a general class or if it is specific to table cells.  The spacing around it being the key consideration there, I think.  If it's specific to table cells, it should probably be a modifier class on `.nx-cell`.

Also, @glenhunter do you have a link to the relevant mockup so we can know if this is meeting the need there?

lib/src/base-styles/_nx-grid.scss
avnvbhatta: > > The vertical lines actually _do_ have a border radius to them in sketch, which is problematic as we can't do that on a border.
> 
> You're right. Not sure how we would want to go about this.
> Thoughts about this @alexalyons126, @glenhunter, @entroform ?

As per discussion, the `border-radius` is barely visible on a `2px` line, so we are fine with current implementation of having no `border-radius`.

lib/src/components/__tests__/SimpleComponents.test.tsx
```diff
@@ -343,3 +344,9 @@ describe('NxSystemNotice.Container', function() {
     expect(shallow(<NxSystemNotice.Container/>)).toMatchSelector('div.nx-system-notice-container');
   });
 });
+
+describe('NxDivider', function() {
+  it('makes a <hr> tag with an nx-divider class', function() {
```
entroform, line 349:
```suggestion
  it('makes a <hr> tag with a nx-divider class', function() {
```

lib/src/base-styles/_nx-divider.scss
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-divider {
+  background-color: var(--nx-swatch-indigo-90);
+  border-style: none;
+  margin: 24px 0;
```
entroform, line 10: Just wondering if we should be using space variable like `var(--nx-spacing-6x)` here?

lib/src/components/__tests__/SimpleComponents.test.tsx
rpokorny: Actually we settled long ago on using "an" in this case (which comes up many many times in the RSC)

entroform: > Actually we settled long ago on using "an" in this case (which comes up many many times in the RSC)

Ah thanks! Good to know 👍

lib/src/base-styles/_nx-divider.scss
avnvbhatta: > var(--nx-spacing-6x)

Added here: 00ac26997d924478e12602e8b0eaa42dca71be95

entroform (approved):

avnvbhatta:
> > > I am also fine with leaving this out completely if nobody else sees the point. The problem is easy enough to solve with custom classes.
> > 
> > 
> > It's not that I don't see the point it's that I don't even see what this could _be_ in practice. If I were asked to implement this I wouldn't know what I was supposed to be implementing. If you or @avnvbhatta want to implement it so I can see what you're talking about I'd be happy to look at it.
> 
> `<td class="nx-cell nx-divider__vertical">` would display a darker two pixel border on the left (or right) edge instead of the standard lighter `1px` border. I looked through the gallery and it's hard to see any other place where such a style might be used excepting (maybe) some of the page headers and maybe some of the wider versions of `nx-card`.
> 
> Note that I see this as only changing border attributes, no spacing would change.

As per discussion, we are not moving forward with implementing vertical borders. Hence, this PR does not and should not contain any vertical border implementation besides mentioning the `.nx-grid` class.
All logic to implement vertical borders has been removed.

rpokorny (approved):

PR #477, opened 2021-09-07
Fix Active State on Dropdowns - RSC-681

entroform:
https://issues.sonatype.org/browse/RSC-681?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel


entroform:
Button active state has a gray background:
button-active-state

Dropdown toggle active state does not:
dropdown-active-state

Currently the default behaviour is that the toggle uses `--nx-color-interactive-background-active`, should I override it to use `transparent` as shown in the mockup?

avnvbhatta (approved):
LGTM

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -43,8 +43,16 @@
     }
   }
 
+  &:active {
+    background-color: transparent;
```
rpokorny, line 47: I suspect this change is incorrect but I'm waiting to hear back from @alexalyons126 [here](https://sonatype.slack.com/archives/CJUF9G7T2/p1631128816043100?thread_ts=1631114504.042400&cid=CJUF9G7T2)

entroform: > I suspect this change is incorrect but I'm waiting to hear back from @alexalyons126 [here](https://sonatype.slack.com/archives/CJUF9G7T2/p1631128816043100?thread_ts=1631114504.042400&cid=CJUF9G7T2)

Thanks! I did reach out regarding this topic. Apparently the Dropdown button behaviour is following the same pattern as the form fields and not button so it remains transparent. I will further confirm this with @alexalyons126 and get back to you.

rpokorny (approved):

PR #478, opened 2021-09-07
Add focus style to Dropdown items - RSC-688

entroform:
https://issues.sonatype.org/browse/RSC-688


avnvbhatta (approved):
🚀

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -56,13 +56,22 @@
   @include nx-text-helpers.truncate-ellipsis;
   @include nx-text-helpers.regular;
 
-  border: none;
+  border: 1px solid transparent;
```
rpokorny, line 59: This adds to the height of the item (it is now 34px instead of 32).  To fix this you can either subtract 1 from the padding or, preferably IMO, use an inset outline instead of a border.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -56,13 +56,22 @@
...
   line-height: var(--nx-line-height);
   padding: var(--nx-spacing-base) var(--nx-spacing-4x);
   text-decoration: none;
+
+  &:focus {
+    border: 1px solid var(--nx-color-interactive-border-focus);
+    box-shadow: var(--nx-box-shadow-focus);
+
+    &, &:hover, &:active {
+      background-color: transparent;
```
rpokorny, line 72: I'm not sure this is desirable.  The mockups tend not to have combinations of states like focus+hover or focus+clicked, but I would assume we want both of those states to do what they usually do.  It seems odd to have an element not visually respond to hover or click just because it is already focused.  @alexalyons126 what do you think?

lib/src/components/NxDropdown/NxDropdown.scss
entroform: > This adds to the height of the item (it is now 34px instead of 32). To fix this you can either subtract 1 from the padding or, preferably IMO, use an inset outline instead of a border.

I liked the inset outline idea but TIL that inset in outline doesn't mean the line is drawn inside; It creates an "inset" 3d effect instead. I subtracted 1px from the padding in this commit. Let me know what your thoughts:
c14e789eedf25680dd3cf41dd75086ec2a3a3dac

rpokorny: > I liked the inset outline idea but TIL that inset in outline doesn't mean the line is drawn inside; It creates an "inset" 3d effect instead.

Oh sorry I didn't mean that fugliness, I meant https://github.com/sonatype/sonatype-react-shared-components/blob/master/lib/src/base-styles/_nx-pagination.scss#L32

entroform: Ah thanks, I updated it here:
ab2756aa36df67e2bd9f331dcf949566b0c9f436

lib/src/components/NxDropdown/NxDropdown.scss
entroform: > I'm not sure this is desirable. The mockups tend not to have combinations of states like focus+hover or focus+clicked, but I would assume we want both of those states to do what they usually do. It seems odd to have an element not visually respond to hover or click just because it is already focused. @alexalyons126 what do you think?

As discussed and confirmed by @alexalyons126, the default hover and active background colors should be added back when in focus state.
Fixed in this commit:
303c2c53ee001bc84189c39268c7a2621f2f26e4

rpokorny (approved):

PR #479, opened 2021-09-07
Add propTypes to NxTransferList - RSC-671 

entroform:
https://issues.sonatype.org/browse/RSC-671

### DIFF
--- a/lib/src/components/NxTransferList/NxTransferList.tsx
+++ b/lib/src/components/NxTransferList/NxTransferList.tsx
@@ -4,7 +4,7 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React, { FC, FormEvent, useMemo } from 'react';
+import React, { FormEvent, useMemo } from 'react';
 import classnames from 'classnames';
 import { faPlusCircle, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
 import { toLower, filter, includes, groupBy, partial, identity, prop, pipe } from 'ramda';
@@ -19,7 +19,7 @@ export { Props, propTypes, DataItem } from './types';
 
 import './NxTransferList.scss';
 
-function TransferListItem<T extends string | number>(props: TransferListItemProps<T>) {
+function TransferListItem<T extends string | number = string>(props: TransferListItemProps<T>) {
   const { checked, id, displayName, onChange: onChangeProp } = props;
 
   function onChange(evt: FormEvent<HTMLInputElement>) {
@@ -44,7 +44,7 @@ const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-const NxTransferList: FC<Props<string | number>> = <T extends string | number>(props: Props<T>) => {
+const NxTransferList = <T extends string | number>(props: Props<T>) => {
   const {
     allItems,
     selectedItems,
@@ -139,12 +139,7 @@ const NxTransferList: FC<Props<string | number>> = <T extends string | number>(p
           }
           <div className="nx-transfer-list__item-list">
             { visibleAvailableItems.map(
-                i => (
-                  <TransferListItem key={i.id}
-                                    checked={false}
-                                    onChange={onChange as TransferListItemProps<string | number>['onChange']}
-                                    { ...i } />
-                ))
+                i => <TransferListItem<T> key={i.id} { ...{ checked: false, onChange, ...i } } />)
             }
           </div>
           <div className="nx-transfer-list__footer">
@@ -166,12 +161,7 @@ const NxTransferList: FC<Props<string | number>> = <T extends string | number>(p
           }
           <div className="nx-transfer-list__item-list">
             { visibleSelectedItems.map(
-                i => (
-                  <TransferListItem key={i.id}
-                                    checked={true}
-                                    onChange={onChange as TransferListItemProps<string | number>['onChange']}
-                                    { ...i } />
-                ))
+                i => <TransferListItem<T> key={i.id} { ...{ checked: true, onChange, ...i } } />)
             }
           </div>
           <div className="nx-transfer-list__footer">
--- a/lib/src/components/NxTransferList/NxTransferList.tsx
+++ b/lib/src/components/NxTransferList/NxTransferList.tsx
@@ -137,7 +137,7 @@ export default function NxTransferList<T extends string | number>(props: Props<T
           }
           <div className="nx-transfer-list__item-list">
             { visibleAvailableItems.map(
-                i => <TransferListItem<T> key={i.id} checked={false} onChange={onChange} { ...i } />)
+                i => <TransferListItem key={i.id} checked={false} onChange={onChange} { ...i } />)
             }
           </div>
           <div className="nx-transfer-list__footer">
--- a/lib/src/components/NxTransferList/NxTransferList.tsx
+++ b/lib/src/components/NxTransferList/NxTransferList.tsx
@@ -15,7 +15,7 @@ import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import NxFieldset from '../NxFieldset/NxFieldset';
 import NxOverflowTooltip from '../NxTooltip/NxOverflowTooltip';
 
-export { Props, propTypes, DataItem } from './types';
+export { Props, DataItem } from './types';
 
 import './NxTransferList.scss';
 
@@ -42,7 +42,7 @@ const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-function NxTransferList<T extends string | number>(props: Props<T>) {
+export default function NxTransferList<T extends string | number>(props: Props<T>) {
   const {
     allItems,
     selectedItems,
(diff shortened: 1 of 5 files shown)


entroform:
`eslint-plugin-react` :  `react/prop-types` seem to ignore missing .propTypes if props is already typed in Typescript. There seems to be no options for me to change this behaviour in the plugin.

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,11 +38,13 @@ function TransferListItem<T extends string | number>(props: TransferListItemProp
   );
 }
 
+TransferListItem.propTypes = transferListItemPropTypes;
```
rpokorny, line 41: As an internal component `TransferListItem` does not need prop types. Only externally visible components need them.

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,11 +38,13 @@ function TransferListItem<T extends string | number>(props: TransferListItemProp
...
+
 const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${n === 1 ? '' : 's'} ${kind}`,
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-export default function NxTransferList<T extends string | number>(props: Props<T>) {
+const NxTransferList = <T extends string | number>(props: Props<T>) => {
```
rpokorny, line 47: Why this change?

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -137,7 +139,7 @@ export default function NxTransferList<T extends string | number>(props: Props<T
           }
           <div className="nx-transfer-list__item-list">
             { visibleAvailableItems.map(
-                i => <TransferListItem key={i.id} checked={false} onChange={onChange} { ...i } />)
+                i => <TransferListItem<T> key={i.id} { ...{ checked: false, onChange, ...i } } />)
```
rpokorny, line 142: Why this change?

lib/src/components/NxTransferList/types.ts
```diff
@@ -39,3 +50,25 @@ export interface Props<T extends string | number = string> extends StatefulProps
...
+    ]).isRequired,
+    displayName: PropTypes.string.isRequired
+  })).isRequired as Validator<DataItem<string | number>[]>,
+  selectedItems: PropTypes.instanceOf(Set).isRequired as Validator<Set<string | number>>,
+  availableItemsLabel: PropTypes.node,
+  selectedItemsLabel: PropTypes.node,
+  availableItemsCountFormatter: PropTypes.func as Validator<(n: number) => string>,
+  selectedItemsCountFormatter: PropTypes.func as Validator<(n: number) => string>,
+  showMoveAll: PropTypes.bool,
+  onChange: PropTypes.func as Validator<(newSelected: Set<string | number>) => void>,
```
rpokorny, line 68: These casts shouldn't be necessary

lib/src/components/NxTransferList/types.ts
```diff
@@ -39,3 +50,25 @@ export interface Props<T extends string | number = string> extends StatefulProps
...
 }
+
+export const propTypes: ValidationMap<Props<string | number>> = {
+  allItems: PropTypes.arrayOf(PropTypes.shape({
+    id: PropTypes.oneOfType([
+      PropTypes.string.isRequired,
+      PropTypes.number.isRequired
+    ]).isRequired,
+    displayName: PropTypes.string.isRequired
+  })).isRequired as Validator<DataItem<string | number>[]>,
```
rpokorny, line 61: Unecessary cast?

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > As an internal component `TransferListItem` does not need prop types. Only externally visible components need them.

Ah, that makes sense. Will remove it.

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > Why this change?

Ops, I changed it out of habit when I removed the `export default`, will change it back to use function.

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > Why this change?

I was getting `is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes` error and that fixed it.
Now that we're removing propTypes from TransferListItem it looks like it's working again. Going to change it back.

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > As an internal component `TransferListItem` does not need prop types. Only externally visible components need them.

Thanks again, I've removed the TransferListItem.propTypes in this commit:
0f0c1c1799991ec9d4c9e64eca29b9c86c4061c0

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > Why this change?
This was changed back in this commit:
0f0c1c1799991ec9d4c9e64eca29b9c86c4061c0

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > Why this change?

This was changed back in this commit:
0f0c1c1799991ec9d4c9e64eca29b9c86c4061c0

lib/src/components/NxTransferList/types.ts
entroform: > These casts shouldn't be necessary

Thanks! This was removed in this commit:
0f0c1c1799991ec9d4c9e64eca29b9c86c4061c0

lib/src/components/NxTransferList/types.ts
entroform: > Unecessary cast?

Thanks! I was missing an `.isRequired`.
This was fixed in this commit:
0f0c1c1799991ec9d4c9e64eca29b9c86c4061c0

lib/src/components/NxTransferList/types.ts
rpokorny: > Thanks! This was removed in this commit:

And the other ones above, on `availableItemsCountFormatter` and `selectedItemsCountFormatter`

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -9,17 +9,17 @@ import classnames from 'classnames';
...
 
-import { Props, TransferListItemProps, DataItem, FilterFn } from './types';
+import { Props, propTypes, TransferListItemProps, DataItem, FilterFn } from './types';
 import NxFilterInput from '../NxFilterInput/NxFilterInput';
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import NxFieldset from '../NxFieldset/NxFieldset';
 import NxOverflowTooltip from '../NxTooltip/NxOverflowTooltip';
 
-export { Props, DataItem } from './types';
+export { Props, propTypes, DataItem } from './types';
```
rpokorny, line 18: No need to export this since it's attached to the component function

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -42,7 +42,7 @@ const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-export default function NxTransferList<T extends string | number>(props: Props<T>) {
+function NxTransferList<T extends string | number>(props: Props<T>) {
```
rpokorny, line 45: No need for this change, you can still export here and then assign the propTypes to it below.  See `NxPagination` for example.

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -159,7 +159,7 @@ export default function NxTransferList<T extends string | number>(props: Props<T
           }
           <div className="nx-transfer-list__item-list">
             { visibleSelectedItems.map(
-                i => <TransferListItem key={i.id} checked={true} onChange={onChange} { ...i } />)
+                i => <TransferListItem key={i.id} checked={true} onChange={onChange} {...i} />)
```
rpokorny, line 162: A few minor extraneous changes here

lib/src/components/NxTransferList/types.ts
entroform: Thanks! I updated the interface for those two to accept `null` so that I'm able to remove those casts:
8df61c7ba4c2ad10415a869de9993da4bfe5b9af

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: Ah, I was using NxTooltip as a reference and it exported them.
Removed in this commit:
8df61c7ba4c2ad10415a869de9993da4bfe5b9af

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: oo thanks didn't know that; this is fixed in the following commit:
8df61c7ba4c2ad10415a869de9993da4bfe5b9af

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: ops! For future reference, is the style to have extra space in between brackets when spread is use?
This is fixed here:
8df61c7ba4c2ad10415a869de9993da4bfe5b9af

rpokorny: > For future reference, is the style to have extra space in between brackets when spread is use?

Strictly speaking eslint is our source of truth there.  It wasn't complaining so your change wasn't wrong just extraneous.  That said, I personally prefer to have these spaces there within curly braces but not within square braces.  Be aware that other projects at Sonatype have somewhat different formatting rules though.

rpokorny (approved):
Before you can merge you'll need to get a passing CI build and before you can get a passing CI build you'll need to bump the version numbers.  You should probably merge from master before that so you are doing the most correct version bump (and because that's another requirement for merging the PR)

PR #481, opened 2021-09-09
change error button colours RSC-678

glenhunter:
https://issues.sonatype.org/browse/RSC-678


lib/src/base-styles/_nx-btn.scss
```diff
@@ -79,24 +79,24 @@
 }
 
 .nx-btn--error {
-  background-color: var(--nx-swatch-red-45);
-  border-color: var(--nx-swatch-red-45);
+  background-color: var(--nx-swatch-red-40);
+  border-color: var(--nx-swatch-red-40);
   color: var(--nx-swatch-white);
 
   &:hover, &:focus, &:active {
```
rpokorny, line 86: Shouldn't `:focus` and `:active` be removed from this selector?

rpokorny:
What is hover+focus supposed to look like?

lib/src/base-styles/_nx-btn.scss
glenhunter: > Shouldn't `:focus` and `:active` be removed from this selector?

Yes, they were there to share the `color` attribute, but that's not even necessary. https://github.com/sonatype/sonatype-react-shared-components/pull/481/commits/1ddadb97ab08d82edf313cd1de4f4c1a392ea935

glenhunter:
> What is hover+focus supposed to look like?

According to @alexalyons126 `:hover` + `:focus` uses the `:hover` background colour and the `:focus` border (wall walk Sept 10)

https://github.com/sonatype/sonatype-react-shared-components/pull/481/commits/1ddadb97ab08d82edf313cd1de4f4c1a392ea935

Note in this commit that I also had to override  `:hover` + `:focus` to get `:active` to look correct.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -79,24 +79,28 @@
 }
 
 .nx-btn--error {
-  background-color: var(--nx-swatch-red-45);
-  border-color: var(--nx-swatch-red-45);
+  background-color: var(--nx-swatch-red-40);
+  border-color: var(--nx-swatch-red-40);
```
rpokorny, line 83: If I'm not mistaken, if this was `transparent`, then `:hover` and `:active` wouldn't need to set `border-color`, which means only `:focus` is overridding it and then the `:hover` and `:focus` blocks could be re-ordered to remove the need for the `:hover:focus` block and thus remove the need for the `:hover:focus:active` block as well.

glenhunter: > If I'm not mistaken, if this was `transparent`, then `:hover` and `:active` wouldn't need to set `border-color`, which means only `:focus` is overridding it and then the `:hover` and `:focus` blocks could be re-ordered to remove the need for the `:hover:focus` block and thus remove the need to specify `:hover:focus:active` rather than just `:active` as well.

Yep that was almost perfect. Just needed the border on `:active` to match Sketch https://github.com/sonatype/sonatype-react-shared-components/pull/481/commits/194217ed9bd1ffbdbc51b84adae392d681d99652

lib/src/base-styles/_nx-btn.scss
```diff
@@ -79,24 +79,22 @@
...
 
-  &:hover, &:focus, &:active {
-    background-color: var(--nx-swatch-red-30);
-    border-color: var(--nx-swatch-red-30);
-    color: var(--nx-swatch-white);
-  }
-
   &:focus {
-    background-color: var(--nx-swatch-red-45);
+    background-color: var(--nx-swatch-red-40);
```
rpokorny, line 87: Can be removed, right?

lib/src/base-styles/_nx-btn.scss
```diff
@@ -79,24 +79,22 @@
...
 
+  &:hover {
+    background-color: var(--nx-swatch-red-35);
+  }
+
   &:active {
-    background-color: var(--nx-swatch-red-60);
-    border-color: var(--nx-swatch-red-60);
+    background-color: var(--nx-swatch-red-65);
+    border-color: var(--nx-swatch-red-65);
```
rpokorny, line 97: Not necessary

rpokorny: > Not necessary

Or are you interpreting the mockups as saying that an active+focused button (which is effectively any active button) should not have the blue border.  If that's the case, you should probably take off the shadow here too.

glenhunter: > > Not necessary
> 
> Or are you interpreting the mockups as saying that an active+focused button (which is effectively any active button) should not have the blue border? If that's the case, you should probably take off the shadow here too.

Confirmed by @alexalyons126  in wall walk.

glenhunter: > > Not necessary
> 
> Or are you interpreting the mockups as saying that an active+focused button (which is effectively any active button) should not have the blue border? If that's the case, you should probably take off the shadow here too.

Box-shadow removed here: https://github.com/sonatype/sonatype-react-shared-components/pull/481/commits/5672f7696e727ba4d80bb51c8cc2676834b35edb

lib/src/base-styles/_nx-btn.scss
glenhunter: > Can be removed, right?

Ah right, yes it can: https://github.com/sonatype/sonatype-react-shared-components/pull/481/commits/5672f7696e727ba4d80bb51c8cc2676834b35edb

lib/src/base-styles/_nx-btn.scss
rpokorny: > Box-shadow removed here: 5672f76

I'd prefer the border color be set back to transparent rather than duplicating the background color.  That way it's more consistent: border is always transparent except when the focus border is there

rpokorny (approved):

lib/src/base-styles/_nx-btn.scss
glenhunter: > > Box-shadow removed here: [5672f76](https://github.com/sonatype/sonatype-react-shared-components/commit/5672f7696e727ba4d80bb51c8cc2676834b35edb)
> 
> I'd prefer the border color be set back to transparent rather than duplicating the background color. That way it's more consistent: border is always transparent except when the focus border is there

Agreed: https://github.com/sonatype/sonatype-react-shared-components/pull/481/commits/0b3cb58ec115d48aaf35451fed42ae14a8e958f0

PR #482, opened 2021-09-10
NxSmallThreatCounter - RSC-666

rpokorny:
https://issues.sonatype.org/browse/RSC-666

I initially went back and forth about whether these should be a new component or just a modifier on `NxThreatCounter`.  As much as I would have liked them to just be a modifier, since they convey the same information, they ultimately work differently enough, including taking a few different props, that it was best to make them completely separate.


entroform (approved):

gallery/src/components/NxSmallThreatCounter/NxSmallThreatCounterPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+const NxSmallThreatCounterCode = require('./NxSmallThreatCounterExample?raw');
+const NxSmallThreatCounterCustomCode = require('./NxSmallThreatCounterCustomExample?raw');
+
+const NxSmallThreatCounterPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <NxP>
+        Similar to <NxCode>NxThreatCounter</NxCode>, <NxCode>NxSmallThreatCounter</NxCode> displays a series of
+        counts of "threats" in different severity categories. This component is more compact than the regular
+        <NxCode>NxThreatCounter</NxCode> and is specifically intended for use within table rows.
```
avnvbhatta, line 24:
```suggestion
        counts of "threats" in different severity categories. This component is more compact than the 
        regular <NxCode>NxThreatCounter</NxCode> and is specifically intended for use within table rows.
```

lib/src/components/NxSmallThreatCounter/NxSmallThreatCounter.tsx
```diff
@@ -0,0 +1,90 @@
...
+      typeof severeCount !== 'number' &&
+      typeof moderateCount !== 'number' &&
+      typeof lowCount !== 'number' &&
+      typeof noneCount !== 'number') {
+    console.warn('No counts have been provided and so nothing will be rendered.');
+    return null;
+  }
+
+  return (
+    <div className={className} { ...attrs }>
```
avnvbhatta, line 70: Although this component is primarily used within tables, I think it would still benefit from accepting `refs`?

rpokorny: > Although this component is primarily used within tables, I think it would still benefit from accepting refs?

The thing is the main point of accepting refs (though theoretically not the only one) is to support tooltips.  But this component as a whole should not have a tooltip since each of it's visible parts has a built-in tooltip

gallery/src/components/NxSmallThreatCounter/NxSmallThreatCounterPage.tsx
rpokorny: ae6d9007

avnvbhatta (approved):

PR #484, opened 2021-09-13
Updated default and hover state styling - RSC-679

entroform:
https://issues.sonatype.org/browse/RSC-679

In addition to NxCheckbox and NxRadio changes, I updated `--nx-color-form-element-border` CSS variable with `--nx-swatch-grey-50` to match Kaleido. This might will change the border color for other form elements as side effect.

### DIFF
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -48,9 +48,7 @@
 
     &:hover {
       .nx-radio-checkbox__control {
-        border-color: var(--nx-swatch-grey-05);
         filter: none;
-        stroke: var(--nx-swatch-grey-05);
       }
 
       .nx-radio-checkbox__input:checked + .nx-radio-checkbox__control {
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -124,8 +124,7 @@ svg.nx-radio-checkbox__control {
     z-index: 0;
   }
 
-  &:checked,
-  &:active:not(:disabled) {
+  &:checked {
     + .nx-radio-checkbox__control {
       background-color: var(--nx-swatch-blue-40);
       border-color: var(--nx-swatch-blue-40);
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -29,7 +29,7 @@
     cursor: pointer;
 
     .nx-radio-checkbox__control {
-      border-color: var(--nx-color-form-element-border);
+      border-color: var(--nx-swatch-grey-05);
       stroke: var(--nx-swatch-grey-05);
     }
   }
@@ -48,7 +48,7 @@
 
     &:hover {
       .nx-radio-checkbox__control {
-        border-color: var(--nx-color-form-element-border);
+        border-color: var(--nx-swatch-grey-05);
         filter: none;
         stroke: var(--nx-swatch-grey-05);
       }
@@ -74,7 +74,7 @@
 
 .nx-radio-checkbox__control {
   background-color: var(--nx-swatch-white);
-  border: var(--nx-border-default);
+  border: 1px solid var(--nx-color-form-element-border);
   flex-shrink: 0;
   margin-right: var(--nx-spacing-2x);
 }
@@ -83,7 +83,7 @@ svg.nx-radio-checkbox__control {
   border-style: none;
   background: none;
   fill: var(--nx-swatch-white);
-  stroke: var(--nx-swatch-indigo-95);
+  stroke: var(--nx-color-form-element-border);
 }
 
 .nx-radio-checkbox__content {
--- a/lib/src/base-styles/_nx-radio-checkbox.scss
+++ b/lib/src/base-styles/_nx-radio-checkbox.scss
@@ -124,7 +124,8 @@ svg.nx-radio-checkbox__control {
     z-index: 0;
   }
 
-  &:checked {
+  &:checked,
+  &:active:not(:disabled) {
     + .nx-radio-checkbox__control {
       background-color: var(--nx-swatch-blue-40);
       border-color: var(--nx-swatch-blue-40);
--- a/lib/src/components/NxCheckbox/NxCheckbox.scss
+++ b/lib/src/components/NxCheckbox/NxCheckbox.scss
@@ -36,6 +36,7 @@ $-width: 16px;
   + .nx-checkbox__box {
     background-color:var(--nx-color-interactive-background-disabled);
     border-color: var(--nx-color-interactive-border-disabled);
+
     .fa-check {
       color: var(--nx-color-text-disabled);
     }
--- a/lib/src/components/NxCheckbox/NxCheckbox.scss
+++ b/lib/src/components/NxCheckbox/NxCheckbox.scss
@@ -32,17 +32,6 @@ $-width: 16px;
   }
 }
 
-.fa-check {
-  visibility: hidden;
-}
-
-.nx-checkbox__input:checked,
-.nx-checkbox__input:active:not(:disabled) {
-  + .nx-checkbox__box .fa-check {
-    visibility: visible;
-  }
-}
-
 .nx-checkbox__input:checked:disabled {
   + .nx-checkbox__box {
     background-color:var(--nx-color-interactive-background-disabled);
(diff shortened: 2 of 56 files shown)


rpokorny:
> This might will change the border color for other form elements as side effect.

You should determine which ones change and whether those changes match kaleido.  One case I see is that this PR changes the hover border on `NxColorPicker` in a way that does not match the spec for that component.  That should be addressed here.

entroform:
> > This might will change the border color for other form elements as side effect.
> 
> You should determine which ones change and whether those changes match kaleido. One case I see is that this PR changes the hover border on `NxColorPicker` in a way that does not match the spec for that component. That should be addressed here.

Thanks, going to look into it and address it. Should I update the ticket to reflect changes affected by `--nx-color-form-element-border`?

rpokorny:
> Thanks, going to look into it and address it. Should I update the ticket to reflect changes affected by --nx-color-form-element-border?

The original purpose of the ticket is what it is (and shouldn't be changed at this point IMO) but if other "bugs" (mismatches with kaleido) get fixed in the process of fixing this one it wouldn't hurt to mention them in a comment on the JIRA ticket

glenhunter:
I think that `NxRadio` needs the same border adjustment(s)

entroform:
> > Thanks, going to look into it and address it. Should I update the ticket to reflect changes affected by --nx-color-form-element-border?
> 
> The original purpose of the ticket is what it is (and shouldn't be changed at this point IMO) but if other "bugs" (mismatches with kaleido) get fixed in the process of fixing this one it wouldn't hurt to mention them in a comment on the JIRA ticket

Fixed other mismatches affected by  `--nx-color-form-element-border` here :)
6b7d61614e64372ad4c8eab6bb09f948817ed8b3

I also noticed that the active state for radios and checkboxes also do no align; I updated it in the above commit.
Let me know if you find any issues with the implementation 👍

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -32,11 +32,22 @@ $-width: 16px;
...
+    visibility: visible;
+  }
+}
+
 .nx-checkbox__input:checked:disabled {
   + .nx-checkbox__box {
     background-color:var(--nx-color-interactive-background-disabled);
     border-color: var(--nx-color-interactive-border-disabled);
-
+
```
glenhunter, line 50:
```suggestion

```

lib/src/components/NxCheckbox/NxCheckbox.tsx
```diff
@@ -56,7 +56,7 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
             {/* Put a space in the box if not checked,
               * in order to provide a consistent vertical-align baseline
               */}
-            { isChecked ? <FontAwesomeIcon icon={faCheck} /> : '\u00A0' }
+            <FontAwesomeIcon icon={faCheck} />
```
glenhunter, line 59: Two things: is this in scope? if so do we still need the comment above?

entroform: > Two things: is this in scope? if so do we still need the comment above?

Hi Glen, I believe so. This change is needed because the icon needs to be visible in the `active` state as well as the`checked` state according to Kaleido. Nice catch, I forgot to remove the comment.

glenhunter: > > Two things: is this in scope? if so do we still need the comment above?
> 
> Hi Glen, I believe so. This change is needed because the icon needs to be visible in the `active` state as well as the`checked` state according to Kaleido. Nice catch, I forgot to remove the comment.

Well technically the `:active` state of a checkbox (does a checkbox even have an `active` state?) could be either of the state of checking it or unchecking it. The checkmark icon should only appear when the checkbox is in the `:checked` state

entroform:
I started a conversation about the active state on checkboxes and radios in the #react-components channel.
It looks like @alexalyons126 will be updating it in Kaleido and once that's done I'll update this PR 👍.

lib/src/components/NxCheckbox/NxCheckbox.tsx
entroform: > > > Two things: is this in scope? if so do we still need the comment above?
> > 
> > 
> > Hi Glen, I believe so. This change is needed because the icon needs to be visible in the `active` state as well as the`checked` state according to Kaleido. Nice catch, I forgot to remove the comment.
> 
> Well technically the `:active` state of a checkbox (does a checkbox even have an `active` state?) could be either of the state of checking it or unchecking it. The checkmark icon should only appear when the checkbox is in the `:checked` state

Yeah, I agree that it should only appear in the `:checked` state. I was confused because in the Kaleido the `:active` state is the same as the `:checked` state. I started a conversation about it in the #react-components channel and it looks like @alexalyons126 will be updating it

entroform: > > > > Two things: is this in scope? if so do we still need the comment above?
> > > 
> > > 
> > > Hi Glen, I believe so. This change is needed because the icon needs to be visible in the `active` state as well as the`checked` state according to Kaleido. Nice catch, I forgot to remove the comment.
> > 
> > 
> > Well technically the `:active` state of a checkbox (does a checkbox even have an `active` state?) could be either of the state of checking it or unchecking it. The checkmark icon should only appear when the checkbox is in the `:checked` state
> 
> Yeah, I agree that it should only appear in the `:checked` state. I was confused because in the Kaleido the `:active` state is the same as the `:checked` state. I started a conversation about it in the #react-components channel and it looks like @alexalyons126 will be updating it

After offline discussions, the decision was to revert back to how it was before. 44c1e9043dad5474c028b5d79b166e6b7e7baaef

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -48,7 +48,7 @@
 
     &:hover {
       .nx-radio-checkbox__control {
-        border-color: var(--nx-color-form-element-border);
+        border-color: var(--nx-swatch-grey-05);
         filter: none;
         stroke: var(--nx-swatch-grey-05);
```
rpokorny, line 53: I don't think border and stroke are actually necessary here at all; they just repeat what's already present via the non-focus `:hover` selector

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -36,7 +36,6 @@ $-width: 16px;
   + .nx-checkbox__box {
     background-color:var(--nx-color-interactive-background-disabled);
     border-color: var(--nx-color-interactive-border-disabled);
-
```
rpokorny, line 39: Extraneous and undesirable change

lib/src/components/NxCodeSnippet/NxCodeSnippet.scss
```diff
@@ -44,7 +44,7 @@
     width: 100%;
 
     .nx-text-input__box {
-      border-color: var(--nx-color-form-element-border);
+      border-color: var(--nx-swatch-grey-05);
```
rpokorny, line 47: @alexalyons126 is this actually what we want? It matches kaleido but effectively what this means is that the text field in an `NxCodeSnippet` would always look the way a _hovered_ text input is supposed to.

lib/src/base-styles/_nx-radio-checkbox.scss
entroform: > I don't think border and stroke are actually necessary here at all; they just repeat what's already present via the non-focus `:hover` selector

Thanks! Removed here:
5c5ea6780bb688265db64fa897b7852cd8c5935e

lib/src/components/NxCheckbox/NxCheckbox.scss
entroform: > Extraneous and undesirable change

Ops, totally missed this. Fixed:
5c5ea6780bb688265db64fa897b7852cd8c5935e

lib/src/components/NxCodeSnippet/NxCodeSnippet.scss
entroform: > > @alexalyons126 is this actually what we want? It matches kaleido but effectively what this means is that the text field in an `NxCodeSnippet` would always look the way a _hovered_ text input is supposed to.
> 
> I reached out to @alexalyons126 on Slack to confirm.

Confirmed with @alexalyons126 that this should match other form fields.
Reverted here:
4c97190c807354f3883f3f2691fe6a38a1f3c2e7

rpokorny (approved):

PR #487, opened 2021-09-17
Custom Icons for Clickable Table Rows - RSC-661

rpokorny:
https://issues.sonatype.org/browse/RSC-661

### DIFF
--- a/gallery/src/components/NxTable/NxTablePage.tsx
+++ b/gallery/src/components/NxTable/NxTablePage.tsx
@@ -11,6 +11,7 @@ import {GalleryDescriptionTile, GalleryExampleTile} from '../../gallery-componen
 
 import NxTableSimpleExample from './NxTableSimpleExample';
 import NxTableClickableExample from './NxTableClickableExample';
+import NxTableClickableCustomExample from './NxTableClickableCustomExample';
 import NxTableSortableExample from './NxTableSortableExample';
 import NxTableLoadingExample from './NxTableLoadingExample';
 import NxTableErrorExample from './NxTableErrorExample';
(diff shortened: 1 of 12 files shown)


lib/src/base-styles/_nx-tables.scss
```diff
@@ -306,10 +317,29 @@
...
+@mixin btn-cell-sizing($left-padding, $right-padding) {
+  margin-left: calc(0px - #{$left-padding});
+  margin-right: calc(0px - #{$right-padding});
+  padding-left: $left-padding;
+  padding-right: $right-padding;
+}
+
+.nx-cell__sort-btn, .nx-cell__row-btn {
+  width: min-content;
+
```
avnvbhatta, line 330: Should we add a `cursor: pointer;` to this class? Right now, hovering over both the sort cell and the row button icons don't show a pointer, making it seem like both the sort cell and the buttons are not clickable.

https://user-images.githubusercontent.com/7110862/134003686-ed73df59-52a5-482a-bb48-3e8b0e800434.mov

https://user-images.githubusercontent.com/7110862/134003713-25930c2c-b3b6-40ae-ae22-5817907ba2f0.mov

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -22,6 +23,7 @@ import NxTablePaginationFilterExample from './NxTablePaginationFilterExample';
 
 const tableSimpleExampleCode = require('./NxTableSimpleExample?raw');
 const tableClickableExample = require('./NxTableClickableExample?raw');
+const tableClickableCustomExample = require('./NxTableClickableCustomExample?raw');
```
avnvbhatta, line 26: This comment refers to line 239-241

The `clickAccessibleLabel` for `NxTable.Row` needs to be modified. It says "The accessible name to set on the click chevron", but in fact the accessible name could be for any other icon besides a chevron as well.

image

lib/src/base-styles/_nx-tables.scss
rpokorny: > Should we add a cursor: pointer; to this class? Right now, hovering over both the sort cell and the row button icons don't show a pointer, making it seem like both the sort cell and the buttons are not clickable.

Good catch. baa93c79

gallery/src/components/NxTable/NxTablePage.tsx
rpokorny: > This comment refers to line 239-241

The clickAccessibleLabel for NxTable.Row needs to be modified. It says "The accessible name to set on the click chevron", but in fact the accessible name could be for any other icon besides a chevron as well.

71f064f1

avnvbhatta (approved):

PR #488, opened 2021-09-17
Add NxTransferList ordering capability - RSC-649

entroform:
https://issues.sonatype.org/browse/RSC-649


entroform:
# Summary of changes:

Added a new `allowReordering` property to NxTransferList component.
If `allowReordering` is true:
- `selectedItems` must receive an Array instead of a Set; otherwise, it will throw an error.
-  You must also pass in both type parameters for `id` and `selectedItems` to NxTransferList, example:
`<NxTransferList<number, number[]> ...`
- When you select an item, the last item you selected appears at the bottom of the second column, instead of following the order of `allItems`.

If `allowReordering` is false:
- It behaves as before and you can must pass in a Set.
- `id` type is default as a string, if it's not, you must pass the type parameter, `<NxTransferList<number> ...`. This is probably a breaking change? I'm going to investigate a way around this.

gallery/src/components/NxTransferList/NxTransferListPage.tsx
```diff
@@ -59,13 +61,31 @@ const NxTransferListPage = () =>
...
               <NxTable.Cell>Yes</NxTable.Cell>
               <NxTable.Cell></NxTable.Cell>
               <NxTable.Cell>
-                A Set containing the ids of the items which are selected – that is, the ones which should appear
-                on the right side of the component. Be sure to always provide a fresh <NxCode>Set</NxCode> instance
-                when updating this prop, rather than mutating the same <NxCode>Set</NxCode>.
+                A Set, or an Array if <NxCode>allowReordering</NxCode> is set to true, containing the ids of the
+                items which are selected – that is, the ones which should
+                appear on the right side of the component. Be sure to always provide a fresh
+                <NxCode>Set</NxCode> or <NxCode>Array</NxCode> instance
```
rpokorny, line 71: Fixes whitespace around the `<NxCode>`
```suggestion
                appear on the right side of the component. Be sure to always provide a
                fresh <NxCode>Set</NxCode> or <NxCode>Array</NxCode> instance
```

gallery/src/components/NxTransferList/NxTransferListPage.tsx
```diff
@@ -59,13 +61,31 @@ const NxTransferListPage = () =>
...
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell>false</NxTable.Cell>
+              <NxTable.Cell>
+                Allow reordering of selected items.
+                When this is enabled, the ordering of the selected list is no longer dictated by
+                the order of <NxCode>allItems</NxCode>. Instead, it is determined by the order in which
+                the user selects the item. The last selected item appears at the bottom of the selected list.
+                Furthermore, if this is set to true, <NxCode>selectedItems</NxCode> must be
+                an <NxCode>Array</NxCode> instead of a <NxCode>Set</NxCode> and you must pass in
+                both the<NxCode>id</NxCode> and <NxCode>selectedItems</NxCode> type
```
rpokorny, line 87:
```suggestion
                both the <NxCode>id</NxCode> and <NxCode>selectedItems</NxCode> type
```

gallery/src/components/NxTransferList/NxTransferListPage.tsx
```diff
@@ -59,13 +61,31 @@ const NxTransferListPage = () =>
...
+                <NxCode>Set</NxCode> or <NxCode>Array</NxCode> instance
+                when updating this prop, rather than mutating the same <NxCode>Set</NxCode> or <NxCode>Array</NxCode>.
+              </NxTable.Cell>
+            </NxTable.Row>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>allowReordering</NxCode></NxTable.Cell>
+              <NxTable.Cell>boolean</NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell>false</NxTable.Cell>
+              <NxTable.Cell>
```
rpokorny, line 80: This description makes no mention of the buttons allowing the _user_ to reorder the items

lib/src/components/NxSearchTransferList/NxSearchTransferList.tsx
```diff
@@ -40,8 +40,8 @@ export default function NxSearchTransferList<T extends string | number>(props: P
       addedCount = addedItems.length,
       addedItemsCountFormatter = addedItemsCountFormatterProp || defaultAddedItemsCountFormatter;
 
-  function onRemoveAll(idsToRemove: Set<T>) {
-    const newAddedItems = reject(({ id }) => idsToRemove.has(id), addedItems);
+  function onRemoveAll(idsToRemove: T[]) {
+    const newAddedItems = reject(({ id }) => idsToRemove.includes(id), addedItems);
```
rpokorny, line 44: I wonder if it's worth it to collect the ids into a Set in a local variable first so that each lookup is O(1) not O(n)

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -19,8 +20,10 @@ const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-export default function NxTransferList<T extends string | number>(props: Props<T>) {
+export default function NxTransferList
+<T extends string | number = string, P extends Set<T> | T[] = Set<T>>(props: Props<T, P>) {
```
rpokorny, line 24: I think this is the first time we've had type parameters this long but the way I'd format this would probably be
```suggestion
export default function NxTransferList<
  T extends string | number = string,
  P extends Set<T> | T[] = Set<T>
>(props: Props<T, P>) {
```

Thoughts?

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +41,63 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
+  const groupedItems = useMemo(() =>
+        !allowReordering
+          ? groupBy(item => selectedItemsArray.includes(item.id) ? 'selected' : 'available', allItems)
+          : {
+            available: allItems.filter(item => !selectedItemsArray.includes(item.id)),
+            selected: selectedItemsArray
+                .map(item => allItems.find(_item => _item.id === item))
+                .filter(item => typeof item !== 'undefined') as DataItem<T>[]
+          },
+      [allItems, selectedItems]
```
rpokorny, line 62:
```suggestion
      [allItems, selectedItems, allowReordering]
```

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +41,63 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
-          [allItems, selectedItems]),
+  const groupedItems = useMemo(() =>
+        !allowReordering
+          ? groupBy(item => selectedItemsArray.includes(item.id) ? 'selected' : 'available', allItems)
+          : {
+            available: allItems.filter(item => !selectedItemsArray.includes(item.id)),
+            selected: selectedItemsArray
+                .map(item => allItems.find(_item => _item.id === item))
+                .filter(item => typeof item !== 'undefined') as DataItem<T>[]
+          },
```
rpokorny, line 61: This ternary is kind of backwards
```suggestion
        allowReordering ? {
            available: allItems.filter(item => !selectedItemsArray.includes(item.id)),
            selected: selectedItemsArray
                .map(item => allItems.find(_item => _item.id === item))
                .filter(item => typeof item !== 'undefined') as DataItem<T>[]
         } : groupBy(item => selectedItemsArray.includes(item.id) ? 'selected' : 'available', allItems),
```

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +41,63 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
 
-  const groupedItems = useMemo(() => groupBy(item => selectedItems.has(item.id) ? 'selected' : 'available', allItems),
-          [allItems, selectedItems]),
+  const groupedItems = useMemo(() =>
+        !allowReordering
+          ? groupBy(item => selectedItemsArray.includes(item.id) ? 'selected' : 'available', allItems)
+          : {
+            available: allItems.filter(item => !selectedItemsArray.includes(item.id)),
+            selected: selectedItemsArray
+                .map(item => allItems.find(_item => _item.id === item))
```
rpokorny, line 59: Repeatedly iterating `allItems` is probably slow.  I'd first build a lookup map from id to item and then use that here

rpokorny, line 59: If you use ramda's `chain` you can combine the mapping and filtering into one operation, and probably avoid the cast too

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > I think this is the first time we've had type parameters this long but the way I'd format this would probably be
> 
> Thoughts?

Thanks! Yeah I like it formatted this way as well!

rpokorny:
See https://github.com/sonatype/sonatype-react-shared-components/pull/498 for a way to get rid of the second type parameter

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +41,63 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
 
-  function onChange(checked: boolean, id: T) {
-    const newSelectedItems = new Set(selectedItems);
+  const handleOnChangeProp = (array: T[]) => onChangeProp((allowReordering ? array : new Set(array)) as P);
 
-    newSelectedItems[checked ? 'add' : 'delete'](id);
+  function onChange(checked: boolean, id: T) {
+    const newSelectedItemsArray = checked
+      ? [...selectedItemsArray, id]
+      : selectedItemsArray.filter(item => item !== id);
```
rpokorny, line 75: https://ramdajs.com/docs/#without would be helpful here and probably in a number of other places

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -19,8 +20,10 @@ const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-export default function NxTransferList<T extends string | number>(props: Props<T>) {
+export default function NxTransferList
```
rpokorny: There's a lot of uses of native array methods like `.map()` and `.filter` here.  I try to be consistent about using ramda's helpers instead of those native methods, _especially_ in files that also use a bunch of other ramda functions.

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +41,63 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
-    onChangeProp(newSelectedItems);
+  function onUnselectAll(idsToRemove: T[]) {
+    const newSelectedItems = selectedItemsArray.filter(item => !idsToRemove.includes(item));
+    handleOnChangeProp(newSelectedItems);
   }
 
-  function onUnselectAll(idsToRemove: Set<T>) {
-    const newSelectedItems = new Set<T>(selectedItems);
+  function onReorderItem(id: T, direction: 1 | -1) {
+    const itemIndex = selectedItemsArray.findIndex(item => item === id);
```
rpokorny, line 90: IMO a number of things would be cleaner if `TransferListItem` were aware of its index and communicated that index to `onReorderItem` rather than us having to look it up here

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -93,10 +122,16 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
-                           filterFn={filterFn} />
+                           filterFn={filterFn}
+                           allowReordering={allowReordering} />
     </div>
   );
 }
 
 NxTransferList.propTypes = propTypes;
+
+NxTransferList.defaultValues = {
```
rpokorny, line 135: When I try to google this ('react "defaultValues"') all I get are results for some third-party form hook.  Are you sure this is a thing? Even if so, it's not something that we use anywhere else.

lib/src/components/NxTransferList/TransferListHalf.tsx
```diff
@@ -27,11 +39,36 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
 
   return (
     <NxOverflowTooltip>
-      <label className="nx-transfer-list__item">
-        <NxFontAwesomeIcon icon={checked ? faTimesCircle : faPlusCircle} />
-        <input className="nx-transfer-list__checkbox" type="checkbox" checked={checked} onChange={onChange} />
-        <span>{displayName}</span>
-      </label>
+      <div className={classnames('nx-transfer-list__item',
+          { 'nx-transfer-list__item--with-reordering': !!showReorderingButtons })}>
```
rpokorny, line 43: Since this `classnames` call is decently complex and multiple lines long, I'd prefer it be pulled out to a variable

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -118,6 +127,43 @@ $-visible-item-count: 10;
   }
 }
 
+.nx-transfer-list__item--with-reordering {
+  .nx-transfer-list__select {
+    border-radius: var(--nx-border-radius);
```
rpokorny, line 132: Isn't this redundant?

lib/src/components/NxTransferList/TransferListHalf.tsx
```diff
@@ -27,11 +39,36 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
...
+        <label className="nx-transfer-list__select">
+          <NxFontAwesomeIcon icon={checked ? faTimesCircle : faPlusCircle} />
+          <input className="nx-transfer-list__checkbox" type="checkbox" checked={checked} onChange={onChange} />
+          <span>{displayName}</span>
+        </label>
+        { showReorderingButtons && (
+          <NxButtonBar className="nx-transfer-list__button-bar">
+            <NxButton variant="icon-only"
+                      className={classnames('nx-transfer-list__button',
```
rpokorny, line 52: Also pull out these `classnames` calls into local vars

lib/src/components/NxTransferList/TransferListHalf.tsx
```diff
@@ -27,11 +39,36 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
...
+                          'nx-transfer-list__button--move-up', {
+                            'disabled': isTopItem
```
rpokorny, line 54: I think we probably want to disable these via attr, not classname.  There's no reason for them to be focusable when disabled I don't think - the tooltips on these aren't helpful on the disabled ones.

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -85,23 +85,32 @@ $-visible-item-count: 10;
...
 
 .nx-transfer-list__item {
   @include nx-container-helpers.container-horizontal;
-  @include nx-text-helpers.truncate-ellipsis;
 
   background-color: var(--nx-swatch-white);
-  border: var(--nx-border-default);
   border-radius: var(--nx-border-radius);
+  display: flex;
+  justify-content: space-between;
```
rpokorny, line 99: What's this for?

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -85,23 +85,32 @@ $-visible-item-count: 10;
...
+  margin: var(--nx-spacing-2x) 0;
+  padding: 0;
```
rpokorny, line 101: Seems redundant

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -85,23 +85,32 @@ $-visible-item-count: 10;
...
   );
 
-
   overflow-y: auto;
   padding: var(--nx-spacing-6x) $-control-box-side-spacing;
   margin-top: var(--nx-spacing-6x);
 }
 
 .nx-transfer-list__item {
   @include nx-container-helpers.container-horizontal;
```
rpokorny, line 94: Probably no longer necessary

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -85,23 +85,32 @@ $-visible-item-count: 10;
...
+}
+
+.nx-transfer-list__select {
+  @include nx-container-helpers.container-horizontal;
+  @include nx-text-helpers.truncate-ellipsis;
+
+  border-radius: var(--nx-border-radius);
+  border: var(--nx-border-default);
   cursor: pointer;
   display: block;
```
rpokorny, line 111: Not needed

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -85,23 +85,32 @@ $-visible-item-count: 10;
...
-  margin: var(--nx-spacing-2x) 0;
   padding: var(--nx-spacing-2x);
+  width: 100%;
```
rpokorny, line 113: This isn't really accurate, I'd rather use `flex: auto`

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > If you use ramda's `chain` you can combine the mapping and filtering into one operation, and probably avoid the cast too

Thanks, this is updated here: d098925532af2c05b55be1e7b1bd141f26e0352a

entroform: > Repeatedly iterating `allItems` is probably slow. I'd first build a lookup map from id to item and then use that here

Thanks, this is updated here: d098925532af2c05b55be1e7b1bd141f26e0352a

gallery/src/components/NxTransferList/NxTransferListPage.tsx
entroform: > This description makes no mention of the buttons allowing the _user_ to reorder the items

Yup, updated the description here: d098925532af2c05b55be1e7b1bd141f26e0352a

gallery/src/components/NxTransferList/NxTransferListPage.tsx
entroform: > 

Removed this line because it's no longer necessary to define the type parameters it seems:
Updated in this commit: d098925532af2c05b55be1e7b1bd141f26e0352a

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -118,6 +127,43 @@ $-visible-item-count: 10;
...
+  padding: var(--nx-spacing-base) var(--nx-spacing-2x);
+}
+
+$-button-size: 32px;
+
+.nx-transfer-list__button {
+  height: $-button-size;
+  padding: 0;
+  margin: 0;
+  width: $-button-size;
```
rpokorny, line 155: These are supposed to be our usual small-style icon-only-buttons right? The right way to make that is with the `small-icon-btn` mixin in `_nx-button-helpers.scss`

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -118,6 +127,43 @@ $-visible-item-count: 10;
...
+
+  &.disabled {
+    &,
+    &:hover,
+    &:active {
```
rpokorny, line 160: Focus still shows up

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
```diff
@@ -80,7 +80,7 @@ describe('TransferListHalf', function() {
     expect(onMoveAll).toHaveBeenCalledTimes(1);
   });
 
-  it('passes onMoveAll a set containing every id if no filterValue is set', function() {
+  it('passes onMoveAll a set containing every id if no filterValue is an array', function() {
```
rpokorny, line 83: This change doesn't make sense.  "set" here was the verb - to _set_ a value.

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -118,6 +127,43 @@ $-visible-item-count: 10;
...
+  border-bottom-left-radius: 0;
+  border-top-left-radius: 0;
+  gap: var(--nx-spacing-base);
+  margin: 0;
+  padding: var(--nx-spacing-base) var(--nx-spacing-2x);
+}
+
+$-button-size: 32px;
+
+.nx-transfer-list__button {
```
rpokorny, line 151: IMO this class doesn't need to exist, you can just select `.nx-btn` within `.nx-transfer-list__button-bar`

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
```diff
@@ -273,7 +277,87 @@ describe('TransferListHalf', function() {
...
+      expect(secondItemMoveDownButton).toExist();
+      expect(secondItemMoveUpButton.hasClass('disabled')).toBeFalsy();
+      expect(secondItemMoveDownButton.hasClass('disabled')).toBeFalsy();
+
+      expect(thirdItem).toExist();
+      expect(thirdItemMoveUpButton).toExist();
+      expect(thirdItemMoveDownButton).toExist();
+      expect(thirdItemMoveUpButton.hasClass('disabled')).toBeFalsy();
+      expect(thirdItemMoveDownButton.hasClass('disabled')).toBeTruthy();
     });
```
rpokorny, line 326: Test that they have the right icons too

gallery/src/components/NxStatefulTransferList/NxStatefulTransferListExample.tsx
```diff
@@ -13,5 +13,5 @@ const items = map(i => ({ id: i, displayName: `Item ${i}` }), range(1, 101));
 export default function NxStatefulTransferListExample() {
   const [selectedItems, setSelectedItems] = useState<Set<number>>(new Set());
 
-  return <NxStatefulTransferList allItems={items} selectedItems={selectedItems} onChange={setSelectedItems} />;
+  return <NxStatefulTransferList<number> allItems={items} selectedItems={selectedItems} onChange={setSelectedItems} />;
```
entroform, line 16:
```suggestion
  return <NxStatefulTransferList allItems={items} selectedItems={selectedItems} onChange={setSelectedItems} />;
```

lib/src/components/NxSearchTransferList/NxSearchTransferList.tsx
entroform: > I wonder if it's worth it to collect the ids into a Set in a local variable first so that each lookup is O(1) not O(n)

Thanks, I think it's worth it: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > https://ramdajs.com/docs/#without would be helpful here and probably in a number of other places

Thanks! Updated here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > There's a lot of uses of native array methods like `.map()` and `.filter` here. I try to be consistent about using ramda's helpers instead of those native methods, _especially_ in files that also use a bunch of other ramda functions.

Refactored here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > IMO a number of things would be cleaner if `TransferListItem` were aware of its index and communicated that index to `onReorderItem` rather than us having to look it up here

Yeah I agree, I've updated it here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/NxTransferList.tsx
entroform: > When I try to google this ('react "defaultValues"') all I get are results for some third-party form hook. Are you sure this is a thing? Even if so, it's not something that we use anywhere else.

Removed it here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/TransferListHalf.tsx
entroform: > Since this `classnames` call is decently complex and multiple lines long, I'd prefer it be pulled out to a variable

Moved it to its own variable here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/TransferListHalf.tsx
entroform: > Also pull out these `classnames` calls into local vars

Moved it to its own variable here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
entroform: > This change doesn't make sense. "set" here was the verb - to _set_ a value.

ops! Fixed it here: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
entroform: > Test that they have the right icons too

Added tests for those in this commit: 774e9702b6a1123bac0705354974edd6087775a5

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
```diff
@@ -273,7 +277,112 @@ describe('TransferListHalf', function() {
...
+    moveDownButton.simulate('click');
+
+    expect(onReorderItem).toHaveBeenCalledWith(0, 1);
+
+    moveUpButton.simulate('click');
+
+    expect(onReorderItem).toHaveBeenCalledWith(1, -1);
+  });
+
+  it('sets the .nx-transfer-list__move-all icon to an x in a circle if isSelected, otherwise a plus in a circle',
```
entroform, line 364: Ops, need to update this!

lib/src/components/NxTransferList/TransferListHalf.tsx
rpokorny: > Moved it to its own variable here: 774e970

I meant move the `classnames` invocation itself.  Like `const classes = classnames('foo', className, { ... });`

That's how we typically do it

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
rpokorny: > Added tests for those in this commit: 774e970

That test's name seems like it's copied from somewhere else

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > These are supposed to be our usual small-style icon-only-buttons right? The right way to make that is with the `small-icon-btn` mixin in `_nx-button-helpers.scss`

I gave that mixin a try but the dimension is different from the spec and adds an additional few pixels to the item height. 🤔 

Screen Shot 2021-09-24 at 12 29 30 PM

Screen Shot 2021-09-24 at 12 32 38 PM

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
entroform: > Ops, need to update this!

Fixed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
entroform: > > Added tests for those in this commit: [774e970](https://github.com/sonatype/sonatype-react-shared-components/commit/774e9702b6a1123bac0705354974edd6087775a5)
> 
> That test's name seems like it's copied from somewhere else

Yup! That was fixed here: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > IMO this class doesn't need to exist, you can just select `.nx-btn` within `.nx-transfer-list__button-bar`

Sounds good! I removed it here: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > Focus still shows up

Fixed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > This isn't really accurate, I'd rather use `flex: auto`

Thanks! I updated it here: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > Not needed

Removed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > Probably no longer necessary

Removed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > Seems redundant

Yup! Removed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > What's this for?

Forgot to remove that: fixed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.tsx
entroform: > I think we probably want to disable these via attr, not classname. There's no reason for them to be focusable when disabled I don't think - the tooltips on these aren't helpful on the disabled ones.

That's a good point: I updated it here: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: > Isn't this redundant?

Yup! Removed: 8da7e29ba73263e6ce9f96e12d36ba9915305d0f

lib/src/components/NxTransferList/TransferListHalf.tsx
entroform: > > Moved it to its own variable here: [774e970](https://github.com/sonatype/sonatype-react-shared-components/commit/774e9702b6a1123bac0705354974edd6087775a5)
> 
> I meant move the `classnames` invocation itself. Like `const classes = classnames('foo', className, { ... });`
> 
> That's how we typically do it

Updated it here: a233e5ac88b0924c468d44ba1c0abc2a8d5a7083 👍

entroform:
> See #498 for a way to get rid of the second type parameter

Thanks! I updated the PR with your suggestion: cad35f0a67ae32b750ecf7df8b90a31bdb2e7143

gallery/src/components/NxTransferList/NxTransferListPage.tsx
```diff
@@ -160,7 +179,8 @@ const NxTransferListPage = () =>
               <NxTable.Cell></NxTable.Cell>
               <NxTable.Cell>
                 Callback for when the user selects or unselects one or more items. Receives the
-                new <NxCode>Set</NxCode> of selected ids as a parameter.
+                new <NxCode>Set</NxCode>, or <NxCode>Array</NxCode> if <NxCode>AllowReordering</NxCode> is true,
```
rpokorny, line 182:
```suggestion
                new <NxCode>Set</NxCode>, or <NxCode>Array</NxCode> if <NxCode>allowReordering</NxCode> is true,
```

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +40,68 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
+
+  const selectedItemsArray = Array.from(selectedItems);
+  const selectedItemsSet = selectedItems instanceof Set ? selectedItems as Set<T> : new Set(selectedItemsArray);
+
   const availableItemsCountFormatter = availableItemsCountFormatterProp || defaultAvailableItemsCountFormatter,
       selectedItemsCountFormatter = selectedItemsCountFormatterProp || defaultSelectedItemsCountFormatter;
 
-  const groupedItems = useMemo(() => groupBy(item => selectedItems.has(item.id) ? 'selected' : 'available', allItems),
-          [allItems, selectedItems]),
+  const allItemsIdToItemlookUp = groupBy(item => item.id.toString(), allItems);
```
rpokorny, line 53:
```suggestion
  const allItemsIdToItemLookUp = groupBy(item => item.id.toString(), allItems);
```

lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -38,39 +40,68 @@ export default function NxTransferList<T extends string | number>(props: Props<T
...
+
+  const groupedItems = useMemo(() =>
+        allowReordering
+          ? {
+            available: reject(item => selectedItemsSet.has(item.id), allItems),
+            selected: chain<T, DataItem<T>>(item => allItemsIdToItemlookUp[item.toString()], selectedItemsArray)
```
rpokorny, line 59: For me it's working without the type annotation

```suggestion
            selected: chain(item => allItemsIdToItemlookUp[item.toString()], selectedItemsArray)
```

rpokorny:
Re-ordering doesn't seem to work when a filter is active.  The buttons just do nothing.

lib/src/components/NxTransferList/__tests__/NxTransferList.test.tsx
```diff
@@ -22,7 +22,9 @@ describe('NxTransferList', function() {
         onSelectedItemsFilterChange: () => {},
         onChange: () => {}
       },
-      getShallow = getShallowComponent<Props<string | number>>(NxTransferList, minimalProps);
+      // eslint-disable-next-line no-spaced-func
+      getShallow = getShallowComponent<Props<string | number, Set<number | string> | (string | number)[]>>
```
rpokorny, line 26: If I'm not mistaken this change is no longer valid

lib/src/components/NxTransferList/__tests__/NxTransferList.test.tsx
```diff
@@ -269,4 +271,101 @@ describe('NxTransferList', function() {
...
+      const items = [
+        { id: 2, displayName: 'bar' },
+        { id: 4, displayName: 'qwerty' },
+        { id: 3, displayName: 'baz' }
+      ];
+
+      expect(secondHalf).toHaveProp('items', items);
+    });
+
+    it('fires onChange with a new Array with the clicked item added at the end'
```
rpokorny, line 326:
```suggestion
    it('fires onChange with a new Array with the clicked item added at the end '
```

lib/src/components/NxTransferList/__tests__/NxTransferList.test.tsx
```diff
@@ -269,4 +271,101 @@ describe('NxTransferList', function() {
       expect(threeSelected.childAt(1)).toHaveProp('footerContent', 'To 3 and beyond!');
     });
   });
+
+  it('throw an error when passing a Set to selectedItems when allowReordering is true', function() {
+    expect(() => getShallow({ allowReordering: true }))
```
rpokorny, line 276: This should no longer compile.  If it does, the props changes didn't work as expected

lib/src/components/NxTransferList/types.ts
```diff
@@ -14,53 +14,69 @@ export type FilterFn<T extends string | number = string> = (d: DataItem<T>[]) =>
...
-  allItems: PropTypes.arrayOf(PropTypes.shape({
-    id: PropTypes.oneOfType([
-      PropTypes.string.isRequired,
-      PropTypes.number.isRequired
-    ]).isRequired,
-    displayName: PropTypes.string.isRequired
-  }).isRequired).isRequired,
-  selectedItems: PropTypes.instanceOf<Set<string | number>>(Set).isRequired,
+// allItems, selectedItems, onAvailableItemsFilterChange, onSelectedItemsFilterChange
+// are excluded in the propTypes due to clash with parametric Props;
```
rpokorny, line 77: We should check whether this is still the case now that the types have been redone

lib/src/components/NxTransferList/TransferListHalf.scss
rpokorny: > I gave that mixin a try but the dimension is different from the spec and adds an additional few pixels to the item height.

I doubt it was deliberate that these buttons be smaller than all of our other small icon-only buttons.  You should check with @alexalyons126 about that.  For reference, generally our small icon-only buttons are 34x34 including the border.  Looks like this mockup has them at 32x32.

entroform:
> Re-ordering doesn't seem to work when a filter is active. The buttons just do nothing.

Thanks for finding this!
I've got a working implementation in this commit: c1f9f1f29542985e93f7345f1af4eaf152f3e952
I'm going to take some time to clean it up and investigate the props change issue.

lib/src/components/NxTransferList/types.ts
entroform: > We should check whether this is still the case now that the types have been redone

I was able to bring 2 back without errors in this commit:
ced6b8e23042c0886a1255efc67fb2f75f92812f

lib/src/components/NxTransferList/__tests__/NxTransferList.test.tsx
entroform: > This should no longer compile. If it does, the props changes didn't work as expected

Sadly this did compile 😭

lib/src/components/NxTransferList/__tests__/NxTransferList.test.tsx
entroform: > If I'm not mistaken this change is no longer valid

Yup! This was fixed here: c1f9f1f29542985e93f7345f1af4eaf152f3e952

lib/src/components/NxTransferList/__tests__/NxTransferList.test.tsx
rpokorny: > Sadly this did compile sob

Actually that might just be a problem with `getShallow`.  If I change `NxTransferListWithReorderingExample` to use a Set instead of an Array, it fails appropriately

rpokorny:
> Thanks for finding this!
I've got a working implementation in this commit: c1f9f1f
Further cleaned up: 5e5713b

I would say that part of fixing this requires having a discussion with design about how it should behave in this case, because it's not at all clear to me what would be most expected.  What it does currently is it swaps the reordered item with its adjacent visible item. 

For an example, observe the following screenshots.  First, I start with a few selected items and no filter:
image

Then, I filter on "2" which hides all but two of the selected items:
image

Then, I reorder the first item downwards:
image

Then, I clear the filter and observe that the item I moved down hasn't truly just moved down one spot, but has moved down three spots _and the other item that had been visible when filtering has moved **up** three spots_ (they swapped places):
image

entroform:
> > Thanks for finding this!
> > I've got a working implementation in this commit: [c1f9f1f](https://github.com/sonatype/sonatype-react-shared-components/commit/c1f9f1f29542985e93f7345f1af4eaf152f3e952)
> > Further cleaned up: [5e5713b](https://github.com/sonatype/sonatype-react-shared-components/commit/5e5713b92df85bb18cc4564c34d3f3de627c58a2)
> 
> I would say that part of fixing this requires having a discussion with design about how it should behave in this case, because it's not at all clear to me what would be most expected. What it does currently is it swaps the reordered item with its adjacent visible item.
> 
> For an example, observe the following screenshots. First, I start with a few selected items and no filter: image
> 
> Then, I filter on "2" which hides all but two of the selected items: image
> 
> Then, I reorder the first item downwards: image
> 
> Then, I clear the filter and observe that the item I moved down hasn't truly just moved down one spot, but has moved down three spots _and the other item that had been visible when filtering has moved **up** three spots_ (they swapped places): image

Yeah, thanks for pointing that out. I assumed (which I shouldn't have) and thought that was the natural expected behaviour. I will bring it up with design in the #react-components channel.

entroform:
> > Thanks for finding this!
> > I've got a working implementation in this commit: [c1f9f1f](https://github.com/sonatype/sonatype-react-shared-components/commit/c1f9f1f29542985e93f7345f1af4eaf152f3e952)
> > Further cleaned up: [5e5713b](https://github.com/sonatype/sonatype-react-shared-components/commit/5e5713b92df85bb18cc4564c34d3f3de627c58a2)
> 
> I would say that part of fixing this requires having a discussion with design about how it should behave in this case, because it's not at all clear to me what would be most expected. What it does currently is it swaps the reordered item with its adjacent visible item.
> 
> For an example, observe the following screenshots. First, I start with a few selected items and no filter: image
> 
> Then, I filter on "2" which hides all but two of the selected items: image
> 
> Then, I reorder the first item downwards: image
> 
> Then, I clear the filter and observe that the item I moved down hasn't truly just moved down one spot, but has moved down three spots _and the other item that had been visible when filtering has moved **up** three spots_ (they swapped places): image

Updated the PR so the move up and down buttons are now disabled when there is filter value:
306a01e484ebaabee1933fae8e81d1b9b158c3a5

rpokorny:
> Updated the PR so the move up and down buttons are now disabled when there is filter value:
306a01e

Should the buttons have a tooltip to explain why they're disabled in this case?

entroform:
> Should the buttons have a tooltip to explain why they're disabled in this case?

I think that's a good idea: I've updated it in this commit:
2fc88b083dc2fb00adaa0466d58bddb9a2d2b106

lib/src/components/NxTransferList/TransferListHalf.tsx
```diff
@@ -48,6 +48,11 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
       }
   );
 
+  const moveUpButtonTitle = isTopItem
+    ? null : isFilteredItem ? 'Reordering is disabled when filtered' : 'Move Up';
```
rpokorny, line 52: IMO nested ternaries are only readable when split over multiple lines such that each line indicates a condition and the result when that condition is true, followed at the end of all the lines by the base case.  Like this:

```
  const moveUpButtonTitle = isTopItem ? null : 
      isFilteredItem ? 'Reordering is disabled when filtered' : 
      'Move Up';
```

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
```diff
@@ -273,7 +277,149 @@ describe('TransferListHalf', function() {
...
+    expect(moveDownButton.prop('disabled')).toBe(true);
+
+    moveDownButton.simulate('click');
+
+    expect(onReorderItem).not.toHaveBeenCalled();
+
+    moveUpButton.simulate('click');
+
+    expect(onReorderItem).not.toHaveBeenCalled();
+  });
```
rpokorny, line 424: Needs a test for the title logic

rpokorny:
Something's wrong with the tooltips; they don't disappear after the mouse leaves.
image

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
entroform: > Needs a test for the title logic

Added test here: 5759f98d182e17a0e9d67472f4f8f68b2f8af0e5

entroform:
> Something's wrong with the tooltips; they don't disappear after the mouse leaves. image

This issue didn't show up on Firefox strangely:
Fixed: 5759f98d182e17a0e9d67472f4f8f68b2f8af0e5

I tried wrapping NxButton in span/div + NxTooltip but the issue is still there. The only way I got it to work is using the 'disabled' class. Let me know if you know a better way around it 😭.

rpokorny:
> I tried wrapping NxButton in span/div + NxTooltip but the issue is still there. The only way I got it to work is using the 'disabled' class. Let me know if you know a better way around it sob.

How about like this: https://github.com/sonatype/sonatype-react-shared-components/pull/512

entroform:
> > I tried wrapping NxButton in span/div + NxTooltip but the issue is still there. The only way I got it to work is using the 'disabled' class. Let me know if you know a better way around it sob.
> 
> How about like this: #512

Ooo thanks, that looks great. Grouping both buttons with tooltip also makes a lot of sense. Thanks for your help!

lib/src/components/NxTransferList/TransferListHalf.scss
```diff
@@ -118,6 +123,45 @@ $-visible-item-count: 10;
...
+  $-button-size: 32px;
+
+  .nx-btn {
+    height: $-button-size;
+    margin: 0;
+    padding: 0;
+    width: $-button-size;
+
+    &[disabled],
+    &.disabled {
```
rpokorny, line 152: I don't think this is needed anymore

rpokorny (approved):

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
rpokorny: > Added test here: 5759f98

:+1: that'll need to be updated though as the approach has changed slightly

rpokorny:
Oh whoops, I rescind my approval as I've just noticed that an important thread is still unresolved: https://github.com/sonatype/sonatype-react-shared-components/pull/488#discussion_r714987126

lib/src/components/NxTransferList/TransferListHalf.scss
entroform: Yup! Updated here along with the test: 885f3cb8f599364a04751640783fbcad7193f3df

lib/src/components/NxTransferList/__tests__/TransferListHalf.test.tsx
entroform: > > Added test here: [5759f98](https://github.com/sonatype/sonatype-react-shared-components/commit/5759f98d182e17a0e9d67472f4f8f68b2f8af0e5)
> 
> 👍 that'll need to be updated though as the approach has changed slightly

Yup test updated: 885f3cb8f599364a04751640783fbcad7193f3df 😄

entroform:
> Oh whoops, I rescind my approval as I've just noticed that an important thread is still unresolved: [#488 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/488#discussion_r714987126)

Thanks! Just confirming that will be updated in this ticket: https://issues.sonatype.org/browse/RSC-735

PR #491, opened 2021-09-20
Change to use NxList component RSC-622

avnvbhatta:
https://issues.sonatype.org/browse/RSC-622

Couple of changes (Applitools reflects these as well)
**- `NxThreatIndicator` Gallery Docs**: found [here.](https://gallery.sonatype.dev/#/pages/NxThreatIndicator)

- Previously, the examples under "Threat Indicators in nx-list" rendered all list items as clickable rows, which is incorrect, as each list item explicitly define how the threat indicator is displayed, for eg. "Indicator on simple list row" should not be clickable, "Indicator on row with actions" should not be clickable, but should have actions etc. 
- The list item "Indicator on row with subtext and actions" is missing subtext, so I have added it in the new gallery doc.

Before:
image
After:
image


entroform (approved):
LGTM! 🚀

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "8.7.0",
+  "version": "8.8.0",
```
rpokorny, line 3: Should definitely be a patch version

rpokorny:
My understanding of this ticket was that it was only about the documentation (ie the `*Page.tsx` files), not the examples

gallery/package.json
avnvbhatta: > Should definitely be a patch version

Updated to correct version.
Check b1871f57592f88bbf1d87580c996934d74a97910

avnvbhatta:
> My understanding of this ticket was that it was only about the documentation (ie the `*Page.tsx` files), not the examples

As per discussion during WW, we will be changing all of the examples to use `NxList` besides one example, "NxViewportSizedAdjacentExample"

gallery/src/guidelines/FontSize/FontSizePage.tsx
```diff
@@ -234,40 +234,40 @@ const FontSizePage = () =>
       <NxP>
         Here is an incomplete list of additional nuances to consider when dealing with inline formatting
       </NxP>
-      <ul className="nx-list nx-list--bulleted">
-        <li className="nx-list__item">
+      <NxList bulleted>
+        <NxList.Item>
           Inline elements do not respond to CSS width, height, margin-top, or margin-bottom properties. Their
```
rpokorny, line 239: Looks like the `NxList.Text` elements are missing here; you should go ahead and add them

gallery/src/guidelines/FormValidation/FormValidationPage.tsx
```diff
@@ -20,27 +20,27 @@ const FormValidationPage = () =>
         This page demonstrates the typical overall approach to communicating form validation matters to the user.
         There are several things to note here:
       </NxP>
-      <ul className="nx-list nx-list--bulleted">
-        <li className="nx-list__item">
+      <NxList bulleted>
+        <NxList.Item>
           Fields are not marked invalid while they are pristine. This is handled by
```
rpokorny, line 25: Missing `NxList.Text`

gallery/src/jsUtilPages/PolicyThreatLevelUtils/PolicyThreatLevelUtilsPage.tsx
```diff
@@ -22,26 +22,26 @@ const PolicyThreatLevelUtilsPage = () =>
         objects and TypeScript types that provide code definitions and mappings between these two concepts.
       </NxP>
 
-      <ul className="nx-list nx-list--bulleted">
-        <li className="nx-list__item">
+      <NxList bulleted>
+        <NxList.Item>
           <strong>Policy Threat Level:</strong> An integer between 0 and 10, inclusive, that describes
```
rpokorny, line 27: Missing `NxList.Text`

gallery/src/pages/AdditionalResources.tsx
```diff
@@ -19,53 +19,53 @@ const AdditionalResources = () =>
       <NxTile.SubsectionHeader>
         <NxH3>Classes / Tutorials</NxH3>
       </NxTile.SubsectionHeader>
-      <ul className="nx-list nx-list--bulleted">
-        <li className="nx-list__item">
+      <NxList bulleted>
+        <NxList.Item>
           <NxTextLink external href="https://reactjs.org/tutorial/tutorial.html">
```
rpokorny, line 24: Missing `NxList.Text` again.  Or do we want to change the documentation of bulleted lists to officially make `NxList.Text` optional in that case? @glenhunter

glenhunter: > Missing `NxList.Text` again. Or do we want to change the documentation of bulleted lists to officially make `NxList.Text` optional in that case? @glenhunter

I think that we specified the use of `NxList.Text` to avoid bare text nodes but it's hard to see that being an issue in a bulleted list.

rpokorny: > I think that we specified the use of NxList.Text to avoid bare text nodes but it's hard to see that being an issue in a bulleted list.

While true, if we're going to officially support it we need to have examples and visual tests of it.

glenhunter: > > I think that we specified the use of NxList.Text to avoid bare text nodes but it's hard to see that being an issue in a bulleted list.
> 
> While true, if we're going to officially support it we need to have examples and visual tests of it.

Should be resolved here: https://github.com/sonatype/sonatype-react-shared-components/pull/491/commits/49e847b39ad47fbb73a4d350ad12678101765537

gallery/src/jsUtilPages/PolicyThreatLevelUtils/PolicyThreatLevelUtilsPage.tsx
glenhunter: > Missing `NxList.Text`

Resolved: https://github.com/sonatype/sonatype-react-shared-components/pull/491/commits/49e847b39ad47fbb73a4d350ad12678101765537

gallery/src/guidelines/FormValidation/FormValidationPage.tsx
glenhunter: > Missing `NxList.Text`

Resolved: https://github.com/sonatype/sonatype-react-shared-components/pull/491/commits/49e847b39ad47fbb73a4d350ad12678101765537

gallery/src/guidelines/FontSize/FontSizePage.tsx
glenhunter: > Looks like the `NxList.Text` elements are missing here; you should go ahead and add them

Resolved: https://github.com/sonatype/sonatype-react-shared-components/pull/491/commits/49e847b39ad47fbb73a4d350ad12678101765537

rpokorny (approved):

PR #494, opened 2021-09-22
Base NxNumberInput off of NxTextInput - RSC-252

rpokorny:


lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -114,6 +114,11 @@ const NxTextInput = forwardRef<HTMLDivElement, Props>(
     }
 );
 
-NxTextInput.propTypes = propTypes;
+const PublicNxTextInput = forwardRef<HTMLDivElement, PublicProps>(function(props, ref) {
```
rpokorny, line 117: This part really should have been done previously, since NxTextInput already had separate PublicProps

PR #497, opened 2021-09-22
Added prop for input attributes - RSC-652

entroform:
https://issues.sonatype.org/browse/RSC-652

### DIFF
--- a/lib/src/components/NxCheckbox/NxCheckbox.tsx
+++ b/lib/src/components/NxCheckbox/NxCheckbox.tsx
@@ -34,6 +34,7 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
             checkboxId,
             overflowTooltip,
             children,
+            checkboxAttributes,
             ...otherProps
           } = props,
           labelClasses = classnames('nx-radio-checkbox', 'nx-checkbox', className, {
@@ -43,15 +44,21 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
           }),
           content = children && <span className="nx-radio-checkbox__content nx-checkbox__content">{children}</span>;
 
+      const {
+        className: checkboxClassName,
+        ...otherCheckboxAttrs
+      } = checkboxAttributes;
+
       return (
         <label { ...otherProps } ref={ref} className={labelClasses}>
           <input type="checkbox"
                  id={checkboxId || undefined}
-                 className="nx-radio-checkbox__input nx-checkbox__input"
+                 className={classnames('nx-radio-checkbox__input', 'nx-checkbox__input', checkboxClassName)}
                  disabled={!!disabled}
                  checked={isChecked}
                  readOnly={!onChange}
-                 onChange={onChange || undefined}/>
+                 onChange={onChange || undefined}
+                 { ...otherCheckboxAttrs } />
           <span className="nx-radio-checkbox__control nx-checkbox__box">
             {/* Put a space in the box if not checked,
               * in order to provide a consistent vertical-align baseline
@@ -67,5 +74,9 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
 );
 
 NxCheckbox.propTypes = propTypes;
+
+NxCheckbox.defaultProps = {
+  checkboxAttributes: {}
+};
 
 export default NxCheckbox;
--- a/lib/src/components/NxCheckbox/types.ts
+++ b/lib/src/components/NxCheckbox/types.ts
@@ -4,8 +4,12 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import { LabelHTMLAttributes } from 'react';
+import { InputHTMLAttributes, LabelHTMLAttributes, Validator } from 'react';
 import * as PropTypes from 'prop-types';
+
+type CheckboxAttributesProp =
+  Omit<InputHTMLAttributes<HTMLInputElement>,
+  'id' | 'disabled' | 'checked' | 'readOnly' | 'onChange'>;
 
 export type Props = Omit<LabelHTMLAttributes<HTMLLabelElement>, 'onChange'> & {
   checkboxId?: string | null;
@@ -13,15 +17,16 @@ export type Props = Omit<LabelHTMLAttributes<HTMLLabelElement>, 'onChange'> & {
   isChecked: boolean;
   disabled?: boolean | null;
   overflowTooltip?: boolean | null;
+  checkboxAttributes: CheckboxAttributesProp;
 };
 
-// In a strictly typescript environment, PropTypes are mostly redundant.  However, they still provide safety when this
+// In a strictly typescript environment, PropTypes are mostly redundant. However, they still provide safety when this
 // project is consumed by javascript projects
 export const propTypes: PropTypes.ValidationMap<Props> = {
   checkboxId: PropTypes.string,
   onChange: PropTypes.func,
   isChecked: PropTypes.bool.isRequired,
   disabled: PropTypes.bool,
-  overflowTooltip: PropTypes.bool
+  overflowTooltip: PropTypes.bool,
+  checkboxAttributes: PropTypes.object as Validator<CheckboxAttributesProp>
 };
-
--- a/lib/src/components/NxRadio/NxRadio.tsx
+++ b/lib/src/components/NxRadio/NxRadio.tsx
@@ -24,6 +24,7 @@ const NxRadio = forwardRef<HTMLLabelElement, Props>(
             children,
             radioId,
             overflowTooltip,
+            radioAttributes,
             ...otherProps
           } = props,
           labelClasses = classnames('nx-radio-checkbox', 'nx-radio', className, {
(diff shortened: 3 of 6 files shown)


gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
```diff
@@ -94,6 +94,21 @@ const NxCheckboxPage = () =>
...
           </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell>checkboxAttributes</NxTable.Cell>
+            <NxTable.Cell>
+              <NxTextLink external href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox">
+                HTML Input Checkbox Attributes
+              </NxTextLink>
+            </NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>
```
rpokorny, line 105: I think the API around `checkboxId` is now confusing - why have to pass in the id that way instead of as part of this prop? I'd suggest marking `checkboxId` as deprecated and making it work with either

gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
```diff
@@ -94,6 +94,21 @@ const NxCheckboxPage = () =>
...
+              NxCheckbox also supports any html attribute that's normally supported by the
+              <NxCode>input</NxCode> element except for <NxCode>id</NxCode>, <NxCode>disabled</NxCode>,
```
rpokorny, line 107: The wording here makes it sound like `NxCheckbox` doesn't support externally setting `disabled`.  It does, just not through this prop.  I'd rephrase this description to be less copy/pasted from our usual native attr docs and clearer for the situation at hand.

lib/src/components/NxCheckbox/NxCheckbox.tsx
```diff
@@ -43,15 +45,26 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
           }),
           content = children && <span className="nx-radio-checkbox__content nx-checkbox__content">{children}</span>;
 
+      const {
+        className: checkboxClassName,
+        ...unfilteredChecboxAttributes
```
rpokorny, line 50:
```suggestion
        ...unfilteredCheckboxAttributes
```

lib/src/components/NxCheckbox/__tests__/NxCheckbox.test.tsx
```diff
@@ -144,4 +144,11 @@ describe('NxCheckbox', function() {
     expect(component).toHaveProp('id', 'foo');
     expect(component).toHaveProp('htmlFor', 'baz');
   });
+
+  it('pass attributes into checkbox input element', function() {
```
rpokorny, line 148: "it pass attributes into checkbox input element" is not grammatically correct

lib/src/components/NxCheckbox/__tests__/NxCheckbox.test.tsx
```diff
@@ -144,4 +144,11 @@ describe('NxCheckbox', function() {
...
+    expect(getShallowComponent({ checkboxAttributes: { name: 'Garfield'} })
+        .find('input')).toHaveProp('name', 'Garfield');
+    expect(getShallowComponent({ checkboxAttributes: { className: 'cat', name: 'Garfield'} })
+        .find('input')).toHaveClassName('cat');
+  });
```
rpokorny, line 153: I'd test what happens when something like `disabled` or the id is specified in multiple ways: in the top-level props and also in `checkboxAttributes`.

rpokorny:
I feel like it would be both clearer and more consistent if these new props were all called `inputAttributes` rather than `checkboxAttributes` or `radioAttributes`

gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
entroform: > I think the API around `checkboxId` is now confusing - why have to pass in the id that way instead of as part of this prop? I'd suggest marking `checkboxId` as deprecated and making it work with either

I've updated it in this commit:
4c796beb23b05878976a76dfdcf72610774430aa

gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
entroform: > The wording here makes it sound like `NxCheckbox` doesn't support externally setting `disabled`. It does, just not through this prop. I'd rephrase this description to be less copy/pasted from our usual native attr docs and clearer for the situation at hand.

Thanks I tried to rephrase it here: 4c796beb23b05878976a76dfdcf72610774430aa

lib/src/components/NxCheckbox/__tests__/NxCheckbox.test.tsx
entroform: > "it pass attributes into checkbox input element" is not grammatically correct

Fixed: 4c796beb23b05878976a76dfdcf72610774430aa

lib/src/components/NxCheckbox/__tests__/NxCheckbox.test.tsx
entroform: > I'd test what happens when something like `disabled` or the id is specified in multiple ways: in the top-level props and also in `checkboxAttributes`.

Updated the tests here: 4c796beb23b05878976a76dfdcf72610774430aa

entroform:
> I feel like it would be both clearer and more consistent if these new props were all called `inputAttributes` rather than `checkboxAttributes` or `radioAttributes`

Thanks! Yeah I thought of using `inputAttributes` as well, but then went with more specific prop name when I saw `checkboxId` and `radioId`.

I updated it to use `inputAttributes` in this commit: 4c796beb23b05878976a76dfdcf72610774430aa

lib/src/components/NxCheckbox/__tests__/NxCheckbox.test.tsx
```diff
@@ -145,10 +146,25 @@ describe('NxCheckbox', function() {
...
+        disabled: false,
+        className: 'checkbox-classname',
+        checked: false
+      }
+    });
+
+    expect(component.find('input')).toHaveProp('id', 'garfield');
+    expect(component.find('input')).toHaveProp('name', 'garfield');
+    expect(component.find('input')).toHaveProp('disabled', true);
+    expect(component.find('input')).toHaveClassName('checkbox-classname');
```
rpokorny, line 167: Wouldn't hurt to also check that the `input` doesn't have `label-classname`

lib/src/components/NxCheckbox/types.ts
```diff
@@ -4,24 +4,34 @@
...
   checkboxId?: string | null;
   onChange?: (() => void) | null;
   isChecked: boolean;
   disabled?: boolean | null;
   overflowTooltip?: boolean | null;
+  inputAttributes?: InputAttributesProp;
+};
+
+// For testing
+export type PropsWithAnyInputAttributes = Props & {
```
rpokorny, line 24: I don't think you need this; you can get away with a cast in the test like this:
```
    const component = getShallowComponent({
      checkboxId: 'not-garfield',
      disabled: true,
      isChecked: true,
      className: 'label-classname',
      inputAttributes: {
        id: 'garfield',
        name: 'garfield',
        disabled: false,
        className: 'checkbox-classname',
        checked: false
      } as Props['inputAttributes']
    });
```

lib/src/components/NxCheckbox/__tests__/NxCheckbox.test.tsx
entroform: > Wouldn't hurt to also check that the `input` doesn't have `label-classname`

Agreed! Added test in this commit: d9cf5af435f57153f63578b8eec1a151a520a8cc

gallery/src/components/NxCheckbox/NxCheckboxPage.tsx
```diff
@@ -34,7 +34,16 @@ const NxCheckboxPage = () =>
             <NxTable.Cell>checkboxId</NxTable.Cell>
             <NxTable.Cell>string</NxTable.Cell>
             <NxTable.Cell>No</NxTable.Cell>
-            <NxTable.Cell>An id to identify the checkbox</NxTable.Cell>
+            <NxTable.Cell>
+              An id to identify the checkbox.
+
+              <NxInfoAlert>
```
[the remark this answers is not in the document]
entroform: > We usually use warning-level alerts for deprecations

Ah, I was using NxAccordion as an example.
Updated to warning-level alerts here: d9cf5af435f57153f63578b8eec1a151a520a8cc

lib/src/components/NxCheckbox/types.ts
entroform: > I don't think you need this; you can get away with a cast in the test like this:
> 
> ```
>     const component = getShallowComponent({
>       checkboxId: 'not-garfield',
>       disabled: true,
>       isChecked: true,
>       className: 'label-classname',
>       inputAttributes: {
>         id: 'garfield',
>         name: 'garfield',
>         disabled: false,
>         className: 'checkbox-classname',
>         checked: false
>       } as Props['inputAttributes']
>     });
> ```

Fixed: d9cf5af435f57153f63578b8eec1a151a520a8cc

rpokorny (approved):

PR #498, opened 2021-09-23
Avoid second type parameter - RSC-649

rpokorny:
By making two separate interfaces, one for when `allowReordering` is true and one for when it isn't, this avoids the need for the second type parameter

### DIFF
--- a/lib/src/components/NxTransferList/NxTransferList.tsx
+++ b/lib/src/components/NxTransferList/NxTransferList.tsx
@@ -20,8 +20,7 @@ const defaultItemsCountFormatter = (kind: string) => (n: number) => `${n} item${
     defaultAvailableItemsCountFormatter = defaultItemsCountFormatter('available'),
     defaultSelectedItemsCountFormatter = defaultItemsCountFormatter('transferred');
 
-export default function NxTransferList
-<T extends string | number = string, P extends Set<T> | T[] = Set<T>>(props: Props<T, P>) {
+export default function NxTransferList <T extends string | number = string>(props: Props<T>) {
   const {
     allowReordering,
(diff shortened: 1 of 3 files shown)


lib/src/components/NxTransferList/NxTransferList.tsx
```diff
@@ -65,7 +64,14 @@ export default function NxTransferList
...
 
-  const handleOnChangeProp = (array: T[]) => onChangeProp((allowReordering ? array : new Set(array)) as P);
+  const handleOnChangeProp = (array: T[]) => {
+    if (allowReordering) {
+      (onChangeProp as (newSelected: T[]) => void)(array);
+    }
+    else {
+      (onChangeProp as (newSelected: Set<T>) => void)(new Set(array));
+    }
+  };
```
rpokorny, line 74: I had hoped this would get rid of the cast here but instead it made it worse.  It seems typescript is not able to fully track the ramifications of the `if (allowReordering)` guard here.  I'm not sure if that's because we're nested pretty far down into the component here, or if it's something about the way react's props work.

entroform: > I had hoped this would get rid of the cast here but instead it made it worse. It seems typescript is not able to fully track the ramifications of the `if (allowReordering)` guard here. I'm not sure if that's because we're nested pretty far down into the component here, or if it's something about the way react's props work.

Hmm yeah :(. I tried something similar, but got stuck at `Argument of type 'Set<T> | T[]' is not assignable to parameter of type 'Set<T> & T[]'`

rpokorny:
These changes were manually integrated into the target branch.

PR #500, opened 2021-09-27
Redesign nx-global-header - RSC-714

avnvbhatta:
https://issues.sonatype.org/browse/RSC-714


lib/src/base-styles/_nx-global-header.scss
```diff
@@ -9,11 +9,12 @@
 
 .nx-global-header {
   align-items: center;
-  border-bottom: var(--nx-border-default);
+  background-color: #FFFFFF;
   display: flex;
   grid-area: header; // see nx-page's grid-template
-  min-height: 32px;
-  padding: var(--nx-spacing-6x);
+  margin-bottom: var(--nx-spacing-6x);
```
rpokorny, line 15: This spacing is already implemented in the main content area and sidebar.  The problem with implementing it here like this is the scrollbars.
image

lib/src/base-styles/_nx-global-header.scss
```diff
@@ -9,11 +9,12 @@
...
+  min-height: 42px;
+  padding: 0 var(--nx-spacing-6x);
 
   &:empty {
```
rpokorny, line 19: This case is missing in the mockup, but I recall @alexalyons126 saying it should just be an empty white rectangle rather than disappearing like it used to.  That should be added to the mockups

lib/src/base-styles/_nx-global-header.scss
```diff
@@ -22,6 +23,7 @@
 
   .nx-back-button {
```
rpokorny, line 24: Back button text currently wraps
image

rpokorny:
I don't see a mockup for when there are no action buttons.  This would be needed to know whether or not the vertical divider should appear in that case. @alexalyons126

lib/src/base-styles/_nx-global-header.scss
```diff
@@ -9,11 +9,12 @@
...
 .nx-global-header {
   align-items: center;
-  border-bottom: var(--nx-border-default);
+  background-color: #FFFFFF;
   display: flex;
   grid-area: header; // see nx-page's grid-template
-  min-height: 32px;
-  padding: var(--nx-spacing-6x);
+  margin-bottom: var(--nx-spacing-6x);
+  min-height: 42px;
```
rpokorny, line 16:
```suggestion
  height: 42px;
```

lib/src/base-styles/_nx-global-header.scss
```diff
@@ -9,11 +9,12 @@
 
 .nx-global-header {
   align-items: center;
-  border-bottom: var(--nx-border-default);
+  background-color: #FFFFFF;
```
rpokorny, line 12:
```suggestion
  background-color: var(--nx-swatch-white);
```

lib/src/base-styles/_nx-global-header.scss
avnvbhatta: > This spacing is already implemented in the main content area and sidebar. The problem with implementing it here like this is the scrollbars. image

The extra spacing has been removed.
Check 65ed208c842eb774b0c4687a8f86793aa07abae4

lib/src/base-styles/_nx-global-header.scss
avnvbhatta: > This case is missing in the mockup, but I recall @alexalyons126 saying it should just be an empty white rectangle rather than disappearing like it used to. That should be added to the mockups

When empty, the header is simply a white rectangle.
Check 79ce920b0c66bb1e6aa69204af90740a1c3044ca

lib/src/base-styles/_nx-global-header.scss
avnvbhatta: > Back button text currently wraps image

Added wrapping to back button text.
Check dca4a3867315315becea917cf424910123da4ee7

lib/src/base-styles/_nx-global-header.scss
```diff
@@ -6,22 +6,21 @@
...
-    display: none;
-  }
+  height: 42px;
+  padding: 0 var(--nx-spacing-6x);
 
   .nx-back-button {
+    @include nx-text-helpers.truncate-ellipsis;
     margin: 0;
+    order: -1; // ensures back button is always on the left
+    padding-right: var(--nx-spacing-8x);
```
glenhunter, line 23: Minor but IMO this should be margin.

glenhunter (approved):

lib/src/base-styles/_nx-global-header.scss
avnvbhatta: > Minor but this should be margin.

Changed here: 1b137a86439f5b0738a5ac2d0339fa4e8949c486

rpokorny:
> Also added a order: -1; property to the back button to ensure that it always stays on the left hand side of the header, regardless of the order of NxBackButton and .nx-global-header__actions.
Check update here: 8e5e0ff

The problem with adding gratuitous flexibility like that is that now you need a visual test for it 😁.  I'd rather just require the elements to be in the right order in the DOM; we do that in plenty of other places.

rpokorny (approved):
+1 except for my recommendation to remove the `order: -1` thing.

avnvbhatta:
> > Also added a order: -1; property to the back button to ensure that it always stays on the left hand side of the header, regardless of the order of NxBackButton and .nx-global-header__actions.
> > Check update here: [8e5e0ff](https://github.com/sonatype/sonatype-react-shared-components/commit/8e5e0ffa7f77b872058d631026c482e6c2c9504f)
> 
> The problem with adding gratuitous flexibility like that is that now you need a visual test for it 😁. I'd rather just require the elements to be in the right order in the DOM; we do that in plenty of other places.

I guess it's simple enough that users will know right away if the order is incorrect. Saves from requiring a visual test as well.
Changed here: 7a663e76ebb925a5ff62e5fd2f9c7d3bd6a72a45

PR #501, opened 2021-09-27
input type=number documentation RSC-252

glenhunter:
https://issues.sonatype.org/browse/RSC-252

Per discussion a comment was left on the story and the `NxTextInput` docs were updated with information about the number input and its issues with browser support and React.

This PR shows the documentation additions to `NxTextInput`


gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -192,6 +192,33 @@ const NxTextInputPage = () =>
...
+        </NxAccordion.Header>
+        <NxP>
+          <NxCode>NxTextInput</NxCode> using <NxCode>type="number"</NxCode> has been an oft requested addition to the
+          RSC but there are problems with its implementation at the browser level that make its use within RSC
+          too complicated to justify further development at this time.
+        </NxP>
+        <NxP>
+          In Chrome it works mostly as expected, however Safari and Firefox allow the user to input non-number
+          characters but do not recognize those characters as values. If a user inputs "aaaa" or 123a"
+          {' '}<NxCode>onChange</NxCode> receives "" as the value in both of those invalid cases. It should be
```
avnvbhatta, line 207:
```suggestion
          , <NxCode>onChange</NxCode> receives "" as the value in both of those invalid cases. It should be
```

gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -192,6 +192,33 @@ const NxTextInputPage = () =>
...
+          setting that as the value stored in the state, which goes in as the value of the &lt;input&gt; in the next
+          render, which should clear the input. Normally you'd expect that as soon as you enter an invalid character
+          after the number the input should have cleared. But for some reason that doesn't happen. We have traced
```
avnvbhatta, line 210:
```suggestion
          after the number, the input should have cleared. But for some reason that doesn't happen. We have traced
```

gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -192,6 +192,33 @@ const NxTextInputPage = () =>
...
+          through the code and all of the state handling in our examples and components is correct. But for some
+          reason, when that empty string gets passed back into the &lt;input&gt; in NxTextInput, it just ignores it.
+          The whole react workflow around <NxCode>type="number"</NxCode> seems to be broken, not to mention our RSC
+          validation workflow.
+        </NxP>
+        <NxP>
+          At this point if an input that only accepts numbers is required it is strongly suggested that the developer
```
avnvbhatta, line 217:
```suggestion
          At this point if an input that only accepts numbers is required, it is strongly suggested that the developer
```

gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -192,6 +192,33 @@ const NxTextInputPage = () =>
...
+          In Chrome it works mostly as expected, however Safari and Firefox allow the user to input non-number
+          characters but do not recognize those characters as values. If a user inputs "aaaa" or 123a"
+          {' '}<NxCode>onChange</NxCode> receives "" as the value in both of those invalid cases. It should be
+          setting that as the value stored in the state, which goes in as the value of the &lt;input&gt; in the next
+          render, which should clear the input. Normally you'd expect that as soon as you enter an invalid character
+          after the number the input should have cleared. But for some reason that doesn't happen. We have traced
+          through the code and all of the state handling in our examples and components is correct. But for some
+          reason, when that empty string gets passed back into the &lt;input&gt; in NxTextInput, it just ignores it.
+          The whole react workflow around <NxCode>type="number"</NxCode> seems to be broken, not to mention our RSC
+          validation workflow.
```
avnvbhatta, line 214:
```suggestion
          reason, when that empty string gets passed back into the &lt;input&gt; in <NxCode>NxTextInput</NxCode>, 
          it just ignores it. The whole React workflow around <NxCode>type="number"</NxCode> seems to be broken, 
          not to mention our RSC validation workflow.
```

gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -192,6 +192,33 @@ const NxTextInputPage = () =>
           </NxTable.Row>
         </NxTable.Body>
       </NxTable>
+      <NxStatefulAccordion defaultOpen={false}>
+        <NxAccordion.Header>
+          <h2 className="nx-accordion__header-title">Regarding &lt;input type="number"&gt;</h2>
```
rpokorny, line 197: Makes me realize we should probably add a component for this

glenhunter: > Makes me realize we should probably add a component for accordion titles

https://issues.sonatype.org/browse/RSC-728

PR #504, opened 2021-09-29
Fixed Firefox focus artifact - RSC-702

entroform:
https://issues.sonatype.org/browse/RSC-702

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "8.7.7",
+  "version": "8.7.8",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "8.7.7",
+  "version": "8.7.8",
   "description": "Sonatype shared UI components and utilities written in React",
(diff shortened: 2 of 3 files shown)


rpokorny (approved):
Seems fixed but while testing it I'm getting an odd thing where the left-side list overall is in the tab order for some reason, in between the filter input and the first item.  Does anyone else see that?  That would be a separate bug

rpokorny:
> Seems fixed but while testing it I'm getting an odd thing where the left-side list overall is in the tab order for some reason

Oh I think it's because it's scrollable.  Is that just a firefox thing, that scroll boxes get added to the tab order?

entroform:
> > Seems fixed but while testing it I'm getting an odd thing where the left-side list overall is in the tab order for some reason
> 
> Oh I think it's because it's scrollable. Is that just a firefox thing, that scroll boxes get added to the tab order?

Hmm interesting. Yeah it is just a Firefox thing it seems. Is this something we should add for accessibility on other browsers? 🤔  https://accessibilityinsights.io/info-examples/web/scrollable-region-focusable/

rpokorny:
> Hmm interesting. Yeah it is just a Firefox thing it seems. Is this something we should add for accessibility on other browsers? thinking https://accessibilityinsights.io/info-examples/web/scrollable-region-focusable/

From that page (emphasis mine):

> Instead, focus must be moved either to the scrollable element itself **or to a child element** within the scrollable element. Then the content can be scrolled using the keyboard’s arrow keys.

In chrome, when one of the list items is focused, the arrow keys do scroll the list.  So I think we're fine.

PR #510, opened 2021-09-30
Improvement Day: jest-image-snapshot

rpokorny:
This PR moves entirely off of applitools for RSC visual testing, and switches to jest-image-snapshot.

Advantages:
* Removes dependency on a third-party service which has less-than-perfect uptime/responsiveness.  
* Total build time seems to be somewhat improved.  
* Could also be considered a proof of concept for how other Sonatype projects might make a simiilar switch resulting in significant vendor cost savings.
* Build is simpler - no more network of docker containers running parallel chrome instances
* The browser automation API which works well with jest-image-snapshot - Puppeteer - has much better documentation than webdriverio.

Downsides: 
* potentially increased costs of running our CI servers due to increased CPU usage associated with doing the image comparisons ourselves
* No real "ignore regions" - ignored elements need to be hidden with CSS before taking the screenshot
* Screenshots are currently stored in git and take a nontrivial amount of space which will grow over time.  Further adjustments to our approach may be possible in order to address this, such as using git-lfs
* No "intelligent" fuzzy matching.  jest-image-snapshot is currently configured to only pass on exact match, though it can be configured to accept a certain percentage of mismatched pixels

A note about storage size: the changes introduced in this branch add about 10MB to the repository, most of which is the current screenshots.  In order to minimize this number, I ran some git commands after everything was working in order to squash the screenshots into a single commit (and thus a single instance of each screenshot).  Before doing this, many screenshots had multiple versions in the git history as I worked out issues and the total size of this branch of about 40MB.  Repository size, both on a per branch basis and overall, is something that will need to be watched closely with this approach.


Dockerfile
```diff
@@ -9,7 +9,8 @@ FROM docker-all.repo.sonatype.com/node:12
 
 RUN apt-get update && \
     apt-get autoremove -y && apt-get clean -y && \
-    apt-get install jq -y
+    apt-get install jq -y && \
+    apt-get install chromium libatk-bridge2.0-0 libxkbcommon0 libgbm1 -y
```
rpokorny, line 13: We use puppeteer for running an automated browser.  Although puppeteer installs its own copy of chromium, we have to make sure the necessary system libraries are in place

Jenkinsfile
```diff
@@ -148,19 +122,8 @@ dockerizedBuildPipeline(
   },
   onFailure: {
     githubStatusUpdate('failure')
+    archiveArtifacts(artifacts: 'gallery/test-results/**/*')
```
rpokorny, line 125: This makes any visual diffs that come up on Jenkins accessible as build artifacts

gallery/src/main.scss
```diff
@@ -6,6 +6,11 @@
  */
 @use 'sass:math';
 
+@font-face {
+  font-family: "Inconsolata";
```
rpokorny, line 10: This is the font face that the gallery code examples specify.  We weren't providing it, so it turns out we've been getting a system-dependent font face there all along.  That creates differences between local snapshots and those on jenkins.

gallery/src/main.scss
```diff
@@ -15,12 +20,6 @@
   width: 400px;
 }
 
-.gallery-example {
-  pre {
-    font-size: var(--nx-font-size-s);
```
rpokorny, line 20: IMO Inconsolata is too small at this size.  Additionally it turns out this was affecting our `nx-pre` example which was definitely not intended.

lib/src/base-styles/_nx-form-select.scss
```diff
@@ -16,6 +17,7 @@
 }
 
 .nx-form-select {
+  @include nx-text-helpers.regular;
```
rpokorny: User-agent styles specify a font for `<select>` elements which were were previously failing to override.

gallery/src/main.scss
glenhunter: > This is the font face that the gallery code examples specify. We weren't providing it, so it turns out we've been getting a system-dependent font face there all along. That creates differences between local snapshots and those on jenkins.

I'm not sure what's going on there but Courier Prime was selected as our monospace font here: https://issues.sonatype.org/browse/RSC-176?jql=text%20~%20%22monospace%22

rpokorny: > I'm not sure what's going on there but Courier Prime was selected as our monospace font here:

The syntax highlighter themes have their own opinions about what font they want to use, which they set with no regard for our preferences.  The theme we had been using for a long time specified Inconsolata, but we hadn't been providing that font, and so the examples were actually rendering differently on different OSes/browsers.

That said, we have actually switched to a different syntax highlighter theme in the time since I wrote the above comment, and that new theme does not use Inconsolata, but rather specifies a handful of proprietary fonts.  That means it would delibreately look a bit different on different OSes, which isn't workable for visual tests, so I used a `!important` to override it back to Inconsolata.  Per your comment though, I probably should have overridden it to Courier Prime instead and gotten rid of Inconsolata since it's no longer what's desired either by the highlighter theme nor by us.  I will make that change

rpokorny: a524a1c9acf4

glenhunter:
As discussed in Slack, differences in font rendering are creating hundreds of false positives. We'll need to figure out a solution to this.

gallery/visualtests/nxCheckbox.js
```diff
@@ -20,64 +29,43 @@ describe('NxCheckbox', function() {
...
     it(`has a blue background and white checkmark with a light blue outer border
       when clicked, focused, and hovered`, async function() {
       const focusSelector = `${selector} input`,
-          [targetElement, focusElement] = await Promise.all([browser.$(selector), browser.$(focusSelector)]);
+          [targetElement, focusElement] = await waitAndGetElements(selector, focusSelector);
+          //targetElementBox = targetElement.boundingBox(),
+          //targetElementCenter = {
+            //x: targetElementBox.x + (targetElementBox.width / 2),
+            //y: targetElementBox.y + (targetElementBox.height / 2)
+          //};
```
avnvbhatta, line 62: Can probably remove these.

gallery/visualtests/nxCheckbox.js
```diff
@@ -20,64 +29,43 @@ describe('NxCheckbox', function() {
...
-        await browser.eyesRegionSnapshot(null, Target.region(targetElement));
-      }
-      finally {
-        // click again to reset the state
-        await targetElement.click();
-      }
+      await focusElement.focus();
+      await targetElement.hover();
+
+      await checkScreenshot(targetElement);
```
avnvbhatta, line 68: Do we not need to reset by clicking away?

gallery/visualtests/nxDropdown.js
```diff
@@ -54,49 +56,36 @@ describe('NxDropdown', function() {
...
       await button.click();
     });
 
     it('has links that look right', async function() {
-      const targetElement = await browser.$(selector);
+      const [targetElement] = await waitAndGetElements(selector);
 
-      await targetElement.scrollIntoView({ block: 'center' });
-      await targetElement.moveTo({ xOffset: -10, yOffset: -10 });
+      //await targetElement.scrollIntoView({ block: 'center' });
```
avnvbhatta, line 67: Can probably remove this as well.

gallery/visualtests/nxRadio.js
```diff
@@ -22,68 +29,38 @@ describe('NxRadio', function() {
...
-        await browser.eyesRegionSnapshot(null, Target.region(targetElement));
-      }
-      finally {
-        // click again to reset the state
-        await otherElement.click();
-      }
+      await moveMouseAway();
+      await focusElement.focus();
+
+      await checkScreenshot(targetElement);
```
avnvbhatta, line 51: Do we not need to reset state?

gallery/visualtests/nxRadio.js
```diff
@@ -22,68 +29,38 @@ describe('NxRadio', function() {
...
-        await browser.eyesRegionSnapshot(null, Target.region(targetElement));
-      }
-      finally {
-        // click again to reset the state
-        await otherElement.click();
-      }
+      await focusElement.focus();
+      await targetElement.hover();
+
+      await checkScreenshot(targetElement);
```
avnvbhatta, line 63: Same as above, do we not need to reset the state?

gallery/visualtests/nxTabs.js
```diff
@@ -44,21 +50,13 @@ describe('NxTabs', function() {
...
-      const targetElement = await browser.$(modalSelector);
+      const [closeModalBtn, targetElement] = await waitAndGetElements(closeModalBtnSelector, modalSelector);
 
-      try {
-        await browser.eyesRegionSnapshot(null, Target.region(targetElement));
-      }
-      finally {
-        await closeModalBtn.click();
-      }
+      await checkScreenshot(targetElement);
```
avnvbhatta, line 59: Do we not need to dismiss the modal?

gallery/visualtests/testUtils.js
```diff
@@ -4,152 +4,251 @@
...
+          finally {
+            await blurElement(focusElement);
+          }
+        };
+      },
+
+      clickTest(elementSelector, clickSelector = elementSelector) {
+        return async () => {
+          const [targetElement, clickElement] = await waitAndGetElements(elementSelector, clickSelector),
+              { x, y, width, height }= await clickElement.boundingBox();
```
avnvbhatta, line 231: Missing space before =

lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "9.15.0",
+  "version": "9.15.1",
```
avnvbhatta, line 3: Probably more than a patch.

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "9.15.0",
+  "version": "9.15.1",
```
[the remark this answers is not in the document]
rpokorny: > Feels like this is definitely more than a patch

Keep in mind the version numbers are really only for the lib.  Gallery-only changes ideally wouldn't bump the version at all, but having CI set up the way we do makes it worthwhile.

gallery/visualtests/nxCheckbox.js
rpokorny: > Do we not need to reset by clicking away?

Observe you `setupBrowser` works.  Each test gets a fresh page via `beforeEach`

gallery/visualtests/nxRadio.js
rpokorny: > Same as above, do we not need to reset the state?

https://github.com/sonatype/sonatype-react-shared-components/pull/510#discussion_r839957163

lib/package.json
rpokorny: > Probably more than a patch.

https://github.com/sonatype/sonatype-react-shared-components/pull/510#discussion_r839954303

gallery/visualtests/testUtils.js
rpokorny: > Missing space before =

2c8e49dc

avnvbhatta (approved):
LGTM!! 🚀

entroform (approved):
LGTM!

PR #513, opened 2021-10-04
Spike: POCs for multiple charting libraries (Not for merging)

rpokorny:
In this PR I add simple pie chart examples built in three react charting libraries: Nivo, Recharts, and Victory.  Of the three, I prefer nivo as it appears to have the most configuration options and the best documentation.  Unfortunately, none of these options really supported any sort of styling via CSS, though recharts included enough information in the DOM to make it _somewhat_ possible. 

While reviewing please take a look at the documentation for each project as it is a significant differentiator.  

Feel free to speak up if you think I should add additional libraries, additional chart types, or additional specific options for these pie charts.

This PR also adds some code to make the RSC selectable color values available in JS.  Whatever we move forward with, we will probably need to keep that and also do something similar for the threat level colors.


glenhunter:
Seems weird that labelling issues like this keep appearing, seems like a basic thing for the library to solve.

Screen Shot 2021-10-08 at 2 30 17 PM

glenhunter:
> > Yeah I didn't really find one that seemed immune to that sort of thing
> 
> FWIW at least some of the libraries (nivo in particular, I recall) have an option to hide labels on pie slices below a certain size. So that could "solve" the problem but I feel like it's a questionable solution

Yeah, but better than the overlapping. I couldn't get the hiding to work in the example - maybe because the slice in question is vertical?

glenhunter:
Of these three Nivo is definitely my preference, the documentation is excellent.

PR #516, opened 2021-10-05
Fix tile header positioning when action buttons are present - RSC-657

rpokorny:
https://issues.sonatype.org/browse/RSC-657


entroform (approved):
LGTM!

PR #517, opened 2021-10-05
Added NxSearchDropdown escape key behaviour - RSC-695 

entroform:
https://issues.sonatype.org/browse/RSC-695

### DIFF
--- a/lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
+++ b/lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
@@ -93,9 +93,7 @@ function NxSearchDropdownRender<T extends string | number = string>(
     if (event.key === 'Escape') {
       onSearchTextChange('');
 
-      if (!event.defaultPrevented) {
-        event.preventDefault();
-      }
+      event.preventDefault();
     }
   };
 
--- a/lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
+++ b/lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
@@ -4,7 +4,7 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React, { FocusEvent, Ref, useRef } from 'react';
+import React, { FocusEvent, KeyboardEventHandler, Ref, useRef } from 'react';
 import classnames from 'classnames';
 import { partial } from 'ramda';
 
@@ -89,6 +89,16 @@ function NxSearchDropdownRender<T extends string | number = string>(
     onSearch(value.trim());
   }
 
+  const handleKeyDown: KeyboardEventHandler = event => {
+    if (event.key === 'Escape') {
+      onSearchTextChange('');
+
+      if (!event.defaultPrevented) {
+        event.preventDefault();
+      }
+    }
+  };
+
   return (
     <div ref={ref} className={className} onFocus={handleComponentFocus} { ...attrs }>
       <NxFilterInput ref={filterRef}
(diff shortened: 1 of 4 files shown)


lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
```diff
@@ -89,6 +89,16 @@ function NxSearchDropdownRender<T extends string | number = string>(
     onSearch(value.trim());
   }
 
+  const handleKeyDown: KeyboardEventHandler = event => {
+    if (event.key === 'Escape') {
```
rpokorny, line 93: I _think_ this is fine nowadays but I recall in the past different browsers had different ways to check for escape, so be sure this works in everything we currently support.

lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
```diff
@@ -89,6 +89,16 @@ function NxSearchDropdownRender<T extends string | number = string>(
...
+      onSearchTextChange('');
+
+      if (!event.defaultPrevented) {
+        event.preventDefault();
```
rpokorny, line 97: Why is this needed? Even if it is, is there a point in wrapping it in the conditional?

rpokorny: > Why is this needed? Even if it is, is there a point in wrapping it in the conditional?

I guess it would be needed if this was within a modal, to prevent esc from also closing the modal

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -318,4 +318,37 @@ describe('NxSearchDropdown', function() {
...
+    expect(onSearchTextChange).not.toHaveBeenCalled();
+
+    filterInput.simulate('keyDown', keyDownEvent);
+
+    expect(escPreventDefault).toHaveBeenCalled();
+    expect(onSearchTextChange).toHaveBeenCalledWith('');
+
+    onSearchTextChange.mockClear();
+
+    dropdownMenu.simulate('keyDown', keyDownEvent);
```
rpokorny, line 350: Isn't the dropdown not even rendered at this point?

entroform: > Isn't the dropdown not even rendered at this point?

I think it's still rendered because `searchText` did not change and that's what toggles the dropdownMenu

lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
entroform: > I _think_ this is fine nowadays but I recall in the past different browsers had different ways to check for escape, so be sure this works in everything we currently support.

Good point. I did look at NxDropdown implementation and it also only checks for `key === 'Escape'`. Do you think that's a safe assumption to say that it works with what we currently supports 🤔 ?
https://caniuse.com/keyboardevent-key

lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
entroform: > > Why is this needed? Even if it is, is there a point in wrapping it in the conditional?
> 
> I guess it would be needed if this was within a modal, to prevent esc from also closing the modal

Yeah, the conditional is probably not needed, removed it here: 62efc9de1d1e3d19052e42ae14e214afbab3bcd4

lib/src/components/NxSearchDropdown/NxSearchDropdown.tsx
rpokorny: > Good point. I did look at NxDropdown implementation and it also only checks for key === 'Escape'. Do you think that's a safe assumption to say that it works with what we currently supports thinking ?

Yeah following what we do in other components should be adequate

rpokorny (approved):

PR #518, opened 2021-10-05
Updated NxDropdown icon size to match Kaleido - RSC-682

entroform:
https://issues.sonatype.org/browse/RSC-682

I wasn't sure if we should size it in CSS using `font-size` or use the sizings from FontAwesome.

I decided to go with the FontAwesome sizings because it felt more deliberate:
https://fontawesome.com/v5.15/how-to-use/on-the-web/styling/sizing-icons


rpokorny (approved):
I think doing it this way is ok, and it's certainly easier than handling the vertical alignment ourselves.

PR #522, opened 2021-10-05
Remove pristine styling. - RSC-690

entroform:
https://issues.sonatype.org/browse/RSC-646

### DIFF
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -235,9 +235,6 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-form-element-border">
               The standard color for the borders of form fields (such as text inputs) in their default state.
             </ColorDocRow>
-            <ColorDocRow colorVar="--nx-color-form-element-border-pristine">
-              The standard color for the borders of form fields (such as text inputs) in their pristine state.
-            </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-interactive-background-selected">
               The standard color for the backgrounds of interactive elements (such as clickable list rows) when they
               are selected.
--- a/gallery/visualtests/nxFormSelect.js
+++ b/gallery/visualtests/nxFormSelect.js
@@ -16,7 +16,7 @@ describe('nx-form-select', function() {
   });
 
   describe('Simple NxFormSelect', function() {
-    it('has a light border when pristine', simpleTest(selector));
+    it('has a dark border when pristine', simpleTest(selector));
 
     it('has a dark border when not pristine', async function() {
       const targetElement = await browser.$(selector);
--- a/gallery/visualtests/nxTextInput.js
+++ b/gallery/visualtests/nxTextInput.js
@@ -28,7 +28,7 @@ describe('NxTextInput', function() {
   });
 
   describe('Simple NxTextInput', function() {
-    it('has a light border when pristine', simpleTest(simpleComponentSelector));
+    it('has a dark border when pristine', simpleTest(simpleComponentSelector));
 
     it('has a dark border when hovered',
         hoverTest(simpleComponentSelector, getInputElementSelector(simpleComponentSelector)));
@@ -52,7 +52,7 @@ describe('NxTextInput', function() {
   });
 
   describe('Validatable NxTextInput', function() {
-    it('has a light border when pristine', simpleTest(validatableComponentSelector));
+    it('has a dark border when pristine', simpleTest(validatableComponentSelector));
 
     it('has validation styles when valid', async function() {
       const inputSelector = getInputElementSelector(validatableComponentSelector),
--- a/lib/src/base-styles/_nx-text-input.scss
+++ b/lib/src/base-styles/_nx-text-input.scss
@@ -14,12 +14,6 @@
   max-width: 100%;
 
   &, &.pristine, &.pristine.valid {
-    &:hover {
-      .nx-text-input__box {
-        border-color: var(--nx-color-form-element-border);
-      }
-    }
-
     &:focus-within {
       .nx-text-input__box {
         border-color: var(--nx-color-interactive-border-focus);
--- a/lib/src/components/NxFilterInput/NxFilterInput.scss
+++ b/lib/src/components/NxFilterInput/NxFilterInput.scss
@@ -31,8 +31,3 @@
     margin-right: 8px;
   }
 }
-
-.nx-filter-input--empty .nx-text-input__box {
-  // empty filter inputs should be styled the same as pristine text inputs
-  border-color: var(--nx-color-form-element-border-pristine);
-}
--- a/lib/src/scss-shared/_nx-form-validation-helpers.scss
+++ b/lib/src/scss-shared/_nx-form-validation-helpers.scss
@@ -25,10 +25,6 @@
 }
 
 @mixin pristine() {
-  .nx-text-input__box {
-    border-color: var(--nx-color-form-element-border-pristine);
-  }
-
   .nx-icon--invalid, .nx-icon--valid {
     display: none;
   }
--- a/lib/src/base-styles/_nx-colors.scss
+++ b/lib/src/base-styles/_nx-colors.scss
@@ -13,7 +13,6 @@
   --nx-color-site-background: var(--nx-swatch-indigo-97);
 
   --nx-color-form-element-border: var(--nx-swatch-grey-50);
-  --nx-color-form-element-border-pristine: var(--nx-swatch-indigo-90);
 
   --nx-color-interactive-background-selected: var(--nx-swatch-blue-90);
   --nx-color-interactive-background-active: var(--nx-swatch-blue-95);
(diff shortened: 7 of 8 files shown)


gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -235,9 +235,6 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-form-element-border">
               The standard color for the borders of form fields (such as text inputs) in their default state.
             </ColorDocRow>
-            <ColorDocRow colorVar="--nx-color-form-element-border-pristine">
-              The standard color for the borders of form fields (such as text inputs) in their pristine state.
```
rpokorny, line 239: Removing the actual variable would be a breaking change.  Instead, mark it as deprecated and explain that we no longer use it.

gallery/visualtests/nxFormSelect.js
```diff
@@ -16,7 +16,7 @@ describe('nx-form-select', function() {
   });
 
   describe('Simple NxFormSelect', function() {
-    it('has a light border when pristine', simpleTest(selector));
+    it('has a dark border when pristine', simpleTest(selector));
```
rpokorny, line 19: The distinction between this test and the one that follows is no longer necessary.  I'd have just one, implemented using `simpleTest`, named `it('has a dark border by default'`.  And then maybe rename the hover test to say "darker" border.

gallery/visualtests/nxTextInput.js
```diff
@@ -28,7 +28,7 @@ describe('NxTextInput', function() {
   });
 
   describe('Simple NxTextInput', function() {
-    it('has a light border when pristine', simpleTest(simpleComponentSelector));
+    it('has a dark border when pristine', simpleTest(simpleComponentSelector));
```
rpokorny, line 31: Similar to what I said on the `NxFormSelect` tests, there no longer needs to be a distinction between this test and the (mis-named) "non-empty" test.

lib/src/base-styles/_nx-text-input.scss
```diff
@@ -14,12 +14,6 @@
   max-width: 100%;
 
   &, &.pristine, &.pristine.valid {
-    &:hover {
-      .nx-text-input__box {
-        border-color: var(--nx-color-form-element-border);
```
rpokorny, line 19: Hover styling seems broken; pretty sure this removal is the reason why.

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -31,8 +31,3 @@
     margin-right: 8px;
   }
 }
-
-.nx-filter-input--empty .nx-text-input__box {
-  // empty filter inputs should be styled the same as pristine text inputs
```
rpokorny, line 36: This is still true - we want the light colored border on filter inputs when they are empty.

lib/src/scss-shared/_nx-form-validation-helpers.scss
```diff
@@ -25,10 +25,6 @@
 }
 
 @mixin pristine() {
```
rpokorny, line 27: Shouldn't this whole mixin be removed?

lib/src/components/NxFilterInput/NxFilterInput.scss
entroform: > This is still true - we want the light colored border on filter inputs when they are empty.

I double checked with Kaleido and @alexalyons126 on this and they mentioned that it should be the default border color (grey-50) even when it's empty

lib/src/base-styles/_nx-text-input.scss
entroform: > Hover styling seems broken; pretty sure this removal is the reason why.

Thanks, yeah I bought it back and changed it from to `--nx-color-form-element-border` to `grey-05` in this commit: 867bfa96276d741db73a1dc17d89e8ba58d04042

gallery/visualtests/nxFormSelect.js
entroform: > The distinction between this test and the one that follows is no longer necessary. I'd have just one, implemented using `simpleTest`, named `it('has a dark border by default'`. And then maybe rename the hover test to say "darker" border.

Thanks, updated: 867bfa96276d741db73a1dc17d89e8ba58d04042

gallery/src/styles/CssVariables/CssVariablesPage.tsx
entroform: > > Updated to deprecated: [867bfa9](https://github.com/sonatype/sonatype-react-shared-components/commit/867bfa96276d741db73a1dc17d89e8ba58d04042)
> 
> > **and explain that we no longer use it.**

Sorry, forgot to add the explanation: Let me know if this is sufficient: e44e2525b03f6b0729c7fcc58a68e78834958efa

lib/src/components/NxFilterInput/NxFilterInput.scss
entroform: > > I double checked with Kaleido and @alexalyons126 on this and they mentioned that it should be the default border color (grey-50) even when it's empty
> 
> I find that very surprising considering previous conversations. @alexalyons126 just created this behavior for filter inputs about a month an a half ago and I recall asking whether the removal of the pristine state styling would impact that, and she confirmed that it would not. Please double check.

I double checked with @alexalyons126 and it's confirmed that filter input shouldn't have a different border color when it's empty.

gallery/src/styles/CssVariables/CssVariablesPage.tsx
rpokorny: > Sorry, forgot to add the explanation: Let me know if this is sufficient: e44e252

I feel like it's a little bit confusing the way you currently have it.  The simplest fix I think would be to separate the description of what the variable is for out of the explanation of why it's deprecated.  i.e. move the second sentence out of the alert.

rpokorny:
> I removed the entire border color tests because it seems to be a duplicate of the ones in text input.

You needed to keep and rename at least one of them as a test of what a "normal" NxFilterInput looks like.  Right now you've only got the search icon variation and the disabled variation.

rpokorny:
> Per the conversation that recently happened in Slack, NxFormSelect has some logic around pristine state tracking that only existed for the sake of this styling that is being removed. So I think that extra logic should be removed as much as possible without becoming a breaking change. That is, I think NxFormSelect should basically ignore the isPristine prop (but still actively avoid passing it to the native <select> to avoid console warnings), the documentations for NxFormSelect isPristine should be removed, and the form select state helpers should be deprecated and replaced with functions that just wrap their parameter in an object (e.g. value => ({ value }))

If you'd rather do that all as a separate PR under the same JIRA ticket I'd be fine with that

gallery/src/components/NxFormSelect/NxFormSelectPage.tsx
```diff
@@ -70,7 +68,7 @@ const NxFormSelectPage = () =>
         <NxP>
           The <NxCode>nxFormSelectStateHelpers</NxCode> object includes the following recommended state helper
```
rpokorny, line 69: IMO these should be marked as deprecated, and it should be explained that they used to do more but are now unneeded and not recommended for new use.

lib/src/components/NxFormSelect/types.ts
```diff
@@ -9,11 +9,11 @@ import { SelectHTMLAttributes } from 'react';
 
 export interface StateProps {
   value: string | number;
-  isPristine: boolean;
+  isPristine?: boolean | null;
```
rpokorny, line 12: This can probably be removed entirely

entroform: > This can probably be removed entirely

Removed it and also updated the propTypes.
I tried removing the propTypes entirely since that was the only attribute but it erroring out because of classNames.
Adding the value propType fixed it. Let me know what you think:
326a216

gallery/src/components/NxFormSelect/NxFormSelectPage.tsx
rpokorny: > I also thought of adding it to the isPristine attribute, let me know if this makes sense!

I think what I had suggested above is that the documentation for `isPristine `just be removed.  Deprecated is fine too though

lib/src/components/NxFormSelect/types.ts
entroform: > > > Removed it and also updated the propTypes.
> > > I tried removing the propTypes entirely since that was the only attribute but it erroring out because of classNames.
> > > Adding the value propType fixed it. Let me know what you think:
> > > [326a216](https://github.com/sonatype/sonatype-react-shared-components/commit/326a216be0343c14e875d4d94d34e4b89f057be9)
> > 
> > 
> > Seems to work like this: #532
> 
> Oh sorry, I mistook eslint error for type error: I'm getting `'className' is missing in props validation`
> 
> Adding `props: Props` to NxFormSelect fixed it. Thanks going to merge and update the PR.

Updated: 1501540e78d886d857f4b62a90ea8a73b332fb9c

PR #524, opened 2021-10-05
Updated icon only button disabled style. - RSC-699

entroform:
https://issues.sonatype.org/browse/RSC-699

### DIFF
--- a/lib/src/base-styles/_nx-btn.scss
+++ b/lib/src/base-styles/_nx-btn.scss
@@ -130,4 +130,12 @@
   .nx-icon {
     height: 1em;
   }
+
+  &.disabled, &[disabled] {
+    &, &:hover, &:focus {
+      background-color: transparent;
+      border-color: transparent;
+      box-shadow: none;
+    }
+  }
 }


rpokorny:
Where are these styles actually demonstrated?

entroform:
> Where are these styles actually demonstrated?

I don't think they are demonstrated anywhere in the gallery icon button page.
I can add that as part of this ticket!

entroform:
> Where are these styles actually demonstrated?



> Where are these styles actually demonstrated?

Updated the example in this commit: 4a0df535dae6b4d47e648c0102e92cbd0cd8fe3d

gallery/src/components/NxButton/NxButtonIconOnlyExample.tsx
```diff
@@ -5,15 +5,16 @@
...
 
 import { NxButton, NxFontAwesomeIcon, NxButtonBar } from '@sonatype/react-shared-components';
 
 const NxButtonIconOnlyExample = () =>
   <NxButtonBar>
-    <NxButton title="Alert" variant="icon-only"><NxFontAwesomeIcon icon={faExclamationTriangle}/></NxButton>
+    <NxButton variant="icon-only" title="Alert"><NxFontAwesomeIcon icon={faExclamationTriangle}/></NxButton>
     <NxButton variant="icon-only" title="Edit"><NxFontAwesomeIcon icon={faEdit}/></NxButton>
     <NxButton variant="icon-only" title="Save"><NxFontAwesomeIcon icon={faSave}/></NxButton>
+    <NxButton variant="icon-only" title="Delete" disabled><NxFontAwesomeIcon icon={faTrashAlt}/></NxButton>
```
rpokorny, line 17: The tooltip for this button is getting stuck open.  Additionally I don't think disabled buttons necessarily need tooltips since they can't be focused.  Maybe just an `aria-label` instead.

rpokorny: > Totally missed that, updated it here: c9b7446

Really I think there are multiple cases that need to be shown here.  We do want the attribute to work, and we want to demonstrate it.  It's also the case that attribute-disabled icon-only buttons don't require `title`s, so maybe we should show an attr-disabled button with no title.  But then it's also the case that the caller might _want_ a tooltip on the button, so we should perhaps additionally show the the way to do that is with the disabled classname.

gallery/src/components/NxButton/NxButtonPage.tsx
```diff
@@ -126,7 +126,8 @@ export default function NxButtonPage() {
                           id="nx-button-icon-only-example"
                           liveExample={NxButtonIconOnlyExample}
                           codeExamples={nxButtonIconOnlyCode}>
-        An example of a button containing only an icon. For accessibility purposes it is important to use the
+        An example of buttons containing only an icon and with one that is in a disabled state.
```
rpokorny, line 129: This sentence is no longer accurate.  Also I think it would make sense to explain here when you would use each of the different ways of disabling, and that the `title` is not required (and doesn't really work right) on attr-disabled icon buttons

rpokorny: > Thanks for pointing that out!
I've updated it here, let me know what you think: a7465be

The last sentence could be confusing as-is.  Maybe explain that the title is important on all enabled buttons but that it may or may not be on a disabled button depending on the situation.

PR #525, opened 2021-10-06
Fix usage of bulleted list in docs - RSC-654 RSC-745

rpokorny:
https://issues.sonatype.org/browse/RSC-654

Also implements https://issues.sonatype.org/browse/RSC-745, the removal of semi-bold font weight from top-level bulleted text.


entroform (approved):
LGTM!

PR #530, opened 2021-10-08
Updated button and small icon button size. - RSC-735

entroform:
https://issues.sonatype.org/browse/RSC-735

Confirmed with @alexalyons126 that the button sizes should also match Kaliedo (40px including border), currently it is 42px in the RSC.


rpokorny:
Full size icon buttons are still 42px wide

rpokorny:
It appears that form fields also have this same issue.  Fixing it for buttons but not form fields causes some slight mis-alignments:
image

PR #542, opened 2021-10-18
Tartan Page Header - RSC-581

entroform:
https://issues.sonatype.org/browse/RSC-581


rpokorny:
My assumption had been that all aspects of the header's placement which aren't shown in the mockup would be remaining unchanged.  That is, that the header content would be horizontally aligned with the edge of the sidebar and that the header would remaining stationary when scrolling

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const plaidVillain = require('./plaid-villain.svg');
+const tartanLogotype = require('./tartan-logotype.svg');
+
+import './PageHeader.scss';
+
+const PageHeader = () => (
+  <header className="nx-gallery-page-header">
+    <div className="nx-gallery-page-header__inner">
+      <h1 className="nx-gallery-page-header__logo">
```
rpokorny, line 19: This shouldn't be an `h1`.  The `.nx-page-title` within the main content area is the `h1` on a given page.

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <a className="nx-gallery-page-header__home-link"
```
rpokorny, line 20: Exactly how much of the header is supposed to be the home link?

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+           href="#"
+           aria-label="home">
```
rpokorny, line 22: This prevents a screenreader from reading any of the content within this element.  It just reads the whole thing as "home".

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <img className="nx-gallery-page-header__logo-image"
+               src={plaidVillain}
+               alt="A cute but villainous Tartan logo" />
```
rpokorny, line 25: After re-reading [WHATWG's guidance on `alt`](https://html.spec.whatwg.org/multipage/images.html#alt), I still think this image should have an empty alt attribute, while the other image should have "Sonatype Tartan" as its alt attribute (as it does).  The following excerpts from the HTML standard are especially relevant:

> A corollary to this is that the alt attribute's value should never contain text that could be considered the image's caption, title, or legend

> If the logo is being used next to the name of the entity that it represents, then the logo is supplemental, and its alt attribute must instead be empty.

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <img className="nx-gallery-page-header__logotype-image"
+               src={tartanLogotype}
+               alt="Sonatype Tartan" />
+          <span className="nx-gallery-page-header__name">
+            React Shared Component Library
+          </span>
+        </a>
+      </h1>
+      <div className="nx-gallery-page-header__version">
```
rpokorny, line 34: There is an intention that this text be baseline-aligned with the "React Shared Component Library" text.  I think it would be a lot simpler to tie them together in that way if these two text elements were siblings rather than having the name within the `.nx-gallery-page-header__logo`.

gallery/src/PageHeader/tartan-logotype.svg
```diff
@@ -0,0 +1 @@
...9.18,183.43,31.08,186.92,31.08Zm13.5,0a8.77,8.77,0,0,0,6.72-2.76v2.24h4.53V16.74c0-5.61-4-7.56-8.67-7.56a12.36,12.36,0,0,0-8.75,3.37l1.85,3.15a8.52,8.52,0,0,1,6.17-2.63c2.89,0,4.87,1.47,4.87,3.89v3.11a8.6,8.6,0,0,0-6.72-2.68c-3.41,0-7.25,2-7.25,6.78C193.17,28.79,197,31.08,200.42,31.08ZM202.18,28c-2.54,0-4.44-1.47-4.44-3.71s1.9-3.76,4.44-3.76a6.07,6.07,0,0,1,5,2.16v3.15A6.07,6.07,0,0,1,202.18,28Zm19.88,2.59V16.35a7.19,7.19,0,0,1,5.35-2.76,6.42,6.42,0,0,1,1.42.13V9.23a8.89,8.89,0,0,0-6.77,3.54V9.7h-4.53V30.56Zm16.69.52A6,6,0,0,0,243,29.74l-1.08-3.41a2.94,2.94,0,0,1-2,.73c-1.29,0-2-1-2-2.46V13.67h4.22v-4h-4.22V4H233.4V9.7H230v4h3.45v12C233.4,29.18,235.26,31.08,238.75,31.08Zm13.49,0A8.79,8.79,0,0,0,259,28.32v2.24h4.53V16.74c0-5.61-4.05-7.56-8.67-7.56a12.36,12.36,0,0,0-8.75,3.37l1.85,3.15a8.52,8.52,0,0,1,6.17-2.63C257,13.07,259,14.54,259,17v3.11a8.6,8.6,0,0,0-6.73-2.68c-3.4,0-7.24,2-7.24,6.78C245,28.79,248.84,31.08,252.24,31.08ZM254,28c-2.54,0-4.44-1.47-4.44-3.71s1.9-3.76,4.44-3.76a6.07,6.07,0,0,1,5,2.16v3.15A6.07,6.07,0,0,1,254,28Zm19.88,2.59V16a6.84,6.84,0,0,1,5.35-2.81c2.54,0,4.18,1.08,4.18,4.32v13H288V15.88c0-4.28-2.25-6.7-6.73-6.7a9.85,9.85,0,0,0-7.33,3.37V9.7h-4.53V30.56Z"/></svg>
```
rpokorny, line 1: I'm not sure how much it really matters since we have the `alt` tag, but it might be nice if this svg had a [`<title>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) containing its text.

rpokorny: > I'm not sure how much it really matters since we have the alt tag, but it might be nice if this svg had a <title> containing its text.

Oh or maybe not as it sounds like that'll generate a tooltip

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -0,0 +1,69 @@
+.nx-gallery-page-header {
+  background-color: var(--nx-swatch-white);
+  border-bottom: 1px solid var(--nx-swatch-grey-90);
+  position: sticky;
+  top: 0;
+  width: 100vw;
```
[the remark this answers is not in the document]
entroform: > This should use the `old-header` slot in the container's grid template just like the other header did. Potentially, this element should even use the `nx-page-header` class.

Thanks! Updated it here: c400966dcd9ada6ecae2cc9d764dcb3363e9fe22

gallery/src/PageHeader/PageHeader.tsx
entroform: > This shouldn't be an `h1`. The `.nx-page-title` within the main content area is the `h1` on a given page.

Ah I see, updated it here: c400966dcd9ada6ecae2cc9d764dcb3363e9fe22

gallery/src/PageHeader/PageHeader.tsx
entroform: > Exactly how much of the header is supposed to be the home link?

Talked it over with Jenni, and updated it here: c400966dcd9ada6ecae2cc9d764dcb3363e9fe22

gallery/src/PageHeader/PageHeader.tsx
entroform: > This prevents a screenreader from reading any of the content within this element. It just reads the whole thing as "home".

Removed it here: c400966dcd9ada6ecae2cc9d764dcb3363e9fe22

gallery/src/PageHeader/PageHeader.tsx
entroform: > After re-reading [WHATWG's guidance on `alt`](https://html.spec.whatwg.org/multipage/images.html#alt), I still think this image should have an empty alt attribute, while the other image should have "Sonatype Tartan" as its alt attribute (as it does). The following excerpts from the HTML standard are especially relevant:
> 
> > A corollary to this is that the alt attribute's value should never contain text that could be considered the image's caption, title, or legend
> 
> > If the logo is being used next to the name of the entity that it represents, then the logo is supplemental, and its alt attribute must instead be empty.

Hmm, okay, makes sense to follow the guidance: updated it here: c400966dcd9ada6ecae2cc9d764dcb3363e9fe22

gallery/src/PageHeader/PageHeader.tsx
entroform: > There is an intention that this text be baseline-aligned with the "React Shared Component Library" text. I think it would be a lot simpler to tie them together in that way if these two text elements were siblings rather than having the name within the `.nx-gallery-page-header__logo`.

Yeah it was difficult before because `name` was inside of the home-link and the `version` is not. But that's been changed here: c400966dcd9ada6ecae2cc9d764dcb3363e9fe22

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -0,0 +1,53 @@
+.nx-page-header {
```
rpokorny, line 1: The gallery's header should have both the `.nx-page-header` class and its own gallery-specific class. The gallery stylesheets should only style the latter, and only add the minimum necessary in addition to the `.nx-page-header` styles.  For instance I wouldn't expect to see the `grid-area` being repeated here.

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -19,7 +19,6 @@ $-product-version-row-height: 1fr;
 
 .nx-page-header {
   box-sizing: content-box;
-
```
rpokorny, line 22: Seems extraneous

entroform: > Seems extraneous

Fixed: c33b4595748df2748e4b30c298492217e78dfe08

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -0,0 +1,50 @@
+.nx-gallery-page-header {
```
rpokorny, line 1: The gallery's standard classname prefix is just `gallery`, not `nx-gallery`

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  &__inner {
```
rpokorny, line 3: I feel like there's still a serious miscommunication here.  The new header should be taking advantage of the existing styles in RSC to the maximum extent that they are applicable, and should be doing so by using all relevant RSC classes (`.nx-page-header__inner` for instance). Then the gallery should add its own classes with their own style overrides as-needed.  We shouldn't be repeating styles that RSC's `.nx-page-header` and friends already have solved here.

rpokorny:
Sketch mockup still has some odd spacing FYI, inconsistent with @entroform 's description in Slack of what the intended layout is:
image

gallery/src/PageHeader/PageHeader.scss
entroform: > The gallery's standard classname prefix is just `gallery`, not `nx-gallery`

Thanks, updated: 188ca23b4a36433fcc202fd279284a0dafbf5dbf

gallery/src/PageHeader/PageHeader.scss
entroform: > I feel like there's still a serious miscommunication here. The new header should be taking advantage of the existing styles in RSC to the maximum extent that they are applicable, and should be doing so by using all relevant RSC classes (`.nx-page-header__inner` for instance). Then the gallery should add its own classes with their own style overrides as-needed. We shouldn't be repeating styles that RSC's `.nx-page-header` and friends already have solved here.

This was my bad, I updated it here: 188ca23b4a36433fcc202fd279284a0dafbf5dbf

entroform:
> Sketch mockup still has some odd spacing FYI, inconsistent with @entroform 's description in Slack of what the intended layout is: image

We updated the header again with new dimensions:

Screen Shot 2021-11-03 at 3 42 43 PM

The Sketch file is still in the process of being updated, will update you on this!

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -0,0 +1,41 @@
+.gallery-page-header {
+  box-sizing: border-box;
+
+  $-logo-image-size: 40px;
```
rpokorny, line 4: Minor, no need to use scss private variable syntax ('$-`) for local variables

rpokorny (approved):

PR #544, opened 2021-10-18
Spike: Bar Charts POC - RSC-755

entroform:
https://issues.sonatype.org/browse/RSC-755


entroform:
> Seems like overall, we continue to be in favor of nivo

Yup! For me the order preference is:
1) Nivo
2) Recharts
3) Victory

PR #554, opened 2021-10-21
SPIKE: line charts RSC-754

glenhunter:
https://issues.sonatype.org/browse/RSC-754

Again I felt that Nivo was the definite front runner here, not just because of the documentation, but also for the data layout. I really disliked Victory's requirement to lay out the data with `x,y` coordinates in the line declarations. Like Nivo, Recharts kept the data separate and I can't fault much about the implementation but it's just not as fully featured, or as nice looking (IMO) as Nivo.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -65,6 +65,7 @@
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@nivo/core": "^0.73.0",
     "@nivo/pie": "^0.73.0",
+    "@nivo/line": "^0.73.0",
     "@nivo/tooltip": "^0.73.0",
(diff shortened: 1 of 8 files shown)


rpokorny:
>  I really disliked Victory's requirement to lay out the data with x,y coordinates in the line declarations.

Could you elaborate on this? It kind of seems the same in your nivo example.  I'd be interested to see if either library has an option like what you do with recharts where you tell it what key in the object is what.  Even if they don't though, I don't see this as a significant issue.  Having to transform the data into a model that makes sense for its view before you pass it to the view seems like a pretty reasonable thing to have to do.

PR #555, opened 2021-10-21
Updated NxTags to match Kaleido - RSC-718

entroform:
https://issues.sonatype.org/browse/RSC-718

### DIFF
--- a/lib/src/components/NxTag/NxTag.scss
+++ b/lib/src/components/NxTag/NxTag.scss
@@ -10,7 +10,6 @@
 .nx-tag {
   @include nx-text-helpers.semi-bold;
   @include nx-container-helpers.container-horizontal;
-  @include nx-text-helpers.truncate-ellipsis;
 
   align-items: center;
   background-color: var(--nx-selectable-color-dark);
@@ -27,7 +26,6 @@
   padding: 0 var(--nx-spacing-4x) 0 var(--nx-spacing-4x);
   position: relative;
   user-select: none;
-  overflow: visible;
   // Uppercase Min so SCSS ignores it and uses CSS min, not SCSS min
   max-width: Min(100%, 320px);
 }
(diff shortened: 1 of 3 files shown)


rpokorny:
I'm getting a dark border on focused selectable tags, which isn't supposed to be there
image

entroform:
> I'm getting a dark border on focused selectable tags, which isn't supposed to be there image

I double checked this with Jenni:
So when it is selected + focused + hovered, there shouldn't be any black border,
but in this case when it is **unselected** + focused + hovered it retains the black border styling from the hover state.

rpokorny:
> I double checked this with Jenni:
So when it is selected + focused + hovered, there shouldn't be any black border,
but in this case when it is unselected + focused + hovered it retains the black border styling from the hover state.
Jenni will update the Kaleido mockup to explicitly specify^

OK but I took the screenshot above when I was merely focused, not also hovering.  Additionally the black border is present in both the selected and unselected focus states.

entroform:
> OK but I took the screenshot above when I was merely focused, not also hovering. Additionally the black border is present in both the selected and unselected focus states.

Ah sorry, for some reason this is only happening in Chrome, I fixed it here:

95b86d66c141dff4869b9e4a09dc3fcbfc7ee69a

lib/src/components/NxTag/NxTag.scss
```diff
@@ -33,25 +33,33 @@
...
 
   &:hover {
     border-color: var(--nx-color-interactive-border-hover);
   }
 
   &:focus {
-    border: 1px solid var(--nx-color-interactive-border-focus);
-    box-shadow: var(--nx-box-shadow-focus);
-    outline: none;
+    &:focus-visible {
```
rpokorny, line 43: I'm not under the impression that this extra bit of selector is needed here.  Disabling the native outline works directly on `:focus`

rpokorny (approved):
+1 except for latest comment

PR #566, opened 2021-11-04
NxTag keyboard access RSC-771 

glenhunter:
https://issues.sonatype.org/browse/RSC-771

I converted the `<div>` tags to `<label>` and added a hidden checkbox to the selectable tags.


entroform:
I noticed a few failing tests in`NxTag.test.tsx`:
some of them just need changing the selector from 'div' to 'label' and changing simulate from 'click' to 'change'.

lib/src/components/NxTag/NxTag.tsx
```diff
@@ -25,10 +25,10 @@ const NxTag: FunctionComponent<Props> =
 
       return (
         <NxOverflowTooltip>
-          <div className={tagClasses} {...attrs}>
+          <label className={tagClasses} {...attrs}>
```
entroform, line 28: I wonder if this could be a breaking change:
In the doc it mentioned that it accepts HTML div attributes: https://gallery.sonatype.dev/#/pages/NxTag
I wonder if wrapping the label in div and moving the classname and attrs back to the div might fix it.

rpokorny: > I wonder if this could be a breaking change:
In the doc it mentioned that it accepts HTML div attributes: https://gallery.sonatype.dev/#/pages/NxTag
I wonder if wrapping the label inside div and moving the classname and attrs back to the div might fix it.

I suspect `div` attrs are a subset of `label` attrs.  If that's true this wouldn't be a breaking change.

lib/src/components/NxTag/NxTag.scss
```diff
@@ -61,6 +61,13 @@
   }
 }
 
+.nx-tag__input {
```
rpokorny, line 64: I thought we were going with just `appearance: none; -moz-appearance: none;` for this situation these days.  Which brings back my other point that we should really make a mixin for this, that way we can stop second-guessing the right way to do it in each situation.

lib/src/components/NxTag/NxTag.scss
```diff
@@ -39,7 +39,7 @@
     border-color: var(--nx-color-interactive-border-hover);
   }
 
-  &:focus {
+  &:focus-within {
```
rpokorny, line 42: Why this change? The effect that I would have expected from this change would be that the focus border only shows up due to keyboard nav and not mouse clicks.  However that's not what I'm seeing, the focus still also shows up on mouse clicks just like if we were still using `:focus` here.

lib/src/components/NxTag/NxTag.tsx
```diff
@@ -48,14 +48,12 @@ export const NxSelectableTag: FunctionComponent<SelectableProps> =
             <NxFontAwesomeIcon icon={isSelected ? faTimesCircle : faPlusCircle} className="nx-tag__action" />;
 
       return (
-        <NxTag tabIndex={0}
-               role="switch"
```
rpokorny, line 52: Based on its definition (https://www.w3.org/TR/wai-aria/#switch) I feel like this role might still be a good idea

lib/src/components/NxTag/NxTag.scss
glenhunter: > I thought we were going with just `appearance: none; -moz-appearance: none;` for this situation these days. Which brings back my other point that we should really make a mixin for this, that way we can stop second-guessing the right way to do it in each situation.

I don't recall that decision and it doesn't work in Safari. 
Screen Shot 2021-11-05 at 1 24 55 PM

lib/src/components/NxTag/NxTag.scss
glenhunter: > Why this change? The effect that I would have expected from this change would be that the focus border only shows up due to keyboard nav and not mouse clicks. However that's not what I'm seeing, the focus still also shows up on mouse clicks just like if we were still using `:focus` here.

The focus styling is applied to the surrounding parent element but it's the `input` that receives the focus so we need to use `:focus-within`

lib/src/components/NxTag/NxTag.tsx
glenhunter: > Based on its definition (https://www.w3.org/TR/wai-aria/#switch) I feel like this role might still be a good idea

Yes: https://github.com/sonatype/sonatype-react-shared-components/pull/566/commits/90758fbf2cfb7327222c9f02b17e385c4620a728

lib/src/components/NxTag/NxTag.scss
rpokorny: > The focus styling is applied to the surrounding parent element but it's the input that receives the focus so we need to use :focus-within

Oh of course, I was thinking of that other new focus thing, `:focus-visible` was it?

lib/src/components/NxTag/NxTag.scss
rpokorny: > I don't recall that decision and it doesn't work in Safari.

Seems line `-webkit-appearance` is also necessary.  See `TransferListHalf.scss`

glenhunter:
> I noticed a few failing tests in`NxTag.test.tsx`:
> some of them just need changing the selector from 'div' to 'label' and changing simulate from 'click' to 'change'.

Fixed here: https://github.com/sonatype/sonatype-react-shared-components/pull/566/commits/f76af5f9c01305011db06d790d9c8f8ec029df12

lib/src/components/NxTag/NxTag.tsx
glenhunter: > > I wonder if this could be a breaking change:
> > In the doc it mentioned that it accepts HTML div attributes: https://gallery.sonatype.dev/#/pages/NxTag
> > I wonder if wrapping the label inside div and moving the classname and attrs back to the div might fix it.
> 
> I suspect `div` attrs are a subset of `label` attrs. If that's true this wouldn't be a breaking change.

The only attributes of `div` are the global attributes. the `label` also includes the global attributes.

lib/src/components/NxTag/NxTag.tsx
rpokorny: > Yes: 90758fb

The switch role requires the `aria-checked` attr according to the spec (though I'm not sure why it really needs to in this case)

lib/src/components/NxTag/NxTag.scss
glenhunter: > > I don't recall that decision and it doesn't work in Safari.
> 
> Seems line `-webkit-appearance` is also necessary. See `TransferListHalf.scss`

`@mixin` created and implemented https://github.com/sonatype/sonatype-react-shared-components/pull/566/commits/eb767c8d2e953f88253039f4938c67c2fe9967b9

lib/src/components/NxTag/NxTag.tsx
glenhunter: > > Yes: [90758fb](https://github.com/sonatype/sonatype-react-shared-components/commit/90758fbf2cfb7327222c9f02b17e385c4620a728)
> 
> The switch role requires the `aria-checked` attr according to the spec (though I'm not sure why it really needs to in this case)

Well it's got it now: https://github.com/sonatype/sonatype-react-shared-components/pull/566/commits/a0ec7fb731847f9020a08368f91254e1fe100fdb

lib/src/scss-shared/_nx-focus-border-helpers.scss
```diff
@@ -15,3 +15,10 @@
     height: calc(#{$height} + 6px);
     width: calc(#{$width} + 6px);
 }
+
+@mixin hide-checkbox {
```
rpokorny, line 19: This doesn't really have to do with focus, I'd rather it be in its own file since I don't think it fits cleanly in any of the existing ones.

rpokorny (approved):
+1 aside from the location of the mixin

PR #573, opened 2021-11-09
AbstractDropdown - RSC-674

entroform:
https://issues.sonatype.org/browse/RSC-674


lib/src/components/NxSegmentedButton/__tests__/NxSegmentedButton.test.tsx
```diff
@@ -141,66 +146,73 @@ describe('NxSegmentedButton', function() {
...
     expect(onToggleOpen).not.toHaveBeenCalled();
 
-    component.find('.nx-segmented-btn__main-btn').simulate('click');
+    component.find('.nx-segmented-btn__main-btn').at(0).simulate('click');
     expect(onToggleOpen).not.toHaveBeenCalled();
 
-    component.find('.nx-segmented-btn__dropdown-btn').simulate('click');
-    expect(onToggleOpen).toHaveBeenCalled();
+    component.find('.nx-segmented-btn__dropdown-btn').at(0).simulate('click');
+    expect(onToggleOpen).not.toHaveBeenCalled();
```
entroform, line 168: I added a `not` here but not 100% sure if it makes sense; we don't want this to be called when the component is disabled right?

lib/src/components/NxDropdown/AbstractDropdown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    ...attrs
+  } = props;
+
+  const {
+    onKeyDown,
+    onToggleCollapse,
+    menuRef,
+    toggleRef,
+    onMenuClosing
+  } = useDropdownEvents(
```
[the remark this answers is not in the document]
entroform: > There's not really a reason to continue to have the events separated out into this hook if it's only used in one place, I'd move all that logic in here directly and get rid of the hook

Yeah that makes sense: Updated it here: 4951b202112da870e8cf078f827f34ca804972a2

lib/src/components/NxDropdown/AbstractDropdown.tsx
```diff
@@ -0,0 +1,152 @@
...
+      if (!defaultPrevented) {
+        onToggleCollapseProp();
+      }
+    }
+  };
+
+  function onToggleCollapse() {
+    // When closing, this is handled by handlDocumentClick. So only do it here when opening
+    // (i.e. when not already open)
+    !isOpen && onToggleCollapseProp && onToggleCollapseProp();
```
rpokorny, line 107: I'm not a huge fan of using `&&` short-circuiting for side-effects, I'd rather use a normal `if` block

lib/src/components/NxDropdown/AbstractDropdown.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+    /* eslint-enable @typescript-eslint/no-non-null-assertion */
+  }
+
+  return (
+    <div ref={ref}
+         className={className}
+         onKeyDown={onKeyDown}
+         {...attrs}
+    >
```
rpokorny, line 141: This seems like it would fit on one line

lib/src/components/NxDropdown/NxDropdown.tsx
```diff
@@ -15,28 +15,23 @@ import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import { wrapTooltipProps } from '../../util/tooltipUtils';
 import './NxDropdown.scss';
 import NxOverflowTooltip from '../NxTooltip/NxOverflowTooltip';
-import NxDropdownMenu from '../NxDropdownMenu/NxDropdownMenu';
-import useDropdownEvents from '../../util/useDropdownEvents';
+
+import AbstractDropdown, {
+  AbstractDropdownToggleElement
+} from './AbstractDropdown';
```
rpokorny, line 21:
```suggestion
import AbstractDropdown, { AbstractDropdownToggleElement } from './AbstractDropdown';
```

lib/src/components/NxDropdown/NxDropdown.tsx
```diff
@@ -50,28 +45,37 @@ const NxDropdown: FunctionComponent<Props> = function NxDropdown(props) {
...
 
   return (
-    <div className={classes} onKeyDown={onKeyDown} {...attrs}>
-      { toggleTooltipProps ? <NxTooltip { ...toggleTooltipProps } >{toggle}</NxTooltip> : toggle }
-      { isOpen && <NxDropdownMenu ref={menuRef} onClosing={onMenuClosing}>{wrappedChildren}</NxDropdownMenu> }
-    </div>
+    <AbstractDropdown className={classes}
+                      isOpen={isOpen}
+                      disabled={disabled}
+                      toggleElement={toggleElement}
```
rpokorny, line 69: I think I'd rather this prop on `AbstractDropdown` be named `renderToggleElement`, since its type is a function which takes some properties from the `AbstractDropdown` and returns the actual (JSX) element

lib/src/components/NxDropdown/__tests__/AbstractDropdown.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    expect(button).toExist();
+
+    expect(onToggleCollapse).not.toHaveBeenCalled();
+
+    button.simulate('click');
+
+    expect(onToggleCollapse).toHaveBeenCalled();
+  });
+
+  it('renders children inside NxDropdownMenu when isOpen is true', function() {
```
rpokorny, line 64: For this test to mean anything you also need to check when `isOpen` is false

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -90,8 +93,10 @@ describe('NxDropdown', () => {
 
   it('uses VDOM as label if supplied', function() {
     const label = <NxFontAwesomeIcon icon={faTrash} />,
-        component = getShallowComponent({ label }),
-        button = component.find(NxButton);
+        component = getMountedComponent({ label });
+
+    const dropdown = component.find(AbstractDropdown);
```
rpokorny, line 98: Is there a need for this change? `.find` searches the whole subtree

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -110,16 +115,16 @@ describe('NxDropdown', () => {
       <button id="link3" className="nx-dropdown-button" key="3">Link3</button>,
       <button id="link4" className="nx-dropdown-right-button" key="4">Link4</button>
     ];
-    const component = getShallowComponent({ children, isOpen: true }),
-        menu = component.find(NxDropdownMenu);
+    const component = getMountedComponent({ children, isOpen: true }),
+        menu = component.find(NxDropdownMenu).find('div');
```
rpokorny, line 119:
```suggestion
        menu = component.find('div.nx-dropdown-menu');
```

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -65,6 +65,7 @@ describe('NxStatefulDropdown', () => {
     expect(mounted.find('.nx-dropdown-menu')).not.toExist();
 
     mounted.find(NxButton).simulate('click');
+
```
rpokorny, line 68: extraneous

lib/src/components/NxDropdown/types.ts
```diff
@@ -4,12 +4,32 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import {ReactNode, HTMLAttributes, ReactElement, WeakValidationMap, KeyboardEventHandler} from 'react';
+import {
+  RefObject,
+  ReactNode, HTMLAttributes, ReactElement, WeakValidationMap, KeyboardEventHandler} from 'react';
```
rpokorny, line 9:
```suggestion
import { RefObject, ReactNode, HTMLAttributes, ReactElement, WeakValidationMap, KeyboardEventHandler} from 'react';
```

lib/src/components/NxDropdown/types.ts
```diff
@@ -4,12 +4,32 @@
...
 
 import { NX_BUTTON_VARIANTS, NX_BUTTON_VARIANT_TYPE } from '../NxButton/types';
 import { TooltipConfigProps, tooltipPropTypesShape } from '../../util/tooltipUtils';
 
+export interface AbstractDropdownToggleElementProps {
+  onToggleCollapse: (() => void);
+  toggleRef: RefObject<HTMLButtonElement>;
+}
+
+export type AbstractDropdownToggleElement = (props: AbstractDropdownToggleElementProps) => ReactNode;
```
rpokorny, line 20: I'd like to draw a bigger distinction between a render function, which is what this is, and a react component proper.  It's worth noting that I think you made the right choice in making this not be an actual component, since it'll typically be a nested function which gets re-instantiated on each render of the parent, which would confuse react and probably degrade performance, as it would think its an entirely new type of component each time.  That said, you've currently got it set up to be an awful lot _like_ a component, which can confuse the reader a little bit.  I think the distinction would be clearer if rather than taking a "props" object, this function simply took he `onToggleCollapse` and `toggleRef` as two separate parameters.

lib/src/components/NxDropdown/types.ts
```diff
@@ -4,12 +4,32 @@
...
+
+export type AbstractDropdownProps = Omit<HTMLAttributes<HTMLDivElement>, 'className'> & {
```
rpokorny, line 22: An `interface` with `extends` is supposed to be a little more performant than an intersection type from what I've heard

lib/src/components/NxDropdown/types.ts
```diff
@@ -4,12 +4,32 @@
...
+  className?: string;
```
rpokorny, line 23: If I remove this line and the `Omit` above it still seems to work

lib/src/components/NxIconDropdown/NxIconDropdown.tsx
```diff
@@ -13,27 +13,20 @@ import NxButton from '../NxButton/NxButton';
 import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import './NxIconDropdown.scss';
 import NxOverflowTooltip from '../NxTooltip/NxOverflowTooltip';
-import NxDropdownMenu from '../NxDropdownMenu/NxDropdownMenu';
-import useDropdownEvents from '../../util/useDropdownEvents';
 
-const NxIconDropdown: FunctionComponent<Props> = function NxIconDropdown(props) {
+import AbstractDropdown, { AbstractDropdownToggleElement } from '../NxDropdown/AbstractDropdown';
+
+const NxIconDropdown: FC<Props> = function NxIconDropdown(props) {
```
rpokorny, line 19: Why the change from `FunctionComponent` to `FC`?

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
```diff
@@ -40,38 +41,38 @@ describe('NxIconDropdown', () => {
...
     expect(component.find(NxDropdownMenu)).toExist();
 
-    component = getShallowComponent({ isOpen: false });
+    component = getMountedComponent({ isOpen: false });
     expect(component.find(NxDropdownMenu)).not.toExist();
   });
 
   it('correctly assigns supplied classes', function() {
-    const component = getShallowComponent({ className: 'class1 class2' });
+    const component = getMountedComponent({ className: 'class1 class2' }).find(AbstractDropdown);
```
rpokorny, line 69: This should go down to the div IMO

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
```diff
@@ -92,20 +93,21 @@ describe('NxIconDropdown', () => {
       <button id="link3" className="nx-dropdown-button" key="3">Link3</button>,
       <button id="link4" className="nx-dropdown-right-button" key="4">Link4</button>
     ];
-    const component = getShallowComponent({ children, isOpen: true }),
-        menu = component.find(NxDropdownMenu);
+    const component = getMountedComponent({ children, isOpen: true }),
+        menu = component.find(NxDropdownMenu).find('div').at(0);
```
rpokorny, line 97: Similar to a suggested change I made on `NxDropdown`, I'd use a single selector string here using both the tag name and class name.

lib/src/components/NxSegmentedButton/NxSegmentedButton.tsx
```diff
@@ -30,29 +30,23 @@ const NxSegmentedButton = forwardRef<HTMLDivElement, Props>(
...
-        <div ref={ref} className={classes} onKeyDown={onKeyDown} { ...attrs }>
+      const toggleElement: AbstractDropdownToggleElement = ({ toggleRef, onToggleCollapse }) => (
+        <>
           <NxButton type={type || undefined}
                     variant={variant}
                     className="nx-segmented-btn__main-btn"
                     onClick={onClick}
                     disabled={disabled || undefined}>
-            {buttonContent}
+            { buttonContent }
```
rpokorny, line 49: This isn't generally how we format this

lib/src/components/NxSegmentedButton/__tests__/NxSegmentedButton.test.tsx
```diff
@@ -76,47 +80,47 @@ describe('NxSegmentedButton', function() {
...
 
   it('sets the variant on the child buttons', function() {
-    const primaryComponent = getShallow(),
-        secondaryComponent = getShallow({ variant: 'secondary' });
+    const primaryComponent = getMounted(),
+        secondaryComponent = getMounted({ variant: 'secondary' });
 
-    expect(primaryComponent.find('.nx-segmented-btn__main-btn')).toHaveProp('variant', 'primary');
-    expect(primaryComponent.find('.nx-segmented-btn__dropdown-btn')).toHaveProp('variant', 'primary');
+    expect(primaryComponent.find('.nx-segmented-btn__main-btn').at(0)).toHaveProp('variant', 'primary');
```
rpokorny, line 94: You can include react component names in string selectors here as if they were html tag names.  That would probably be cleaner.

lib/src/components/NxSegmentedButton/__tests__/NxSegmentedButton.test.tsx
rpokorny: > I added a not here but not 100% sure if it makes sense; we don't want this to be called when the component is disabled right?

Hmm that seems to be what the documentation says, and also how it behaves IRL on main.  This leads me to believe that this test is bad.  I think the not-calling-things behavior IRL is based on the fact that a `disabled` button doesn't fire click events in the first place.  But this test is simulating them as being fired, which doesn't accurately reproduce that scenario.  We already have the test below that the buttons are getting the native `disabled` attr, which is probably all we can really do in a unit test.  If we wanted to be extra sure, we could write a functional test for it.

Is this test passing with your change here? If so I don't understand why.

lib/src/components/NxDropdown/AbstractDropdown.tsx
entroform: > I'm not a huge fan of using `&&` short-circuiting for side-effects, I'd rather use a normal `if` block

Updated here: 4112512fc83b79f755931c83d7df51fa75b88751

lib/src/components/NxDropdown/AbstractDropdown.tsx
entroform: > This seems like it would fit on one line

Updated: 86fda305f98b1a5df940fdaeba47240a69561ec0

lib/src/components/NxDropdown/NxDropdown.tsx
entroform: > I think I'd rather this prop on `AbstractDropdown` be named `renderToggleElement`, since its type is a function which takes some properties from the `AbstractDropdown` and returns the actual (JSX) element

Renamed it here: d080533d2c8381e03a063ff766639f8ef30c2d80 :)

lib/src/components/NxDropdown/types.ts
entroform: > I'd like to draw a bigger distinction between a render function, which is what this is, and a react component proper. It's worth noting that I think you made the right choice in making this not be an actual component, since it'll typically be a nested function which gets re-instantiated on each render of the parent, which would confuse react and probably degrade performance, as it would think its an entirely new type of component each time. That said, you've currently got it set up to be an awful lot _like_ a component, which can confuse the reader a little bit. I think the distinction would be clearer if rather than taking a "props" object, this function simply took the `onToggleCollapse` and `toggleRef` as two separate parameters.

Thanks so much, yeah that makes sense: I updated it here: 50b69374e3165bcd5ab1febc84be2c3716211632

lib/src/components/NxDropdown/types.ts
entroform: > An `interface` with `extends` is supposed to be a little more performant than an intersection type from what I've heard

Oo, I didn't know that. Thanks, updated: fcc0e190116829f7cd732cca279e3d6f789f5f69

lib/src/components/NxIconDropdown/NxIconDropdown.tsx
entroform: I think, when I was playing around with the code, it went over the max-len and I changed it to FC to remove the error temporarily but forgot to change it back or something like that. I changed it back here: a3e9a119a3904ddcecaa7ab82bcc38bcc6ccfcd3

lib/src/components/NxSegmentedButton/NxSegmentedButton.tsx
entroform: > This isn't generally how we format this

Fixed: a3e9a119a3904ddcecaa7ab82bcc38bcc6ccfcd3

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > extraneous

Fixed: a3e9a119a3904ddcecaa7ab82bcc38bcc6ccfcd3

lib/src/components/NxDropdown/__tests__/AbstractDropdown.test.tsx
entroform: > For this test to mean anything you also need to check when `isOpen` is false

Thanks, fixed: 00d164cbbfcaa3c8509a0ae741d5574560f0699c

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > Is there a need for this change? `.find` searches the whole subtree

I forgot to remove this, fixed here: 49aa3582e3f83104b1fe96ee9d123b417d169b53

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
entroform: > This should go down to the div IMO

Updated: 49aa3582e3f83104b1fe96ee9d123b417d169b53

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
entroform: > Similar to a suggested change I made on `NxDropdown`, I'd use a single selector string here using both the tag name and class name.

Updated it here: 49aa3582e3f83104b1fe96ee9d123b417d169b53

lib/src/components/NxSegmentedButton/__tests__/NxSegmentedButton.test.tsx
```diff
@@ -126,9 +130,10 @@ describe('NxSegmentedButton', function() {
     expect(onToggleOpen).not.toHaveBeenCalled();
 
     act(() => {
-      component!.find('button.nx-segmented-btn__dropdown-btn').getDOMNode().dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
+      component!.find('button.nx-segmented-btn__dropdown-btn').at(0)
```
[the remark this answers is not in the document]
entroform: > Why is the `at(0)` necessary here? We are already narrowing it down to the native element due to the `button` part of the selector

Fixed it here: 76eb7e2a8d3dcc466b79be54c6df3aa7d6485d0a

lib/src/components/NxSegmentedButton/__tests__/NxSegmentedButton.test.tsx
entroform: > > I added a not here but not 100% sure if it makes sense; we don't want this to be called when the component is disabled right?
> 
> Hmm that seems to be what the documentation says, and also how it behaves IRL on main. This leads me to believe that this test is bad. I think the not-calling-things behavior IRL is based on the fact that a `disabled` button doesn't fire click events in the first place. But this test is simulating them as being fired, which doesn't accurately reproduce that scenario. We already have the test below that the buttons are getting the native `disabled` attr, which is probably all we can really do in a unit test. If we wanted to be extra sure, we could write a functional test for it.
> 
> Is this test passing with your change here? If so I don't understand why.

Yeah this test is passing strangely (with the change here) and would fail if I remove the `not` 🤔

rpokorny: > Yeah this test is passing strangely (with the change here) and would fail if I remove the not thinking

Well, regardless I think the test is invalid and should be removed.  It's testing what happens when a disabled dropdown receives a click event, and that's just not something that happens.

entroform: > > Yeah this test is passing strangely (with the change here) and would fail if I remove the not thinking
> 
> Well, regardless I think the test is invalid and should be removed. It's testing what happens when a disabled dropdown receives a click event, and that's just not something that happens.

Removed: 8bd59c0065da56673b1f11d0db45bad388ccbdea

lib/src/components/NxSegmentedButton/__tests__/NxSegmentedButton.test.tsx
entroform: > You can include react component names in string selectors here as if they were html tag names. That would probably be cleaner.

Hmm I can't seem to get the selector to work `NxButton.nx-segmented-btn__main-btn` (without the `.at(0)`).
It returns an empty node for me.

rpokorny: > Hmm I can't seem to get the selector to work NxButton.nx-segmented-btn__main-btn (without the .at(0)).
It returns an empty node for me.

Oh it's because of forwardRef ugliness. It seems to work if you write it as `ForwardRef(NxButton).nx-segmented-btn__main-btn`

Btw for the future you can add lines like `console.log(primaryComponent.debug());` to diagnose this sort of thing

entroform: > > Hmm I can't seem to get the selector to work NxButton.nx-segmented-btn__main-btn (without the .at(0)).
> > It returns an empty node for me.
> 
> Oh it's because of forwardRef ugliness. It seems to work if you write it as `ForwardRef(NxButton).nx-segmented-btn__main-btn`
> 
> Btw for the future you can add lines like `console.log(primaryComponent.debug());` to diagnose this sort of thing

Thanks so much, I didn't know you can write it like `ForwardRef(NxButton)`! and the `.debug()` that's very helpful.
Fixed in this commit: 0242c3a2dac0d099ed8fe2a9d1f51f89a6d4633f

rpokorny (approved):

PR #574, opened 2021-11-18
Added missing NxCollapsibleItemsChild test - RSC-778

entroform:
https://issues.sonatype.org/browse/RSC-778

### DIFF
--- a/lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
+++ b/lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
@@ -24,10 +24,6 @@ describe('NxCollapsibleItems', function() {
 
   it('is aliased as NxTreeView', function() {
     expect(NxCollapsibleItems).toBe(NxTreeView);
-  });
-
-  it('has NxCollapsibleItemsChild that is aliased as NxTreeViewChild', function() {
-    expect(NxCollapsibleItemsChild).toBe(NxTreeViewChild);
     expect(NxCollapsibleItems.Child).toBe(NxTreeView.Child);
   });
 
--- a/lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
+++ b/lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
@@ -8,7 +8,7 @@
 import React from 'react';
 
 import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
-import NxCollapsibleItems, { NxCollapsibleItemsChild, Props } from '../NxCollapsibleItems';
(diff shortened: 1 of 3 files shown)


lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
```diff
@@ -26,6 +26,11 @@ describe('NxCollapsibleItems', function() {
     expect(NxCollapsibleItems).toBe(NxTreeView);
   });
 
+  it('has NxCollapsibleItemsChild that is aliased as NxTreeViewChild', function() {
+    expect(NxCollapsibleItemsChild).toBe(NxTreeViewChild);
```
rpokorny, line 30: `NxCollapsibleItemsChild` isn't publicly exported, so it doesn't need to be tested and shouldn't be the one mentioned in the test name

entroform: > `NxCollapsibleItemsChild` isn't publicly exported, so it doesn't need to be tested and shouldn't be the one mentioned in the test name

Oh it is publicly exported. We added the Child sub-component on the rename NxTreeView ticket and we still export the individual component so it's not a breaking change (I think).

rpokorny: > Oh it is publicly exported. We added the Child sub-component on the rename NxTreeView ticket and we still export the individual component so it's not a breaking change (I think).

`NxTreeViewChild` is still exported but `NxCollapsibleItemChild` isn't.  It is only externally available as `NxCollapsibleItems.Child`

lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
```diff
@@ -24,10 +24,6 @@ describe('NxCollapsibleItems', function() {
 
   it('is aliased as NxTreeView', function() {
     expect(NxCollapsibleItems).toBe(NxTreeView);
-  });
-
-  it('has NxCollapsibleItemsChild that is aliased as NxTreeViewChild', function() {
-    expect(NxCollapsibleItemsChild).toBe(NxTreeViewChild);
```
rpokorny: This isn't quite what I meant.  We want `expect(NxCollapsibleItems.Child).toBe(NxTreeViewChild);`

entroform: > This isn't quite what I meant. We want `expect(NxCollapsibleItems.Child).toBe(NxTreeViewChild);`

Thanks, updated it here: 0df577d035319b98ebd84048cd7bf2d151392346

rpokorny (approved):

PR #579, opened 2021-11-25
Updated NxCode and NxTextLink color to match Kaleido - RSC-761

entroform:
https://issues.sonatype.org/browse/RSC-761

I also updated the padding of NxCode to match Sketch (Kaleido) by making the border inclusive.
And added examples of NxTextLink inside NxCode.


gallery/src/components/NxTextLink/NxTextLinkExternalExample.tsx
```diff
@@ -6,10 +6,18 @@
...
-  </NxP>;
+  <>
+    <NxP>
+      The quick brown fox <NxTextLink external href="https://html.spec.whatwg.org/multipage/">MDN</NxTextLink> jumped
+      over the lazy dog.
+    </NxP>
+    <NxP>
+      Learn more about{' '}
+      <NxCode>
+        <NxTextLink external href="https://www.a11yproject.com/">a11y</NxTextLink>
```
rpokorny, line 20: Good thinking including this case, however I think it definitely needs to be checked with design.  I strongly suspect that how it is right now isn't what they'll desire, as the alignment between the icon and the background doesn't look good.
image

lib/src/base-styles/_nx-colors.scss
```diff
@@ -29,7 +29,7 @@
   --nx-color-text-disabled: var(--nx-color-interactive-border-disabled); // match disabled text and border
 
   --nx-color-link: var(--nx-swatch-blue-40);
-  --nx-color-link-focus: var(--nx-swatch-teal-45);
+  --nx-color-link-focus: var(--nx-swatch-blue-30);
```
rpokorny, line 32: It's a little hard to tell, since sketch won't let me click the text in the focus example (I can only click the focus border), but I'm pretty sure the text color of a focused link is supposed to be the same as the text color of an ordinary link.  As a result this variable should be deprecated (it can't be removed for backwards compat reasons) and set to `--nx-color-link`.

lib/src/scss-shared/_nx-text-helpers.scss
```diff
@@ -80,9 +80,9 @@
   @include monospace;
   background-color: var(--nx-swatch-grey-95);
   border: 1px solid var(--nx-swatch-grey-90);
-  border-radius: var(--nx-border-radius);
+  border-radius: 4px;
```
rpokorny, line 83: I'm seeing some mockup issues here.  For one thing, I'm realizing that there is no navigable mockup for plain `nx-code` in kaliedo.  I had to do a search for "code" to find it under the symbols, here: https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/eKOOqjm .  That said, there are aspects of that symbol which do not match the nx-code + nx-text-link mockup but presumably should, notably the border radius.

rpokorny:
We have not, so far, established a rule for whether we support `nx-text-link` within `nx-code`, or the opposite, or both.  You only have examples for the former, but I am wondering if we need examples of the latter as well or instead.  Relatedly, I wonder if we need examples (and designs) for cases where the inner element is only a subsection of an outer one.  For instance, I can imagine a stretch of mostly-normal text which is a link and which has an `nx-code` as just part of it without being the entirety of the link.  The opposite is also imaginable though more of a stretch IMO: an `.nx-code` where only part of its contents are a link.

lib/src/scss-shared/_nx-text-helpers.scss
entroform: Hmm I just noticed that the way Firefox calculates the height of inline elements  is different from Chrome. So the height might be different by around 0.5 to 1 px depending on the typeface. Getting the height matching with Sketch is going to be difficult unless we turn this into `inline-block`.
Additionally, outline-offset scales the border-radius so it's hard to match it with the mockup 🤔 . We could approximate it to get it visually close, I'm going to investigate a way around this.

lib/src/base-styles/_nx-colors.scss
entroform: >  It's a little hard to tell, since sketch won't let me click the text in the focus example (I can only click the focus border), but I'm pretty sure the text color of a focused link is supposed to be the same as the text color of an ordinary link. As a result this variable should be deprecated (it can't be removed for backwards compat reasons) and set to --nx-color-link.

Thanks for catching this, I updated it here: c9bd26d419b0f5529a872cbd7938929541e26672

gallery/src/components/NxTextLink/NxTextLinkExternalExample.tsx
entroform: Good thinking including this case, however I think it definitely needs to be checked with design. I strongly suspect that how it is right now isn't what they'll desire, as the alignment between the icon and the background doesn't look good.

The icon alignment have been fixed here: 41dd3cf609d38fc97bafb710a1a1c6007289155a
Still working on alignment with design on updating Kaleido but we're very close.

entroform:
> We have not, so far, established a rule for whether we support `nx-text-link` within `nx-code`, or the opposite, or both. You only have examples for the former, but I am wondering if we need examples of the latter as well or instead. Relatedly, I wonder if we need examples (and designs) for cases where the inner element is only a subsection of the outer one. For instance, I can imagine a stretch of mostly-normal text which is a link and which has an `nx-code` as just part of it without that code being the entirety of the link. The opposite is also imaginable though more of a stretch IMO: an `.nx-code` where only part of its contents are a link.

Thanks for catching this:
Added the examples here: 49ecd2646bc243c8175f2d76e5ef1217fb16d805.

After talking with the designers it looks like we agreed to add support for both (NxCode inside NxTextLink and NxTextLink inside NxCode). The change is that the text color for when NxCode is inside NxTextLink should match the link color.

The Sketch update for ^ is still in progress.

rpokorny:
> After talking with the designers it looks like we agreed to add support for both (NxCode inside NxTextLink and NxTextLink inside NxCode). The change is that the text color for when NxCode is inside NxTextLink should match the link color.

That covers the partial cases well but I think we need the whole cases as well: where an `nx-code` contains nothing but an `NxTextLink` and vice-versa.  And we need these for both internal and external links.

lib/src/base-styles/_nx-globals.scss
```diff
@@ -18,19 +18,45 @@ $-nx-scrollable-default-height: 400px;
...
+    .nx-code {
+      color: var(--nx-color-link-focus);
+    }
+  }
+
+  .nx-icon {
+    // Offsets the icon down to make it
+    // more optically aligned vertically
+    // to the middle.
+    margin-bottom: -1px;
```
rpokorny, line 59: FA icons have a custom vertical-align value defined in ems.  I think that any custom adjustment of vertical positioning of FA icons that we do should also be done with vertical-align to keep things simple.  I also think it should be done in ems so that it works at different font sizes.

lib/src/base-styles/_nx-globals.scss
```diff
@@ -18,19 +18,45 @@ $-nx-scrollable-default-height: 400px;
...
     outline: 1px solid var(--nx-color-interactive-border-focus);
-    outline-offset: 2px;
+    outline-offset: 4px;
+
+    .nx-code {
+      color: var(--nx-color-link-focus);
+    }
+  }
+
+  .nx-icon {
```
rpokorny, line 55: Should this apply to any `.nx-icon` within a link or just the external link icon?

entroform: > Should this apply to any .nx-icon within a link or just the external link icon?

Thanks that's a good point: Updated it to target that icon specifically here:
d887eb2500280aec071ede3466f809ce6a022191

lib/src/base-styles/_nx-globals.scss
entroform: > FA icons have a custom vertical-align value defined in ems. I think that any custom adjustment of vertical positioning of FA icons that we do should also be done with vertical-align to keep things simple. I also think it should be done in ems so that it works at different font sizes.

Updated to use vertical-align and em here: Should this apply to any .nx-icon within a link or just the external link icon?

entroform:
> > After talking with the designers it looks like we agreed to add support for both (NxCode inside NxTextLink and NxTextLink inside NxCode). The change is that the text color for when NxCode is inside NxTextLink should match the link color.
> 
> That covers the partial cases well but I think we need the whole cases as well: where an `nx-code` contains nothing but an `NxTextLink` and vice-versa. And we need these for both internal and external links.

I updated the examples here:
14f17385ad27b28414ba81644fcf84d698b850d3

I noticed that NxCode inside NxTextLink, the icon is outside the NxCode boundary. Will bring this up with the designers.

Screen Shot 2021-12-07 at 3 00 58 PM

rpokorny:
> I noticed that NxCode inside NxTextLink, the icon is outside the NxCode boundary. Will bring this up with the designers.

That would be next to impossible to fix, I suspect if that's a problem we should jut prohibit that order of nesting

entroform:
> That would be next to impossible to fix, I suspect if that's a problem we should jut prohibit that order of nesting

Yeah agreed.
So after discussions with the designers, it was decided that we recommend `NxTextLink` inside `NxCode ` instead of the the other way around: I updated the descriptions and examples text here: 2b064017627497dcd8b0e73adeb71c658a2b8209

Let me know your thoughts! Also the Sketch document have been updated but it looks like it may be down 🤔

rpokorny:
I'm still seeing inconsistencies in the mockups.  `nx-code` on its own page is 22px tall but nx-code on the nx-text-link page is 21px tall

entroform:
> I'm still seeing inconsistencies in the mockups. `nx-code` on its own page is 22px tall but nx-code on the nx-text-link page is 21px tall

The mockup, Text Link page, have been updated to have 22px tall nx-code! :)

rpokorny:
> The mockup, Text Link page, have been updated to have 22px tall nx-code! :)

Ok now I'm seeing that the nx-text-link focus borders are not the same height - 24px normally but 26px when wrapping an nx-code.  Further, I'm seeing lots of variation in the height of the focus boxes in the various live examples.  This presumably goes back to the different

lib/src/base-styles/_nx-globals.scss
```diff
@@ -18,19 +18,45 @@ $-nx-scrollable-default-height: 400px;
...
+  .nx-code {
+    color: var(--nx-color-link);
+  }
+
   &:hover {
     color: var(--nx-color-link-hover);
+
+    .nx-code {
+      color: var(--nx-color-link-hover);
+    }
```
rpokorny, line 30:
```suggestion
  &:hover {
    &, .nx-code {
      color: var(--nx-color-link-hover);
    }
```

rpokorny:
FYI I'm getting a non-rectangular outline on external links.  Probably not anything we can do about that. 
image

gallery/src/components/NxTextLink/NxTextLinkPage.tsx
```diff
@@ -91,14 +91,24 @@ export default function NxTabsPage() {
                           liveExample={NxTextLinkInternalExample}
                           codeExamples={NxTextLinkInternalExampleCode}>
         A simple <NxCode>NxTextLink</NxCode> to an internal page.
+
+        In the case of link and code that fully overlaps, we recommend that <NxCode>NxTextLink</NxCode>{' '}
+        is nested inside <NxCode>NxCode</NxCode> instead of the the other way around.
```
rpokorny, line 96: Should probably mention the reason why

entroform:
> FYI I'm getting a non-rectangular outline on external links. Probably not anything we can do about that. image

Hmm weird, I'm not seeing this on Chrome and Firefox 🤔. Could it be a linux thing? I did see this issue before though prior to adjusting the icon size.

lib/src/base-styles/_nx-globals.scss
```diff
@@ -18,19 +18,45 @@ $-nx-scrollable-default-height: 400px;
...
+    .nx-code {
+      color: var(--nx-color-link-focus);
+    }
+  }
+
+  .nx-icon.fa-external-link-alt {
+    // Offsets the icon down to make it
+    // more optically aligned vertically
+    // to the middle.
+    vertical-align: -0.16em;
```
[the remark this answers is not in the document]
entroform: > How'd you arrive at 0.16 as the ideal value?

I arrived at the value by playing around until it looks most optically centred with the text but unfortunately the icon stroke weight doesn't really match with the font and scale 🤔

rpokorny:
Active state text color seems to be broken

lib/src/base-styles/_nx-globals.scss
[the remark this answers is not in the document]
rpokorny: > I arrived at the value by playing around until it looks most optically centred with the text but unfortunately the icon stroke weight doesn't really match with the font and scale thinking

As seen in my screenshot on the other comment about the non-rectangular outline, the icon still pretty much touches the nx-code border.  So I'm wondering if it should be moved down further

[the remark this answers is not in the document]
rpokorny: > So I'm wondering if it should be moved down further

Or we could try making the icon a smaller font-size, which would also address some of your other concerns.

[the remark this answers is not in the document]
entroform: > > So I'm wondering if it should be moved down further
> 
> Or we could try making the icon a smaller font-size, which would also address some of your other concerns.

Yeah that's a good idea 👍. I'm consulting with the designers regarding this and the mockup issues

entroform:
> Active state text color seems to be broken

Fixed here: c1f1179ce5c6c7a2b9ece06afc9ed1becae03445

lib/src/base-styles/_nx-globals.scss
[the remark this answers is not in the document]
entroform: > > So I'm wondering if it should be moved down further
> 
> Or we could try making the icon a smaller font-size, which would also address some of your other concerns.

I adjusted the icon size here: 943a11e48c9c8fd5ead6618f8d26d371ffb0352c

entroform:
> > The mockup, Text Link page, have been updated to have 22px tall nx-code! :)
> 
> Ok now I'm seeing that the nx-text-link focus borders are not the same height - 24px normally but 26px when wrapping an nx-code. Further, I'm seeing lots of variation in the height of the focus boxes in the various live examples. This presumably goes back to the different


I consulted with Jason and we updated the heights so they're all consistent. Additionally, added some padding here so that the calculated inline height of the monospace font and the body text is equal.
One of the discrepancy now in the mockup is the horizontal spacing with the icons and the focus border, this is in the work.
943a11e48c9c8fd5ead6618f8d26d371ffb0352c

lib/src/base-styles/_nx-globals.scss
```diff
@@ -18,19 +18,44 @@ $-nx-scrollable-default-height: 400px;
...
+      color: var(--nx-color-link-hover);
+    }
   }
 
   &:focus {
-    border-radius: var(--nx-border-radius);
+    // outline-offset scales the border-radius so this
+    // was changed to 1px to get it visually close to
+    // var(--nx-border-radius) or (6px).
+    border-radius: 1px;
```
rpokorny, line 35: I suspect this needs to be tweaked now that the `outline-offset` has been decreased

lib/src/base-styles/_nx-globals.scss
```diff
@@ -18,19 +18,44 @@ $-nx-scrollable-default-height: 400px;
...
+    .nx-code {
+      color: var(--nx-color-link-focus);
+    }
+  }
+
+  &:active {
+    color: var(--nx-color-link-active);
+
+    .nx-code {
+      color: var(--nx-color-link-active);
```
rpokorny, line 49: Can use the `&, .nx-code {` pattern in these two places as well

lib/src/base-styles/_nx-globals.scss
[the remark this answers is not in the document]
rpokorny: > I adjusted the icon size here: 943a11e

With the smaller icon size I don't think we need the vertical-align customization.  It's making no discernible difference on my screen at normal zoom.

rpokorny:
> I'm still seeing inconsistencies in the mockups. nx-code on its own page is 22px tall but nx-code on the nx-text-link page is 21px tall

There's still an inconsistency in sketch here.  On nx-code's own page it is still 22px tall but currently on the nx-text-link page it is 24px tall

lib/src/base-styles/_nx-globals.scss
entroform: > I suspect this needs to be tweaked now that the `outline-offset` has been decreased

Yup! Totally forgot about this, thanks!
abda55ee13b4fb3a177f2b74d570d49eb66e59e3

lib/src/base-styles/_nx-globals.scss
entroform: > Can use the `&, .nx-code {` pattern in these two places as well

Cleaned it up here: abda55ee13b4fb3a177f2b74d570d49eb66e59e3

lib/src/base-styles/_nx-globals.scss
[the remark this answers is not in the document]
entroform: > > I adjusted the icon size here: [943a11e](https://github.com/sonatype/sonatype-react-shared-components/commit/943a11e48c9c8fd5ead6618f8d26d371ffb0352c)
> 
> With the smaller icon size I don't think we need the vertical-align customization. It's making no discernible difference on my screen at normal zoom.

Thanks! I removed it here: 784c8879019874be629808e13eb6c547e73afc46

entroform:
> > I'm still seeing inconsistencies in the mockups. nx-code on its own page is 22px tall but nx-code on the nx-text-link page is 21px tall
> 
> There's still an inconsistency in sketch here. On nx-code's own page it is still 22px tall but currently on the nx-text-link page it is 24px tall

Reached out to Jason and they're in the progress of updating it 👍

rpokorny (approved):

PR #580, opened 2021-11-30
Fixed page title truncation - RSC-813

entroform:
https://issues.sonatype.org/browse/RSC-813


lib/src/base-styles/_nx-page.scss
```diff
@@ -25,6 +25,7 @@
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
+    max-width: max-content;
```
rpokorny, line 28: `unset` would probably be simpler

entroform: > 

Thanks! I updated it here: 1ba2b3704241535dd70661b3e257907df46c96b6

rpokorny (approved):

PR #582, opened 2021-12-02
Spike: Checking accessibility properties at the linter level

rpokorny:
https://issues.sonatype.org/browse/RSC-431

This PR is for one of three subtasks aimed at looking into automated testing of our accessibility.  This PR adds an eslint plugin that verifies a variety of rules primarily around aria usage.  The full set of rules can be seen here: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#supported-rules.  I've stuck to the recommended set of rules as defined by the `"plugin:jsx-a11y/recommended"` configuration, except that I've added some additional clarification to give it necessary knowledge of some of our components.

Pros:
* Simple to add to the project, likely simpler than adding and maintaining unit or functional tests
* Seems thorough in catching standards compliance issues with regards to aria tags
* Like any eslint plugin, all rules are individually configurable

Debatable:
* Given our existing eslint setup, the plugin applies not just to the lib but also to the gallery.  This required making some examples a bit more "real" than they were previously

Cons:
* There were a decent number of special cases where rules had to be disabled. In some cases, this was due to real-world screenreaders needing the exact opposite of what the spec says. (Example: spec says don't add redundant roles to elements, screenreader doesn't know the built-in role of certain newer HTML elements and so we must add the role). Note however that this would be just as much of a problem with any other validator that is as thorough as this one.
* Concerns about severe opinionation/attitude on the part of the plugin author (discussed more in Slack) which could potentially lead to the plugin trying to enforce things that aren't really part of any standard and which we don't agree with.  Generally this shouldn't cause blockers though since the rules are individually configurable.

All in all I'd say I'm in favor of this approach though we should be aware that it will need to be overridden from time to time.

Note that there are a handlful of places in this PR where I did make changes to aria attributes.  I checked these changes in ChromeVox and make sure they did not cause anything to become worse.  I would appreciate it if a mac user checked them in VoiceOver as well.


gallery/src/GalleryNav.tsx
```diff
@@ -125,7 +125,6 @@ function GalleryNav() {
       <NxFilterInput placeholder="Search RSC…"
                      value={filter}
                      onChange={setFilter}
-                     autoFocus
```
rpokorny, line 128: Consensus opinion in the wild is that autofocus is bad for a11y.  @JenniKobylski has confirmed that we will stick to that opinion.

gallery/src/components/NxDropdown/NxDropdownLinksExample.tsx
```diff
@@ -10,26 +10,24 @@ import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons';
...
-      onClick = () => { alert('click'); };
+  const [isOpen, onToggleCollapse] = useToggle(false);
 
   return (
     <NxDropdown label="Links!"
                 className="extra-class"
                 isOpen={isOpen}
                 onToggleCollapse={onToggleCollapse}>
-      <a className="nx-dropdown-link disabled">
+      <a href="#/" className="nx-dropdown-link disabled">
```
rpokorny, line 20: Plugin enforces that links are actual links. A lot of our examples were a bit lazy on this point and needed updating.

Note also that the plugin doesn't consider `#` to be good enough as a URL.  This is one of the plugin authors opinions that I find contentious.  I'm familiar with the bootstrap-style code where `href="#"` is used as an inaccessible cop out, which is what the plugin is trying to block, but there are other ways to use it and that's not how we were using it here.  At any rate though, it was easy to work around as our ui router treats `#` and `#/` the same.

gallery/src/components/NxDropdown/NxDropdownScrollingExample.tsx
```diff
@@ -9,50 +9,52 @@ import React from 'react';
...
       </a>
-      <a href="#" onClick={onClick} className="nx-dropdown-button">
+      <a href="#/" className="nx-dropdown-button">
         Text Link 6
       </a>
-      <a href="#" className="disabled nx-dropdown-button">
+      <a href="#/"
+         onClick={evt => { evt.preventDefault(); }}
+         className="disabled nx-dropdown-button"
+         aria-disabled="true">
```
rpokorny, line 39: Aside from the href, the improvements on this element weren't actually things that came from the plugin; just my own tidying up.

gallery/src/components/NxSubmitMask/NxSubmitMaskCustomMessageExample.tsx
```diff
@@ -17,8 +17,10 @@ export default function NxSubmitMaskCustomMessageExample() {
   }
 
   return (
-    <div onKeyUp={handleKeyUp} className="gallery-submit-mask-area">
-      <NxButton onClick={setMaskVisible.bind(null, true)}>Click here to create a NxSubmitMask</NxButton>
+    <div className="gallery-submit-mask-area">
+      <NxButton onKeyUp={handleKeyUp} onClick={setMaskVisible.bind(null, true)}>
```
rpokorny, line 21: This keyup handler is kind of hacky, it's just to give the user a way to get out of the example.  When it was on the div, the plugin wanted the div to have a role.  I couldn't really think of one and it works just as well on the button, so...

gallery/src/styles/NxList/NxListClickableLinksExample.tsx
```diff
@@ -39,8 +43,11 @@ const NxListClickableLinksExample = () =>
         <NxFontAwesomeIcon icon={faAngleRight} className="nx-chevron" />
       </a>
     </li>
-    <li className="nx-list__item nx-list__item--clickable" aria-disabled="true">
-      <a href="#/pages/nx-alert" className="nx-list__link disabled">
+    <li className="nx-list__item nx-list__item--clickable">
+      <a href="#/pages/nx-alert"
+         onClick={evt => { evt.preventDefault(); }}
+         className="nx-list__link disabled"
+         aria-disabled="true">
```
rpokorny, line 50: The main difference here is that the plugin pointed out that `aria-disabled` isn't valid on the list item itself. But it seems to be valid on the link.

lib/.eslintrc
```diff
@@ -148,7 +149,10 @@
     "react/no-unescaped-entities": "off",
     "react/jsx-no-target-blank": "off",
     "react/jsx-fragments": ["error", "syntax"],
-    "react/display-name": "off"
+    "react/display-name": "off",
+    "jsx-a11y/label-has-associated-control": [ 2, {
+      "controlComponents": ["NxTextInput", "NxStatefulTextInput", "NxFilterInput"]
```
rpokorny, line 154: This rule verifies that `<label>` elements either refer to an `<input>` by id or contain an input.  Since this is static analysis, the plugin can't automatically know whether a given react component within the `<label>` will render an `<input>`.  Here we tell it which components do.

lib/.eslintrc
```diff
@@ -1,158 +1,9 @@
 {
```
rpokorny, line 1: This needed to be split up into two files: the base file which is effectively exactly what this used to be, and then this file which extends the base and ads the a11y linting to it.  This was done so that the unit test linting configuration can extend the base file and thereby avoid a11y linting, which IMO would just get in the way for unit tests.

entroform (approved):
LGTM!

avnvbhatta (approved):
While I agree on not confirming to non-standard rules set by the author, I do believe that the pros outweigh the cons. I think this will be a decent addition.

PR #587, opened 2021-12-10
NxDateInput proptypes errors - RSC-829

rpokorny:
https://issues.sonatype.org/browse/RSC-829

To fix this while still enforcing the rules that we want around public usage of `NxTextInput` I decided to split all of `NxTextInput`'s hidden functionality into a separate component that is not exposed outside of RSC.  `NxTextInput` itself is now a simple wrapper around that component which has more constrained typescript types and propTypes (the internal component doesn't have any propTypes).  A similar change was made to `NxStatefulTextInput`


entroform (approved):
Thanks, this looks good to me!

PR #588, opened 2021-12-10
Spike: a11y functional testing - RSC-430

rpokorny:
https://issues.sonatype.org/browse/RSC-430

This PR adds accessibility functional tests to some pages and fixes the issues that arose.  Enabling this testing on the remaining pages would be a large amount of additional work that can be undertaken if/after this approach is accepted; it is not necessary for demonstrating how this works.


entroform (approved):
LGTM!

PR #589, opened 2021-12-13
Add Gainsight PX to the gallery - RSC-782

rpokorny:
https://issues.sonatype.org/browse/RSC-782

### DIFF
--- a/gallery/src/telemetry.js
+++ b/gallery/src/telemetry.js
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+export function init() {
+  // Filled in via the webpack EnvironmentPlugin
+  /* eslint-disable-next-line no-undef */
+  const apiKey = process.env.PX_API_KEY;
+
+  if (apiKey) {
+    // This blob of code is provided by the gainsight admin interface to initialize PX. For legal reasons, it
+    // presumably cannot be de-obfuscated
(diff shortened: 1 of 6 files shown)


gallery/src/telemetry.js
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+export function init() {
+  // Filled in via the webpack EnvironmentPlugin
+  /* eslint-disable-next-line no-undef */
+  const apiKey = process.env.PX_API_KEY;
```
rpokorny, line 11: To verify that this works, you can search for the API key value in the `bundle.js` that was produced by the latest jenkins build, and see that it is present.

gallery/webpack.config.js
```diff
@@ -115,6 +116,9 @@ module.exports = function(env = { production: false }) {
       new MiniCssExtractPlugin({
         filename: 'gallery.css'
       }),
+      new webpack.EnvironmentPlugin({
+        'PX_API_KEY': ''
```
rpokorny, line 120: This sets the value within our code to an empty string unless the `PX_API_KEY` environment variable is set during the build, in which case the env var gets passed in as the value in the code.

gallery/src/telemetry.js
glenhunter: > To verify that this works, you can search for the API key value in the `bundle.js` that was produced by the latest jenkins build, and see that it is present. Once this code is merged to main, we will be able to see it in action on gallery.sonatype.dev

I was able to verify that the API Key is present in the `bundle.js`

glenhunter (approved):

PR #591, opened 2021-12-13
SPIKE - RSC-815 Intelligent Dropdown Placement

avnvbhatta:
https://issues.sonatype.org/browse/RSC-815


lib/src/components/NxDropdownMenu/NxDropdownMenu.tsx
```diff
@@ -19,8 +19,9 @@ export { Props };
  */
 /* eslint-disable-next-line react/prop-types */
 const NxDropdownMenu = forwardRef<HTMLDivElement, Props>(function NxDropdownMenu(props, ref) {
-  const { onClosing, className: classNameProp, ...attrs } = props,
-      className = classnames('nx-dropdown-menu', classNameProp);
+  const { onClosing, className: classNameProp, openTop, openRight, ...attrs } = props,
+      className = classnames('nx-dropdown-menu', classNameProp, {'nx-dropdown-menu__open-top': openTop},
+          {'nx-dropdown-menu__open-right': openRight});
```
rpokorny, line 24:
```suggestion
      className = classnames('nx-dropdown-menu', classNameProp, {
        'nx-dropdown-menu__open-top': openTop,
        'nx-dropdown-menu__open-right': openRight
      });
```

lib/src/util/useDynamicDropdownPlacement.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @param childrenCount the number of items the dropdown menu contains to dynamically allocate spacing.
+ * @return a [boolean, boolean] if the dropdown menu should be placed on top and on the right respectively.
+ */
+
+const DROPDOWN_ITEM_HEIGHT = 32; //32px is the height of an item in the dropdown
+const ICON_ONLY_DROPDOWN_WIDTH = 250; //250px is the width of the dropdown
+
+const useDynamicDropdownPlacement = (
+  dropdownDivRef: RefObject<HTMLDivElement>,
+  childrenCount: number,
```
rpokorny, line 24: Rather than computing the number of JSX children and then doing math that makes assumptions about the styling, why not just measure the div?

lib/src/util/useDynamicDropdownPlacement.ts
```diff
@@ -0,0 +1,63 @@
...
+) => {
+  const [openTop, setOpenTop] = useState<boolean>(false);
+  const [openRight, setOpenRight] = useState<boolean>(false);
+
+  // The event listeners look for changes to be made to the dropdown position if the user
+  // changes the window size by either scrolling the page or resizing the page.
+  // childrenCount is added as a dependency because in components like NxSearchDropdown, the height of the
+  // menu depends on the number of results/menu items shown in the component.
+  useEffect(() => {
+    window.addEventListener('scroll', determinePlacement);
```
rpokorny, line 34: I think we should get design input on whether it should move when scrolling occurs.  Additionally this implementation is limited in that it won't handle scrolling of containers other than the viewport itself

lib/src/util/useDynamicDropdownPlacement.ts
```diff
@@ -0,0 +1,63 @@
...
+    window.addEventListener('resize', determinePlacement);
```
rpokorny, line 35: Consider `useResizeObserver` to check for resizes of the element's container, not just the viewport

rpokorny:
The more I think about this the more different cases I see that we need to determine whether or not they need special behavior.  Here's a list of what I can think of so far:

1. When the dropdown is at the bottom of the document, such that the menu would either be inaccessible by scrolling or would itself increase the scroll length of the page
2. When `#1` doesn't apply, but the dropdown is currently close to the bottom of the _viewport_ due to scrolling, such that its menu would open below the end of the viewport and would need to be scrolled to get to
3. When the dropdown is at the bottom of some container such that its menu overflows out of the bottom of the container, but is still visible and interactive
4. When the dropdown is at the bottom of some container such that its menu overflows out of the bottom of the container, but the container hides the overflowing part making it inaccessible to the user
5. When the dropdown is in a scroll container and the scroll position has the dropdown currently close to the bottom of the container, such that the container would need to be scrolled downwards to see the dropdown
6. When an icon dropdown is close to the left edge of the window such that its menu is cut off when opened
7. When an icon dropdown is close to the left edge of its container such that the menu overflows out the side of the container, but the container does not cut it off.
8. When an icon dropdown is close to the left edge of its container such that the menu overflows out the side of the container, but the container `does` cut it off via hiding overflow.
9. When an icon dropdown menu becomes cut off due to horizontal scrolling of the window
10. When an icon dropdown menu becomes cut off due to horizontal scrolling of its container.

Some of these are more feasible to address than others, and some are more important to address than others.  The most important ones IMO, which are also some of the most doable, are 1 and 6.  Of the rest, many are more about convenience than  breakage (does the user need to scroll to see the menu), and some would be very difficult (trying to figure out whether a parent container is or isn't cutting off an overflowing menu)

PR #595, opened 2021-12-16
Dropdown Overlap - RSC-834

avnvbhatta:
https://issues.sonatype.org/browse/RSC-834

### DIFF
--- a/lib/src/base-styles/_nx-page-layout.scss
+++ b/lib/src/base-styles/_nx-page-layout.scss
@@ -106,6 +106,7 @@
   overflow-y: auto;
   padding: var(--nx-spacing-6x) var(--nx-spacing-6x);
   white-space: normal;
+  z-index: 0;
 }
 
 .nx-html--page-scrolling {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "9.5.1",
+  "version": "9.5.2",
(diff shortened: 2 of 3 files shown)


lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -106,6 +106,7 @@
   overflow-y: auto;
   padding: var(--nx-spacing-6x) var(--nx-spacing-6x);
   white-space: normal;
+  z-index: 0;
```
rpokorny, line 109: I think this should probably also be added to `.nx-page-sidebar` so that a dropdown in a sidebar doesn't have the same problem

rpokorny (approved):
+1 but with some hesitation.  This is the kind of change that could cause some obscure bug somewhere completely unrelated to dropdowns.  This should be tested carefully and obviously any applitools differences should be taken seriously.

avnvbhatta:
> +1 but with some hesitation. This is the kind of change that could cause some obscure bug somewhere completely unrelated to dropdowns. This should be tested carefully and obviously any applitools differences should be taken seriously.

I agree. I mean if we haven't heard about this issue from downstream users, so maybe a safer approach is to update the `z-index` of the header?

rpokorny:
> I agree. I mean if we haven't heard about this issue from downstream users, so maybe a safer approach is to update the z-index of the header?

That's probably just as dangerous.  It could result in the header appearing in front of modals or something.  Actually, this fix might cause that too but in a different way.  Maybe do some testing _before_ you merge it.

PR #597, opened 2021-12-16
fix overflow bug RSC-833

glenhunter:
https://issues.sonatype.org/browse/RSC-833

I thought there were be a better solution than brute forcing the input width but I couldn't find one.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "9.5.2",
+  "version": "9.5.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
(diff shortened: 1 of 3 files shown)


rpokorny:
I think `min-width: 0` is the correct solution.  You're running into the "`min-width: auto` means don't shrink below my natural width" problem, combined with (I believe) the browser's default size for `<input>`

glenhunter:
> I think `min-width: 0` is the correct solution. You're running into the "`min-width: auto` means don't shrink below my natural width" problem, combined with (I believe) the browser's default size for `<input>`

Thanks, that was it: https://github.com/sonatype/sonatype-react-shared-components/pull/597/commits/f0b0b558798dea44c56778711b1deba95b165e53

avnvbhatta (approved):

rpokorny (approved):

PR #598, opened 2021-12-16
Fixed NxSegmentedButton height - RSC-836

entroform:
https://issues.sonatype.org/browse/RSC-836

### DIFF
--- a/lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
+++ b/lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
@@ -69,8 +69,8 @@
       border-bottom-style: none;
       border-left-style: none;
 
-      padding-top: calc(var(--nx-spacing-2x) + 1px);
-      padding-bottom: calc(var(--nx-spacing-2x) + 1px);
+      padding-top: var(--nx-spacing-2x);
+      padding-bottom: var(--nx-spacing-2x);
       padding-left: calc(var(--nx-spacing-6x) + 1px);
     }
   }
@@ -92,9 +92,9 @@
       border-right-style: none;
       border-bottom-style: none;
 
-      padding-top: calc(var(--nx-spacing-2x) + 1px);
+      padding-top: var(--nx-spacing-2x);
       padding-right: calc(var(--nx-spacing-3x) + 1px);
-      padding-bottom: calc(var(--nx-spacing-2x) + 1px);
+      padding-bottom: var(--nx-spacing-2x);
     }
   }
 }
--- a/lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
+++ b/lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
@@ -47,7 +47,7 @@
   padding-right: calc(var(--nx-spacing-3x) - 1px);
 
   > .nx-icon {
-    font-size: 1.5em;
+    font-size: 1.2em;
   }
 }
 
--- a/lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
+++ b/lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
@@ -43,8 +43,8 @@
   // lock the line-height here so that it is not affected by the font-size increase on the icon
   line-height: 1em;
   margin: 0;
-  padding-left: var(--nx-spacing-3x);
-  padding-right: var(--nx-spacing-3x);
+  padding-left: calc(var(--nx-spacing-3x) - 1px);
+  padding-right: calc(var(--nx-spacing-3x) - 1px);
 
   > .nx-icon {
     font-size: 1.5em;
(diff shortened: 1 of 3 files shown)


lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -69,8 +69,8 @@
       border-bottom-style: none;
       border-left-style: none;
 
-      padding-top: calc(var(--nx-spacing-2x) + 1px);
-      padding-bottom: calc(var(--nx-spacing-2x) + 1px);
+      padding-top: var(--nx-spacing-2x);
+      padding-bottom: var(--nx-spacing-2x);
       padding-left: calc(var(--nx-spacing-6x) + 1px);
```
rpokorny, line 74: I suspect this and the `+1` on the other half's `padding-right` also aren't supposed to be here anymore.  It would be weird to include the border in the spacing on one axis but not the other.

entroform: > I suspect this and the `+1` on the other half's `padding-right` also aren't supposed to be here anymore. It would be weird to include the border in the spacing on one axis but not the other.

Fixed here: ff880780732be0bf892fef562bbd689c617c28d9

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -43,8 +43,8 @@
   // lock the line-height here so that it is not affected by the font-size increase on the icon
   line-height: 1em;
   margin: 0;
-  padding-left: var(--nx-spacing-3x);
-  padding-right: var(--nx-spacing-3x);
+  padding-left: calc(var(--nx-spacing-3x) - 1px);
+  padding-right: calc(var(--nx-spacing-3x) - 1px);
```
avnvbhatta, line 47: Why do we need the `-1 px`?

entroform: > Why do we need the `-1 px`? I think we can get rid all of the addition and subtraction?

Oh the `-1px` is so border is inclusive as part of the padding.

rpokorny: > Oh the -1px is so border is inclusive as part of the padding.

This is on the dropdown arrow part of the button.  Checking sketch closely, the arrow itself isn't even the same size in the implementation vs the mockup so that needs to be fixed before we can really know what's supposed to be done here.

rpokorny: > This is on the dropdown arrow part of the button. Checking sketch closely, the arrow itself isn't even the same size in the implementation vs the mockup so that needs to be fixed before we can really know what's supposed to be done here.

That said I do suspect at least the right-side `-1` is correct here.  The style it's overridding also uses `-1` and it makes sense with the border.  The inner distance (left padding) though, I'm not sure about.

entroform: > > This is on the dropdown arrow part of the button. Checking sketch closely, the arrow itself isn't even the same size in the implementation vs the mockup so that needs to be fixed before we can really know what's supposed to be done here.
> 
> That said I do suspect at least the right-side `-1` is correct here. The style it's overridding also uses `-1` and it makes sense with the border. The inner distance (left padding) though, I'm not sure about.

I updated the icon size here: 307103d22bc47cf81b89614255a161921bf83652

The total width is now 36 (12px padding with the borders + 12px icon) compared to 38px (13px padding with the borders + 12px icon) in the mockup.
I think the mockup might not be updated as intended: the border is not inclusive as part of the horizontal padding but the vertical padding is 🤔. I'll consult with the designers!

rpokorny: > I think the mockup might not be updated as intended: the border width is not inclusive as part of the horizontal padding but the vertical padding is thinking. I'll consult with the designers!

The mockups are a little odd in regards to the divider line between the two parts of the button.  For one thing, the spacing appears to be different in primary buttons vs secondary!  But on secondary buttons it appears that he spacing to right of the divider is a standard distance while the spacing to the left of the divider is one less.  Which seems inconsistent to me.

entroform: > > I think the mockup might not be updated as intended: the border width is not inclusive as part of the horizontal padding but the vertical padding is thinking. I'll consult with the designers!
> 
> The mockups are a little odd in regards to the divider line between the two parts of the button. For one thing, the spacing appears to be different in primary buttons vs secondary! But on secondary buttons it appears that he spacing to right of the divider is a standard distance while the spacing to the left of the divider is one less. Which seems inconsistent to me.

I've consulted with Jason with this and they updated the Sketch file so it's all consistent.

rpokorny (approved):

PR #603, opened 2021-12-23
Updated none text color to white of NxSmallThreatCounter - RSC-839

entroform:
https://issues.sonatype.org/browse/RSC-839

Reached out to Jason, they will update the mockup soon.


rpokorny (approved):
+1 assuming a designer has confirmed that this is the desired fix

PR #604, opened 2022-01-03
Change Unspecified Color - RSC-835

avnvbhatta:
https://issues.sonatype.org/browse/RSC-835


rpokorny (approved):
+1 as long as all applitools diffs seem right

PR #605, opened 2022-01-04
Source Code Display Collapsed by Default - RSC-841

avnvbhatta:
https://issues.sonatype.org/browse/RSC-841

### DIFF
--- a/gallery/src/CodeExample.tsx
+++ b/gallery/src/CodeExample.tsx
@@ -33,7 +33,7 @@ const CodeExample: FunctionComponent<Props> =
       <section className="gallery-example-code">
         { language &&
           <h3 className="nx-h3 nx-tile__section-header">
-            Code Snippet ({language.toUpperCase()}):
+            {language.toUpperCase()}:
           </h3>
         }
         <SyntaxHighlighter language={language || 'tsx'} style={atomDark}>
--- a/gallery/src/CodeExample.tsx
+++ b/gallery/src/CodeExample.tsx
@@ -28,16 +28,14 @@ const removeLicense = (content: string) =>
 
 const CodeExample: FunctionComponent<Props> =
   function CodeExample({ content, language }): ReactElement<Props> {
-    const licenseStrippedContent = removeLicense(content),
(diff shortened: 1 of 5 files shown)


gallery/src/CodeExample.tsx
```diff
@@ -28,16 +28,14 @@ const removeLicense = (content: string) =>
...
     return (
       <section className="gallery-example-code">
-        <h3 className="nx-h3 nx-tile__section-header">
-          Code Snippet
-          {headerLanguagePart}
-          :
-        </h3>
+        { language &&
+          <h3 className="nx-h3 nx-tile__section-header">
+            Code Snippet ({language.toUpperCase()}):
```
rpokorny, line 36: Would it make more sense now if it just said the language name instead of preceding it with "Code Snippet"?

gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -120,16 +120,16 @@ export const GalleryExampleTile: FunctionComponent<GalleryExampleTileProps> =
           }
         </NxTile.Content>
 
-        { collapseCodeExample ?
+        { uncollapseCodeExample ?
+          <NxTile.Content>{codeExampleElements}</NxTile.Content> :
```
rpokorny, line 124: Is there a reason to still have the option to uncollapse?

avnvbhatta: > Is there a reason to still have the option to uncollapse?

Not really now that I think about it. Removed here: 67a94c06376d861325202cdcc7e1b5b93aec5bb9

rpokorny (approved):

gallery/src/CodeExample.tsx
avnvbhatta: > > Would it make more sense now if it just said the language name instead of preceding it with "Code Snippet"?
> 
> I'm fine with either. Any preference @entroform and @glenhunter ?

Removed here: 3447c8ab275c0221c3e3f4f17bea6f713895ef4a

PR #606, opened 2022-01-05
Added custom NxList style when it is nested inside NxAccordion - RSC-824

entroform:
https://issues.sonatype.org/browse/RSC-824

### DIFF
--- a/gallery/src/components/NxAccordion/NxAccordionPage.tsx
+++ b/gallery/src/components/NxAccordion/NxAccordionPage.tsx
@@ -12,10 +12,12 @@ import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-compon
 import NxAccordionSimpleExample from './NxAccordionExample';
 import NxAccordionComplexExample from './NxAccordionComplexExample';
 import NxAccordionTertiaryButtonExample from './NxAccordionTertiaryButtonExample';
+import NxAccordionWithNxListExample from './NxAccordionWithNxListExample';
 
 const NxAccordionSimpleCode = require('./NxAccordionExample?raw'),
     NxAccordionComplexCode = require('./NxAccordionComplexExample?raw'),
-    NxAccordionTertiaryButtonCode = require('./NxAccordionTertiaryButtonExample?raw');
+    NxAccordionTertiaryButtonCode = require('./NxAccordionTertiaryButtonExample?raw'),
+    NxAccordionWithNxListCode = require('./NxAccordionWithNxListExample?raw');
 
 const NxAccordionPage = () =>
   <>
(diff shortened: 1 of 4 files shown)


lib/src/components/NxAccordion/NxAccordion.scss
```diff
@@ -81,4 +81,14 @@
   border-top: 1px solid #e5e8f4;
   margin: 0 var(--nx-spacing-6x);
   padding: var(--nx-spacing-6x) 0;
+
+  .nx-list:first-child {
+    margin-top: calc(var(--nx-spacing-6x) * -1);
+
+    .nx-list__item:first-child {
+      &, > .nx-list__link, & > .nx-list__btn {
```
avnvbhatta, line 89: Do we need to specify these selectors?

entroform: > Do we need to specify these selectors?

It looks like we need it, without these selectors, the accordion divider line looks darker and thicker when the items are `NxList.LinkItem` or `NxList.ButtonItem`.

lib/src/components/NxAccordion/NxAccordion.scss
```diff
@@ -81,4 +81,14 @@
   border-top: 1px solid #e5e8f4;
   margin: 0 var(--nx-spacing-6x);
   padding: var(--nx-spacing-6x) 0;
+
+  .nx-list:first-child {
+    margin-top: calc(var(--nx-spacing-6x) * -1);
```
rpokorny, line 86: To keep it connected and make the purpose of this line more clear, the top padding of the accordion content area should probably be extracted into a scss variable

lib/src/components/NxAccordion/NxAccordion.scss
rpokorny: > It looks like we need it, without these selectors, the accordion divider line looks darker and thicker when the items are NxList.LinkItem or NxList.ButtonItem.

We should have gallery examples of these cases then so they can have visual tests.  I'd say just put all three lists (normal, button, and link) in the same example as three separate accordions in a row

rpokorny:
Needs visual tests

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -160,6 +162,14 @@ const NxAccordionPage = () =>
       A simple example of an <NxCode>NxAccordion</NxCode>.
     </GalleryExampleTile>
 
+    <GalleryExampleTile title="Example with nested NxList"
+                        id="nx-accordion-nested-nx-list-example"
+                        defaultCheckeredBackground={true}
+                        liveExample={NxAccordionWithNxListExample}
+                        codeExamples={NxAccordionWithNxListCode}>
+      An example of an <NxCode>NxAccordion</NxCode> with nested <NxCode>NxList</NxCode>.
```
rpokorny, line 170: I think it would be worthwhile to add a brief mention of this be a special stylistic case which is why this example exists

entroform: > I think it would be worthwhile to add a brief mention of this be a special stylistic case which is why this example exists

Good point! I added an extra description here: dad83b7bca34e011e27cd8acb6202874e227dd82

lib/src/components/NxAccordion/NxAccordion.scss
entroform: > To keep it connected and make the purpose of this line more clear, the top padding of the accordion content area should probably be extracted into a scss variable

Updated it here! 27f715e95bfb758837314828d2958a46ad3e369f

avnvbhatta (approved):

rpokorny (approved):

PR #607, opened 2022-01-06
SPIKE merge examples in tabs RSC-842

glenhunter:
https://issues.sonatype.org/browse/RSC-842

This is a spike to demonstrate changes to the Gallery layout. The goal is to consolidate stateless, stateful, React, and HTML examples on one page using tabs. An additional goal is to reduce the length of the pages by showing the usage/props/etc in a tab separate from the examples.

Please look at the NxAccordion and NxTable pages (under components) for examples.

### DIFF
--- a/gallery/src/components/NxAccordion/NxAccordionPage.tsx
+++ b/gallery/src/components/NxAccordion/NxAccordionPage.tsx
@@ -50,7 +50,7 @@ export default function NxAccordionPage() {
               NxAccordion renders a panel with an always-visible header and a collapsible/expandable body section.
               This is analogous to the HTML 5 <NxCode>&lt;details&gt;</NxCode> element (which it is
               implemented on top of). There are three related components: <NxCode>NxAccordion</NxCode> itself,
-              <NxCode>NxAccordion.Header</NxCode> which represents the header content, and
+              {' '}<NxCode>NxAccordion.Header</NxCode> which represents the header content, and
               {' '}<NxCode>NxAccordion.Title</NxCode> which is a convenience component for the header title. All other
               children of <NxCode>NxAccordion</NxCode> aside from
               the <NxCode>Header</NxCode> and the <NxCode>Title</NxCode> are rendered in the collapsible section.
@@ -77,14 +77,14 @@ export default function NxAccordionPage() {
             <NxList bulleted>
               <NxList.Item>
                 Displaying critical system information or a primary action to be taken on the page. (for example,
-                <a href="/components/alert">alerts</a>, confirmation or cancellation buttons).
+                {' '}<a href="/components/alert">alerts</a>, confirmation or cancellation buttons).
               </NxList.Item>
               <NxList.Item>
                 Displaying navigation elements such as <a href="/components/tabs">tabs</a>.
               </NxList.Item>
               <NxList.Item>
                 Displaying links pointing to sections of the same page, instead use a
-                <a href="/components/list">list</a>.
+                {' '}<a href="/components/list">list</a>.
               </NxList.Item>
               <NxList.Item>
                 Creating hierarchy levels by nesting them within each other. If you need to add hierarchy to the
--- a/gallery/src/components/NxAccordion/NxAccordionPage.tsx
+++ b/gallery/src/components/NxAccordion/NxAccordionPage.tsx
@@ -4,184 +4,265 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React from 'react';
-import { NxTable, NxInfoAlert, NxCode, NxTextLink, NxP, NxH3, NxTile } from '@sonatype/react-shared-components';
+import React, {useState} from 'react';
+import {
+  NxTable,
+  NxInfoAlert,
+  NxCode,
+  NxTextLink,
+  NxP,
+  NxH3,
+  NxTile,
+  NxTabs,
+  NxTabList,
+  NxTab,
+  NxTabPanel,
+  NxH2,
+  NxList
+} from '@sonatype/react-shared-components';
 
 import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
 
 import NxAccordionSimpleExample from './NxAccordionExample';
 import NxAccordionComplexExample from './NxAccordionComplexExample';
 import NxAccordionTertiaryButtonExample from './NxAccordionTertiaryButtonExample';
+import NxStatefulAccordionExample from '../NxStatefulAccordion/NxStatefulAccordionExample';
 
 const NxAccordionSimpleCode = require('./NxAccordionExample?raw'),
     NxAccordionComplexCode = require('./NxAccordionComplexExample?raw'),
-    NxAccordionTertiaryButtonCode = require('./NxAccordionTertiaryButtonExample?raw');
+    NxAccordionTertiaryButtonCode = require('./NxAccordionTertiaryButtonExample?raw'),
+    NxStatefulAccordionCode = require('../NxStatefulAccordion/NxStatefulAccordionExample?raw');
 
-const NxAccordionPage = () =>
-  <>
-    <GalleryDescriptionTile>
-      <NxP>
-        NxAccordion renders a panel with an always-visible header and a collapsible/expandable body section.
-        This is analogous to the HTML 5 <NxCode>&lt;details&gt;</NxCode> element (which it is
-        implemented on top of). There are three related components: <NxCode>NxAccordion</NxCode> itself,
-        <NxCode>NxAccordion.Header</NxCode> which represents the header content, and
-        {' '}<NxCode>NxAccordion.Title</NxCode> which is a convenience component for the header title. All other
-        children of <NxCode>NxAccordion</NxCode> aside from
-        the <NxCode>Header</NxCode> and the <NxCode>Title</NxCode> are rendered in the collapsible section.
+export default function NxAccordionPage() {
+  const [activeTabId, setActiveTabId] = useState(0);
 
-        Note that this component is stateless – its open state must be tracked externally.
-        See <NxCode>NxStatefulAccordion</NxCode> for a version which tracks its own open state.
-      </NxP>
-      <NxTile.Subsection>
-        <NxTile.SubsectionHeader>
-          <NxH3>NxAccordion</NxH3>
-        </NxTile.SubsectionHeader>
-        <NxTable>
-          <NxTable.Head>
-            <NxTable.Row>
-              <NxTable.Cell>Name</NxTable.Cell>
-              <NxTable.Cell>Type</NxTable.Cell>
-              <NxTable.Cell>Required</NxTable.Cell>
-              <NxTable.Cell>Description</NxTable.Cell>
-            </NxTable.Row>
-          </NxTable.Head>
-          <NxTable.Body>
-            <NxTable.Row>
-              <NxTable.Cell>onToggle</NxTable.Cell>
-              <NxTable.Cell>(() =&gt; void)</NxTable.Cell>
-              <NxTable.Cell>No</NxTable.Cell>
-              <NxTable.Cell>
-                A function which gets called when the accordion collapse/expand state is toggled.
+  return (
+    <>
+      <NxTabs activeTab={activeTabId} onTabSelect={setActiveTabId}>
+        <NxTabList aria-label="Table Component Examples">
+          <NxTab>Usage</NxTab>
+          <NxTab>Stateless Examples</NxTab>
+          <NxTab>Stateful Examples</NxTab>
+        </NxTabList>
+        <NxTabPanel>
+          <GalleryDescriptionTile>
+            <NxP>
+              NxAccordion renders a panel with an always-visible header and a collapsible/expandable body section.
+              This is analogous to the HTML 5 <NxCode>&lt;details&gt;</NxCode> element (which it is
+              implemented on top of). There are three related components: <NxCode>NxAccordion</NxCode> itself,
+              <NxCode>NxAccordion.Header</NxCode> which represents the header content, and
+              {' '}<NxCode>NxAccordion.Title</NxCode> which is a convenience component for the header title. All other
+              children of <NxCode>NxAccordion</NxCode> aside from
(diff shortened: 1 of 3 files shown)


gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -4,184 +4,265 @@
...
-      <NxP>
-        NxAccordion renders a panel with an always-visible header and a collapsible/expandable body section.
-        This is analogous to the HTML 5 <NxCode>&lt;details&gt;</NxCode> element (which it is
-        implemented on top of). There are three related components: <NxCode>NxAccordion</NxCode> itself,
-        <NxCode>NxAccordion.Header</NxCode> which represents the header content, and
-        {' '}<NxCode>NxAccordion.Title</NxCode> which is a convenience component for the header title. All other
-        children of <NxCode>NxAccordion</NxCode> aside from
-        the <NxCode>Header</NxCode> and the <NxCode>Title</NxCode> are rendered in the collapsible section.
+export default function NxAccordionPage() {
+  const [activeTabId, setActiveTabId] = useState(0);
```
rpokorny, line 37: We have `NxStatefulTabs` to simplify this

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -4,184 +4,265 @@
...
-            <NxTable.Row>
-              <NxTable.Cell>onToggle</NxTable.Cell>
-              <NxTable.Cell>(() =&gt; void)</NxTable.Cell>
-              <NxTable.Cell>No</NxTable.Cell>
-              <NxTable.Cell>
-                A function which gets called when the accordion collapse/expand state is toggled.
+  return (
+    <>
+      <NxTabs activeTab={activeTabId} onTabSelect={setActiveTabId}>
+        <NxTabList aria-label="Table Component Examples">
```
rpokorny, line 42: I don't think an `aria-label` is needed here - this section has a header (the page title) right above the tabs

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -4,184 +4,265 @@
...
+            </NxP>
+            <NxH3>Stateful Accordion</NxH3>
+            <NxP>
+              <NxCode>NxStatefulAccordion</NxCode> is a wrapper
+              around <NxCode>NxAccordion</NxCode> which tracks its own toggle state. It accepts
+              the same props as <NxCode>NxAccordion</NxCode>, except that instead
+              of <NxCode>open</NxCode>, it accepts <NxCode>defaultOpen</NxCode> which
+              provides the initial toggle state.
+            </NxP>
+            <NxH2>Guidelines</NxH2>
```
rpokorny, line 69: I don't think it's valid either in an HTML-outline sense or in a kaleido sense to have a second h2 partway down a tile

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -4,184 +4,265 @@
...
+            <NxList bulleted>
+              <NxList.Item>Displaying and grouping additional information.</NxList.Item>
+              <NxList.Item>Adding granular control over the information on a given page.</NxList.Item>
+              <NxList.Item>Shortening pages to reduce scrolling.</NxList.Item>
+            </NxList>
+            <NxH3>When Not to Use</NxH3>
+            <NxList bulleted>
+              <NxList.Item>
+                Displaying critical system information or a primary action to be taken on the page. (for example,
+                {' '}<a href="/components/alert">alerts</a>, confirmation or cancellation buttons).
```
rpokorny, line 80: I know this is a spike but these should ultimately be `NxTextLink`s

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -4,184 +4,265 @@
...
+                content use a <a href="/components/tree">tree</a>.
+              </NxList.Item>
+              <NxList.Item>
+                Displaying a set of visual components following the same style, prefer using Collapsable.(TODO: Add
+                link to collapsable).
+              </NxList.Item>
+            </NxList>
+            <NxTile.Subsection>
+              <NxTile.SubsectionHeader>
+                <NxH3>NxAccordion</NxH3>
```
rpokorny, line 100: Why are the subheaders above this point not also in tile subsections?

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -48,486 +90,660 @@ const paginationCodeExamples = [
...
+              <li className="nx-list__item">
+                city sign plastic gang math- wristwatch industrial grade
+              </li>
+              <li className="nx-list__item">
+                film drone advert tank-traps DIY monofilament industrial grade. office car kanji cartel monofilament
+              </li>
+            </ul>
+            <NxTile.Subsection>
+              <NxTile.SubsectionHeader>
+                <NxH3>NxTable Props and Classes</NxH3>
```
rpokorny, line 141: I don't see any docs about classes here

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -48,486 +90,660 @@ const paginationCodeExamples = [
...
+            </NxP>
+          </GalleryDescriptionTile>
+        </NxTabPanel>
+
+        <NxTabPanel>
+
+          <GalleryExampleTile title="Simple Example"
+                              liveExample={NxTableSimpleExample}
+                              codeExamples={tableSimpleExampleCode}
+                              collapseCodeExample>
```
rpokorny, line 572: If you get the latest from main I believe this prop is no longer needed/valid

gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -48,486 +90,660 @@ const paginationCodeExamples = [
...
+        </NxTabPanel>
+        <NxTabPanel>
+          <GalleryExampleTile title="NX Table Simple Example"
+                              htmlExample={NxTableSimpleCode}
+                              codeExamples={NxTableSimpleCode}
+                              collapseCodeExample>
+            A simple, static demonstration of <NxCode>nx-table</NxCode> styles.
+          </GalleryExampleTile>
+
+          <GalleryExampleTile title="NX Table Truncation and Wrapping Example"
```
rpokorny, line 672: I feel like this doesn't entirely address the fact that different aspects of the styling are shown in the HTML examples than in the React examples, which isn't always obvious to users.  Consider this example here: it shows how truncation and text wrapping work within a table, which is important information whether the table is constructed in react or plain HTML.  Does the reorganization make that any easier to find than it was previously?  I'm not sure if we need a naming adjustment to the tabs, or maybe another tab, but I feel like we need a way to make it clearer to both React users and HTML users where they should look for "styling variations"

gallery/src/components/NxAccordion/NxAccordionPage.tsx
```diff
@@ -4,184 +4,265 @@
...
-              <NxTable.Cell>No</NxTable.Cell>
-              <NxTable.Cell>
-                A function which gets called when the accordion collapse/expand state is toggled.
+  return (
+    <>
+      <NxTabs activeTab={activeTabId} onTabSelect={setActiveTabId}>
+        <NxTabList aria-label="Table Component Examples">
+          <NxTab>Usage</NxTab>
+          <NxTab>Stateless Examples</NxTab>
+          <NxTab>Stateful Examples</NxTab>
```
avnvbhatta, line 45: Perhaps these should be called "React Stateless Examples" and "React Stateful Examples"? Considering in the `NxTable` page we have "React Examples" and "HTML Examples".

glenhunter: > Perhaps these should be called "React Stateless Examples" and "React Stateful Examples"? Considering in the `NxTable` page we have "React Examples" and "HTML Examples".

Sure: https://github.com/sonatype/sonatype-react-shared-components/pull/607/commits/c5e04ef3037e9ad665ef83c6b03f0d625d92119f

gallery/src/components/NxAccordion/NxAccordionPage.tsx
glenhunter: > I don't think an `aria-label` is needed here - this section has a header (the page title) right above the tabs

https://github.com/sonatype/sonatype-react-shared-components/pull/607/commits/1da396f0ab254dc6305be33e1a44bfee47fc4613

gallery/src/components/NxAccordion/NxAccordionPage.tsx
glenhunter: > I don't think it's valid either in an HTML-outline sense or in a kaleido sense to have a second h2 partway down a tile

https://github.com/sonatype/sonatype-react-shared-components/pull/607/commits/cc91e7c1fa1737f501917c78fa45f8119d43f9d7

gallery/src/components/NxAccordion/NxAccordionPage.tsx
glenhunter: > Why are the subheaders above this point not also in tile subsections?

https://github.com/sonatype/sonatype-react-shared-components/pull/607/commits/cc91e7c1fa1737f501917c78fa45f8119d43f9d7

gallery/src/components/NxTable/NxTablePage.tsx
glenhunter: > I don't see any docs about classes here

https://github.com/sonatype/sonatype-react-shared-components/pull/607/commits/8d000e0c81863ac792ca4a7acee0dbef9b54de0c

gallery/src/components/NxTable/NxTablePage.tsx
glenhunter: > If you get the latest from main I believe this prop is no longer needed/valid

https://github.com/sonatype/sonatype-react-shared-components/pull/607/commits/5a6b331a6becbaa1b60dc91b70bd8a059fa8f696

gallery/src/components/NxTable/NxTablePage.tsx
glenhunter: > I feel like this doesn't entirely address the fact that different aspects of the styling are shown in the HTML examples than in the React examples....

Take a look at this branch: https://github.com/sonatype/sonatype-react-shared-components/tree/RSC-842-SPIKE-tabs--NxTable-variations

I added a "Variations" tab for the extra layouts, and then consolidated the examples which had both React and HTML code.

rpokorny: > I added a "Variations" tab for the extra layouts, and then consolidated the examples which had both React and HTML code.

Hmm that means that for examples that have two implementations shown, only one implementation is actually driving the live example, correct? I worry about the other example being incorrect or getting out of sync.

glenhunter: > > I added a "Variations" tab for the extra layouts, and then consolidated the examples which had both React and HTML code.
> 
> Hmm that means that for examples that have two implementations shown, only one implementation is actually driving the live example, correct? I worry about the other example being incorrect or getting out of sync.

I think that if we default to showing the React example then the risk is quite low. The React examples still use the same HTML/CSS behind the scenes as the solely HTML/CSS examples.

rpokorny: > I think that if we default to showing the React example then the risk is quite low.

I have to disagree, I think that the risk is quite high.  The HTML examples in this scenario are going completely untested.

> The React examples still use the same HTML/CSS behind the scenes as the solely HTML/CSS examples.

This assumption is where the risk lies.  We need a way to verify that it remains true

glenhunter: > I have to disagree, I think that the risk is quite high. The HTML examples in this scenario are going completely untested.

In almost all cases they are going completely untested now. If you look at the visual tests we are testing the React page for `NxTable`, but not the HTML examples. This seems to be true but reversed for `nx-list.

rpokorny: > In fact they are going completely untested now. If you look at the visual tests we are testing the React page for NxTable, but not the HTML examples. This seems to be true but reversed for nx-list.

That's partially true.  The `nx-table` visual tests include certain examples from both the `NxTable` page and the `nx-table` page (as well as `nx-table-container.  But even the ones that aren't getting automated checks are at least visible on the page and get manually tested.

An idea that I have to improve this is that when we want to show the same thing in both react and plain HTML examples, we could have a tabbed interface or something that switches not just which example code is shown but also which live example is shown - so there's still a way to see the result of the plain HTML example.

That said, another idea is to just minimize the number of plain HTML examples. Have one basic one and then rely exclusively on React examples to show all the different options.  The only issue I see with this is when the React and HTML versions have distinct ways of doing the same thing.  For instance, the different button variants are activated using classes when you're in plain HTML, but have a special `variant` prop when you're in React.

glenhunter: > Hmm that means that for examples that have two implementations shown, only one implementation is actually driving the live example, correct? I worry about the other example being incorrect or getting out of sync.

Another way to do this would be to have tabs for the examples as well. We don't have very many components that have both HTML & React examples: Tables, Lists, Alerts, and Buttons the last two which I worry could be out of date.

glenhunter: > That said, another idea is to just minimize the number of plain HTML examples.

I don't see how this is much different than what's there now except that we aren't showing the code. There's still no testing to see if there is any divergence.

glenhunter: > An idea that I have to improve this is that when we want to show the same thing in both react and plain HTML examples, we could have a tabbed interface or something that switches not just which example code is shown but also which live example is shown - so there's still a way to see the result of the plain HTML example.

Yeah we're on the same wavelength here. I'm not sure if the example code we have now will support this.

glenhunter: > But even the ones that aren't getting automated checks are at least visible on the page and get manually tested.

As you said above, "The assumption is where the risk lies." When was the last time anybody compared the `nx-alert` page to the `NxAlert` page?

glenhunter:
Looking at the complex examples I don't think we need to have HTML examples for the sortable columns because that example requires React for the icons.

avnvbhatta (approved):

glenhunter (commented):
+1 for the recent URL changes, I really like it, very clean

PR #614, opened 2022-01-24
Server-side Rendering Initial Support - RSC-852

rpokorny:
https://issues.sonatype.org/browse/RSC-852

This ticket makes the necessary changes to get RSC in general working with serverside rendering in a Next.js application.  For the moment, the Next application in which manual testing was done is included in the PR so that reviewers can see it working.  However, I intend to remove that application before this is merged and to then create some sort of automated testing infrastructure for this feature in a follow-on ticket.  So far, not all RSC components have been explicitly tested so it is possible that individual components still have problems.  I consider the output of this ticket to be a "soft support" that is sufficient for the known downstream team that needs this feature to get started on their work.  Once more thorough testing is added, then I'd say we can consider the support official by e.g. adding it to the readme.

Note that a passing Jenkins build won't be possible until the next app is deleted, due to the license check.


lib/package.json
```diff
@@ -9,20 +9,24 @@
     "directory": "lib"
   },
   "scripts": {
-    "build": "npm-run-all -p lint compile-ts compile-scss copy-scss copy-package-json copy-assets",
+    "build": "npm-run-all output-frontend output-server",
```
rpokorny, line 12: The basic idea here is that we create a new output folder called `dist/server` which includes another copy of all of the RSC output files, but compiled in such a way that they are compatible with next.js.  At this point the only difference between the two is that the `server` build uses commonjs imports instead of ES6.

lib/src/util/importImage.ts
```diff
@@ -0,0 +1,27 @@
...
+ * out statically. This is likely predicated on the basename being itself statically computable.
+ */
+export default function importImage(basename: string): string {
+  const imageImport = require(`../assets/img/${basename}`);
+
+  if (typeof imageImport === 'string') {
+    return imageImport;
+  }
+  else {
+    return imageImport.src ?? imageImport.default.src;
```
rpokorny, line 25: Essentially, when you import an image in a next.js app, you don't get a string path like we're used to.  Instead you get an object that has that path as its `src` attribute as well as a number of other things that help next's `Image` tag do optimization.

server/next.config.js
```diff
@@ -0,0 +1,5 @@
+/** @type {import('next').NextConfig} */
+
+const withTM = require('next-transpile-modules')(['@sonatype/react-shared-components']);
```
rpokorny, line 3: This is the key thing that next.js apps need to do to use RSC.   By default, next does not apply any sort of webpack/transpilation logic to external dependency code, and so things like image imports and scss imports in RSC cause build errors.  This `next-transpile-modules` plugin tells next to apply transpilation to the speciffied dependency packages. Apparently it has a performance cost, which is why it should only be used on the packages that really need it.

server/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "name": "server",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@sonatype/react-shared-components": "file:../lib/dist",
```
rpokorny, line 11: Note: This cannot use `link:` because node follows symlinks when importing.  What this means is that when an RSC module imports, say, `react`, even though from next's perspective that RSC module is at `server/node_modules/@sonatype/react-shared-components/...`, and you'd want it to resolve to `server/node_modules/react`, the `react-shared-components` folder there is a symlink to `lib/dist` and so node will actually resolve the react import to `lib/node_modules/react`.  This leads to "duplicate react instances" errors when using any hooks.

Of course, in a real app that isn't hosted in the same repo as the RSC source code, you wouldn't be using either `link:` or `file:`, so this doesn't really matter outside of our own testing

server/styles/Home.module.css
```diff
@@ -0,0 +1,116 @@
+.container {
```
rpokorny, line 1: This is all auto-generated and is part of the next app that will be removed before merging.  Ignore

lib/package.json
rpokorny: > The basic idea here is that we create a new output folder called dist/server which includes another copy of all of the RSC output files, but compiled in such a way that they are compatible with next.js. At this point the only difference between the two is that the server build uses commonjs imports instead of ES6.

Actually, this has all been removed as I realized that with the next plugin we're using it's unnecessary.  Next can run on the exact same build output as frontend usage.

server/styles/Home.module.css
rpokorny: > This is all auto-generated and is part of the next app that will be removed before merging. Ignore

It's now removed

server/next.config.js
DarthHater: Supposedly next 12 uses a rust compiler for everything that's wicked fast, so it might all even out? It'll be interesting to see if that's the case.

rpokorny: > Supposedly next 12 uses a rust compiler for everything that's wicked fast, so it might all even out? It'll be interesting to see if that's the case.

I wouldn't even say the configuration seen here is particularly slow.  Just that there's a reason they don't transpile your entire `node_modules` dir by default

DarthHater (approved):
I think I'm a loose approval on this, but none of these changes look horrible nor scary, and the explanations help (my only suggestion is add some of them to the README when you get to a supported state)

PR #615, opened 2022-01-24
Hosting multiple versions of the gallery - RSC-810

entroform:
# This PR adds support for selecting different versions of the RSC and the gallery.

https://issues.sonatype.org/browse/RSC-810

https://jenkins.ci.sonatype.dev/job/uxui/job/publish-gallery-with-versions-to-s3/6/

Notes from Joel:
1) This job will receive whatever the version number is being built from the upstream job
2) It will upload that version under versions/<version number> AND versions/latest
3) Then it will read all the versions
4) Once it has the versions, it will stick them into a hosted-versions.json file and drop it at the root level (so it would be retrieved from https://gallery.sonatype.dev/hosted-versions.json

This PR also includes a version selector added to the main gallery header.
This selector is populated from http://gallery.sonatype.dev/versions/latest/hosted-versions.json, so "older" gallery versions (that have this version select component) will still have the updated hosted versions list.
This also means that users will not be able to use select versions after navigating to a gallery page without the version selector.


gallery/webpack.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = function(env = { production: false }) {
     entry: './main.tsx',
     output: {
       filename: 'bundle.js',
-      publicPath: '/'
+      publicPath: ''
```
rpokorny, line 28: Usually `./` is the way to go here

entroform: > Usually `./` is the way to go here

Changed it here: d06c6ce171050742d5d82b746fb11965a9f1c55e

gallery/hosted-versions.json
```diff
@@ -0,0 +1,6 @@
+[
+  "9.14.0",
```
entroform, line 2: These will be updated before merge.

gallery/package.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "9.16.1",
+  "version": "9.17.0",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
     "start": "npm-run-all lint webpack-dev-server",
     "build": "npm-run-all lint webpack copy-assets",
-    "copy-assets": "cpx \"{src/index.html,src/assets/images/favicon*}\" dist",
+    "copy-assets": "cpx \"{src/index.html,src/assets/images/favicon*}\" dist; cpx hosted-versions.json dist",
```
rpokorny, line 9: We try to maintain platform independence in our buld scripts, so a semi-colon is a no-go.  Use `npm-run-all` instead

rpokorny: > We try to maintain platform independence in our buld scripts, so a semi-colon is a no-go. Use npm-run-all instead

Or really I feel like this could just be combined into one `cpx` call by adding the new file to the existing glob

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+ */
+import React, { useEffect, ChangeEventHandler, useState } from 'react';
+import { NxFormSelect } from '@sonatype/react-shared-components';
+
+import packageJson from '../../package.json';
+
+import './HostedVersionsSelect.scss';
+
+const CURRENT_VERSION = packageJson.version;
+const HOSTED_VERSIONS_JSON_URL = 'https://gallery.sonatype.dev/versions/latest/hosted-versions.json';
```
rpokorny, line 15: I'm not super comfortable with the idea that this is hardcoded especially when the gallery is often run on other hosts

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const getHostedVersion = () => fetch(HOSTED_VERSIONS_JSON_URL, {
+  headers: {
+    'Content-Type': 'application/json',
```
rpokorny, line 19: I don't think this header is needed.  As a `GET` it has no request body and so no need for a Content-Type

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [isLoadingHostedVersions, setIsLoadingHostedVersions] = useState(false);
+  const [hostedVersions, setHostedVersions] = useState([CURRENT_VERSION]);
+
+  const onChange: ChangeEventHandler<HTMLSelectElement> = event =>
+    window.location.href = `/versions/${event.target.value}`;
+
+  useEffect(() => {
+    setIsLoadingHostedVersions(true);
+    getHostedVersion()
+        .then((hostedVersions) => {
```
rpokorny, line 39: async/await might be slightly nicer looking

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+          setIsLoadingHostedVersions(false);
+        })
+        .catch(() => {
+          setHostedVersions([CURRENT_VERSION]);
+          setIsLoadingHostedVersions(false);
+        });
+  }, []);
+
+  return (
+    <div className="hosted-versions-select">
```
rpokorny, line 50: Classnames here should start with `gallery-`

gallery/package.json
```diff
@@ -73,6 +73,7 @@
     "core-js": "^3.8.3",
     "es6-object-assign": "^1.1.0",
     "es6-set": "^0.1.5",
+    "json-loader": "^0.5.7",
```
rpokorny, line 76: This doesn't seem right.  The gallery was already doing a json import before this PR

lib/package.json
```diff
@@ -28,7 +28,7 @@
     "test-watch": "jest --watch",
     "test-watch-debug": "node --inspect-brk node_modules/.bin/jest --runInBand --watchAll",
     "clean": "rimraf dist junit.xml",
-    "version": "versync -b sync -a"
+    "version": "versync -b sync -a; node ../update-hosted-versions.js"
```
rpokorny, line 31: Use `npm-run-all`.  Also not sure if invoking node like this is completely safe or not

gallery/src/PageHeader/HostedVersionsSelect.tsx
rpokorny: > I'm not super comfortable with the idea that this is hardcoded especially when the gallery is often run on other hosts

Since this file is being fetched at runtime anyway I'm wondering if it should even be managed by the RSC build.  What if the jenkins job that deploys gallery.sonatype.dev had a step where it appended the latest version number to an s3 bucket or something, and that buck was what go fetched here?

rpokorny:
Since this whole thing will only really work on gallery.sonatype.dev, I'd prefer there be a build flag of some sort to enable it, such that it is not enabled during our local builds.

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > > I'm not super comfortable with the idea that this is hardcoded especially when the gallery is often run on other hosts
> 
> Since this file is being fetched at runtime anyway I'm wondering if it should even be managed by the RSC build. What if the jenkins job that deploys gallery.sonatype.dev had a step where it appended the latest version number to an s3 bucket or something, and that buck was what go fetched here?

Hmm, I guess one benefit of having it hardcoded here is that we can control or make changes to the versions listed through as part of our codebase (not entirely sure if that's entirely useful, perhaps filtering the versions we want users to be able to navigate to etc).
But I see how having the Jenkins build job handle the appending having s3 to a file is nice too. I'll reach out to Joel to explore that idea.

rpokorny: > Hmm, I guess one benefit of having it hardcoded here is that we can control or make changes to the versions listed through as part of our codebase (not entirely sure if that's entirely useful, perhaps filtering the versions we want users to be able to navigate to etc).

That's part of what doesn't seem right to me.  Which versions are available is a concern of the gallery.sonatype.dev deployment specifically and not of this codebase more generally.  Really, even a list of available versions in an S3 bucket  is iffy as there's nothing really guaranteeing that that file is in sync with the reality of what versions are actually being hosted.  Ideally the list of versions would somehow be generated from what's actually available on the server

entroform: > > Hmm, I guess one benefit of having it hardcoded here is that we can control or make changes to the versions listed through as part of our codebase (not entirely sure if that's entirely useful, perhaps filtering the versions we want users to be able to navigate to etc).
> 
> That's part of what doesn't seem right to me. Which versions are available is a concern of the gallery.sonatype.dev deployment specifically and not of this codebase more generally. Really, even a list of available versions in an S3 bucket is iffy as there's nothing really guaranteeing that that file is in sync with the reality of what versions are actually being hosted. Ideally the list of versions would somehow be generated from what's actually available on the server

Hmm, I see your point, I'll reach out to Joel and see how we can get a generated a list of of hosted versions.

gallery/package.json
entroform: > > We try to maintain platform independence in our buld scripts, so a semi-colon is a no-go. Use npm-run-all instead
> 
> Or really I feel like this could just be combined into one `cpx` call by adding the new file to the existing glob

Thanks! This is no longer needed and have been removed with the new implementation

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > I don't think this header is needed. As a `GET` it has no request body and so no need for a Content-Type

Updated here: 452ee649b08bf6659d64b8b0410767d0dc26ef30

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > async/await might be slightly nicer looking

Cleaned up here: e62b4b36d53a86e07b2a51316138afb54c716b0c

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Classnames here should start with `gallery-`

Fixed here: d3041d6fa05deb382ed71116327f25d8b5e5ffc6

gallery/package.json
entroform: > This doesn't seem right. The gallery was already doing a json import before this PR

Ops, I thought I had removed this.
Removed: 212a504adf72913462c9aac2c11bf991dc949cc3

lib/package.json
entroform: > Use `npm-run-all`. Also not sure if invoking node like this is completely safe or not

This was no longer needed and removed in the current implementation.

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > > > Hmm, I guess one benefit of having it hardcoded here is that we can control or make changes to the versions listed through as part of our codebase (not entirely sure if that's entirely useful, perhaps filtering the versions we want users to be able to navigate to etc).
> > 
> > 
> > That's part of what doesn't seem right to me. Which versions are available is a concern of the gallery.sonatype.dev deployment specifically and not of this codebase more generally. Really, even a list of available versions in an S3 bucket is iffy as there's nothing really guaranteeing that that file is in sync with the reality of what versions are actually being hosted. Ideally the list of versions would somehow be generated from what's actually available on the server
> 
> Hmm, I see your point, I'll reach out to Joel and see how we can get a generated a list of of hosted versions.

We updated our approach! Instead of having the truth of the list of versions in the codebase, we have generate a json file from the directory listing after the last version is uploaded:

https://gallery.sonatype.dev/hosted-versions.json

Jenkinsfile
```diff
@@ -134,8 +134,9 @@ dockerizedBuildPipeline(
   testResults: ['lib/junit.xml', 'gallery/test-results/junit.xml'],
   onSuccess: {
     githubStatusUpdate('success')
-    if (env.BRANCH_NAME == 'main') {
-      build job:'/uxui/publish-gallery-to-s3', propagate: false, wait: false, parameters: [
+    if (env.BRANCH_NAME == 'RSC-810_gallery-versions') {
```
rpokorny, line 137: I assume this needs to be changed back to main before the merge? Don't forget

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "10.0.0",
+  "version": "10.1.0",
```
rpokorny, line 3: Should be a patch version

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }, []);
+
+  return (
+    <div className="gallery-hosted-versions-select">
+      <label className="gallery-hosted-versions-select__label" htmlFor="gallery-hosted-versions-select">
+        Version
+      </label>
+      {
+        !isLoadingHostedVersions && (
+          <NxFormSelect className="gallery-hosted-versions-select__select nx-form-select--short"
```
rpokorny, line 71: Would it be more semantically correct for this to be an `NxDropdown` full of links?

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [isLoadingHostedVersions, setIsLoadingHostedVersions] = useState(false);
+  const [hostedVersions, setHostedVersions] = useState([CURRENT_VERSION]);
+
+  const onChange: ChangeEventHandler<HTMLSelectElement> = event =>
+    window.location.href = `/versions/${event.target.value}`;
+
+  useEffect(() => {
+    setIsLoadingHostedVersions(true);
+
+    const loadHostedVersions = async () => {
```
rpokorny, line 34: Why is `loadHostedVersions` a separate function rather than having this logic inline?

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const CURRENT_VERSION = packageJson.version;
+const HOSTED_VERSIONS_JSON_URL = 'https://gallery.sonatype.dev/hosted-versions.json';
+
+const getHostedVersions = () => fetch(HOSTED_VERSIONS_JSON_URL).then((response) => {
+  if (!response.ok) {
+    throw new Error('Failed to load hosted versions.');
+  }
+  return response.json();
+});
```
rpokorny, line 22: This can also use async/await

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+    window.location.href = `/versions/${event.target.value}`;
+
+  useEffect(() => {
+    setIsLoadingHostedVersions(true);
+
+    const loadHostedVersions = async () => {
+      try {
+        const hostedVersions = await getHostedVersions();
+
+        const sortedVersions = hostedVersions.filter((v: string) => v !== 'latest').sort((a: string, b: string) => {
```
rpokorny, line 38: Ramda's `without` would be slightly cleaner than `filter`

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { NxFormSelect } from '@sonatype/react-shared-components';
+
+import packageJson from '../../package.json';
+
+import './HostedVersionsSelect.scss';
+
+const CURRENT_VERSION = packageJson.version;
+const HOSTED_VERSIONS_JSON_URL = 'https://gallery.sonatype.dev/hosted-versions.json';
+
+const getHostedVersions = () => fetch(HOSTED_VERSIONS_JSON_URL).then((response) => {
```
rpokorny, line 17: If this function had an explicit return type you probably wouldn't need as many type annotations in `useEffect`

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  useEffect(() => {
+    setIsLoadingHostedVersions(true);
+
+    const loadHostedVersions = async () => {
+      try {
+        const hostedVersions = await getHostedVersions();
+
+        const sortedVersions = hostedVersions.filter((v: string) => v !== 'latest').sort((a: string, b: string) => {
+          const aParts = a.split('.');
```
rpokorny, line 39: Consider an external package for this logic.  https://www.npmjs.com/package/compare-versions looks about right

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+      catch {
+        setHostedVersions([CURRENT_VERSION]);
+        setIsLoadingHostedVersions(false);
+      }
+    };
+
+    loadHostedVersions();
+  }, []);
+
+  return (
```
rpokorny, line 64: When I run this branch locally, the dropdown appears but does not work correctly.  We won't want that to be the case on `main`

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,88 @@
...
+ */
+import React, { useEffect, ChangeEventHandler, useState } from 'react';
+import { NxFormSelect } from '@sonatype/react-shared-components';
+
+import packageJson from '../../package.json';
+
+import './HostedVersionsSelect.scss';
+
+const CURRENT_VERSION = packageJson.version;
+const HOSTED_VERSIONS_JSON_URL = 'https://gallery.sonatype.dev/hosted-versions.json';
```
rpokorny, line 15: Not sure how I feel about `gallery.sonatype.dev` being hard-coded.  I wonder if it should either be an environment variable, or if the URL should just have the path part (`/hosted-versions.json`) along with logic to hide the dropdown if the file isn't found.

gallery/src/PageHeader/HostedVersionsSelect.scss
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ 
+.gallery-hosted-versions-select {
+  display: flex;
```
rpokorny, line 9: Should this really be flex? I'd think we'd want the "Version" text baseline-aligned with the dropdown (which it isn't currently) and could just use whitespace or margin for the space inbetween, rather than needing `gap`.

gallery/src/PageHeader/HostedVersionsSelect.scss
```diff
@@ -0,0 +1,16 @@
...
+  align-items: center;
+  gap: var(--nx-spacing-2x);
+
+  &__label {
+    font-size: 14px;
```
rpokorny, line 14: Did this sizing come from a designer?

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -27,21 +27,15 @@
...
-
   &__name {
+    color: var(--nx-swatch-grey-30);
     font-size: 18px;
     font-weight: 600;
     margin-right: var(--nx-spacing-4x);
     min-width: max-content;
   }
 
   &__version {
```
rpokorny, line 38: Ah this makes me notice that the old version number is now gone.  What's expected to happen when running locally? As I mentioned in another thread currently we get the dropdown but it doesn't work.

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Why is `loadHostedVersions` a separate function rather than having this logic inline?

This is to avoid useEffect complaining: `'() => Promise<void>' is not assignable to parameter of type 'EffectCallback'`

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Consider an external package for this logic. https://www.npmjs.com/package/compare-versions looks about right

Changed to use compare-versions here: 4e1cbb1e169011bf6cf4bd1f9055bb5011c1f75a

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > This can also use async/await

Fixed here: 457e155bea405b56213e7592b3be25760726364b

gallery/package.json
entroform: > Should be a patch version

Updated here: 8401f44a8e328618e7a323886e36bbd9fd25e71c

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Would it be more semantically correct for this to be an `NxDropdown` full of links?

Yup, updated here: ff6bd9ba0dcea4559103c7febdd8deae03169c76

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Ramda's `without` would be slightly cleaner than `filter`

Fixed as part of: 4e1cbb1e169011bf6cf4bd1f9055bb5011c1f75a

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Not sure how I feel about `gallery.sonatype.dev` being hard-coded. I wonder if it should either be an environment variable, or if the URL should just have the path part (`/hosted-versions.json`) along with logic to hide the dropdown if the file isn't found.

Yup! I updated it here: 18aab428bcb0badf4d66b08b695127dd78b9a43e
Let me know what you think!

gallery/src/PageHeader/HostedVersionsSelect.scss
entroform: > Should this really be flex? I'd think we'd want the "Version" text baseline-aligned with the dropdown (which it isn't currently) and could just use whitespace or margin for the space inbetween, rather than needing `gap`.

This was cleaned up here: 18aab428bcb0badf4d66b08b695127dd78b9a43e

gallery/src/PageHeader/PageHeader.scss
entroform: > Ah this makes me notice that the old version number is now gone. What's expected to happen when running locally? As I mentioned in another thread currently we get the dropdown but it doesn't work.

When running locally it will not show the dropdown: 18aab428bcb0badf4d66b08b695127dd78b9a43e

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > When I run this branch locally, the dropdown appears but does not work correctly. We won't want that to be the case on `main`

Fixed: 18aab428bcb0badf4d66b08b695127dd78b9a43e

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > If this function had an explicit return type you probably wouldn't need as many type annotations in `useEffect`

I think this is cleaned up here:
457e155bea405b56213e7592b3be25760726364b
18aab428bcb0badf4d66b08b695127dd78b9a43e

gallery/src/PageHeader/HostedVersionsSelect.scss
entroform: > Did this sizing come from a designer?

It approved by a designer but we changed it here recently:
18aab428bcb0badf4d66b08b695127dd78b9a43e

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+import './HostedVersionsSelect.scss';
+
+const CURRENT_VERSION = packageJson.version;
+const HOSTED_VERSIONS_JSON_PATH = '/hosted-versions.json';
+
+// eslint-disable-next-line no-undef
+const PRODUCTION = process.env.PRODUCTION;
+
+const getHostedVersions = async () => {
+  const response = await fetch(HOSTED_VERSIONS_JSON_PATH);
```
rpokorny, line 23: Maybe don't attempt the fetch if `!PRODUCTION`

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [hasError, setHasError] = useState(false);
+  const [hostedVersions, setHostedVersions] = useState([CURRENT_VERSION]);
+
+  useEffect(() => {
+    setIsLoadingHostedVersions(true);
+
+    const loadHostedVersions = async () => {
+      try {
+        const hostedVersions = await getHostedVersions();
+        const sortedVersions = without(['latest'], hostedVersions).sort(compareVersions).reverse();
```
rpokorny, line 43: Minor: slightly more efficient
```suggestion
        const sortedVersions = without(['latest'], hostedVersions).sort(pipe(compareVersions, negate));
```

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <div className="gallery-hosted-versions-select">
+      <span className="gallery-hosted-versions-select__version">
+        Version { (!PRODUCTION || hasError) && CURRENT_VERSION }
+      </span>
+      {
+        PRODUCTION && !isLoadingHostedVersions && !hasError && (
+          <NxStatefulDropdown className="gallery-hosted-versions-select__dropdown nx-dropdown--short"
+                              label={CURRENT_VERSION}
+          >
```
rpokorny, line 66: Does this not fit on one line?

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+            {
+              hostedVersions.map(
+                  version => (
+                    <a key={version} href={`versions/${version}`} className="nx-dropdown-button">
```
rpokorny, line 70: If I'm already on `https://gallery.sonatype.dev/versions/10.0.0` and I click a link (say, to 10.0.1) with an href structured like this, won't that take me to `https://gallery.sonatype.dev/versions/versions/10.0.1`?

gallery/src/PageHeader/HostedVersionsSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }, []);
+
+  return (
+    <div className="gallery-hosted-versions-select">
+      <span className="gallery-hosted-versions-select__version">
+        Version { (!PRODUCTION || hasError) && CURRENT_VERSION }
+      </span>
+      {
+        PRODUCTION && !isLoadingHostedVersions && !hasError && (
+          <NxStatefulDropdown className="gallery-hosted-versions-select__dropdown nx-dropdown--short"
```
rpokorny, line 64: Where can I see/test this dropdown now that you've got it so that it doesn't appear locally?

gallery/webpack.config.js
```diff
@@ -16,8 +16,15 @@ module.exports = function(env = { production: false }) {
...
       destination: 'webpack-modules',
       includePackageJsons: true
+    }),
+    new webpack.DefinePlugin({
+      'process.env.PRODUCTION': JSON.stringify(true)
     })
-  ] : [];
+  ] : [
+    new webpack.DefinePlugin({
+      'process.env.PRODUCTION': JSON.stringify(false)
```
rpokorny, line 25: From https://webpack.js.org/guides/production/ :
> Since webpack v4, specifying [mode](https://webpack.js.org/configuration/mode/) automatically configures [DefinePlugin](https://webpack.js.org/plugins/define-plugin) for you:

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > If I'm already on `https://gallery.sonatype.dev/versions/10.0.0` and I click a link (say, to 10.0.1) with an href structured like this, won't that take me to `https://gallery.sonatype.dev/versions/versions/10.0.1`?

Ah, that's meant to be `/versions/...`

gallery/webpack.config.js
entroform: > tomatically configures [DefinePlugin](https://webpack.js.org/plugins/define-plugin) for you:

Thanks! fixed here: 6934bd3e8e37d07275c29148132f54d7299a3ab8

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Maybe don't attempt the fetch if `!PRODUCTION`

Yup! Fixed here:
351f30f526a5ccb21e4d7096551ff71fa5891914
5a69dbe0903a773243dd8c96f84247997ef79ca9

gallery/src/PageHeader/HostedVersionsSelect.tsx
entroform: > Where can I see/test this dropdown now that you've got it so that it doesn't appear locally?

Currently, the only way to test it locally is to manually change
`HOSTED_VERSIONS_JSON_PATH` from  `/hosted-versions.json` to `https://gallery.sonatype.dev/hosted-versions.json` and change `IS_PRODUCTION` to `true`.

Should I create a flag for this to test locally? I was also planning to do a build and I think it should upload to that /versions so we can test it live

rpokorny: > I was also planning to do a build and I think it should upload to that /versions so we can test it live

Oh that should work

entroform: > > I was also planning to do a build and I think it should upload to that /versions so we can test it live
> 
> Oh that should work

Looks to be working here! https://gallery.sonatype.dev/versions/10.0.3/#/

rpokorny:
When I try the versions deployed to gallery.sonatype.dev, the dropdown initializes to 9.14.0 no matter which version I go to.  I'm not sure what the state of those old versions is currently, are they expected to work correctly at the moment?

rpokorny (approved):

entroform:
> When I try the versions deployed to gallery.sonatype.dev, the dropdown initializes to 9.14.0 no matter which version I go to. I'm not sure what the state of those old versions is currently, are they expected to work correctly at the moment?

Ah the other versions that's not 10.0.3 hosted there are older builds so dropdown is no longer working properly

PR #617, opened 2022-01-26
Fix NxTile header height and spacings issues - RSC-823

entroform:
https://issues.sonatype.org/browse/RSC-823


rpokorny:
Incorrect in the case where this a title and subtitle but no action buttons.  I believe instead of changing the height of the entire header, you need to change the height of its first grid row.

rpokorny:
> Incorrect in the case where this a title and subtitle but no action buttons. I believe instead of changing the height of the entire header, you need to change the height of its first grid row.

Going along with that, I'm not sure the `min-height` on the header overall is still necessary.  For that matter, I'm not super clear on why it was necessary previously.

avnvbhatta (approved):

rpokorny:
I didn't see my comments above get addressed

entroform:
> I didn't see my comments above get addressed

Sorry about that: here's a new PR addressing your comments:
https://github.com/sonatype/sonatype-react-shared-components/pull/674

PR #618, opened 2022-01-27
NxOverflowTooltip nextjs support - RSC-864

rpokorny:
https://issues.sonatype.org/browse/RSC-864

All tests still pass if we just use `useEffect` instead of `useLayoutEffect`, and it makes next happy.


DarthHater (approved):
Looks very simple, no feedback!

PR #619, opened 2022-02-01
NxList Overflow Fix - RSC-865

avnvbhatta:
https://issues.sonatype.org/browse/RSC-865

### DIFF
--- a/lib/src/base-styles/_nx-lists.scss
+++ b/lib/src/base-styles/_nx-lists.scss
@@ -10,6 +10,7 @@
 
 .nx-list {
   margin: var(--nx-spacing-4x) 0 var(--nx-spacing-6x) 0;
+  overflow-wrap: anywhere;
   padding: 0;
 
   &, > ul, > ol {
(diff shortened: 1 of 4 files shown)


lib/src/base-styles/_nx-lists.scss
```diff
@@ -10,6 +10,7 @@
 
 .nx-list {
   margin: var(--nx-spacing-4x) 0 var(--nx-spacing-6x) 0;
+  overflow-wrap: anywhere;
```
rpokorny, line 13: Safari still doesn't support this.  It does support `word-break: break-word` which is the more extreme version that will break words in order to fill lines whether it needs to or not.  We probably don't want that by default but I thought it was worth discussing.

avnvbhatta: > Safari still doesn't support this. It does support `word-break: break-word` which is the more extreme version that will break words in order to fill lines whether it needs to or not. We probably don't want that by default but I thought it was worth discussing.

Acknowledged. This has been discussed during WW. Shall proceed with this change knowing that it will still not be fixed in Safari.

rpokorny (approved):

PR #622, opened 2022-02-04
Allow autogenerated IDs to work in serverside rendering  - RSC-872 RSC-867

rpokorny:
https://issues.sonatype.org/browse/RSC-867
https://issues.sonatype.org/browse/RSC-872

This PR fixes the two issues that were discovered with using autogenerated IDs in serverside rendering - whether explicitly or through RSC components that use them internally.

The first issue is that we were using `crypto` to generate random values for the IDs.  This is no longer the case as random IDs can't be used in SSR at all, which leads to the next issue:

The autogenerated IDs generated serverside and clientside must match.  Therefore we have switched from random IDs to sequential integer IDs.  In order for these to match client and server side, several things must be true:

* The serverside render and initial client-side render must be identical - this should already be true is any properly constructed SSR app.
* The serverside render and initial client-side render must execute everything in the same order.  This is an assumption we just have to take on faith.
* The serverside render must reset the sequential integer counter for each page render (or rather, use separate counters for each render).  The only way that I've seen to do this is to have the consuming code wrap their entire page (or the RSC-using parts of it, at least) in a new component that uses React Context to manage separate counter instances for each page load.  This PR introduces the `NxStableUniqueIdContext` component for this purpose.

See 8fd468f314 for a next.js app that uses `NxStableUniqueIdContext`


gallery/src/jsUtilPages/ServerSideRendering/ServerSideRenderingPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <GalleryDescriptionTile>
+      <NxP>
+        RSC strives to be compatible with server-side React rendering frameworks such as Next.js. One hurdle to this
+        compatibility is the usage of autogenerated IDs. As a consumer of RSC, you might have such IDs in your own code,
+        and additionally such IDs are common within the internals of RSC components. Often, these IDs are used in order
+        to link elements together within the ARIA accessibility APIs. These IDs present a challenge to the
+        <em>hydration</em> step of server-side rendering. It is during this step that the browser executes the
+        first <em>client-side</em> react render and attempts to match it up with the server-rendered HTML already
+        present within the document. If there are any mismatches, errors are thrown. Therefore it is important that
+        any autogenerated ids result in the same value on the client as they do on the server.
```
avnvbhatta, line 27:
```suggestion
        any autogenerated IDs result in the same value on the client as they do on the server.
```

gallery/src/jsUtilPages/ServerSideRendering/ServerSideRenderingPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+const ServerSideRenderingPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <NxP>
+        RSC strives to be compatible with server-side React rendering frameworks such as Next.js. One hurdle to this
+        compatibility is the usage of autogenerated IDs. As a consumer of RSC, you might have such IDs in your own code,
+        and additionally such IDs are common within the internals of RSC components. Often, these IDs are used in order
+        to link elements together within the ARIA accessibility APIs. These IDs present a challenge to the
+        <em>hydration</em> step of server-side rendering. It is during this step that the browser executes the
+        first <em>client-side</em> react render and attempts to match it up with the server-rendered HTML already
```
avnvbhatta, line 25:
```suggestion
        first <em>client-side</em> React render and attempts to match it up with the server-rendered HTML already
```

gallery/src/jsUtilPages/ServerSideRendering/ServerSideRenderingPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <NxP>
+        As of the time of writing, the only way to get such consistent autogenerated IDs is to derive them from
+        matching, predictable sequences of values on both the client and server. Most simply, this means using
+        an incrementing counter. However, since the server process is long lived and typically renders many pages
+        over time, care must be taken to ensure that the server uses a separate incrementing counter for each
+        page render. That way, for each page the counter starts at 0 on both the client and server.
+      </NxP>
+      <NxP>
+        In order to facilitate these separate per-page counters, RSC provides
+        an <NxCode>NxStableUniqueIdContext</NxCode> component. Whenever using RSC within serverside rendering,
```
avnvbhatta, line 38:
```suggestion
        an <NxCode>NxStableUniqueIdContext</NxCode> component. Whenever using RSC within server-side rendering,
```

gallery/src/jsUtilPages/ServerSideRendering/ServerSideRenderingPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+        to link elements together within the ARIA accessibility APIs. These IDs present a challenge to the
+        <em>hydration</em> step of server-side rendering. It is during this step that the browser executes the
+        first <em>client-side</em> react render and attempts to match it up with the server-rendered HTML already
+        present within the document. If there are any mismatches, errors are thrown. Therefore it is important that
+        any autogenerated ids result in the same value on the client as they do on the server.
+      </NxP>
+      <NxP>
+        As of the time of writing, the only way to get such consistent autogenerated IDs is to derive them from
+        matching, predictable sequences of values on both the client and server. Most simply, this means using
+        an incrementing counter. However, since the server process is long lived and typically renders many pages
```
avnvbhatta, line 32:
```suggestion
        an incrementing counter. However, since the server process is long-lived and typically renders many pages
```

avnvbhatta (approved):

PR #629, opened 2022-02-25
Update text color NxCollapsibleItems - RSC-862

avnvbhatta:
https://issues.sonatype.org/browse/RSC-862


lib/src/components/NxCollapsibleItems/NxCollapsibleItems.scss
```diff
@@ -129,7 +129,7 @@ $-form-child-left-padding: calc(#{$-trigger-side-padding} + #{$-triangle-box-wid
   border: 1px solid transparent;
   border-radius: var(--nx-border-radius);
   box-sizing: border-box;
-  color: var(--nx-color-text);
+  color: var(--nx-color-text-dark);
```
rpokorny, line 132: Was the intent to make this change everywhere or only when the collapsible items are in a sidebar.  There's a distinct set of styles for that case at the bottom of this file

avnvbhatta: > Was the intent to make this change everywhere or only when the collapsible items are in a sidebar. There's a distinct set of styles for that case at the bottom of this file

I looked at the updated mockup and it looks like it's intended to be everywhere?

rpokorny (approved):

PR #637, opened 2022-03-02
NxGrid Automated A11y- RSC-914

avnvbhatta:
https://issues.sonatype.org/browse/RSC-914

### DIFF
--- a/gallery/src/styles/NxGrid/NxGridExamples.tsx
+++ b/gallery/src/styles/NxGrid/NxGridExamples.tsx
@@ -31,11 +31,9 @@ const NxGridExamples = () =>
       An example of an <NxCode>nx-grid</NxCode> containing two cells each of which scroll. As is typical in this
       case, a horizontal keyline is added above the row of cells.
 
-      <p>
-        Since this example renders a scrollable area without any interactive elements, <NxCode>tabindex="0"</NxCode> has
-        been added to make the element part of the normal tab flow. It is not necessary to
-        add <NxCode>tabindex</NxCode> if the scrollable area contains any interactive elements.
-      </p>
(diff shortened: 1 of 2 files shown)


avnvbhatta:
> I'd add a brief note, maybe to the example desription, explaining why the tabIndex was added (and that it wouldn't be necessary if the scrollable areas contained interactive elements).

A brief description has been added. Check: cb9ea08bfd47a78ced835210e0f127c425d7b9e7

gallery/src/styles/NxGrid/NxGridExamples.tsx
```diff
@@ -30,6 +30,12 @@ const NxGridExamples = () =>
                         codeExamples={NxGridScrollingCode}>
       An example of an <NxCode>nx-grid</NxCode> containing two cells each of which scroll. As is typical in this
       case, a horizontal keyline is added above the row of cells.
+
+      <p>
```
rpokorny, line 34: The entire `GalleryExampleTile` children actually ends up in a `<p>` so this creates a nested `<p>` which is invalid.  Unfortunately `GalleryExampleTile` doesn't have a way to have multiple paragraphs of children, so generally all the text just gets stuffed in the one paragraph.

avnvbhatta: > The entire `GalleryExampleTile` children actually ends up in a `<p>` so this creates a nested `<p>` which is invalid. Unfortunately `GalleryExampleTile` doesn't have a way to have multiple paragraphs of children, so generally all the text just gets stuffed in the one paragraph.

Ah ok. Adjusted to remove the `<p>` tag: 07622c38f0a3402540c88a01685ce52de9f49228

rpokorny (approved):

PR #644, opened 2022-03-02
NxPageTitle Automated a11y fixes - RSC-918

avnvbhatta:
https://issues.sonatype.org/browse/RSC-918


rpokorny:
Oh it's just mad about the checkered backgrounds.  I'd add an easy way for the visual tests to turn them off and then the `a11yTest` should pass without having to disable rules

avnvbhatta:
> Oh it's just mad about the checkered backgrounds. I'd add an easy way for the visual tests to turn them off and then the `a11yTest` should pass without having to disable rules. We'll need that on other pages too.

Changed to use the query param here: e06904ba367ef8fcce65fec2fad854e5c8f4b1ac

rpokorny (approved):

PR #656, opened 2022-03-08
Add URL path to uncheck "show checkered background" - RSC-941

avnvbhatta:
https://issues.sonatype.org/browse/RSC-941

With this PR, can navigate to a page and add the /no-checkered-background at the of the URL to uncheck all the "show checkered background" checkboxes.

Example URL: /pages/Page%20Title/no-checkered-background

### DIFF
--- a/gallery/src/gallery-components/GalleryTiles.tsx
+++ b/gallery/src/gallery-components/GalleryTiles.tsx
@@ -19,6 +19,8 @@ import {
 } from '@sonatype/react-shared-components';
 import { GalleryTileFooter } from './GalleryTileFooter';
 
+import { useLocation } from 'react-router-dom';
+
 interface PropsWithRequiredChildren {
   children: ReactNode;
 }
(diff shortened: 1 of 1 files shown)


avnvbhatta:
> I'd try to make it a little more consistent with the gallery's other URL flags: camelCased and as a query param not a path param

Changed to use camelCase here: 943ee711c43bc0e937bd38b859dbc81e8bc69c75

avnvbhatta:
> Just wondering if we should document this new path behaviour in the README as well.

Do you think there'll be use cases of this outside of our visual tests? Personally I don't think this will be a recurring thing, so I wouldn't add it.

entroform:
> > Just wondering if we should document this new path behaviour in the README as well.
> 
> Do you think there'll be use cases of this outside of our visual tests? Personally I don't think this will be a recurring thing, so I wouldn't add it.

Oh that's true, maybe a comment in the code for context?

gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -87,7 +89,12 @@ export const GalleryExampleTile: FunctionComponent<GalleryExampleTileProps> =
           defaultCheckeredBackground
         } = props,
 
-        [checkeredBackground, toggleCheckeredBackground] = useToggle(defaultCheckeredBackground || false),
+        location = useLocation(),
+
+        overrideCheckeredBackground = location.pathname.includes('noCheckeredBackground'),
```
rpokorny, line 94: This is still making it a path parameter.  See `NxAccordionPage` for how to make it a query parameter

avnvbhatta: > This is still making it a path parameter. See `NxAccordionPage` for how to make it a query parameter

Oh I completely missed the query params part. Changed here: c46c0ff64b7d09eb24486ef35694e8dfe79a8613

rpokorny (approved):

PR #665, opened 2022-03-16
NxTable a11y Fixes - RSC-954

avnvbhatta:
https://issues.sonatype.org/browse/RSC-954

### DIFF
--- a/gallery/src/components/NxTable/NxTablePaginationExample.tsx
+++ b/gallery/src/components/NxTable/NxTablePaginationExample.tsx
@@ -22,7 +22,13 @@ const tableData: Row[] = [
   {name: 'Karen', country: 'Australia'},
   {name: 'Paul', country: 'UK'},
   {name: 'Raul', country: 'Argentina'},
-  {name: 'Maria', country: 'Spain'}
+  {name: 'Maria', country: 'Spain'},
+  {name: 'Karim', country: 'France'},
+  {name: 'Leonardo', country: 'Italy'},
+  {name: 'Cristiano', country: 'Portugal'},
+  {name: 'Lionel', country: 'Argentina'},
+  {name: 'Antonio', country: 'Italy'},
+  {name: 'Joe', country: 'USA'}
 ];
 
 const NxTablePaginationExample = () => {
--- a/gallery/visualtests/nxTable.js
+++ b/gallery/visualtests/nxTable.js
@@ -162,5 +162,6 @@ describe('NxTable', function() {
     });
   });
 
-  it('passes a11y checks', a11yTest());
+  //For clickable rows, we add an empty <th> so that the chevron/icons are rendered on the last column
+  it('passes a11y checks', a11yTest(builder => builder.disableRules('empty-table-header')));
 });


gallery/src/components/NxTable/NxTablePaginationExample.tsx
```diff
@@ -22,7 +22,13 @@ const tableData: Row[] = [
   {name: 'Karen', country: 'Australia'},
   {name: 'Paul', country: 'UK'},
   {name: 'Raul', country: 'Argentina'},
-  {name: 'Maria', country: 'Spain'}
+  {name: 'Maria', country: 'Spain'},
```
avnvbhatta, line 25: Was getting a cannot have duplicate aria-label errors for the pagination labels. Both the pagination and the pagination and filtering example rendered 2 pages, so both of the examples had "pagination, page 1 of 2". I added more rows to get around the error.

rpokorny: > Was getting something similar to a "cannot have duplicate aria-label" errors for the pagination labels. Both the pagination and the pagination and filtering example rendered 2 pages, so both of the examples had "pagination, page 1 of 2". To get around the error, I added more rows so that the pagination examples had 3 pages and the labels were different.

That sounds like a legitimate problem though and working around it just in this specific case doesn't really solve it.

The real issue here is that the pagination is a `<nav>`, which gives it a role of `navigation` by default, and that's a landmark role.  All landmarks of the same role must be distinguishable by accessible name.  I don't think pagination should be a landmark, so we probably need to change either the `role` or the element tagname.  I'm not sure which would be better to change.  Looking through the history, it seems that it was originally a `<div>` and was changed to a `<nav>` during some a11y-minded work.  I'm not sure though if that change actually made some difference or was just a "this seems right" kind of thing.

rpokorny: > The real issue here is that the pagination is a <nav>, which gives it a role of navigation by default, and that's a landmark role. All landmarks of the same role must be distinguishable by accessible name. I don't think pagination should be a landmark, so we probably need to change either the role or the element tagname. I'm not sure which would be better to change. Looking through the history, it seems that it was originally a <div> and was changed to a <nav> during some a11y-minded work. I'm not sure though if that change actually made some difference or was just a "this seems right" kind of thing.

Hmm, I just checked how google does it in their search results and they use `<div role="navigation">`.  But in their case, the navigation is more of a global thing, the search results are the whole page that that control goes to the next page.  `NxPagination` might be used in multiple situations, some like that and some not so much (like when it pages a table that's just part of the page).

rpokorny: > Hmm, I just checked how google does it in their search results and they use <div role="navigation">. But in their case, the navigation is more of a global thing, the search results are the whole page that that control goes to the next page. NxPagination might be used in multiple situations, some like that and some not so much (like when it pages a table that's just part of the page).

`<nav>` supports the following aria roles:
* `navigation`: what it already has.  Not necessarily the best, as it makes it a landmark and also the descriptions of this role generally talk about links, like if each pagination item actually navigated the page URL, which isn't the case for us
* `menu`: seems too vague
* `menubar`: a horizontal menu akin to the File, Edit, etc dropdowns across the top of a desktop application.  Also doesn't seem right for us
* `none`/`presentation`: removes from screenreader support, definitely not what we want
* `tablist`: tabs, also not what we want.  

None of these seem to quite match our use case.  Maybe we need to do more research on what role this should have, possibly moving away from it being a `<nav>`

avnvbhatta: > > Hmm, I just checked how google does it in their search results and they use . But in their case, the navigation is more of a global thing, the search results are the whole page that that control goes to the next page. NxPagination might be used in multiple situations, some like that and some not so much (like when it pages a table that's just part of the page).
> 
> `<nav>` supports the following aria roles:
> 
> * `navigation`: what it already has.  Not necessarily the best, as it makes it a landmark and also the descriptions of this role generally talk about links, like if each pagination item actually navigated the page URL, which isn't the case for us
> * `menu`: seems too vague
> * `menubar`: a horizontal menu akin to the File, Edit, etc dropdowns across the top of a desktop application.  Also doesn't seem right for us
> * `none`/`presentation`: removes from screenreader support, definitely not what we want
> * `tablist`: tabs, also not what we want.
> 
> None of these seem to quite match our use case. Maybe we need to do more research on what role this should have, possibly moving away from it being a `<nav>`


With the new pagination changes, only needed to modify to have unique aria-label: fc6a1260e6e0023df491595ad992cf20f17ec

gallery/visualtests/nxTable.js
```diff
@@ -162,5 +162,6 @@ describe('NxTable', function() {
     });
   });
 
-  it('passes a11y checks', a11yTest());
+  //For clickable rows, we add an empty <th> so that the chevron/icons are rendered on the last column
+  it('passes a11y checks', a11yTest(builder => builder.disableRules('empty-table-header')));
```
rpokorny, line 166: I feel like we could probably fix this rather than ignoring the error.  Maybe add an `aria-label` of something like "row selection" to the table header?

avnvbhatta: > 

I don't think that would fix the issue. Even if we did add an `aria-label`, the content of the `<th>` would still be blank as in the "Incorrect markup solutions" on:
https://dequeuniversity.com/rules/axe/4.3/empty-table-header?application=webdriverio

rpokorny: >  don't think that would fix the issue. Even if we did add an aria-label, the content of the <th> would still be blank as in the "Incorrect markup solutions" on:
https://dequeuniversity.com/rules/axe/4.3/empty-table-header?application=webdriverio

Oh that's a shame.  Still though, we are legitimately breaking this rule (it's not just a buggy rule or something) so we should really try to find a way not to rather than ignoring it

avnvbhatta: > > don't think that would fix the issue. Even if we did add an aria-label, the content of the  would still be blank as in the "Incorrect markup solutions" on:
> > https://dequeuniversity.com/rules/axe/4.3/empty-table-header?application=webdriverio
> 
> Oh that's a shame. Still though, we are legitimately breaking this rule (it's not just a buggy rule or something) so we should really try to find a way not to rather than ignoring it

I've implemented as discussed in WW: 292b9323eea8fa9be840faedc08d445d5601bff4

Based on [this article](https://www.a11yproject.com/posts/how-to-hide-content/)

lib/src/base-styles/_nx-globals.scss
```diff
@@ -47,3 +47,14 @@ $-nx-scrollable-default-height: 400px;
 .nx-blockquote {
   @extend %nx-blockquote;
 }
+
+//Class that allows making content visible only to screen readers but not visually
```
rpokorny, line 51: I've never understood why online examples make this so complicated.  You really don't need to deal with clipping or even absolute positioning, you can just make the element 0x0 and then either make overflow hidden or make the text transparent.

rpokorny: > I did a little bit of research on this and turns out it's not really recommended to set an element's width or height to 0px. Check out [this article](https://webaim.org/techniques/css/invisiblecontent/).

Huh, that's annoyingly inconsistent - the intent is the same whether the content is hidden via clipping or by making the element smaller, so it seems like it ought to be treated the same by screenreaders.  But let's stick with the clip approach then

lib/src/components/NxTable/__tests__/NxTableCell.test.tsx
```diff
@@ -90,7 +90,10 @@ describe('NxTableCell', function () {
         });
 
         expect(component).toMatchSelector('th.nx-cell.nx-cell--header');
-        expect(component.children()).not.toExist();
```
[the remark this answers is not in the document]
avnvbhatta: > The name of this test needs to be updated

Updated here: Changed here: 561e7b33df46898d6c5f37a2adcf7f74d79ab0a8

lib/src/base-styles/_nx-globals.scss
```diff
@@ -47,3 +47,14 @@ $-nx-scrollable-default-height: 400px;
...
+.nx-visually-hidden {
```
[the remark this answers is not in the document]
avnvbhatta: > I'm interested in what the rest of the team thinks, but I would tend to prefer applying the relevant styles to `.nx-cell--header.nx-cell--row-btn` or the appropriate child thereof, rather than making this utility class. I try to avoid making classes that are named after their visual effect, and prefer classes that designate ad-hoc semantics of what they're applied to.

Changed here: 561e7b33df46898d6c5f37a2adcf7f74d79ab0a8

lib/src/components/NxTable/NxTableCell.tsx
```diff
@@ -96,10 +96,11 @@ const NxTableCell = function NxTableCell(props: NxTableCellProps) {
...
       {
-        isSortable ? cellSortingContents :
-        rowBtnIcon ? rowBtnCellContents :
-        children
-      }
+          isSortable ? cellSortingContents :
+          rowBtnIcon ? rowBtnCellContents :
+          children
+        }
+      {isHeader && rowBtnIcon && <span>Select Row</span>}
```
rpokorny, line 103: Probably better to give this span a class and use that in the CSS

lib/src/components/NxTable/NxTableCell.tsx
```diff
@@ -96,10 +96,11 @@ const NxTableCell = function NxTableCell(props: NxTableCellProps) {
...
     <Tag className={classes} aria-sort={ariaSort} {...attrs}>
       {
-        isSortable ? cellSortingContents :
-        rowBtnIcon ? rowBtnCellContents :
-        children
-      }
+          isSortable ? cellSortingContents :
+          rowBtnIcon ? rowBtnCellContents :
+          children
+        }
```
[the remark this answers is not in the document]
avnvbhatta: > Is this indentation right?

Fixed here: 8797e5690fea8222b4fc812549cc3e7ad319ced6

lib/src/components/NxTable/NxTableCell.tsx
avnvbhatta: > Probably better to give this span a class and use that in the CSS

Changed here: 8797e5690fea8222b4fc812549cc3e7ad319ced6

rpokorny: > Changed here: https://github.com/sonatype/sonatype-react-shared-components/commit/8797e5690fea8222b4fc812549cc3e7ad319ced6

Again, classes should be about what the thing is not how it looks.  So I was expecting something like `nx-cell__row-btn-header-text` as the name here.

lib/src/base-styles/_nx-tables.scss
```diff
@@ -268,6 +268,12 @@
   }
 }
 
+//Allows making content visible only to screen readers but not visually
+//Appllied to the <th> when dealing with clickable rows
```
rpokorny, line 272:
```suggestion
//Applied to the <th> when dealing with clickable rows
```

lib/src/base-styles/_nx-tables.scss
```diff
@@ -268,6 +268,12 @@
...
+.nx-cell--header.nx-cell--row-btn > .nx-cell__row-btn-header-text {
```
rpokorny, line 273: It seems likely that the parent part of this selector is no longer necessary
```suggestion
.nx-cell__row-btn-header-text {
```

rpokorny (approved):

PR #668, opened 2022-03-16
NxTransferList a11y fixes - RSC-957

avnvbhatta:
https://issues.sonatype.org/browse/RSC-957

### DIFF
--- a/lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
+++ b/lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
@@ -50,8 +50,8 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
       }
   );
 
-  const moveUpButtonTitle = isTopItem || isFilteredItem ? null : 'Move Up';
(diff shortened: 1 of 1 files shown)


lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
```diff
@@ -68,13 +68,15 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
             <NxButton type="button"
                       variant="icon-only"
                       title={moveUpButtonTitle}
+                      aria-label={moveUpButtonTitle || `${moveUpButtonTitle} (disabled)`}
```
rpokorny, line 71: This shouldn't be necessary, the `title` should be the accessible name

avnvbhatta: > This shouldn't be necessary, the `title` should be the accessible name

Oh right. I think this is a better approach: 3c8882aff25a511c08e893521cf0a9bf8adfd6ad

rpokorny (approved):
I was under the impression that disabled buttons didn't need accessible names but I guess I was wrong

PR #670, opened 2022-03-18
New NxProgressBar Component - RSC-861

entroform:
https://issues.sonatype.org/browse/RSC-861


rpokorny:
I think a few more examples might be necessary - mainly, inline examples for success and error.  To really cover all the bases, what you could do is reorganize the examples so that each layout type (e.g. each variation currently shown in the first example) is a separate example, and within each of those have a 0% bar, a partial-progress bar, an error bar, and a success bar.

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxProgressBar } from '@sonatype/react-shared-components';
+
+export default function NxProgressBarExample() {
+  return (
+    <>
+      <div>
+        <NxProgressBar max={50}
```
rpokorny, line 14: This one should have some surrounding inline content to demonstrate its alignment.  It is also missing the horizontal margin specified in sketch.
image

rpokorny, line 14: According to https://www.w3.org/TR/wai-aria-1.2/#progressbar all progressbars must have accessible names.  This one doesn't, and ChromeVox skips over it without reading anything or even stopping on it.

gallery/src/components/NxProgressBar/NxProgressBarSuccessExample.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <NxProgressBar value={100}
+                       variant="small"
+                       showCounter={true}
+        />
+        <NxProgressBar value={100}
+                       label="10 of 10 of files transferred."
+                       showCounter={true}
+        />
+        <NxProgressBar value={100}
+                       label="10 of 10 of files transferred."
```
rpokorny, line 23:
```suggestion
                       label="10 of 10 files transferred."
```

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+
+  .nx-counter {
+    border-color: var(--nx-swatch-red-40);
+    background-color: var(--nx-swatch-red-95);
+  }
+}
+
+.nx-progress-bar__label {
+  align-items: center;
+  display: flex;
```
rpokorny, line 48: I don't feel like this needs to be flex, it's all just text-based elements that can lay out inline

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+  gap: var(--nx-spacing-2x);
+  grid-area: label;
```
rpokorny, line 50: What's this for?

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+$normal-height: 25px;
+
+progress.nx-progress-bar__progress[value] {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
+
+  border: 0;
+
+  height:$normal-height;
```
rpokorny, line 66:
```suggestion
  height: $normal-height;
```

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+  grid-area: label;
+
+  .nx-icon {
+    margin: 0 0 0 var(--nx-spacing-2x);
+  }
+}
+
+$normal-height: 25px;
+
+progress.nx-progress-bar__progress[value] {
```
rpokorny, line 59: Why is `[value]` part of all of these selectors?  For that matter why is `progress` part of these selectors?

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+  border: 0;
+
+  height:$normal-height;
+  width: 100%;
+}
+
+// Track
+progress.nx-progress-bar__progress[value],
+progress.nx-progress-bar__progress[value]::-webkit-progress-bar {
+  background: var(--nx-swatch-grey-95);
```
rpokorny, line 73: Should be indigo-95 according to sketch

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+  border-radius: $normal-height;
```
rpokorny, line 74: Should be half the height

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+// Indicator
+progress.nx-progress-bar__progress[value]::-moz-progress-bar {
+  background: var(--nx-swatch-teal-50);
+}
+
+progress.nx-progress-bar__progress[value]::-webkit-progress-value {
+  background: var(--nx-swatch-teal-50);
```
rpokorny, line 109: Should be teal-60 according to sketch.  Also these two blocks can be combined to avoid repeating the color

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import { Props, propTypes } from './types';
+
+import './NxProgressBar.scss';
+
+const NxProgressBar = forwardRef<HTMLProgressElement, Props>(
+    function NxProgressBar(props, ref) {
+      const {
+        hasError,
+        inlineCounter = false,
```
rpokorny, line 21: We generally can't use this syntax since it only considers `undefined` and not `null`.  We have to be able to handle both due to the way optional `PropTypes` work

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+      const progressId = useUniqueId('nx-progress-bar__progress');
+
+      const renderCounter = () => {
+        if (!showCounter || variant === 'inline') {
+          return null;
+        }
+
+        return (
+          <div className="nx-progress-bar__counter">
```
rpokorny, line 42: What's the purpose of this wrapper element?

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+        if (percentage === 100 && (labelSuccess || label)) {
+          return (
+            <>
+              <NxFontAwesomeIcon icon={faCheckCircle} />
+              <span className="nx-progress-bar__label-text">{labelSuccess || label}</span>
+            </>
+          );
+        }
+
+        return <span className="nx-progress-bar__label-text">label</span>;
```
rpokorny, line 71:
```suggestion
        return <span className="nx-progress-bar__label-text">{label}</span>;
```

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import { useUniqueId } from '../../util/idUtil';
+import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import { Props, propTypes } from './types';
+
+import './NxProgressBar.scss';
+
+const NxProgressBar = forwardRef<HTMLProgressElement, Props>(
+    function NxProgressBar(props, ref) {
+      const {
```
rpokorny, line 19: Extra native props including `className` and `id` should be handled

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+        expect(getShallowComponent({ value: 0 }).find(counterSelector)).toHaveText('0%');
+        expect(getShallowComponent({ value: 50 }).find(counterSelector)).toHaveText('50%');
+        expect(getShallowComponent({ value: 100 }).find(counterSelector)).toHaveText('100%');
+      }
+  );
+
+  it('sets the correct success className when progress is 100%',
+      function() {
+        const component = getShallowComponent({ value: 0 });
+        const successComponent = getShallowComponent({ value: 100 });
```
rpokorny, line 55: Probably should test this with a custom max as well

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+        expect(successLabelComponent).toContainMatchingElement(NxFontAwesomeIcon);
+        expect(successLabelComponent.find(labelTextSelector)).toHaveText('success label');
+      }
+  );
+
+  it('sets the correct error className when hasError is true',
+      function() {
+        const component = getShallowComponent({ value: 100 });
+        const errorComponent = getShallowComponent({ value: 100, hasError: true });
```
rpokorny, line 79: Why is `value` being set in this test?

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+        return <span className="nx-progress-bar__label-text">label</span>;
+      };
+
+      const renderCounterAndLabel = () => {
+        if (variant === 'inline') {
+          return null;
+        }
+
+        return (
+          <label htmlFor={progressId} className="nx-progress-bar__label">
```
rpokorny, line 80: Could the `<label>` be the wrapping element rather than a sibling? Then you wouldn't need to deal with the id

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > What's this for?

Ops, removed: 038fa7afcecbb1202fa5858f30d2bebf690f2cd0

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > Should be indigo-95 according to sketch

Fixed: 12b27041fdcedd016145c83e35724bc0802e1aed

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > Should be teal-60 according to sketch. Also these two blocks can be combined to avoid repeating the color

The color is fixed here: 12b27041fdcedd016145c83e35724bc0802e1aed

I had these two combined before, but for some reason that broke the styling and the only fix was having them separated like that. Going to investigate

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > Should be half the height

Fixed: 2f176c9ee65a1e4feebe909db0b133920add2e29

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > What's the purpose of this wrapper element?

Removed: e7b1944aa769a038f569858f8b2f13f52bca69aa

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > Extra native props including `className` and `id` should be handled

Fixed: 19a5769c8bc492f419e07360d86b750655c95708

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > Could the `<label>` be the wrapping element rather than a sibling? Then you wouldn't need to deal with the id

Made that change here: d59382ac8739c3784a9a9d2da10103eab3e3cc9b

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > We generally can't use this syntax since it only considers `undefined` and not `null`. We have to be able to handle both due to the way optional `PropTypes` work

Updated it here: 06102db95349206668c0fbd1a50442d09ea010d1

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: > According to https://www.w3.org/TR/wai-aria-1.2/#progressbar all progressbars must have accessible names. This one doesn't, and ChromeVox skips over it without reading anything or even stopping on it.

I made label required here: 1794de009617bd465405ca7af6f8ec11824f0f72
When `variant` is "small", "inline", or `inlineCounter` is true (the label is hidden), the label is assigned to aria-label on the progress element. I also changed the label tag to span when variant is "inline".

Not entirely sure if this is a good solution 🤔 let me know your thoughts.

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > Why is `[value]` part of all of these selectors? For that matter why is `progress` part of these selectors?

I removed progress as part of this d59382ac8739c3784a9a9d2da10103eab3e3cc9b.

I tried removing [value] before but it seems to break the progress styling 🤔

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
rpokorny: > I made label required here: https://github.com/sonatype/sonatype-react-shared-components/commit/1794de009617bd465405ca7af6f8ec11824f0f72
> When variant is "small", "inline", or inlineCounter is true (the label is hidden), the label is assigned to aria-label on the progress element. I also changed the label tag to span when variant is "inline".
>
> Not entirely sure if this is a good solution thinking let me know your thoughts.


That seems good to me.  The only case I'm not sure about is if the progress bar is next to some text or something that makes it redundant from a screenreader perspective.  In that case what I would have thought would make sense would be to include no label on the progress bar and to give it a role of `none` (or `presentation` which is the same thing).  However, looking at https://w3c.github.io/html-aria/#el-progress it doesn't seem like that role is allowed on this element which surprises me.  I thought it was allowed on every element.

rpokorny: > In that case what I would have thought would make sense would be to include no label on the progress bar and to give it a role of none (or presentation which is the same thing). However, looking at https://w3c.github.io/html-aria/#el-progress it doesn't seem like that role is allowed on this element which surprises me. I thought it was allowed on every element.

Looking at https://www.w3.org/TR/wai-aria/#aria-hidden , it seems like maybe `aria-hidden` is the correct way to accomplish that

lib/src/components/NxProgressBar/NxProgressBar.scss
rpokorny: > I had these two combined before, but for some reason that broke the styling and the only fix was having them separated like that. Going to investigate

Very strange.  I'm seeing that too

rpokorny: > Very strange. I'm seeing that too

Judging from [this](https://stackoverflow.com/questions/70550606/is-there-a-reason-why-css-pseudo-elementnon-standard-cannot-be-combined-with-c), I believe what's happening is any part of the selector includes a psuedo-element that Chrome doesn't recognize (like `::-moz-...`), Chrome will error out parsing the entire compound selector and not consider the half that it could have understood

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: > > In that case what I would have thought would make sense would be to include no label on the progress bar and to give it a role of none (or presentation which is the same thing). However, looking at https://w3c.github.io/html-aria/#el-progress it doesn't seem like that role is allowed on this element which surprises me. I thought it was allowed on every element.
> 
> Looking at https://www.w3.org/TR/wai-aria/#aria-hidden , it seems like maybe `aria-hidden` is the correct way to accomplish that

Thanks! Made a change here so that aria-hidden is set to true when progress is accompanied by text: a32c4f9f05950afecb13c94e71382d9c6413f817

Additionally, I made `labelError` to be required when `hasError` is true: 6e94da26a79a5bb93d1197df875142b3456e44ac

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > > I tried removing [value] before but it seems to break the progress styling thinking
> 
> I just tried and removing it made no difference, everything still seemed to work fine

Oh ops I tried removing the wrong one. I removed it as part of this commit: 6e94da26a79a5bb93d1197df875142b3456e44ac

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > I don't feel like this needs to be flex, it's all just text-based elements that can lay out inline

yup! removed it here: 98c00140628196d4cb2eb0182c85efbba11d5fd3

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: Updated the description here: 2a4f6adc7ec8ee17bb5293d17c89b853248a9451

rpokorny: > Thanks! Made a change here so that aria-hidden is set to true when progress is accompanied by text: https://github.com/sonatype/sonatype-react-shared-components/commit/a32c4f9f05950afecb13c94e71382d9c6413f817

That's not what I was suggesting.  What I was suggesting is that `label` not be a required prop, because it is conceivable that an inline progress bar might be used adjacent to some other text or something which makes the label redundant.  In that scenario, the progress component should maybe not be labelled and be hidden from the screenreader.  Maybe we just shouldn't worry about it though.  It's an edge case, it's not the end of the world if we don't do it (it'll just read some redundant content), and I'm not certain the solution we have in mind is standards complaint - does a hidden element get out of needing a label?

rpokorny: > That's not what I was suggesting. What I was suggesting is that label not be a required prop, because it is conceivable that an inline progress bar might be used adjacent to some other text or something which makes the label redundant. In that scenario, the progress component should maybe not be labelled and be hidden from the screenreader. Maybe we just shouldn't worry about it though. It's an edge case, it's not the end of the world if we don't do it (it'll just read some redundant content), and I'm not certain the solution we have in mind is standards complaint - does a hidden element get out of needing a label?

Going back to what you had in https://github.com/sonatype/sonatype-react-shared-components/commit/1794de009617bd465405ca7af6f8ec11824f0f72 seems adequate to me

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,123 @@
...
+        value,
+        variant,
+        ...otherAttributes
+      } = props;
+
+      const showCounter = showCounterProp ?? true;
+      const max = maxProp ?? 100;
+      const percentage = Math.round(value / max * 100);
+
+      let labelText = label;
```
rpokorny, line 36: Mutable variables can get confusing pretty quickly, especially when they are mutated in helper functions.  I'd very much prefer to keep everything using `const`s

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+      const renderCounter = () => {
```
rpokorny, line 38: All of these functions, always being called exactly once, don't really need to be functions.  They could either be local variables or just included directly in the component's return expression.

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,123 @@
...
+            {renderCounter()}
+            {inlineCounter ? null : renderLabel()}
+          </span>
+        );
+      };
+
+      const classes = classnames('nx-progress-bar', {
+        'nx-progress-bar--inline': variant === 'inline',
+        'nx-progress-bar--small': variant === 'small',
+        'nx-progress-bar--full': variant === 'full',
```
rpokorny, line 95:
```suggestion
      const classes = classnames('nx-progress-bar', `nx-progress-bar--${variant}`, {
```

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,123 @@
...
+        'nx-progress-bar--inline-counter': inlineCounter,
+        'nx-progress-bar--error': hasError,
+        'nx-progress-bar--success': percentage === 100
+      }, className);
+
+      const WrapperElement = variant === 'inline' ? 'span' : 'label';
```
rpokorny, line 101: This needs to be a `label` always.  You can set it to inline in CSS

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,123 @@
...
+      const assignProgressAriaLabel = !!(variant === 'inline' || variant === 'small' || inlineCounter);
+      const progressAriaLabel = assignProgressAriaLabel ? labelText : undefined;
```
rpokorny, line 103: I think this'll be going away in response to some of the other comments, but if not, I'm finding it very confusing and wonder if it could be refactored or given some better variable names to be clearer

lib/src/components/NxProgressBar/types.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+interface PropsWithError extends BaseProps {
+  hasError: true;
+  labelError: string;
+}
+
+interface PropsWithoutError extends BaseProps {
+  hasError?: false | null;
+  labelError?: null;
```
rpokorny, line 30: Do we need both of these props? If the `labelError` is always defined when there's an error, and also only defined when there's an error, then `hasError` is redundant

lib/src/components/NxProgressBar/types.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export type Props = PropsWithError | PropsWithoutError;
+
+export const propTypes: PropTypes.ValidationMap<Props> = {
+  inlineCounter: PropTypes.bool,
+  label: PropTypes.string.isRequired,
+  labelError: (props, propName) =>
+    (props.hasError === true && (!props[propName] || typeof props[propName] !== 'string'))
+      ? new Error(`${propName} is required when hasError is true`) : null,
```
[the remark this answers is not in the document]
entroform: > I'm not sure there's value in checking this both here and within the component. It's a runtime error either way, so I think one should suffice. In the component is what we most typically do

Removed: 59cf2b01a1416a04c9f097e60f60bde9891ba2b6

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: > > That's not what I was suggesting. What I was suggesting is that label not be a required prop, because it is conceivable that an inline progress bar might be used adjacent to some other text or something which makes the label redundant. In that scenario, the progress component should maybe not be labelled and be hidden from the screenreader. Maybe we just shouldn't worry about it though. It's an edge case, it's not the end of the world if we don't do it (it'll just read some redundant content), and I'm not certain the solution we have in mind is standards complaint - does a hidden element get out of needing a label?
> 
> Going back to what you had in [1794de0](https://github.com/sonatype/sonatype-react-shared-components/commit/1794de009617bd465405ca7af6f8ec11824f0f72) seems adequate to me

Reverted here: ec7ea78d46a3723b218f42de4188989fb987e028

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > This needs to be a `label` always. You can set it to inline in CSS

Fixed here: 90ff9ad9be471cf52e92754ad72052e46b897545

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > Mutable variables can get confusing pretty quickly, especially when they are mutated in helper functions. I'd very much prefer to keep everything using `const`s

Thanks, I cleaned it up here: f8899749dc8630480e63265407d430690f5ec6d0

lib/src/components/NxProgressBar/NxProgressBar.tsx
entroform: > All of these functions, always being called exactly once, don't really need to be functions. They could either be local variables or just included directly in the component's return expression.

I cleaned it up here: f8899749dc8630480e63265407d430690f5ec6d0

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
entroform: > Why is `value` being set in this test?

Ops, removed it as part of this commit: 40650085762ac5ec62e6e5927255bc82dc399347

entroform:
> Don't forget to add this new component to the ssr-tests

Added ssr-tests here: 8e41296ee94d4b888a19cec16c59defbf2fbf2a0

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
entroform: > Probably should test this with a custom max as well

Fixed: 40650085762ac5ec62e6e5927255bc82dc399347

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: > This one should have some surrounding inline content to demonstrate its alignment. It is also missing the horizontal margin specified in sketch. image

Fixed as part of this commit: d2db7cf70a0250f16e446e9ff7882131108e315e

entroform:
> I think a few more examples might be necessary - mainly, inline examples for success and error. To really cover all the bases, what you could do is reorganize the examples so that each layout type (e.g. each variation currently shown in the first example) is a separate example, and within each of those have a 0% bar, a partial-progress bar, an error bar, and a success bar.

Reorganized examples here: d2db7cf70a0250f16e446e9ff7882131108e315e

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+        expect(getShallowComponent({ value: 0 }).find(progressSelector)).toHaveProp('max', 100);
+        expect(getShallowComponent({ value: 0, max: 50 }).find(progressSelector)).toHaveProp('max', 50);
+        expect(getShallowComponent({ value: 20, max: 40 }).find(progressSelector)).toHaveProp('value', 20);
+      }
+  );
+
+  it('displays the correct className and hides label when inlineCounter is set to true',
+      function() {
+        const component = getShallowComponent({ value: 100, label: 'hello' });
+        const inlineCounterComponent = getShallowComponent({ value: 100, inlineCounter: true, label: 'hello' });
```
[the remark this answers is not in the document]
entroform: > Should probably test inline in conjunction with success and error

Updated here: 93a20d838f78afff3f3519a47448f51aafbaa2bc

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
rpokorny: > Fixed as part of this commit: https://github.com/sonatype/sonatype-react-shared-components/commit/d2db7cf70a0250f16e446e9ff7882131108e315e

Come to think of it, we probably need a mockup from design to demonstrate exactly what the correct vertical alignment of inline progress bars relative to text is.  What you have now might be good but I'm not certain.

lib/src/components/NxProgressBar/NxProgressBar.scss
rpokorny: > Oh ops I tried removing the wrong one. I removed it as part of this commit: https://github.com/sonatype/sonatype-react-shared-components/commit/6e94da26a79a5bb93d1197df875142b3456e44ac

To be clear, I'm asking about all of them.  When I tried two days ago I removed all of them with no ill effect.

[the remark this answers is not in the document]
rpokorny: > For some strange reason nesting it breaks the progress bar styling on Chrome

This is likely the same issue described in the other thread here: https://github.com/sonatype/sonatype-react-shared-components/pull/670#discussion_r833510296 .  See if you can nest as much as possible while avoiding that issue

gallery/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,189 @@
...
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell><NxCode>labelError</NxCode></NxTable.Cell>
+            <NxTable.Cell>
+              string
+            </NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>
+              If this label is specified the component progress is set to 0% and
+              error styling is applied.
```
rpokorny, line 94: Should probably mention that in addition to those two ancillary effects, this string also gets displayed instead of the `label`

gallery/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,189 @@
...
+                        codeExamples={nxProgressBarSmallExampleCode}>
+      Examples of "small" variant <NxCode>NxProgressBar</NxCode> showing 0% progress,
+      a partial-progress, success (100% progress), and error state progress bar with and without inlineCounter.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="NxProgressBar Full Example"
+                        id="nx-progress-bar-full-example"
+                        liveExample={NxProgressBarFullExample}
+                        codeExamples={nxProgressBarFullExampleCode}>
+      Examples of "full" variant (default) <NxCode>NxProgressBar</NxCode> showing 0% progress,
```
rpokorny, line 175:
```suggestion
      Examples of "full" variant <NxCode>NxProgressBar</NxCode> showing 0% progress,
```

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: > > Fixed as part of this commit: [d2db7cf](https://github.com/sonatype/sonatype-react-shared-components/commit/d2db7cf70a0250f16e446e9ff7882131108e315e)
> 
> Come to think of it, we probably need a mockup from design to demonstrate exactly what the correct vertical alignment of inline progress bars relative to text is. What you have now might be good but I'm not certain.

I'll reach out to Jason!

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > > Oh ops I tried removing the wrong one. I removed it as part of this commit: [6e94da2](https://github.com/sonatype/sonatype-react-shared-components/commit/6e94da26a79a5bb93d1197df875142b3456e44ac)
> 
> To be clear, I'm asking about all of them. When I tried two days ago I removed all of them with no ill effect.

Removed: c523b2aa91b4ef5936961488bee8d6307f897d41

gallery/src/components/NxProgressBar/NxProgressBarExample.tsx
entroform: > > Fixed as part of this commit: [d2db7cf](https://github.com/sonatype/sonatype-react-shared-components/commit/d2db7cf70a0250f16e446e9ff7882131108e315e)
> 
> Come to think of it, we probably need a mockup from design to demonstrate exactly what the correct vertical alignment of inline progress bars relative to text is. What you have now might be good but I'm not certain.

I talked to Jason, and they confirmed that it will be how it is in the example right now. They are working on updating the Sketch file 👍

gallery/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell><NxCode>labelError</NxCode></NxTable.Cell>
+            <NxTable.Cell>
+              string
+            </NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>
+              When specified, error styling is applied, the component progress is set to 0%, and,
+              this string will be displayed instead of <NxCode>label</NxCode>.
```
rpokorny, line 95:
```suggestion
              When specified, error styling is applied, the component progress is set to 0%, and
              this string will be displayed instead of <NxCode>label</NxCode>.
```

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,106 @@
...
+  gap: var(--nx-spacing-2x);
+  margin: var(--nx-spacing-6x) 0;
+
+  &--inline-counter {
+    flex-direction: row;
+  }
+
+  &--inline {
+    display: inline-flex;
+    margin: 0 var(--nx-spacing-6x);
```
rpokorny, line 21: The way these margins are set up causes a minor problem: when in `inline-counter` mode, the counter's right margin is still active, effectively mandating that much extra space at the right end of the component.  This is particularly relevant in full-width mode, where that extra space makes it not-quite-full-width.

My recommendation would be for each of the three second-row elements - the counter, icon, and label - to exclusively use left margin to manage the expected space between them.

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,106 @@
...
+}
+
+.nx-progress-bar__counter-and-label {
+  .nx-icon {
+    margin: 0 var(--nx-spacing-2x);
+  }
+}
+
+.nx-progress-bar__counter {
+  margin: 0 var(--nx-spacing-2x) 0 0;
```
rpokorny, line 51: The way these margins are set up causes a minor problem: when in inline-counter mode, the counter's right margin is still active, effectively mandating that much extra space at the right end of the component. This is particularly relevant in full-width mode, where that extra space makes it not-quite-full-width.

My recommendation would be for each of the three second-row elements - the counter, icon, and label - to exclusively use left margin to manage the expected space between them.

lib/src/components/NxProgressBar/NxProgressBar.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+      return (
+        <label className={classes}>
+          <progress ref={ref}
+                    aria-label={!showLabelElement ? labelText : undefined}
+                    className="nx-progress-bar__progress"
+                    value={labelError ? 0 : value}
+                    max={max}
+                    {...otherAttributes}>
+          </progress>
```
rpokorny, line 65:
```suggestion
                    {...otherAttributes} />
```

lib/src/components/NxProgressBar/NxProgressBar.scss
entroform: > The way these margins are set up causes a minor problem: when in inline-counter mode, the counter's right margin is still active, effectively mandating that much extra space at the right end of the component. This is particularly relevant in full-width mode, where that extra space makes it not-quite-full-width.
> 
> My recommendation would be for each of the three second-row elements - the counter, icon, and label - to exclusively use left margin to manage the expected space between them.

I changed the margins for counter, icon, and label to left margin but I noticed that the spacing between the counter and the icon or label is larger. I thought it might be more difficult to manage that extra spacing with left margin because there are cases where it might be icon + label or just the label itself.

I added the right margin back here but removed it when it is inline-counter in these two commits.
(Also added vertical-align middle to center vertically)
Let me know your thoughts:

cbc48b5d5441963ab729607b574a50e4a57848d6
66f39ea76b30c03c5ba9f4630401f0fc3f0f74ce

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,118 @@
...
+}
+
+
+
+.nx-progress-bar__label-text {
+  margin-left: var(--nx-spacing-2x);
+  vertical-align: middle;
+}
+
+$normal-height: 25px;
```
rpokorny, line 66: This should be a local variable
```suggestion
$-normal-height: 25px;
```

lib/src/components/NxProgressBar/NxProgressBar.scss
```diff
@@ -0,0 +1,118 @@
...
+    margin: 0 0 0 var(--nx-spacing-2x);
+    vertical-align: middle;
+  }
+}
+
+
+
+.nx-progress-bar__label-text {
+  margin-left: var(--nx-spacing-2x);
+  vertical-align: middle;
```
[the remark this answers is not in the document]
entroform: > Why are these vertical-aligns needed? It seems to more closely match the mockup without them

Removed them here: c4259c55d6f2ed89459a2d59b1df71ff27b32007

rpokorny (approved):

PR #672, opened 2022-03-23
NxTransferListHalf Docs - RSC-888

rpokorny:
https://issues.sonatype.org/browse/RSC-888


avnvbhatta (approved):
Minor: missing some periods in the `props` table.

PR #674, opened 2022-03-25
Fixed nx-tile header height - RSC-823

entroform:
https://issues.sonatype.org/browse/RSC-823

https://github.com/sonatype/sonatype-react-shared-components/pull/617

### DIFF
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -40,7 +40,7 @@
   display: grid;
   grid-template:
     "title      tags   actions" 40px
-    "subtitle subtitle _" min-content /
+    "subtitle subtitle ." min-content /
      auto       1fr       auto;
 
   + .nx-tile-content {
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -7,9 +7,6 @@
 @use '../scss-shared/nx-container-helpers';
 @use '../scss-shared/nx-text-helpers';
 
-$-tile-header-height: 40px; // Minimum height
-$-tile-header-height-actions: 43px; // Height with sub-title and/or right actions
-
 /*
   #Name-spaced Tile System
 */
@@ -45,10 +42,9 @@ $-tile-header-height-actions: 43px; // Height with sub-title and/or right action
   // NOTE: without the min-content sizing on the rows it adds extra vertical space when only title and actions are
   // present. I don't have an explanation for this, it doesn't happen with _any_ other combination of header children
   grid-template:
-    "title      tags   actions" min-content
+    "title      tags   actions" 40px
     "subtitle subtitle actions" min-content /
      auto       1fr       auto;
-  min-height: $-tile-header-height;
 
   + .nx-tile-content {
     padding-top: var(--nx-spacing-4x);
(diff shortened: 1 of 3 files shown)


lib/src/base-styles/_nx-tile.scss
```diff
@@ -45,10 +42,9 @@ $-tile-header-height-actions: 43px; // Height with sub-title and/or right action
   // NOTE: without the min-content sizing on the rows it adds extra vertical space when only title and actions are
```
rpokorny, line 42: Since we're getting rid of one of the two `min-content` usages, I'm not 100% sure this comment still fully applies.

lib/src/base-styles/_nx-tile.scss
```diff
@@ -45,10 +42,9 @@ $-tile-header-height-actions: 43px; // Height with sub-title and/or right action
...
   // present. I don't have an explanation for this, it doesn't happen with _any_ other combination of header children
   grid-template:
-    "title      tags   actions" min-content
+    "title      tags   actions" 40px
     "subtitle subtitle actions" min-content /
```
rpokorny, line 46: I don't think we need or want the `actions` area to extend to this row anymore.  Which begs the question, should that last cell in the grid become part of the subtitle, or should it become unused?

lib/src/base-styles/_nx-tile.scss
entroform: > Since we're getting rid of one of the two `min-content` usages, I'm not 100% sure this comment still fully applies.

Thanks, removed it here: 8cf12124868d92ed59719abfba159faadb763ff4

lib/src/base-styles/_nx-tile.scss
entroform: > I don't think we need or want the `actions` area to extend to this row anymore. Which begs the question, should that last cell in the grid become part of the subtitle, or should it become unused?

After talking with Jason, it was decided that it should become unused: I made an amendment here: 7dff084045a5732059fceacaf5037c260e287d8a

I used "_" to indicate unused grid-area, let me know your thoughts!

rpokorny: > I used "_" to indicate unused grid-area, let me know your thoughts!

Actually CSS has a defined way to label unused areas: `.`

entroform: > > I used "_" to indicate unused grid-area, let me know your thoughts!
> 
> Actually CSS has a defined way to label unused areas: `.`

Oh thanks! Changed it here: 62da7e966a7d58dcf57ec57c78e444825fbd33db

lib/src/base-styles/_nx-tile.scss
```diff
@@ -41,14 +38,10 @@ $-tile-header-height-actions: 43px; // Height with sub-title and/or right action
...
   align-items: center;
   display: grid;
-
-  // NOTE: without the min-content sizing on the rows it adds extra vertical space when only title and actions are
-  // present. I don't have an explanation for this, it doesn't happen with _any_ other combination of header children
   grid-template:
-    "title      tags   actions" min-content
-    "subtitle subtitle actions" min-content /
+    "title      tags   actions" 40px
+    "subtitle subtitle ." min-content /
```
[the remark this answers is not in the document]
entroform: > I suspect the `min-content` here can now be `auto`

Changed it to `auto` here: 900b9bc3cade895080ea6f051ed5e3794c0b794e

rpokorny (approved):

PR #677, opened 2022-03-29
Fixed tile title alignment and subtitle spacing - RSC-963

entroform:
https://issues.sonatype.org/browse/RSC-963

### DIFF
--- a/lib/src/base-styles/_nx-tile.scss
+++ b/lib/src/base-styles/_nx-tile.scss
@@ -36,7 +36,7 @@
 }
 
 .nx-tile-header {
-  align-items: top;
+  align-items: start;
   display: grid;
   grid-template:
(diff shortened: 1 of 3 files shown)


lib/src/base-styles/_nx-tile.scss
```diff
@@ -36,7 +36,7 @@
 }
 
 .nx-tile-header {
-  align-items: center;
+  align-items: top;
```
rpokorny, line 39: This isn't a valid value for this property

entroform: > This isn't a valid value for this property

Ops, I fixed it here: 1922addd2f72058347203b17a89c352515e3e9aa

rpokorny (approved):

PR #681, opened 2022-04-04
debounce the link RSC-892

glenhunter:
https://issues.sonatype.org/browse/RSC-892

The issue seems to be a collision between the use of `align items: first baseline` and the checkbox component when it's in a list. I created custom rule for the gallery that uses `align-items: start` for `.nx-list__item` instead.


rpokorny:
While looking at this I notice that the styles written for `.gallery-accessibility-list .nx-checkbox__box` are affecting the filter list at the top in addition to the criteria lists.  Is that intentional?

gallery/src/main.scss
```diff
@@ -128,6 +128,10 @@
     white-space: normal;
   }
 
+  .nx-list__item {
+    align-items: flex-start;
```
rpokorny, line 132: A better alignment can be achieved by overriding `align-items` to `first baseline` on the `.nx-radio-checkbox` instead of this.

glenhunter: > A better alignment can be achieved by overriding `align-items` to `first baseline` on the `.nx-radio-checkbox` instead of this.

Very much not better IMO 
Screen Shot 2022-04-05 at 2 15 18 PM

rpokorny: > Very much not better IMO

I think you must not have done quite what I meant.  It looks like the following for me.  Note that the checkbox text and the link are baseline-aligned which is not the case in this PR currently. 
image

glenhunter: > > Very much not better IMO
> 
> I think you must not have done quite what I meant. It looks like the following for me. Note that the checkbox text and the link are baseline-aligned which is not the case in this PR currently. image

To my eye this is six of one, half a dozen of the other but I fixed the problem above at the same time so: https://github.com/sonatype/sonatype-react-shared-components/pull/681/commits/2d426353c77a62d6762e9df25208442f380cfa7e

glenhunter:
> While looking at this I notice that the styles written for `.gallery-accessibility-list .nx-checkbox__box` are affecting the filter list at the top in addition to the criteria lists. Is that intentional?

https://github.com/sonatype/sonatype-react-shared-components/pull/681/commits/2d426353c77a62d6762e9df25208442f380cfa7e

PR #682, opened 2022-04-05
Fix lint errors - RSC-970

avnvbhatta:
https://issues.sonatype.org/browse/RSC-970


rpokorny:
If we're going to worry about this I'd be in favor of turning on linting for these files during the build

avnvbhatta:
> If we're going to worry about this I'd be in favor of turning on linting for these files during the build

I don't think we need to turn on linting. I think a "code format" was long overdue for some of these files, and I just refactored to make them a bit organized.

gallery/visualtests/nxOverflowTooltip.js
```diff
@@ -8,16 +8,11 @@ const { setupBrowser } = require('./testUtils');
...
-    focusAndHoverTest,
-    hoverTest,
-    simpleTest,
     getPage,
     waitAndGetElements,
     wait,
-    getElements
- , a11yTest } = setupBrowser('#/pages/Overflow%20Tooltip');
+    getElements,
+    a11yTest } = setupBrowser('#/pages/Overflow%20Tooltip');
```
rpokorny, line 15:
```suggestion
    a11yTest
  } = setupBrowser('#/pages/Overflow%20Tooltip');
```

gallery/visualtests/nxOverflowTooltip.js
```diff
@@ -207,6 +191,6 @@ describe('NxOverflowTooltip', function() {
     await waitAndGetElements(tooltipSelector);
 
     //disabling the region rule to get around the "Some page content is not contained by landmarks" for tooltips
-    await a11yTest(builder => builder.disableRules('region'));
+    a11yTest(builder => builder.disableRules('region'));
```
rpokorny, line 194: This doesn't seem right.  I think it needs to be 
```suggestion
    await a11yTest(builder => builder.disableRules('region'))();
```

gallery/visualtests/testUtils.js
```diff
@@ -6,6 +6,7 @@
  */
 const puppeteer = require('puppeteer');
 
+// eslint-disable-next-line no-undef
```
rpokorny, line 9: I assume adding the `node` env to the eslint config would be the more correct way to fix this

gallery/.eslintrc
```diff
@@ -4,6 +4,9 @@
     "require": true,
     "document": true
   },
+  "env": {
+    "jest": true
```
rpokorny, line 8: I'd rather this only be active for linting the tests, which would require a separate `test-lint` script in `package.json` and a separate eslint config file.  That's complex enough that I do think this should be split to a separate ticket after all.

avnvbhatta: > I'd rather this only be active for linting the tests, which would require a separate `test-lint` script in `package.json` and a separate eslint config file. That's complex enough that I do think this should be split to a separate ticket after all.

I removed the eslint changes for this PR and created a tech debt ticket for it.
https://issues.sonatype.org/browse/RSC-978

rpokorny (approved):

PR #692, opened 2022-04-08
Update NxPageHeader Layout - RSC-876

rpokorny:
https://issues.sonatype.org/browse/RSC-876


gallery/src/PageHeader/PageHeader.scss
```diff
@@ -7,7 +7,9 @@
 .gallery-page-header {
   $logo-image-size: 40px;
 
-  box-sizing: border-box;
+  > .nx-page-header__inner {
+    align-items: center;
+  }
```
rpokorny, line 12: Note that these styles are for the gallery's semi-custom page header

rpokorny:
During implementation it was noticed that the gallery's custom page header was 1px shorter than it was supposed to be.  That is fixed here and is the reason for many of the visual test diffs

lib/src/components/NxPageHeader/NxPageHeader.tsx
```diff
@@ -17,6 +17,7 @@ const logoImg = importImage('sonatype-logo-with-hexagon.png');
 function HeaderProductInfo({ name, version }: ProductInfo) {
   return (
     <>
+      <div role="presentation" className="nx-product__divider">|</div>
```
rpokorny, line 20: This is no longer being done as a `::before`` because it made the styling undesirably complicated, as this bar needed a different top margin than the name.

entroform (approved):
LGTM

PR #699, opened 2022-04-14
Improvement Day: git lfs

rpokorny:
This PR adds use of git-lfs to manage the visual test screenshots.  From what I understand, the value of this is that the clone of the repo itself will no longer continue to grow as new screenshot versions are added.  Instead, within the repo itself small text files are stored which point to the images.  During checkout of a given commit, the git-lfs tooling (which is a plugin for git itself) will fetch the binaries that are present _just in that commit_ using separate REST calls, and replaces the text file pointers with the actual binaries in your local checkout.  Essentially, at any given point in time your local checkout will have the real files as they should be in that commit, and your system will still have taken the time to download those exact versions of those files, but you won't have to have every version of every binary file downloaded locally at all times like you normally would with plain git.

Using this branch (or the repo in general once this is merged) will require installation of the separate git-lfs tool.  I'm unclear on how well it integrates with git GUIs, though the GUI that Github themselves provide does support it.  Testing that aspect would be welcome as people review this PR.


entroform (approved):
This looks good to me! I tested it and it looks like it's working.

Just wondering if we should be adding a note about installing git-lfs to the README?
I have not tested it with any git GUIs though because I don't use one.

rpokorny:
> Just wondering if we should be adding a note about installing git-lfs to the README?

Great point.  I've added a mention in the readme in [ce52c83](https://github.com/sonatype/sonatype-react-shared-components/pull/699/commits/ce52c8349958afa9b3f56edb872318947e010627).  Please re-review and lmk if that seems good.

entroform:
> > Just wondering if we should be adding a note about installing git-lfs to the README?
> 
> Great point. I've added a mention in the readme in [ce52c83](https://github.com/sonatype/sonatype-react-shared-components/pull/699/commits/ce52c8349958afa9b3f56edb872318947e010627). Please re-review and lmk if that seems good.

This looks good to me!

PR #704, opened 2022-04-19
Testing ignore script - RSC-980

entroform:


README.md
```diff
@@ -150,8 +150,14 @@ The RSC code is split into two separate codebases: the library itself which live
 ## Building
 
 ### Required Software
-Node 10.x or 12.x.  Node 14.x is known not to work currently.
-yarn 1.21.1
+<!-- Node 10.x or 12.x. Node 14.x is known not to work currently. -->
+Node 12.x
```
rpokorny, line 154: This is out of date even before this PR.  We're on Node 16 now and also support 14

lib/.gitignore
```diff
@@ -7,3 +7,11 @@
...
 npm-debug.log
 junit.xml
+
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
rpokorny, line 17: We have a top-level `.gitignore` which woul be the right place for these entries that are in both `lib` and `gallery`

gallery/package.json
```diff
@@ -68,19 +68,26 @@
...
     "@react-hook/debounce": "^4.0.0",
-    "@sonatype/react-shared-components": "link:../lib/dist",
+    "@sonatype/react-shared-components": "../lib/dist",
     "axios": "^0.21.1",
     "classnames": "^2.2.6",
     "compare-versions": "^4.1.3",
     "core-js": "^3.8.3",
     "es6-object-assign": "^1.1.0",
     "es6-set": "^0.1.5",
+    "link": "^1.4.0",
```
rpokorny, line 78: What's this?

rpokorny:
Could you explain what it is about yarn 2+ that enables the installation of puppeteer without running its install scripts?

rpokorny:
I suspect an alternative to switching our yarn version would be to use puppeteer's environment variables to disable the download and provide it the path to the chromium executable already present on the system. See https://dev.to/maxprogramming/how-to-skip-chromium-download-in-puppeteer-2c24

entroform:
> Could you explain what it is about yarn 2+ that enables the installation of puppeteer without running its install scripts?

So in yarnrc.yml
I set it so it disables script by default:
```
enableScripts: false
```

Then in package.json, I added this configuration to allow scripts on a specific module:
```
 "dependenciesMeta": {
    "puppeteer": {
      "built": true
    }
  },
```

https://yarnpkg.com/configuration/manifest#dependenciesMeta.built

gallery/package.json
entroform: > What's this?

Oh I meant to remove that!

entroform:
> I suspect an alternative to switching our yarn version would be to use puppeteer's environment variables to disable the download and provide it the path to the chromium executable already present on the system. See https://dev.to/maxprogramming/how-to-skip-chromium-download-in-puppeteer-2c24

I'll ses if I can get this to work on another branch!

PR #707, opened 2022-05-04
React Testing Library - RSC-946

entroform:
# React Testing Library

https://issues.sonatype.org/browse/RSC-946

It looks like RTL have a different philosophy in terms of testing.
Testing how the user will interact with the application and avoiding implementation details.
There's also no shallow rendering.

I found it tricky at, first, to select a container or a label with no text.
I discover a way to do this from the playground example, 
without using `test-id`, by querying it from the container.

[Useful playground to test queries](https://testing-playground.com/)

### jest-dom and jest-enzyme matchers conflict

When setting `@testing-library/jest-dom` globally in setupTests it appears that there were conflicts with
`jest-enzyme`, some of the matches have the same name and jest-dom were overriding them causing test failures.
The solution suggested from the Github issues I read was to import the jest-dom on individual RTL modules.

## fireEvent vs userEvent:

https://stackoverflow.com/questions/64006345/react-testing-library-when-to-use-userevent-click-and-when-to-use-fireevent

## Useful Key Maps

https://github.com/testing-library/user-event/blob/main/src/keyboard/keyMap.ts

## Links:
https://stackoverflow.com/questions/54152562/difference-between-enzyme-reacttestutils-and-react-testing-library#:~:text=Enzyme%20allows%20you%20to%20access,access%20to%20the%20implementation%20details.

https://github.com/testing-library/jest-dom#installation


lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import '@testing-library/jest-dom';
```
rpokorny, line 8: And this doesn't mess up other tests if done in this file?

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+describe('NxToggle', function() {
+  const simpleProps: Props = {
+    inputId: 'toggle-id',
+    isChecked: false,
+    onChange: () => {},
+    disabled: undefined,
+    children: 'Enables whales'
+  };
+
+  it('renders a <label> containing a toggle <input> and toggle control', function() {
```
rpokorny, line 23: This doesn't check for the `<label>` and doesn't check that the `switch` is an `<input>`

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  it('adds classes specified with the className prop', function() {
+    const { container } = render(<NxToggle {...simpleProps} className="foo"></NxToggle>);
+
+    const label = container.querySelector('label');
+
+    expect(label).toHaveClass('foo');
+    expect(label).toHaveClass('nx-toggle');
+  });
+
+  it('calls its onChange prop when the input fires a change event', function() {
```
rpokorny, line 43: This test is no longer testing the change event directly, it's testing the click event.  I guess an additional test would be needed to test keyboard activation.  I wonder if there are patterns for that, so we don't have to encode all the ways a control can be activated into each test suite?

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
entroform: > And this doesn't mess up other tests if done in this file?

Yup! .. and only if you reserve this file for RTL tests and not mix it with enzyme tests.

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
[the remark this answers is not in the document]
entroform: > This test name should probably be rephrased a little bit to call out the `switch` role as that's foundational to the way most of the other tests have been written.

Changed the name here: 3f77912c8f3e4b30b756cc5e5cb2537c483c65ed

entroform: > This doesn't check for the `<label>` and doesn't check that the `switch` is an `<input>`

Ops,
I added a separate test to check for that below.
Not entirely sure if it's the recommended way in RTL.
I read somewhere that RTL philosophy is to avoid testing implementation details and about testing how the user would interact with the computer. Not entirely sure if this counts the former 🤔
3f77912c8f3e4b30b756cc5e5cb2537c483c65ed

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
entroform: > This test is no longer testing the change event directly, it's testing the click event. I guess an additional test would be needed to test keyboard activation. I wonder if there are patterns for that, so we don't have to encode all the ways a control can be activated into each test suite?

Turns out there is `fireEvent.change` but it doesn't work with checkboxes.
https://github.com/testing-library/react-testing-library/issues/156

Most of the discussions on the topic is to use fire the click event.
I can't seem to find any other pattern to do both so it does look like we'll have to test keyboard activation separately.

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
rpokorny: > I read somewhere that RTL philosophy is to avoid testing implementation details and about testing how the user would interact with the computer. Not entirely sure if this counts the former thinking

Yeah I'm not sure that philosophy is as valid when testing a component library.  We need to make (and test) guarantees about what HTML elements we're using so that calling code knows what they're constructing and can make determinations about whether it's resulting in valid, semantically-reasonable DOM.

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
rpokorny: It's not clear to me whether it makes a difference in this case, but some of the RTL docs refer to this additional library that sounds like it's somewhat relevant and is something we should be aware of: https://github.com/testing-library/user-event

rpokorny:
I think it would be a good idea to get a relatively senior IQ team member (especially @marushkevych) to review this as well as they have more experience with RTL than we do.

lib/src/setupTests.ts
```diff
@@ -8,6 +8,12 @@
 
 import 'jest-enzyme';
 
+// We can't have @testing-library/jest-dom imported globally
+// because it will overwrite some of the jest-enzyme matchers.
+// Instead, we need to import them individual in the test modules
+// until we fully transitioned to react-testing-library.
+// import '@testing-library/jest-dom';
```
marushkevych, line 15: We also had conflicts issue with `jasmineEnzyme` matchers in IQ.
The way we resolved it is:
- we load `jasmineEnzyme` matchers in `beforeAll`:
   https://github.com/sonatype/insight-brain/blob/71b8750d19838df221e9a76e004bf228aaf0fe51/insight-brain-frontend/src/test/frontend/specRoot.js#L62

- and then for RTL tests we override the matchers in `render` wrapper:
  https://github.com/sonatype/insight-brain/blob/722ced204aeb16ebd4a950e52e63668ed2e8be22/insight-brain-frontend/src/test/frontend/SpecUtil.js#L349

`render` wrapper is recommended practice if you need to perform some additional set-up that should apply to all your RTL tests. We use it to wrap components under test with Redux store provider.

marushkevych (approved):

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
entroform: > > Yeah I'm not sure that philosophy is as valid when testing a component library. We need to make (and test) guarantees about what HTML elements we're using so that calling code knows what they're constructing and can make determinations about whether it's resulting in valid, semantically-reasonable DOM.
> 
> That said I do get the whole idea of querying by role. What if we refactor that test so that it queries the `switch` role, asserts that the resulting element has a `tagName` of `INPUT`, and then uses the `.labels` property to get the label element and asserts its text content? Asserting the accessible name during the role query would be a good idea too - e.g. `getByRole('switch', { name: 'Enables whales' })`

Thanks that's a good idea. I've updated it here: 2c0f887648cfad445b9a03fb8c3ecc5c2705b501

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  it('adds classes specified with the className prop', function() {
+    const { container } = render(<NxToggle {...simpleProps} className="foo"></NxToggle>);
+
+    const label = container.querySelector('label');
+
+    expect(label).toHaveClass('foo');
+    expect(label).toHaveClass('nx-toggle');
+  });
+
+  it('calls its onChange prop when the input fires a change event', async function() {
```
rpokorny, line 55: This test's name is no longer particularly accurate

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    render(<NxToggle {...simpleProps} onChange={onChange}></NxToggle>);
+
+    const checkbox = screen.getByRole('switch');
+
+    expect(onChange).not.toHaveBeenCalled();
+
+    await userEvent.click(checkbox);
+
+    expect(onChange).toHaveBeenCalled();
+  });
```
rpokorny, line 67: I think we need a test for keyboard interaction as well.  

I'm not sure how I feel about that.  The previous testing paradigm relied on us basically saying "this component uses these browser built-in APIs and we just test the interaction with those APIs". We weren't attempting to test the behaviors of those APIs themselves (e.g, what user interactions cause a checkbox's `change` event to fire).  RTL seems to be opposed to that, but I worry that the new approach dramatically increases our testing burden as well as leaves us with a theoretical blindspot regarding non-standard input systems.  For instance, it seems we are now supposed to test everything about the user interacting with a component via mouse, and via keyboard, and via... what else? Things we might not even imagine.  Whereas before, we were just testing more abstractly "what happens when this component is activated" trusting the browser to send that activation signal at the appropriate time not matter what input device the user is using, even if it's one we aren't familiar with.  Thoughts @sonatype/team-plaid ?

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      isChecked: true,
+      className: 'label-classname',
+      inputAttributes: {
+        id: 'garfield',
+        name: 'garfield',
+        disabled: false,
+        className: 'input-classname',
+        checked: false
+      } as Props['inputAttributes']
+    };
```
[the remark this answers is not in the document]
entroform: > I haven't tried it but I think this should work and would be better if it does:

Unfortunately I got this :(

Screen Shot 2022-06-22 at 9 57 46 AM

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
Blacktiger: By mocking out the interactions between the APIs before you could easily have had something broken not realizing that a certain event isn't fired by certain user interactions right? This way requires you to be a bit more explicit about how a component is interacted with, but it also means you know the component will respond correctly under the specific conditions of the test which match up to how a user will use the component.

If you don't want to test the app they way you expect the user to interact with it you could just use [fireEvent](https://testing-library.com/docs/dom-testing-library/api-events/#fireevent) instead. `fireEvent.change` for example should trigger the `onChange` event without worrying about how the user would actually cause that event to fire. The downside to doing that though is you don't necessarily know that your event handler responds to the change event correctly for both a keyboard activation and a mouse or pointer click.

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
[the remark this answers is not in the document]
rpokorny: > Unfortunately I got this :(

That's because we explicitly don't allow `disabled` to be specified on the `inputAttributes` of `NxToggle`, since it's part of the main props.
```
export type InputAttributesProp =
  Omit<InputHTMLAttributes<HTMLInputElement>,
  'disabled' | 'checked' | 'readonly' | 'onChange'>;
```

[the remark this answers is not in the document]
entroform: > > Unfortunately I got this :(
> 
> That's because we explicitly don't allow `disabled` to be specified on the `inputAttributes` of `NxToggle`, since it's part of the main props.
> 
> ```
> export type InputAttributesProp =
>   Omit<InputHTMLAttributes<HTMLInputElement>,
>   'disabled' | 'checked' | 'readonly' | 'onChange'>;
> ```

Updated: ee21ee3bedf888680b74d745104e502641f677ee

lib/src/components/NxToggle/__tests__/NxToggleRTL.test.tsx
rpokorny: > If you don't want to test the app they way you expect the user to interact with it you could just use [fireEvent](https://testing-library.com/docs/dom-testing-library/api-events/#fireevent) instead. fireEvent.change for example should trigger the onChange event without worrying about how the user would actually cause that event to fire. The downside to doing that though is you don't necessarily know that your event handler responds to the change event correctly for both a keyboard activation and a mouse or pointer click.

Does that actually work? I thought what @entroform was demonstrating here was that RTL doesn't work that way

Blacktiger: Should work just fine. We still use `fireEvent.change` in the nxrm codebase though we've been switching to userEvent over time since that more accurately simulates the browser events. For example I have this little helper function for changing form fields. You can see how we had to simulate both target and currentTarget because some handlers were using one or the other (not sure if that still applies, but it did at one point).

```
  static async changeField(fieldSelector, value) {
    fireEvent.change(fieldSelector(), {
      currentTarget: {
        name: fieldSelector().name,
        value
      },
      target: {
        name: fieldSelector().name,
        value
      }
    });
    try {
      await waitFor(() => expect(fieldSelector()).toHaveValue(value));
    } catch(error) {
      throw new Error(`${fieldSelector().name} with value ${fieldSelector().value} did not match the expected value \n ${error.message}`);
    }
  }
```

entroform: > I think we just always used the `click` event to trigger the checkboxes. From that issue it sounds like that might be intentional on the part of the react testing library which is wierd considering the browser supports the change event on checkbox inputs. It might be a limitation of jsdom.
> 
> EDIT: I wonder if a change event sending the spacebar or firing a keypress for spacebar would trigger the checkbox?

Hmm it looks like firing keypress or change event for spacebar didn't trigger the checkbox.
I also tried it using `userEvent.keyboard` and didn't work. I read that it might be JSDOM limitation as well

```
await checkbox.focus();
userEvent.keyboard('{Space}');
```

entroform:
Got `userEvent.keyboard` to work, not sure why it wasn't working before. Looks like it's not a env thing suggested in other threads.

274b5afca599025fc3934407d222bdc7ccfd523e

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
```diff
@@ -5,116 +5,97 @@
...
+    isChecked: false,
+    onChange: () => {},
+    disabled: undefined,
+    children: 'Enables whales'
+  };
 
-  it('renders a <label> containing a toggle <input> and toggle control', function() {
-    const shallowRender = getShallow();
+  it('renders a <label> with an <input>', function() {
+    render(<NxToggle {...simpleProps}></NxToggle>);
```
rpokorny, line 25: It seems to me like we'd still benefit from a wrapper similar to `getShallowComponent` that encapsulates the `render()` call, the name of the component being rendered, and the minimal props.

rpokorny:
`yarn install` gives

```
warning " > @testing-library/user-event@14.1.1" has unmet peer dependency "@testing-library/dom@>=7.21.4".
```

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
```diff
@@ -5,116 +5,97 @@
...
-      <NxToggle { ...simpleProps }>
-        <div className="bar"></div>
-      </NxToggle>
-    );
+  it('adds classes specified with the className prop', function() {
+    const { container } = render(<NxToggle {...simpleProps} className="foo"></NxToggle>);
 
-    expect(render).toContainMatchingElement('label .nx-toggle__content .bar');
-  });
+    const label = container.querySelector('label');
```
rpokorny, line 49: When we're trying to get _the_ top level element that the component renders, is it inappropriate to do something like `container.children[0]`?  I feel like these tests as-written don't do a good job of asserting that the label is the top-level element, which when it comes to testing that custom classes get added, is probably important.

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
```diff
@@ -5,116 +5,97 @@
...
-    component.find('input').simulate('change');
-    expect(onChange).toHaveBeenCalled();
-  });
 
-  it('sets the input as readonly if there is no onChange handler', function() {
-    expect(getShallow().find('input')).toHaveProp('readOnly', false);
-    expect(getShallow({ onChange: undefined }).find('input')).toHaveProp('readOnly', true);
-    expect(getShallow({ onChange: null }).find('input')).toHaveProp('readOnly', true);
-  });
+    await user.click(checkbox);
```
rpokorny, line 66: It's really when anywhere in the label is clicked.  This is especially important since the `<input type="checkbox">` is invisible.

lib/src/setupTests.ts
rpokorny, line 15: Is there any value in adding this comment?

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
entroform: > When we're trying to get _the_ top level element that the component renders, is it inappropriate to do something like `container.children[0]`? I feel like these tests as-written don't do a good job of asserting that the label is the top-level element, which when it comes to testing that custom classes get added, is probably important.

That's a good point: I updated it here: 70a802873cd0a388c381c0dab40dd72929eba491

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
entroform: > It's really when anywhere in the label is clicked. This is especially important since the `<input type="checkbox">` is invisible.

Updated the click test here: 5617387e0a013b6eb0a477e5b48c09008e7305e3

entroform:
> `yarn install` gives
> 
> ```
> warning " > @testing-library/user-event@14.1.1" has unmet peer dependency "@testing-library/dom@>=7.21.4".
> ```

Fixed: 2defa79c7cb1581d9f72b7a3b0961f7311e76edc

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
entroform: > It seems to me like we'd still benefit from a wrapper similar to `getShallowComponent` that encapsulates the `render()` call, the name of the component being rendered, and the minimal props.

Added render wrapper here: 95a4049c1f778e0213ca03d4d8ce4985a6995b00

README.md
```diff
@@ -192,6 +192,14 @@ tests, in which case you may have to hit 'a' in your console window to run the t
 debug as normal.  Be aware of the test filtering options in the console which can cut down on extraneous breakpoint
 hits.
 
+## Writing RTL Unit Test
+
+We are currently transitioning from Enzyme to the [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) so we have both set up in place.
+We also use [jest-dom](https://testing-library.com/docs/ecosystem-jest-dom/) to provide custom DOM element matchers, but it is not imported globally because it will overwrite some of the `jest-enzyme` matchers.
```
rpokorny, line 198: These lines are both long and can/should be broken

lib/src/__testutils__/rtlUtils.tsx
```diff
@@ -0,0 +1,16 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { ComponentType } from 'react';
+import { render, RenderResult } from '@testing-library/react';
+
+import '@testing-library/jest-dom';
+
+export function rtlRender<P>(Component: ComponentType<P>, minimalProps: P) {
+  return function getShallowComponent(additionalProps?: Partial<P>): RenderResult {
```
rpokorny, line 13: Name of this function should be changed

rpokorny (approved):

lib/src/__testutils__/rtlUtils.tsx
entroform: > Name of this function should be changed

Ops! Fixed here: e8d1b39cb63f1262257b2399db9972fec7877230

PR #709, opened 2022-05-10
Spike: Approach for Implementing Dark Mode - RSC-929

rpokorny:
https://issues.sonatype.org/browse/RSC-929

This PR is not to be merged as-is.  At a minimum the placeholder color variable values that I changed here should not go to main.

This seemed almost too easy, let me know if you think there's something I failed to consider.

### DIFF
--- a/lib/src/base-styles/_nx-colors.scss
+++ b/lib/src/base-styles/_nx-colors.scss
@@ -4,6 +4,8 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
+@use '../scss-shared/nx-dark-mode-helpers';
+
 :root {
   --nx-color-border: var(--nx-swatch-indigo-90);
   --nx-color-border-secondary: var(--nx-swatch-indigo-95);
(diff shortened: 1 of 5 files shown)


lib/src/base-styles/_nx-colors.scss
```diff
@@ -46,4 +48,44 @@
 
   // Deprecated variables
   --nx-color-subsection-border: var(--nx-color-border);
+  --nx-color-text-dark: var(--nx-color-text-stark);
+
+  @include nx-dark-mode-helpers.dark-mode {
```
rpokorny, line 53: All of these reassignments are temporary values for the sake of demonstration

glenhunter:
Overall it looks very good to me. I think the biggest question will be organizational: how do we want to list the dark mode colors in the SCSS? A separate SCSS file, a separate section in the existing file(s), inter-leaved with the other colours, or something else?

rpokorny:
> Overall it looks very good to me. I think the biggest question will be organizational: how do we want to list the dark mode colors in the SCSS? A separate SCSS file, a separate section in the existing file(s), inter-leaved with the other colours, or something else?

I was thinking to follow the pattern I established here in `_nx-colors.scss` and `_nx-tile.scss`.  The overrides to the general colors take care of a good chunk of the app.  Anything additional I'd say should be defined in an `@include dark-mode` block adjacent (in the same parent block anyway) to whatever light-mode color it's overriding.

glenhunter:
> > Overall it looks very good to me. I think the biggest question will be organizational: how do we want to list the dark mode colors in the SCSS? A separate SCSS file, a separate section in the existing file(s), inter-leaved with the other colours, or something else?
> 
> I was thinking to follow the pattern I established here in `_nx-colors.scss` and `_nx-tile.scss`. The overrides to the general colors take care of a good chunk of the app. Anything additional I'd say should be defined in an `@include dark-mode` block adjacent (in the same parent block anyway) to whatever light-mode color it's overriding.

That does seem like the method that matches best how we do our other styles.

rpokorny:
This branch has been manually merged to a new feature branch: FEATURE_RSC-928_dark_mode.  Only the changes in this branch that are intended to be part of the real implementation were merged.  The spike demonstration changes were not.

PR #710, opened 2022-05-13
SPIKE: New Approach to Required Form Fields - RSC-1064

rpokorny:
https://issues.sonatype.org/browse/RSC-1064

This spike demonstrates how the design changes around required fields in forms can be implemented.  These are breaking changes that add a new required prop to `NxForm`. That prop is strictly checked at runtime to ensure its use and thus ensure that downstream code code does in fact upgrade all of their forms to the new logic.  Or at least, all of their forms that are build on `NxForm`.

The requirements are as follows:
1. Submit button is no longer disabled when the form is invalid overall
2. If a submit is attempted while the form is invalid, an error message will appear in the form footer telling the user what is wrong (i.e. the same message currently used as the submit button tooltip) and all individual field validation errors will become visible.
3. The error in the footer will disappear as soon as all field are corrected

To accomplish the above, this PR does a few major things:
1. Adds an `isPristine` prop to `NxForm` that behaves similarly to the `isPristine` prop on individual fields: it suppresses validation messages when true.  Upon clicking submit, this prop should be set to false which not only allows the form's overall validation message to appear but also, via a React context, overrides the `isPristine` state on all children, causing all of them to show any applicable validation messages.
2. Adds `isPristine` and `validationErrors` props to `NxFieldset`.  This is intended for cases where a fieldset contains a group of checkboxes, at least one of which must be selected.  This is one of the types of messages that we want to show when the submission is attempted
3. Adds `NxStatefulForm` to encapsulate the `isPristine` logic for `NxForm`.  Note that this component can only be used "single submit" forms.  Forms that can be saved and then remain present for further modification should not use this component as the `isPristine` flag does not get reset on submission.

Additionally, this PR changes the required field styling as prescribed by design.  Optional fields are no longer called out with the italic "Optional" text, and instead required fields are called out with an asterisk ~~(black at this moment, to be updated to red shortly to match the latest mockups)~~

~~One thing this PR does not do, and about which I don't think there's a whole lot we can do, is add the label that design has prescribed for the tops of forms which indicates that the red asterisks shows required fields.  `NxForm` doesn't have any control or special awareness around any header it might contain, and so cannot automatically add such a message.  One thing that could be done would be to add a React component that just hard codes that message~~

This PR also adds a component to encapsulate the standard message that should be included at the tops of forms to define the asterisk.  Note that `NxForm` cannot include this message automatically as it has no knowledge of its header structure.  So instead we just provide a component which contains the appropriate text and require forms to include that component.  In this PR, that is demonstrated on the `nx-tile` form example, though note that that example is not fully fleshed out in other ways (such as actual validation logic).  Making all examples more consistent would be desired when moving this spike to production readiness.

### DIFF
--- a/gallery/src/components/NxForm/NxFormCustomizedExample.tsx
+++ b/gallery/src/components/NxForm/NxFormCustomizedExample.tsx
@@ -65,16 +65,16 @@ export default function NxFormCustomizedExample() {
 
   return (
     <NxStatefulForm onSubmit={onSubmit}
-            onCancel={resetForm}
-            submitMaskState={submitMaskState}
-            className="gallery-custom-form"
-            submitBtnClasses="gallery-custom-form__submit"
-            submitError={submitError}
-            submitErrorTitleMessage="There was an error sending the message."
-            submitBtnText="Send it"
-            submitMaskMessage="Sending…"
-            submitMaskSuccessMessage="Sent!"
-            additionalFooterBtns={additionalFooterBtns}>
+                    onCancel={resetForm}
+                    submitMaskState={submitMaskState}
+                    className="gallery-custom-form"
+                    submitBtnClasses="gallery-custom-form__submit"
+                    submitError={submitError}
+                    submitErrorTitleMessage="There was an error sending the message."
+                    submitBtnText="Send it"
+                    submitMaskMessage="Sending…"
+                    submitMaskSuccessMessage="Sent!"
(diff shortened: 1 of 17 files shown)


entroform:
This looks good! I think we should proceed with this with the examples fleshed out as you mentioned.

entroform (approved):
🚀

PR #713, opened 2022-05-20
Fix NxSearchDropdown items trigger a submit on an enclosing form - RSC-1078

entroform:
https://issues.sonatype.org/browse/RSC-1078


rpokorny (approved):
Add a unit test

PR #719, opened 2022-06-07
Add keynav to NxTabs - RSC-992

entroform:
Ticket:
https://issues.sonatype.org/browse/RSC-992

WAI-ARIA links:
https://www.w3.org/TR/wai-aria-practices-1.2/#tabpanel
https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/

Testing in Firefox requires changing this setting:
https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html


rpokorny:
It looks like you've implemented both patterns shown in the ARIA examples: tab panels showing as you navigate between their respective tabs, and tab panels showing only once you press Enter on their respective tabs.  I'm not sure we needed both of these patterns - the ARIA spec just describes both in otder to cover different variations seen in the wild.  Check with design to see which pattern(s) we actually want to support

gallery/visualtests/nxTabs.js
```diff
@@ -66,5 +89,239 @@ describe('NxTabs', function() {
...
+
+      await page.keyboard.press('ArrowRight');
+
+      expect(await isFocused(fourthTab)).toBe(true);
+      expect(await hasClass(fourthTab, 'active')).toBe(false);
+
+      await page.keyboard.press('ArrowRight');
+
+      expect(await isFocused(firstTab)).toBe(true);
+      expect(await hasClass(firstTab, 'active')).toBe(false);
```
rpokorny, line 171: I'd say that this test and the next one should check which tab panel is visible/active during each step as well

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -40,6 +47,49 @@ const NxTab = function NxTabElement(props: NxTabProps) {
...
+        }
+        case 'Home': {
+          return tabElements?.[0];
+        }
+        case 'End': {
+          return tabElements?.[tabElements.length - 1];
+        }
+        default: {
+          return;
+        }
```
rpokorny, line 75:
```suggestion
        case 'ArrowLeft':
          return tabElements?.[index - 1] ?? tabElements?.[tabElements.length - 1];
        case 'Right':
        case 'ArrowRight':
          return tabElements?.[index + 1] ?? tabElements?.[0];
        case 'Home':
          return tabElements?.[0];
        case 'End':
          return tabElements?.[tabElements.length - 1];
        default:
          return;
```

lib/src/components/NxTabs/types.ts
```diff
@@ -12,12 +12,14 @@ export interface TabContextType {
...
   index: number;
   onTabSelect: (index: number) => void;
+  activationMode?: 'automatic' | 'manual';
 }
 
 export type NxTabsProps = HTMLAttributes<HTMLDivElement> & {
   activeTab?: number | null ;
   onTabSelect: ((index: number) => void);
   children?: ReactNode | null;
+  activationMode?: 'automatic' | 'manual';
```
rpokorny, line 22: Missing in `nxTabsPropTypes`.  Once added there, I suspect you'll find that you need to add `| null` to this type, and that using default value syntax on the destructuring on https://github.com/sonatype/sonatype-react-shared-components/pull/719/files#diff-c904675f258f76689c48e3ed68a619b63b027ed3b6cb84232531d6e53b37bdf0R24 won't work.

rpokorny:
Unit tests?

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -16,13 +16,17 @@
...
-  margin: 0;
+  margin: 0 0 var(--nx-spacing-6x) 0;
   padding: 0;
 }
 
 .nx-tab-panel {
   @include nx-container-helpers.container-vertical;
-  padding-top: var(--nx-spacing-6x);
+
+  &:focus {
```
rpokorny, line 26: I'd say we need designs for the intended focus styling here

rpokorny:
Current tabpanel focus style does not work properly in modals: 
image

lib/src/components/NxModal/NxModal.scss
```diff
@@ -111,7 +111,7 @@
   }
 
   .nx-tab-list {
-    margin: 1px 1px 0 1px;
+    margin: 1px 1px var(--nx-spacing-6x) 1px;
```
rpokorny, line 114: This this is repeating a non-trivial value now, I'd say it's better to just set `margin-top`, `margin-left`, and `margin-right` and leave `margin-bottom` un-overridden here instead.

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -40,6 +47,44 @@ const NxTab = function NxTabElement(props: NxTabProps) {
     }
   }
 
+  function handleKeyDown(event: React.KeyboardEvent<HTMLLIElement>) {
+    if (event.isDefaultPrevented()) {
+      return;
+    }
+
+    const tabElements = event.currentTarget.parentElement?.children;
```
rpokorny, line 55: The code below is written as if there's a possibility that this is `null`ish, but really that can't happen right? I guess if the `<li>` was somehow the root of the document it could, but that shouldn't ever be the case.  Might be worthwhile to use the non-nullish assertion operator (`!`) so that the code below can be simplified.

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -40,6 +47,44 @@ const NxTab = function NxTabElement(props: NxTabProps) {
...
+
+    const getNextElementFromEventKey = (key: string) => {
+      switch (key) {
+        case 'Left':
```
rpokorny, line 59: From MDN:
> [1] Internet Explorer, Edge (16 and earlier), and Firefox (36 and earlier) use "Left", "Right", "Up", and "Down" instead of "ArrowLeft", "ArrowRight", "ArrowUp", and "ArrowDown".

We don't support any of those browser versions, so I don't think these aliases are needed here

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -40,6 +47,44 @@ const NxTab = function NxTabElement(props: NxTabProps) {
...
+    const nextElement = getNextElementFromEventKey(event.key);
+
+    if (nextElement) {
+      event.preventDefault();
+      (nextElement as HTMLElement).focus();
+    }
+  }
+
+  function handleFocus() {
+    if (activationMode === 'automatic' && activeTab !== index) {
```
rpokorny, line 83: It would be easier to read and slightly more efficient if an `isActiveTab` local variable were extracted out for use here and in the `tabIndex` computation below

lib/src/components/NxTabs/NxTab.tsx
entroform: > The code below is written as if there's a possibility that this is `null`ish, but really that can't happen right? I guess if the `<li>` was somehow the root of the document it could, but that shouldn't ever be the case. Might be worthwhile to use the non-nullish assertion operator (`!`) so that the code below can be simplified.

Thanks, cleaned up here: 7dbc588c766bf52503c345f23530850a30eec371

lib/src/components/NxModal/NxModal.scss
entroform: > This this is repeating a non-trivial value now, I'd say it's better to just set `margin-top`, `margin-left`, and `margin-right` and leave `margin-bottom` un-overridden here instead.

Thanks, cleaned up here: 39053712add71530ea8243258991ee5e469ea10f

entroform:
> Current tabpanel focus style does not work properly in modals: image

Weird this looked fine when I last test it. Fixed here: 02b5c12999fd550c4efcb76a79c7d5fa6cc668e0

gallery/visualtests/nxTabs.js
entroform: > I'd say that this test and the next few should check which tab panel is visible/active during each step as well

Updated the tests here: 08d960bd4908a06aea14a276fd7495f6069a2433

lib/src/components/NxTabs/NxTabs.scss
entroform: > I'd say we need designs for the intended focus styling here

Yup! I believe Jason had updated the design in Kaleido

entroform:
> Unit tests?

Since we no longer have that extra prop and all the keyboard and focus tests are done in the visual tests, do we still need additional unit tests?

rpokorny:
A bug:
1. select/activate a tab other than the first tab in a given list (with mouse or keyboard, doesn't matter)
2. press Left to move focus to the previous tab
3. Press Tab

Expected Result: focus goes to the tab panel
Actual Result: focus goes to the activated tab button

entroform:
> A bug:
> 
> 1. select/activate a tab other than the first tab in a given list (with mouse or keyboard, doesn't matter)
> 2. press Left to move focus to the previous tab
> 3. Press Tab
> 
> Expected Result: focus goes to the tab panel Actual Result: focus goes to the activated tab button

Thanks, nice catch!
I fixed it here: 8c028e49477ddd3e2ca4151dacbc1b9e691a0153

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -16,13 +16,18 @@
...
+    box-shadow: var(--nx-box-shadow-focus);
+    outline: solid 1px var(--nx-color-interactive-border-focus);
+    outline-offset: -1px;
```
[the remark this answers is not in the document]
entroform: > This causes the tab focus style to cover up the first pixel of tab panel content in each direction. Do we really want that?

This was a compromise so the outline is visible inside the modal. I confirmed with design and they are okay with this compromise

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -40,16 +46,52 @@ const NxTab = function NxTabElement(props: NxTabProps) {
...
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const tabElements = event.currentTarget.parentElement!.children;
+
+    const getNextElementFromEventKey = (key: string) => {
+      switch (key) {
+        case 'Tab':
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          return event.currentTarget.parentElement!.nextSibling;
```
rpokorny, line 61: This makes assumptions about the DOM structure that make me slightly nervous, since we rely on the calling code to arrange the `NxTabs` family of components correctly.  Conveniently however, we have access to the id of the tab panel via the `aria-controls` attribute of the current tab.  So I recommend using that to do a `document.getElementById` lookup.

PR #720, opened 2022-06-07
Fix visual tests for m1 - RSC-1086

cindyko226:
https://issues.sonatype.org/browse/RSC-1086 
https://github.com/puppeteer/puppeteer/blob/main/docs/api.md

Consult with team SRE and they recommend this solution.
The problem was that we were installing puppeteer chromium outside of the docker container.
We test this on M1 and Intel macbook and the CI. All passed.


rpokorny:
> The problem was that we were installing puppeteer chromium outside of the docker container.

The local build instructions are set up such that that shouldn't have been the case.  The recommended line for running the visual tests in docker locally is this:
```
docker run --rm -it -w /home/jenkins/gallery -v $PWD:/home/jenkins rsc-visualtesting yarn test
```

I.e., run `yarn test` in the `gallery` directory within the docker container. Crucially, the gallery's `yarn test` first runs `check-install` script which runs `yarn install --check-files --frozen-lockfile`.  This was already set up this way because _no_ Macs were working with it otherwise: Running `yarn install` on the host on any mac will download the mac version of chromium, which won't run within the docker container because that is a Linux environment, regardless of whether the CPU architectures match.

So I do not understand how the changes in this PR help, or what wasn't working before if running `yarn install` within the container is all that's necessary.

.dockerignore
```diff
@@ -0,0 +1 @@
+**/node_modules
```
rpokorny, line 1: Is this still needed? What effect does it have?

gallery/visualtests/pageLayouts.js
```diff
@@ -113,7 +113,7 @@ describe('Page Layout', function() {
 
       it('passes a11y checks', async function() {
         // This rule appears to not work correctly on elements that are partially scrolled off the page
-        await a11yTest(builder => builder.disableRules('color-contrast'));
+        await a11yTest(builder => builder.disableRules('color-contrast'), true);
```
rpokorny, line 116: This builder customization is no longer necessary since `a11yTest` does the same thing by default.
```suggestion
        await a11yTest(null, true);
```
Also remove the comment above, and make equivalent changes in the other a11y test below.

.dockerignore
cindyko226: No need, it is removed. 71172291bf8b83e22a640dc120baa19947de6cee

gallery/visualtests/pageLayouts.js
cindyko226: It is updated here ff4855eca50726eb77872778ee9edd74cc7836be.

gallery/visualtests/pageLayouts.js
```diff
@@ -112,8 +112,7 @@ describe('Page Layout', function() {
       });
 
       it('passes a11y checks', async function() {
-        // This rule appears to not work correctly on elements that are partially scrolled off the page
-        await a11yTest(builder => builder.disableRules('color-contrast'));
+        await a11yTest(null, true);
```
rpokorny, line 115: As we discussed on slack, this syntax has actually been wrong all along and these tests essentially aren't doing anything.  I think that's out of scope for this ticket, since I think it affects other places not already being edited here, and may also result in new a11y errors to fix, but did you put in a separate ticket to fix it?

rpokorny (approved):

gallery/visualtests/pageLayouts.js
cindyko226: Yes, I have created a separate ticket for it. https://issues.sonatype.org/browse/RSC-1134

PR #725, opened 2022-06-15
RSC-998 donut chart docs

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-998

Update the description tile inside of the gallery to provide information on the default behaviours of CSS height and display properties, and the options to override the defaults.

### DIFF
--- a/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
+++ b/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+ .gallery-binary-donut-chart-custom {
+    display: block;
+    height: 45px;
+  }
+  
--- a/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
+++ b/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
@@ -8,5 +8,8 @@
  .gallery-binary-donut-chart-custom {
     display: block;
     height: 45px;
+    .nx-binary-donut-chart__arc {
+      stroke: var(--nx-swatch-purple-70);
+    }
   }
   
--- a/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
+++ b/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
@@ -12,4 +12,3 @@
       stroke: var(--nx-swatch-purple-70);
     }
   }
-  
--- a/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.tsx
+++ b/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.tsx
@@ -10,12 +10,24 @@ import { NxBinaryDonutChart } from '@sonatype/react-shared-components';
 export default function NxBinaryDonutChartCustomExample() {
   return (
     <>
-      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom" percent={0} aria-label="0 out of 100 components identified" />
-      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom" percent={15} aria-label="15 out of 100 components identified" />
-      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom" percent={25} aria-label="25 out of 100 components identified" />
-      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom" percent={50} aria-label="50 out of 100 components identified" />
-      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom" percent={90} aria-label="90 out of 100 components identified" />
-      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom" percent={100} aria-label="100 out of 100 components identified" />
+      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom"
+                          percent={0}
+                          aria-label="0 out of 100 components identified" />
+      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom"
+                          percent={15}
+                          aria-label="15 out of 100 components identified" />
+      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom"
+                          percent={25}
+                          aria-label="25 out of 100 components identified" />
+      <NxBinaryDonutChart className="gallery-binary-donut-chart-custom"
+                          percent={50}
+                          aria-label="50 out of 100 components identified" />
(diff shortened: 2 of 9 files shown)


EmilyBermudez88:
The first two commits (dacbc99 and 8103cd1) were from the previous ticket RSC-1084) and accidentally added.

rpokorny:
> The first two commits (https://github.com/sonatype/sonatype-react-shared-components/commit/dacbc9967e5ec177033d2cee71152f0d5b921f79 and https://github.com/sonatype/sonatype-react-shared-components/commit/8103cd1510fe7d06e282396dfc95138e44e9149b) were from the previous ticket RSC-1084) and accidentally added.

Don't worry about that. No one looks at the commit list.  Do make sure the PR diff is correct though (it is here).

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -27,6 +27,18 @@ const NxBinaryDonutChartPage = () =>
       <NxP>
         <NxCode>NxBinaryDonutChart</NxCode> represents a binary donut chart.
       </NxP>
+      <NxP>
+        The <NxCode>NxBinaryDonutChart</NxCode> accepts a default height of 30px (without padding or margin), which
+        can be modified by accessing the CSS height property on the <NxCode>.nx-binary-donut-chart</NxCode> class.
```
rpokorny, line 32: We don't want to advocate for them restyling RSC classes.  Rather they should add a classname of their own and apply their styling overrides to that class.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -27,6 +27,18 @@ const NxBinaryDonutChartPage = () =>
...
+      </NxP>
+      <NxP>
+        By default the CSS property of <NxCode>NxBinaryDonutChart</NxCode> is set to a value of
```
rpokorny, line 35: Is this supposed to say "display property"?

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -27,6 +27,18 @@ const NxBinaryDonutChartPage = () =>
...
+        inline. <NxCode>Display:inline</NxCode> will not generate any line breaks to maintain the flow of the
```
rpokorny, line 36: Code rules take precedence over sentence rules```suggestion
        inline. <NxCode>display:inline</NxCode> will not generate any line breaks to maintain the flow of the
```

**Comment by rpokorny** [comment]
A PR that has no changes to the code shouldn't need changes to the visual test snapshots.  Does your new snapshot pass in jenkins?

**Comment by rpokorny** [comment]
You could consider this to be overkill, but I'm wondering if a new example demonstrating block layout, custom sizing, and even custom colors would be a good idea.

File: gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx (hunk 2 above)
Comment (EmilyBermudez88) [reply gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx:35]: Yes  it is, thanks for the catch

File: gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx (hunk 3 above)
Comment (EmilyBermudez88) [reply gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx:36]: ok, I'll make the change

**Comment by EmilyBermudez88** [comment]
> A PR that has no changes to the code shouldn't need changes to the visual test snapshots. Does your new snapshot pass in jenkins?

I thought so too, but it failed two CI builds citing a diff in the "NxBinary Donut Chart Donut with large hole looks right" . Once I updated the snapshot the CI build passed in Jenkins

**Comment by EmilyBermudez88** [comment]
> You could consider this to be overkill, but I'm wondering if a new example demonstrating block layout, custom sizing, and even custom colors would be a good idea.

I'll leave the final decision up to you, but I'm happy to a customizable example to the page as well.

**Comment by rpokorny** [comment]
> I'll leave the final decision up to you, but I'm happy to a customizable example to the page as well.

Let's do it

**Comment by EmilyBermudez88** [comment]
> > The first two commits ([dacbc99](https://github.com/sonatype/sonatype-react-shared-components/commit/dacbc9967e5ec177033d2cee71152f0d5b921f79) and [8103cd1](https://github.com/sonatype/sonatype-react-shared-components/commit/8103cd1510fe7d06e282396dfc95138e44e9149b)) were from the previous ticket RSC-1084) and accidentally added.
> 
> Don't worry about that. No one looks at the commit list. Do make sure the PR diff is correct though (it is here).

Ok, thanks

**Comment by rpokorny** [comment]
> Would you like a new branch started for that, or am I ok to keep working within the RSC-998 ticket?

To me it's part of the ideal end state for this ticket, and isn't separate

**Comment by EmilyBermudez88** [comment]
> > Would you like a new branch started for that, or am I ok to keep working within the RSC-998 ticket?
> 
> To me it's part of the ideal end state for this ticket, and isn't separate

Perfect, I'll get started.

**Comment by EmilyBermudez88** [comment]
> > Would you like a new branch started for that, or am I ok to keep working within the RSC-998 ticket?
> 
> To me it's part of the ideal end state for this ticket, and isn't separate

The extra GalleryExampleTile has been added, but one question. As of right now, there is a separate SCSS file for the custom example, so the tsx file just has an extra class name attached. If that's the case, only the classname appears as a modification in the Live Example element. Would you rather it be inline styled for readers to see? Or should there be a second live example included? Or can readers discern the meaning for themselves?

File: gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss [hunk 4]
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+ .gallery-binary-donut-chart-custom {
+    display: block;
+    height: 45px;
+    .nx-binary-donut-chart__arc {
+      stroke: var(--nx-swatch-purple-70);
+    }
+  }
+
```
rpokorny, line 15: Trailing whitespace

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.tsx
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxBinaryDonutChart } from '@sonatype/react-shared-components';
+
+export default function NxBinaryDonutChartCustomExample() {
```
rpokorny, line 10: Let's add a visual test for this example

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
EmilyBermudez88: thanks for the catch

rpokorny (approved):
You need to commit the snapshot for your visual test but otherwise looks good.

EmilyBermudez88:
> You need to commit the snapshot for your visual test but otherwise looks good.

updated. Thanks

PR #727, opened 2022-06-21
Rsc 990 meter role donut chart

cindyko226:
https://issues.sonatype.org/browse/RSC-990 

Implement the role='meter' and its corresponding aria attributes.
Accessible Rich Internet Applications (WAI-ARIA) 1.2 https://www.w3.org/TR/wai-aria-1.2/#meter

### DIFF
--- a/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
+++ b/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
@@ -5,7 +5,7 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React from 'react';
-import { NxTable, NxCode, NxP, NxTextLink } from '@sonatype/react-shared-components';
+import { NxTable, NxCode, NxP, NxTextLink, NxH3 } from '@sonatype/react-shared-components';
 
 import { GalleryDescriptionTile, GalleryExampleTile } from '../../gallery-components/GalleryTiles';
 
@@ -59,19 +59,6 @@ const NxBinaryDonutChartPage = () =>
             </NxTable.Cell>
           </NxTable.Row>
           <NxTable.Row>
-            <NxTable.Cell>aria-label</NxTable.Cell>
-            <NxTable.Cell>string</NxTable.Cell>
-            <NxTable.Cell>No</NxTable.Cell>
-            <NxTable.Cell>
-              If the chart is not accompanied by visible text content that contains the same information that the chart
-              conveys, then the chart should either be wrapped in a tooltip or have
-              an <NxCode>aria-label</NxCode> attribute giving it an accessible name which adequately describes its
-              information for non-visual users. If the chart is accompanied by a text description however, such a
-              label would be redundant and the chart is considered a presentational element. In that scenario, it should
-              be given <NxCode>role="presentation"</NxCode>.
-            </NxTable.Cell>
-          </NxTable.Row>
-          <NxTable.Row>
             <NxTable.Cell>SVG <NxCode>&lt;svg&gt;</NxCode> Attributes</NxTable.Cell>
             <NxTable.Cell>
               <NxTextLink external href="https://developer.mozilla.org/en/docs/Web/SVG/Element/svg">
@@ -86,6 +73,38 @@ const NxBinaryDonutChartPage = () =>
           </NxTable.Row>
         </NxTable.Body>
       </NxTable>
+      <NxH3>Accessibility</NxH3>
+      <NxP>
+        <NxCode>NxBinaryDonutChart</NxCode> uses the <NxCode>meter</NxCode> role and must have an
+        accessible name. If the meter has a visible label, it is referenced by <NxCode>aria-labelledby</NxCode>
+        {' '} attribute. Otherwise, the element with role meter has a label provided by <NxCode>aria-label</NxCode>
+        {' '} attribute.
+      </NxP>
+      <NxTable>
+        <NxTable.Head>
+          <NxTable.Row>
+            <NxTable.Cell>Attribute</NxTable.Cell>
+            <NxTable.Cell>Details</NxTable.Cell>
+          </NxTable.Row>
+        </NxTable.Head>
+        <NxTable.Body>
+          <NxTable.Row>
+            <NxTable.Cell><NxCode>aria-labelledby</NxCode></NxTable.Cell>
+            <NxTable.Cell>
+              When the <NxCode>aria-labelledby</NxCode> attribute is used an ID is applied to the
+              HTML element that will be providing the label information, the ID is referenced by
+              <NxCode>aria-labelledby</NxCode>.
+            </NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell><NxCode>aria-label</NxCode></NxTable.Cell>
+            <NxTable.Cell>
+              When the <NxCode>aria-label</NxCode> attribute is used the text is added directly to the
+              attribute.
+            </NxTable.Cell>
+          </NxTable.Row>
+        </NxTable.Body>
+      </NxTable>
     </GalleryDescriptionTile>
 
     <GalleryExampleTile title="Minimal Examples"
@@ -93,8 +112,7 @@ const NxBinaryDonutChartPage = () =>
                         codeExamples={nxBinaryDonutChartMinimalExampleCode}
                         liveExample={NxBinaryDonutChartMinimalExample}>
       Minimal examples of <NxCode>NxBinaryDonutChart</NxCode>s with different values.
-      Some of these charts demonstrate the usage of <NxCode>aria-label</NxCode> to describe the
-      chart contents.
+      These charts demonstrate the usage of <NxCode>aria-label</NxCode> to describe the chart contents.
     </GalleryExampleTile>
 
     <GalleryExampleTile title="Example with no Hole"
--- a/lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
+++ b/lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
@@ -36,8 +36,14 @@ const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
       const largeArc = percent > 50 ? 1 : 0;
 
       return (
-        <svg ref={ref} viewBox="-15 -15 30 30" role="meter" aria-valuemin={0} aria-valuemax={100} 
-             aria-valuenow={percent} className={donutClasses} {...svgAttrs}>
+        <svg ref={ref}
+             viewBox="-15 -15 30 30"
+             role="meter"
+             aria-valuemin={0}
+             aria-valuemax={100}
+             aria-valuenow={percent}
+             className={donutClasses}
+             {...svgAttrs}>
           {/*
             * The first `M` and two `a`s below draw the outer border and the later ones draw the inner border.
             * The reason we need two `a`s for each is because svg gets confused trying to draw a full
--- a/lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
+++ b/lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
@@ -36,7 +36,8 @@ const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
       const largeArc = percent > 50 ? 1 : 0;
 
       return (
-        <svg ref={ref} viewBox="-15 -15 30 30" role="img" className={donutClasses} {...svgAttrs}>
+        <svg ref={ref} viewBox="-15 -15 30 30" role="meter" aria-valuemin={0} aria-valuemax={100} 
+             aria-valuenow={percent} className={donutClasses} {...svgAttrs}>
           {/*
             * The first `M` and two `a`s below draw the outer border and the later ones draw the inner border.
             * The reason we need two `a`s for each is because svg gets confused trying to draw a full
--- a/lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
+++ b/lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
@@ -10,6 +10,22 @@ import NxBinaryDonutChart from '../NxBinaryDonutChart';
 describe('NxBinaryDonutChart', function() {
   const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, {percent: 0});
 
+  it('sets the role to "meter"', function() {
+    expect(getShallowComponent()).toHaveProp('role', 'meter');
+  });
+
+  it('sets the aria-valuemin to 0', function() {
+    expect(getShallowComponent()).toHaveProp('aria-valuemin', 0);
+  });
+
+  it('sets the aria-valuemax to 100', function() {
+    expect(getShallowComponent()).toHaveProp('aria-valuemax', 100);
+  });
+
+  it('sets the aria-valuenow to percent value', function() {
+    expect(getShallowComponent({percent: 30})).toHaveProp('aria-valuenow', 30);
+  });
+
   it('renders an svg with the expected properties', function() {
     expect(getShallowComponent({ percent: 90 })).toHaveClassName('.nx-binary-donut-chart');
     expect(getShallowComponent({ percent: 90 })).toHaveProp('viewBox', '-15 -15 30 30');
--- a/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
+++ b/gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartMinimalExample.tsx
@@ -11,10 +11,10 @@ export default function NxBinaryDonutChartMinimalExample() {
   return (
     <>
       <NxBinaryDonutChart percent={0} aria-label="0 out of 100 components identified" />
-      <NxBinaryDonutChart percent={15} role="presentation" />
-      <NxBinaryDonutChart percent={25} role="presentation" />
+      <NxBinaryDonutChart percent={15} aria-label="15 out of 100 components identified" />
+      <NxBinaryDonutChart percent={25} aria-label="25 out of 100 components identified" />
       <NxBinaryDonutChart percent={50} aria-label="50 out of 100 components identified" />
-      <NxBinaryDonutChart percent={90} role="presentation" />
+      <NxBinaryDonutChart percent={90} aria-label="90 out of 100 components identified" />
       <NxBinaryDonutChart percent={100} aria-label="100 out of 100 components identified" />
     </>
   );


cindyko226:
One failing on the CI that requires updating visual testing.
nx-binary-donut-chart-js-nx-binary-donut-chart-nx-binary-donut-chart-donut-with-large-hole-looks-right-1-diff

rpokorny:
> One failing on the CI that requires updating snapshot.

That's fine in this case.  The donut chart's exact rendering appears to be sensitive to the amount of content on the page above it.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -36,7 +36,14 @@ const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
       const largeArc = percent > 50 ? 1 : 0;
 
       return (
-        <svg ref={ref} viewBox="-15 -15 30 30" role="img" className={donutClasses} {...svgAttrs}>
+        <svg ref={ref}
+             viewBox="-15 -15 30 30"
+             role="meter"
+             aria-valuemin={0}
+             aria-valuemax={100}
```
rpokorny, line 43: I wonder if we should introduce a new optional prop to allow calling code to set a different max.  Oftentimes the donut chart is used to represent some count out of some total, such as "87 out or 95 components identified".  It'd be a shame to lose those real numbers for the accessibility metadata.

cindyko226: Agreed. I will add an optional prop for the different maximum value .

cindyko226: Optional prop added, thanks.

rpokorny: > Optional prop added, thanks.

Don't forget to reference the commit where you've fixed something when you respond saying that it is fixed

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -102,22 +97,54 @@ const NxBinaryDonutChartPage = () =>
           </NxTable.Row>
         </NxTable.Body>
       </NxTable>
+      <NxH3>Accessibility</NxH3>
+      <NxP>
+        <NxCode>NxBinaryDonutChart</NxCode> uses the <NxCode>meter</NxCode> role and must have an
+        accessible name. If the meter has a visible label, it is referenced by <NxCode>aria-labelledby</NxCode>
+        {' '} attribute. Otherwise, the element with role meter has a label provided by <NxCode>aria-label</NxCode>
+        {' '} attribute.
```
rpokorny, line 105: The way this is phrased, saying it "has" `aria-label` or `aria-labelledby`, could be misunderstood to mean that we are somehow taking care of that automatically.  To make it clear that the caller is responsible for this, I'd change it to say "must have".

rpokorny:
Going along with the `maxVal`, I wonder if the `percent` prop should also be adjusted: if it were renamed to `value` or something, and we document that it is out of 100 by default but is out of `maxVal` if `maxVal` is specified, it seems like that would provide the most convenient API for cases where `maxVal` is being used.  We would still also need to support `percent` of course to avoid breaking backwards compatibility, but we could deprecate it.

rpokorny:
We encourage users to set `role="presentation"` on donut charts that are redundant with adjacent visible text content.  `role="presentation"` is not compatible with the other ARIA attributes that you have added.  It might be safest if we only set those attributes when the role is meter (or perhaps, only when the caller has not passed in a role of their own).

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
cindyko226: Got it, the commit for adding optional prop is here 1dacc26140b8acee4ca38610eb64886a5b2821be, thanks.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
cindyko226: The description is updated here 7962483aae75b6263fb0276983d8b3a14e4e9a63, thanks.

cindyko226:
> We encourage users to set role="presentation" on donut charts that are redundant with adjacent visible text content. role="presentation" is not compatible with the other ARIA attributes that you have added. It might be safest if we only set those attributes when the role is meter (or perhaps, only when the caller has not passed in a role of their own).

We also introduce the `aria-labelledby` for the accessible name, if there is a visible text content, it will be referenced by `aria-labelledby`, in that case, do we also need the extra `role="presentation"` as an option?

rpokorny:
> We also introduce the aria-labelledby for the accessible name, if there is a visible text content, it will be referenced by aria-labelledby, in that case, do we also need the extra role="presentation" as an option?

Good point, but after doing a little bit more research I think we might've been wrong to add that recommendation.  From what I'm seeing, when a donut chart is next to visible text and it references that visible text with `aria-labelledby`, a screenreader will read that text out twice: once as the label of the chart and once as text printed on the page.  I suspect we should not have recommended `aria-labelledby` for that situation and that `role="presentation"` is still more correct for that case

rpokorny:
> Good point, but after doing a little bit more research I think we might've been wrong to add that recommendation. From what I'm seeing, when a donut chart is next to visible text and it references that visible text with aria-labelledby, a screenreader will read that text out twice: once as the label of the chart and once as text printed on the page. I suspect we should not have recommended aria-labelledby for that situation and that role="presentation" is still more correct for that case

I may be wrong about that actually.  The main screenreader I use is pretty bad and isn't indicative of how others work, and trying another along with some DOM tweaks I'm getting different results.  Let's stick with what you've got now.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
cindyko226: Updated description here 0ed014303392f3c5ad34ad40b2541725f6c07c0e, thanks.

cindyko226:
> Yes, I mean deprecating the percent external prop in favor of a new external prop called value which is out of maxVal instead of always being out of 100. And then internally we adjust the math so that e.g. arcEndAngle divides by maxVal instead of 100. Basically all of the hardcoded 100s within the component (except the ones related to innerRadiusPercent) would be switched to maxVal

Thanks for the clarification. New prop is added here 32a8b732ed24020ed03e3509936c6858c41902c6, thanks.

lib/src/util/customPropTypes.ts
```diff
@@ -49,3 +49,21 @@ export const optionalPercentNumber: PropTypes.Validator<number> =
     }
     return requiredPercentNumberValidator(props, propName, componentName);
   };
+
+export const requiredValueNumber: PropTypes.Validator<number> =
+  function requiredValueNumberValidator(props: Props, propName: string, componentName: string): Error | null {
+    const value = props[propName];
+
+    if (props['maxVal']) {
+      const max = props['maxVal'];
```
rpokorny, line 58:
```suggestion
    if (props.maxVal) {
      const max = props.maxVal;
```

lib/src/util/customPropTypes.ts
```diff
@@ -49,3 +49,21 @@ export const optionalPercentNumber: PropTypes.Validator<number> =
...
+      if (typeof value === 'number' && value >= 0 && value <= max) {
+        return null;
+      }
+      const err = `${componentName}: prop "${propName}" must be a number between 0 and ${max} inclusive;
+      received ${value}`;
```
rpokorny, line 63: If I'm not mistaken this actually encodes the newline and the indentation into the error string. We don't want that do we?

rpokorny:
New prop looks good, but as I said we still need to support `percent` as a deprecated alternative, to keep this from being a breaking change.

lib/src/util/customPropTypes.ts
cindyko226: You are correct, that does encode the newline and the indentation, I have fixed here 811b214b36b355e0386f216455feb945afaa4b4b

cindyko226:
> New prop looks good, but as I said we still need to support percent as a deprecated alternative, to keep this from being a breaking change.

Forgot about this part, updated here 173f21a3b1761ffd10d63a84cc16fe31e943a721, thanks.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -56,10 +56,31 @@ const NxBinaryDonutChartPage = () =>
...
               Percentage which this donut represents. E.g. when 0 the donut is empty, and as it increases towards 100
               the amount of the donut which is filled in increases.
+              <NxWarningAlert>
+                Deprecated: With the introduction of the <NxCode>value</NxCode> prop, you can now define
+                current value, and the percent will be calculated based on the <NxCode>value</NxCode> prop
+                and the <NxCode>maxVal</NxCode> prop.
+              </NxWarningAlert>
+            </NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
```
rpokorny, line 70: Let's have this rows above the `percent` row so we aren't leading with the deprecated one

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -8,12 +8,32 @@ import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
...
+  it('sets the aria-valuemax to maxVal when maxVal is specified', function() {
+    expect(getShallowComponent({maxVal: 200})).toHaveProp('aria-valuemax', 200);
+  });
+
+  it('sets the aria-valuenow to value', function() {
+    expect(getShallowComponent({value: 30})).toHaveProp('aria-valuenow', 30);
+  });
 
   it('renders an svg with the expected properties', function() {
-    expect(getShallowComponent({ percent: 90 })).toHaveClassName('.nx-binary-donut-chart');
```
rpokorny, line 14: Needs some unit tests for `percent`

lib/src/components/NxBinaryDonutChart/types.ts
```diff
@@ -6,14 +6,18 @@
  */
 import { SVGAttributes } from 'react';
 import * as PropTypes from 'prop-types';
-import { requiredPercentNumber, optionalPercentNumber } from '../../util/customPropTypes';
+import { requiredValueNumber, optionalPercentNumber } from '../../util/customPropTypes';
 
 export interface Props extends SVGAttributes<SVGSVGElement> {
-  percent: number;
+  value: number;
```
rpokorny, line 12: To not break backwards compatibility, `value` needs to be optional.  If you feel up to it you could try playing around with union types to create a type where at least one of `percent` or `value` are required.  See `NxModal`'s types for an example

lib/src/util/customPropTypes.ts
```diff
@@ -49,3 +49,21 @@ export const optionalPercentNumber: PropTypes.Validator<number> =
     }
     return requiredPercentNumberValidator(props, propName, componentName);
   };
+
+export const requiredValueNumber: PropTypes.Validator<number> =
```
rpokorny, line 53: I think this is a little too specific for this file: it relies on there being another prop called `maxVal` which is specifically a donut chart thing.  We don't generally worry that much about custom prop types validators, I'd just use `PropTypes.number` for this the new prop.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
cindyko226: This is updated here b089e64cd333e786b70425b84c1f0e8972797b7e.

lib/src/components/NxBinaryDonutChart/types.ts
cindyko226: This is updated here b089e64cd333e786b70425b84c1f0e8972797b7e.

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
cindyko226: I keep the original percent unit tests and add calculated percent unit tests when value is specified. 99e2bdd95b3f5c95162458b610abb575dbd9ab53

lib/src/util/customPropTypes.ts
cindyko226: This is removed ddb1cf189874e4a07b3cdbb84a00febddd6e2e1d

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -53,13 +53,34 @@ const NxBinaryDonutChartPage = () =>
           </NxTable.Row>
         </NxTable.Head>
         <NxTable.Body>
+          <NxTable.Row>
+            <NxTable.Cell>value</NxTable.Cell>
+            <NxTable.Cell>0 &le; number &le; <NxCode>maxVal</NxCode></NxTable.Cell>
+            <NxTable.Cell>Yes</NxTable.Cell>
```
rpokorny, line 59:
```suggestion
            <NxTable.Cell>Yes, unless <NxCode>percent</NxCode> is specified</NxTable.Cell>
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -5,15 +5,46 @@
...
 import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
-import NxBinaryDonutChart from '../NxBinaryDonutChart';
+import NxBinaryDonutChart, {Props} from '../NxBinaryDonutChart';
 
 describe('NxBinaryDonutChart', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, {percent: 0});
+  const minimalProps: Props = {
+    value: 0,
+    percent: 0
+  };
```
rpokorny, line 14: Only one or the other of these props is supposed to be specified at a time.  I recommend having two separate sets of props: `minimalProps` and `minimalDeprecatedProps`, with a `getShallowComponent`-like function constructed from each.

lib/src/util/customPropTypes.ts
```diff
@@ -49,3 +49,4 @@ export const optionalPercentNumber: PropTypes.Validator<number> =
     }
     return requiredPercentNumberValidator(props, propName, componentName);
   };
+
```
rpokorny, line 52: Extraneous

lib/src/components/NxBinaryDonutChart/types.ts
```diff
@@ -6,14 +6,28 @@
...
+  maxVal?: number | null;
+}
+
+interface CurrentProps extends BaseProps {
+  value: number;
+  percent?: number;
 }
 
+interface DeprecatedProps extends BaseProps {
+  value?: number | null;
```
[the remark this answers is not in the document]
cindyko226: In DeprecatedProps, could you please explain `percent?: never;` in more details, why do we have the type `never` for percent if we still want to support the percent?

When setting the `never` to `value?` in DeprecatedProps, there is a compile error `Type 'undefined' is not assignable to type 'number`, does it mean I would need a custom PropType for value in `PropTypes.ValidationMap<Props>`, e.g. `optionalNullableNumber`?

lib/src/util/customPropTypes.ts
cindyko226: Without this newline, there is an error `Newline required at end of file but not found  eol-last`, I think we do need it based on the configuration `"eol-last": "error"`?

rpokorny: > Without this newline, there is an error Newline required at end of file but not found eol-last, I think we do need it based on the configuration "eol-last": "error"?

The diff here is showing that the file ends in two newlines (the one between lines 51 and 52, and the one at the end of line 52).  It should end in one.

cindyko226: > It also seems like this is in fact fixed? This file is no longer in the PR at all which was the goal.

There are two commits for this file, 
1dcfa28c3b4a8238c4dae01fe1b214f9270f621d I only saw one newline at the end of this file, and I removed it.
b179560e362379fc768750edde5b4d288c9d973f I realized the newline is required, then I added it back.

lib/src/components/NxBinaryDonutChart/NxBinaryDonutChart.tsx
```diff
@@ -13,10 +13,14 @@ import classnames from 'classnames';
 
 const NxBinaryDonutChart = forwardRef<SVGSVGElement, Props>(
     function NxBinaryDonutChart(props, ref) {
-      const { innerRadiusPercent, percent, className, ...svgAttrs } = props;
+      const { innerRadiusPercent, value, percent, className, maxVal, ...svgAttrs } = props;
 
       const donutClasses = classnames('nx-binary-donut-chart', className);
 
+      const max = maxVal ?? 100;
+      const calculatedPercent = value ? Math.round(value / max * 100) : percent as number;
```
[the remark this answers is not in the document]
cindyko226: The decimals should not affect the chart, it is removed here f7801304337b3fb2327dcb92c136350dc0e3a42e

lib/src/components/NxBinaryDonutChart/types.ts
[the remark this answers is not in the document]
cindyko226: > I'm not setting percent to never in DeprecatedProps; I'm doing so in CurrentProps. In DeprecatedProps I'm setting value to never.

Sorry I misread the suggested change block. 

> The type logic that attempts to verify that the PropTypes and Props match does not work with this pattern. You'll have to cast it, as is done in NxModal's types.

This is updated here d9fe9ab54fcb95b0a873ac92fbc0dcca0a97aeb7

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
cindyko226: This is updated here a566b81d0357d211dcaf4fc5849f33401a6d0b0f with two sets of props and corresponding shallowComponents.

lib/src/util/customPropTypes.ts
rpokorny: > There are two commits for this file,
https://github.com/sonatype/sonatype-react-shared-components/commit/1dcfa28c3b4a8238c4dae01fe1b214f9270f621d I only saw one newline at the end of this file, and I removed it.
https://github.com/sonatype/sonatype-react-shared-components/commit/b179560e362379fc768750edde5b4d288c9d973f I realized the newline is required, then I added it back.

The important thing is that it's fixed now (it is), but if you want a deeper understanding, take a look at the way GH presents the diffs for those two commits.  They are not opposites.  The first one removed two newlines (resulting in a file which did not end in a newline and which GH calls out with the 🚫) while the second commit added back a single newline.

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -5,20 +5,73 @@
...
+    expect(DeprecatedPropsComponent()).toHaveProp('aria-valuemax', 100);
+  });
+
+  it('sets the aria-valuemax to maxVal when maxVal is specified', function() {
+    expect(propsComponent({maxVal: 200})).toHaveProp('aria-valuemax', 200);
+    expect(DeprecatedPropsComponent({maxVal: 200})).toHaveProp('aria-valuemax', 200);
+  });
+
+  it('sets the aria-valuenow to value', function() {
+    expect(propsComponent({value: 30})).toHaveProp('aria-valuenow', 30);
```
rpokorny, line 42: Add another line that checks `aria-valuenow` with a non-default `maxVal`

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -39,18 +92,53 @@ describe('NxBinaryDonutChart', function() {
...
+    expect(paths.at(0)).toHaveClassName('.nx-binary-donut-chart__background');
+    expect(paths.at(1)).toHaveClassName('.nx-binary-donut-chart__arc');
+    expect(paths.at(1)).toHaveProp('d', 'M 0 -10.5 A 10.5 10.5 0 0 1 ' + (-10.5 * Math.cos(1.5 * Math.PI))
+        + ' ' + (-10.5 * Math.sin(1.5 * Math.PI)));
+    expect(paths.at(1)).toHaveProp('strokeWidth', 7);
+  });
+
+  it('renders a background path and arc path with expected properties when 0 < calculated percent < 100', function() {
+    const circles = propsComponent({value: 50}).find('circle');
+    const paths = propsComponent({value: 50}).find('path');
```
rpokorny, line 141: A lot of these `value` tests should really check the case with a non-default `maxVal` as well

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
cindyko226: When the `value` is used, it is applied to aria-valuenow whether the maxVal is specified or not.
When the `percent` is used, the calculated value based on `percent ` and default maxVal/ non-default maxVal will be assigned to `aria-valuenow`. There are unit test for these under `percent` describe block. 
2b591aae733e6fe31f0544d52ab760c662ce85cf

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
cindyko226: This is updated here 2b591aae733e6fe31f0544d52ab760c662ce85cf
There are two describe blocks added for the use of `value` and `percent` respectively. In each block, there are unit tests for whether the maxVal is specified or not.
Since we do have test for percent between 0 and 100 `'renders a background path and arc path with expected properties when 0 < percent < 100'`, I think the following two tests are repeated, so I deleted them.
- renders an arc circle and its background with the expected properties when percent is 72
- renders an arc circle and its background with the expected properties when percent is 11

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -5,129 +5,291 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
-import NxBinaryDonutChart from '../NxBinaryDonutChart';
+import NxBinaryDonutChart, {Props} from '../NxBinaryDonutChart';
 
 describe('NxBinaryDonutChart', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent(NxBinaryDonutChart, {percent: 0});
+  describe('with minimalProps', function() {
```
rpokorny, line 11:
```suggestion
  describe('with non-deprecated props', function() {
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -5,129 +5,291 @@
...
-    expect(withNegativeInnerRadius.find('.nx-binary-donut-chart__arc')).toHaveProp('strokeWidth', 14);
-    expect(withExcessInnerRadius.find('.nx-binary-donut-chart__arc')).toHaveProp('d',
-        `M 0 -14 A 14 14 0 0 1 ${-14 * Math.cos(1.5 * Math.PI)} 14`);
-    expect(withExcessInnerRadius.find('.nx-binary-donut-chart__arc')).toHaveProp('strokeWidth', 0);
-
-    expect(withNegativeInnerRadiusAnd100Percent.find('.nx-binary-donut-chart__arc')).toHaveProp('r', 7);
-    expect(withNegativeInnerRadiusAnd100Percent.find('.nx-binary-donut-chart__arc')).toHaveProp('strokeWidth', 14);
-    expect(withExcessInnerRadiusAnd100Percent.find('.nx-binary-donut-chart__arc')).toHaveProp('r', 14);
-    expect(withExcessInnerRadiusAnd100Percent.find('.nx-binary-donut-chart__arc')).toHaveProp('strokeWidth', 0);
+  describe('with minimalDeprecatedProps', function() {
```
rpokorny, line 176:
```suggestion
  describe('with deprecated props', function() {
```

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -5,129 +5,291 @@
...
+    it('renders an svg with the expected properties', function() {
+      expect(getShallowComponent({value: 90})).toHaveClassName('.nx-binary-donut-chart');
+      expect(getShallowComponent({value: 90})).toHaveProp('viewBox', '-15 -15 30 30');
+      expect(getShallowComponent({value: 90})).toMatchSelector('svg');
+    });
 
-  it('renders only a background circle path percent is negative', function() {
-    const circles = getShallowComponent({ percent: -50}).find('circle');
-    const paths = getShallowComponent({ percent: -50}).find('path');
+    it('renders only the background path with the expected properties at zero percent', function() {
```
rpokorny, line 43: A bunch of these test names are still talking about percent rather than value

rpokorny (approved):
+1 aside from a few test naming comments

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
cindyko226: The test names here are under describe `'with non-deprecated props'` block, it tests at certain `percent` calculated based on value and default/non-default maxVal. But it might not be clear, I will change it to `'renders only the background path with the expected properties at zero value'`.

cindyko226: This is updated here f82507e6b150c4c93b8fcfba96d44f31ca682ddb

rpokorny:
> There is the same error shown on CI again, is it ok to go ahead updating the snapshot?

Yes

PR #728, opened 2022-06-21
Fix dropdown toggle icon color - RSC-973

rjp99:
https://issues.sonatype.org/browse/RSC-973


entroform:
Another approach that we saw from nx-tree is to only apply the dark color to nx-icons that is not a threat indicator, `.nx-icon:not(.nx-threat-indicator)`, but we are not sure if there are potentially other icons that uses custom color.

gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -306,6 +319,13 @@ const NxDropdownPage = () =>
       This dropdown contains more complex JSX in its label.
     </GalleryExampleTile>
 
+    <GalleryExampleTile title="Custom label with NxThreatIndicator example"
+                        id="nx-dropdown-with-nx-threat-indicator-example"
+                        liveExample={NxDropdownWithNxThreatIndicatorExample}
+                        codeExamples={nxDropdownWithNxThreatIndicatorExampleCode}>
+      An example of a dropdown with <NxCode>NxThreatIndicator</NxCode>.
```
rpokorny, line 326: Might want to call out that `nx-dropdown-icon` is _not_ used here and _is_ used in the example above.

gallery/src/components/NxDropdown/NxDropdownWithNxThreatIndicatorExample.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { NxDropdown, useToggle, NxThreatIndicator } from '@sonatype/react-shared-components';
+
+function NxDropdownWithNxThreatIndicatorExample() {
+  const [isOpen, onToggleCollapse] = useToggle(false),
+      onClick = () => { alert('click'); },
+      labelElement = <><NxThreatIndicator threatLevelCategory="moderate" /><span>Moderate</span></>;
+
+  return (
+    <NxDropdown label={labelElement}
+                className="extra-class"
```
rpokorny, line 18: This seems extraneous to the purpose of the example.

gallery/visualtests/nxDropdown.js
```diff
@@ -90,6 +90,25 @@ describe('NxDropdown', function() {
...
+
+      await button.click();
+    });
+
+    it('looks right', async function() {
+      const [targetElement] = await waitAndGetElements(selector);
+
+      await moveMouseAway();
+
+      await checkScreenshot(targetElement, 251, 252);
```
rpokorny, line 107: It looks like a few px could/should be trimmed off of the bottom of this to minimize the risk of unrelated changes to the documentation page causing diffs.

gallery/src/components/NxDropdown/NxDropdownWithNxThreatIndicatorExample.tsx
rjp99: Which part? The dropdown example?

rpokorny: > Which part? The dropdown example?

The `className`.  It's not doing anything or demonstrating anything, right?

rjp99:
@rpokorny I updated the image size, but wasn't able to run the visual tests or update the visual test with the new image.

entroform:
> > Another approach that we saw from nx-tree is to only apply the dark color to nx-icons that is not a threat indicator, .nx-icon:not(.nx-threat-indicator)
> 
> Where are you saying you saw that code? I'm not finding it. I think when `NxTree` was in-progress it worked like that for a time, but not recently.

Ah sorry, so in nx-tree implementation, it sets the color of nx-icon to grey-05 by default, unless the icon also has the class name `.nx-tree__colored-icon` (`.nx-icon:not(.nx-tree__colored-icon)`).

I was wondering if it's nicer if we do something similar. Instead of creating a new class name for the consuming developers to assign to icons inside nx dropdown, what if by default we set the icon color to grey-05, but make an exception for nx-threat-indicators `.nx-icon:not(.nx-threat-indicator)` (This is assuming nx-threat-indicator is the only icons with custom colors).

rpokorny:
> Ah sorry: So in nx-tree implementation, it sets the color of .nx-icon to grey-05 by default, unless the icon also has the class name .nx-tree__colored-icon (.nx-icon:not(.nx-tree__colored-icon)).
>
> I was wondering if it's nicer if we do something similar. Instead of creating a new class name for the consuming developers to assign to icons inside NxDropdown, what if, by default we set the .nx-icon color to grey-05, but make an exception for nx-threat-indicators .nx-icon:not(.nx-threat-indicator) (This is assuming nx-threat-indicator will be the only icons with custom colors inside NxDropdown).

Ah! So it's doing basically the opposite of what this PR is doing.  Yes, I think we should be consistent about it, and switch this one to be more like that

rjp99:
> > Ah sorry: So in nx-tree implementation, it sets the color of .nx-icon to grey-05 by default, unless the icon also has the class name .nx-tree__colored-icon (.nx-icon:not(.nx-tree__colored-icon)).
> > I was wondering if it's nicer if we do something similar. Instead of creating a new class name for the consuming developers to assign to icons inside NxDropdown, what if, by default we set the .nx-icon color to grey-05, but make an exception for nx-threat-indicators .nx-icon:not(.nx-threat-indicator) (This is assuming nx-threat-indicator will be the only icons with custom colors inside NxDropdown).
> 
> Ah! So it's doing basically the opposite of what this PR is doing. Yes, I think we should be consistent about it, and switch this one to be more like that

We made the change here:
f0f6966a017d1b7b89c1932d3391193950aa2935

rpokorny:
> We made the change here:
https://github.com/sonatype/sonatype-react-shared-components/commit/f0f6966a017d1b7b89c1932d3391193950aa2935

Oh, it is important IMO though that it be a new class just for this purpose, rather than writing the styles specifically for `.nx-threat-indicator`.  We don't want to make the assumption that that will be the only colored icon that might ever get used here.

entroform:
> > We made the change here:
> > [f0f6966](https://github.com/sonatype/sonatype-react-shared-components/commit/f0f6966a017d1b7b89c1932d3391193950aa2935)
> 
> Oh, it is important IMO though that it be a new class just for this purpose, rather than writing the styles specifically for `.nx-threat-indicator`. We don't want to make the assumption that that will be the only colored icon that might ever get used here.

Hmm, what are your thoughts on creating `nx-icon--colored` modifier and setting that inside `NxThreatIndicator` icon and other colored icons in the future? We can document that modifier nx-icon documentation and note it in NxTree and NxDropdown just incase devs want to create custom colored icons, but it is set by default for all RSC colored icons. It could be a better DX for consuming developers, so they don't have to worry about applying a classname to RSC components when they are being composed.

rpokorny:
> Hmm, what are your thoughts on creating nx-icon--colored modifier and setting that inside NxThreatIndicator icon and other colored icons in the future? We can document that modifier nx-icon documentation and note it in NxTree and NxDropdown just incase devs want to create custom colored icons, but it is set by default for all RSC colored icons. It could be a better DX for consuming developers, so they don't have to worry about applying a classname to RSC components when they are being composed.

That sounds like a decent idea but I'd say doing that properly is significantly out of scope for this ticket - a new documentation page, modifying `NxTree`, figuring out what to do about the existing approach in terms of breaking changes, etc.  Let's do it the existing way for now and then as either another ticket or on Improvement Day someone can explore the approach that you're suggesting.

rpokorny (approved):

gallery/src/components/NxDropdown/NxDropdownPage.tsx
entroform: > Might want to call out that `nx-dropdown-icon` is _not_ used here and _is_ used in the example above.

We added it here: f7052b6057df0e5d4a37409ddf30b9e601d5be42

entroform:
@rpokorny Hi Ross, we went back to our original implementation, (using `nx-dropdown-icon` ) because we noticed that it was overriding disabled state icon buttons, the reset button on dropdown with filter, and external link icon color. After this discovery, we thought the original solution might be the simpler way to go. Let me know what you think!

rpokorny:
> @rpokorny Hi Ross, we went back to our original implementation, (using nx-dropdown-icon ) because we noticed that it was overriding disabled state icon buttons, the reset button on dropdown with filter, and external link icon color. After this discovery, we thought the original solution might be the simpler way to go. Let me know what you think!

That's fine I guess. Should it be `nx-dropdown-icon` or something like `nx-dropdown__icon` though?

gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -258,6 +260,16 @@ const NxDropdownPage = () =>
                 any <NxCode>.nx-dropdown-right-button</NxCode>s.
               </NxTable.Cell>
             </NxTable.Row>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>nx-dropdown-icon</NxCode></NxTable.Cell>
+              <NxTable.Cell>
+                Set custom icons to the correct color.
+              </NxTable.Cell>
+              <NxTable.Cell>
+                This class should be applied to custom icons that are not colored{' '}
```
rpokorny, line 269: "are not colored" might not be the clearest phrasing. Maybe "do not have a built-in color"

rpokorny (approved):

gallery/src/components/NxDropdown/NxDropdownPage.tsx
entroform: > "are not colored" might not be the clearest phrasing. Maybe "do not have a built-in color"

Updated: 34e67258d5977d2fd3d77619b29f01a416f2bfaf

PR #729, opened 2022-06-22
Disallow Specific Props Combo on NxButton - RSC 1057

achen5:
https://issues.sonatype.org/browse/RSC-1057

While working on this ticket, I discovered that when `NxTransferList` has `allowReordering` set to true, its "Move Up" and "Move Down" buttons use native tooltips. I believe these buttons should use MUI tooltips tips instead of native ones. That way, the buttons' tooltips are consistent with the [button documentation](https://gallery.sonatype.dev/#/pages/Button). I created a ticket for this [here](https://issues.sonatype.org/browse/RSC-1113).


lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
```diff
@@ -69,14 +69,14 @@ function TransferListItem<T extends string | number = string>(props: TransferLis
             <NxButton type="button"
                       variant="icon-only"
                       title={moveUpButtonTitle}
-                      disabled={isFilteredItem || isTopItem}
+                      className={isFilteredItem || isTopItem ? 'disabled' : ''}
```
achen5, line 72: Updated `TransferListItem`'s `NxButton`s such that it now uses the `disabled` class instead of the `disabled` attribute. As such, tooltips are now enabled for its disabled buttons.

gallery/src/components/NxButton/NxButtonPage.tsx
```diff
@@ -59,7 +59,9 @@ export default function NxButtonPage() {
                 A string to render as a tooltip and accessibility label for the button. This is generally not necessary
                 for buttons that include text content, but icon-only buttons should use this to make the button's
                 meaning clear in all contexts. Omitting this prop when using an icon-only button is deprecated and will
-                become unsupported in a future release.
+                become unsupported in a future release. If <NxCode>NxButton</NxCode> contains both
+                {' '}<NxCode>title</NxCode> and <NxCode>disabled</NxCode> props, <NxCode>NxButton</NxCode> will throw an
```
entroform, line 63: I wonder if it is a good idea to wrap this inside `NxWarningAlert`

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "10.2.6",
+  "version": "10.2.7",
```
entroform, line 3: Does this count as a breaking change? @rpokorny

rpokorny: > Does this count as a breaking change? @rpokorny

I wouldn't say so.  This combination was never intended to work, we're just being more explicit about it now since it was discovered that it caused _especially_ bad behavior

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
rpokorny: > Consequently, native tooltips are now enabled for its disabled buttons.

We don't want native tooltips on anything.

gallery/package.json
rpokorny: > Ah thanks, that makes sense +1

I thought I left an additional comment on this but now I don't see it.  After looking through the PR diff I changed my mind.  Because we used to have an example of the combination that we are now prohibiting, effectively it was formerly supported.  So removing it would be a breaking change after all.

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
achen5: > We don't want native tooltips on anything.

Just to confirm. Should `TransferListItem`'s disabled "Move Up", "Move Down" buttons have MUI tooltips instead of native ones? Or should those disabled buttons have no tooltip at all?

achen5: > The rules for tooltips and disabling should be as follows:
> 
> When not disabled at all:
> 
> * When `title` is provided to `NxButton`, the button gets an MUI tooltip
> * When `title` is not provided, the button gets no tooltip
> 
> When disabled via `prop`:
> 
> * When `title` is provided, throw an error
> * When `title` is not provided, the button gets no tooltip
> * The button's `onClick` handler does not get called
> 
> When disabled via class:
> 
> * When `title` is provided to `NxButton`, the button gets an MUI tooltip
> * When `title` is not provided, the button gets no tooltip
> * The button's `onClick` handler _does_ get called when the button is activated. It is up to the handler to do nothing in this case.  Whether this could/should be improved or adjusted could be debated, but this is how it is now and changing it is out of scope.
> 
> That the transfer list ordering buttons are currently misbehaving (on `main` too, not just this branch) is also out of scope, it is a bug relating to the logic around nested `NxTooltip`s. I'm not certain whether that should be fixed in `NxTransferListHalf` or in `NxButton` or even `NxTooltip` itself. But figuring that out should be a separate effort and ticket.

Through Slack, Ross and I agreed that the disabled buttons should have tooltips. So the current solution using the `disabled` class stays. The disabled buttons will use native tooltips for now, but they will be updated to MUI tooltips in a future update. I'm going to consider this conversation as resolved.

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -320,6 +320,18 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-validation-invalid">
               The color used on validatable form fields when they are invalid.
             </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-alert-info">
```
rpokorny, line 323: These changes shouldn't be showing up in this diff, something is merged wrong or out of date.  To fix, it should work to merge main into the feature branch and then merge the feature branch into this branch.  And then in the future, avoid merging main directly into this branch.

lib/src/components/NxButton/NxButton.tsx
```diff
@@ -27,11 +27,12 @@ const NxButton = forwardRef<HTMLButtonElement, Props>(
             </button>
           ),
           wrapInTooltip = title && !disabled && !alreadyHasTooltip;
-
+      if (disabled && title) {
+        throw new TypeError('NxButton cannot contain both the \'disabled\' and \'title\' props.');
```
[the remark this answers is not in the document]
achen5: Added a unit test here: [41e96ae](https://github.com/sonatype/sonatype-react-shared-components/pull/729/commits/41e96ae97a10b570c738a312f316ec28bf16ff7d)

gallery/package.json
achen5: > I thought I left an additional comment on this but now I don't see it. After looking through the PR diff I changed my mind. Because we used to have an example of the combination that we are now prohibiting, effectively it was formerly supported. So removing it would be a breaking change after all.

Bumped version: [5a68caf](https://github.com/sonatype/sonatype-react-shared-components/pull/729/commits/5a68cafba991e7ec342b701a762f9f14733198f9)

And per Ross's request, I updated this branch such that it merges to `FEATURE_11.0` once its approved.

gallery/src/components/NxButton/NxButtonPage.tsx
achen5: > I wonder if it is a good idea to wrap this inside `NxWarningAlert`

Wrapped text in a warning in this commit: [5d26c0b](https://github.com/sonatype/sonatype-react-shared-components/pull/729/commits/5d26c0bcaeffd6f8334a903ec5fbf5a5d759d326)

rpokorny (approved):

PR #731, opened 2022-06-22
RSC-1063 Allow textarea for NxCodeSnippet to be smaller

EmilyBermudez88:
JIRA
https://issues.sonatype.org/browse/RSC-1063

NxCodeSnippet is currently using the attributes attached to inputProps to define the number of rows. We can keep this as is, or use the separate row prop I added in here. Either way, if the user wants the text area to be smaller than 9 rows, then a separate class with a newly defined min-height needs to be added to override it's base styles. 

I haven't gone any further on this because I wanted feedback prior to moving further with the new prop.


EmilyBermudez88:
> > I haven't gone any further on this because I wanted feedback prior to moving further with the new prop.
> 
> So should this PR be a draft? It looks like it was originally but then you marked it as ready?

Yes, that was my mistake. Sorry

rpokorny:
IMO the `inputProps` should just be a passthrough, and adding `rows` as a top-level prop, as you have done, is the right way to go :+1:

gallery/src/components/NxCodeSnippet/NxCodeSnippet.scss
```diff
@@ -0,0 +1,5 @@
+.nx-code-snippet-custom-sizing {
+    textarea {
+        min-height: 25px;
```
rpokorny, line 3: What I'm picturing is when `NxCodeSnippet` receives a defined value for `rows`, that it adds a class to itself that removes that `min-height` entirely in addition to passing the `rows` prop to the `textarea`.  So there wouldn't be a need for custom styles by downstream code such as this example.

EmilyBermudez88: Thinking out loud, so I would still need an scss file and setting min-height to initial to scrap the 220px value in base styles, and add this class via classnames if rows is defined instead of having the class manually added by the user

EmilyBermudez88:
> IMO the `inputProps` should just be a passthrough, and adding `rows` as a top-level prop, as you have done, is the right way to go 👍

Ok great. Thanks for clarifying the choice to go with

gallery/src/components/NxCodeSnippet/NxCodeSnippet.scss
rpokorny: > Thinking out loud, so I would still need an scss file and setting min-height to initial to scrap the 220px value in base styles, and add this class via classnames if rows is defined instead of having the class manually added by the user

Well not a new scss file.  This would go in the existing `NxCodeSnippet.scss`

EmilyBermudez88: sounds good, I'll start working on the changes

gallery/src/components/NxCodeSnippet/NxCodeSnippetPage.tsx
```diff
@@ -112,13 +123,20 @@ export default function NxCodeSnippetPage() {
         copied to the clipboard.
       </GalleryExampleTile>
 
+      <GalleryExampleTile title="Custom Sizing Example"
+                          id="nx-code-snippet-custom-sizing-example"
+                          liveExample={NxCodeSnippetSizingExample}
+                          codeExamples={nxCodeSnippetSizingCode}>
+        An <NxCode>NxCodeSnippet</NxCode> demonstrating custom sizing with the optional <NxCode>rows</NxCode> prop.
```
rpokorny, line 130: The last example demonstrates the `rows` prop too, meaning that IMO the real value of this example is that it demonstrates using `rows` to create a single-line textarea.  That's what wasn't cleanly possible before this PR.  I think the name and description on this example should be oriented around that aspect.

gallery/visualtests/nxCodeSnippet.js
```diff
@@ -20,11 +20,13 @@ describe('NxCodeSnippet', function() {
   } = setupBrowser('#/pages/Code%20Snippet');
 
   const multipleSnippetSelector = '#nx-code-snippet-simple-example .gallery-example-live',
+      sizingSnippetSelector = '#nx-code-snippet-custom-sizing-example .gallery-example-live',
       complexSnippetExample = '#complex-nx-code-snippet',
       textareaSelector = `${complexSnippetExample} textarea`,
       copyBtnSelector = `${complexSnippetExample} .nx-btn`;
 
   it('looks right including distance to other components', simpleTest(multipleSnippetSelector));
+  it('looks right', simpleTest(sizingSnippetSelector));
```
rpokorny, line 29: Needs a better name.  How about using anested  `describe` block to narrow down which "it" we're talking about:

```
describe('NxCodeSnippet', function() {
...
  describe('with rows=1', function() {
    it('renders with a one-line textarea', simpleTest(sizingSnippetSelector));
  });
});
```

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -81,21 +83,26 @@ export const PrivateNxTextInput = forwardRef<HTMLDivElement, Props>(
         }
       }
 
+      // let classes
+      // rows? console.log('hey') : console.log('nope');
+      // rows?  classes= 'nx-text-input__input custom';
+
```
rpokorny, line 89:
```suggestion
```

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -48,6 +49,7 @@ export const PrivateNxTextInput = forwardRef<HTMLDivElement, Props>(
       const isTextArea = type === 'textarea',
           element = isTextArea ? 'textarea' : 'input',
           typeAttr = isTextArea ? undefined : (type || 'text'),
+          classes = rows && isTextArea ? 'nx-text-input__input textarea-custom-sizing' : 'nx-text-input__input',
```
rpokorny, line 52: I think this should be managed within `NxCodeSnippet`.  That's where the styling for this class is defined, and additionally we aren't trying to add behavior to `NxTextInput`, just `NxCodeSnippet`.

rpokorny, line 52: Use the `classnames` utility for conditional classes.  See a few lines below this for an example.

rpokorny:
Actually, thinking more about this I'm wondering if an almost entirely different approach would be cleaner.  What if textarea inputs only set the `min-height` when the `rows` attribute is unspecified? That is, What if the `textarea.nx-text-input__input` block in `_nx-text-input.scss` was rewritten like this:

```
textarea.nx-text-input__input {
  min-height: 220px;
  padding: 0;
  resize: vertical;

  &[rows] {
    min-height: unset;
    resize: none;
  }
}
```

Then we wouldn't need JS logic around `rows` and adding classnames at all, which means there'd be no reason for `rows` to be elevated to a top-level prop, which means no API change at all.  It'd be just a styling update and the new example.

Note that in my code above I also disabled user resizing when `rows` is set.  I suspect that's something we want to do otherwise the user can drag the resizer on our single-line example and get this:
image

gallery/visualtests/nxCodeSnippet.js
EmilyBermudez88: Sounds good. I grouped both the simpleTest()s together inside the parent describe('NxCodeSnippet'), but should the hover and focus tests conducted on complexSnippetExample also be included inside that group?

EmilyBermudez88:
> Actually, thinking more about this I'm wondering if an almost entirely different approach would be cleaner. What if textarea inputs only set the `min-height` when the `rows` attribute is unspecified? That is, What if the `textarea.nx-text-input__input` block in `_nx-text-input.scss` was rewritten like this:
> 
> ```
> textarea.nx-text-input__input {
>   min-height: 220px;
>   padding: 0;
>   resize: vertical;
> 
>   &[rows] {
>     min-height: unset;
>     resize: none;
>   }
> }
> ```
> 
> Then we wouldn't need JS logic around `rows` and adding classnames at all, which means there'd be no reason for `rows` to be elevated to a top-level prop, which means no API change at all. It'd be just a styling update and the new example.
> 
> Note that in my code above I also disabled user resizing when `rows` is set. I suspect that's something we want to do otherwise the user can drag the resizer on our single-line example and get this: image

I really like this idea. As you say, it's a lot cleaner of an approach and would function in the same way. I'll get started making those changes and re-instating rows as an attribute under inputProps instead of a top-level prop

lib/src/components/NxTextInput/NxTextInput.tsx
EmilyBermudez88: removed, thanks

gallery/src/components/NxCodeSnippet/NxCodeSnippetPage.tsx
EmilyBermudez88: updated in 4e1e760

gallery/visualtests/nxCodeSnippet.js
rpokorny: > Sounds good. I grouped both the simpleTest()s together inside the parent describe('NxCodeSnippet'), but should the hover and focus tests conducted on complexSnippetExample also be included inside that group?

Oh you misunderstood my example a little bit.  The outer `describe` was intended to be the top-level one already present in the file.  So I was saying all of the other tests should be left as they are, and the new test should be wrapped in `describe('with rows=1'`

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "10.2.6",
+  "version": "10.2.8",
```
rpokorny, line 3: Why is it bumped two versions?

rpokorny:
Many of the new screenshots look the same as the old ones. Are the new ones actually needed?

gallery/package.json
EmilyBermudez88: Main has been updated since I began the ticket, so I was updating to the newest version according to main (which is currently 10.2.7)

lib/src/components/NxTextInput/NxTextInput.tsx
```diff
@@ -87,7 +87,7 @@ export const PrivateNxTextInput = forwardRef<HTMLDivElement, Props>(
             * presentational concern
             */}
           {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions*/}
-          <div className="nx-text-input__box" onClick={setFocusToInput}>
+          <div className="nx-text-input__box custom" onClick={setFocusToInput}>
```
rpokorny, line 90: What's this?

EmilyBermudez88:
> Many of the new screenshots look the same as the old ones. Are the new ones actually needed?

I struggled to find the difference too, but found it. The resize option was removed for when the rows attribute is present, so the updated snapshots remove the resizing lines in the bottom right hand corner of the textarea

lib/src/components/NxTextInput/NxTextInput.tsx
EmilyBermudez88: Sorry, that was when I was testing something and meant to be removed. Thanks for catching it

EmilyBermudez88: removed in 87b9ab3

gallery/visualtests/nxCodeSnippet.js
EmilyBermudez88: > > Sounds good. I grouped both the simpleTest()s together inside the parent describe('NxCodeSnippet'), but should the hover and focus tests conducted on complexSnippetExample also be included inside that group?
> 
> Oh you misunderstood my example a little bit. The outer `describe` was intended to be the top-level one already present in the file. So I was saying all of the other tests should be left as they are, and the new test should be wrapped in `describe('with rows=1'`

Thanks for the clarification. Updated in c69f56c and a813df6

rpokorny (approved):

PR #734, opened 2022-06-28
Improve Performance of Tooltips in Large Transfer Lists - RSC-1110

rpokorny:
https://issues.sonatype.org/browse/RSC-1110

The key to the fix here is that tooltips are not fully initialized on first render - they are initially given an empty string as their `title`, which causes MUI's tooltip library to do substantially less work.  Then, asynchronously, they are given their real titles, and this is done in batches.  The batching is important as without it there is a large amount of react overhead and it executes the update of each tooltip separately, and this really adds up when there are many tooltips.  Luckily, react added a function called `unstable_batchedUpdates` that groups work together and cuts down on that overhead.  Regarding batch size, there is a tradeoff to be had between keeping the page from freezing up noticeably and getting all of the tooltips initialized in a timely manner.  A batch size of 100 seemed to work decently well on my system.  I also use `requestIdleCallback` to attempt to get this work scheduled when it is best for the browser, though admittedly the runtime of each batch tends to overrun the amount of idle time that the browser wants to provide by a large percentage (the devtools performance view can show this).  Oh well.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "10.2.8",
+  "version": "10.2.9",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "10.2.8",
+  "version": "10.2.9",
   "description": "Sonatype shared UI components and utilities written in React",
   "main": "index.js",
   "repository": {


lib/src/components/NxTooltip/updateBatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+ * on React overhead.  Note that adjusting BATCH_SIZE involves a tradeoff between minimizing react overhead
+ * by using large batches, and keeping the page reasonably interactive. The current BATCH_SIZE is tuned to the
+ * workloads and needs of NxTooltip (and NxOverflowTooltip) specifically.
+ */
+export default function batch(work: () => void) {
+  queue.push(work);
+
+  if (queue.length >= BATCH_SIZE) {
+
+    //queue is full; dispatch work
```
entroform, line 37:
```suggestion
    // queue is full; dispatch work
```

lib/src/components/NxTooltip/updateBatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    timeoutHandle = window.setTimeout(() => {
+      dispatch();
+      timeoutHandle = null;
+    }, MAX_WORK_TIME);
+  }
+}
+
+function dispatch() {
+  const q = queue;
+  requestIdleCallback(() => { execute(q); });
```
entroform, line 57: Just wondering why we queue is not just passed in directly into execute here.

rpokorny: > Just wondering why we queue is not just passed in directly into execute here.

Because in the time between when `requestIdleCallback` is called and the time the anonymous function executes, `queue` may have been updated to refer to a different list instance.  `q` just keeps a reference to whichever list `queue` was when `dispatch` was called

gallery/visualtests/nxOverflowTooltip.js
```diff
@@ -27,7 +27,8 @@ describe('NxOverflowTooltip', function() {
       maxBeforeOverflowStr = 'WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW',
       descendantOverflowItemSelector = `${descendantOverflowListSelector} .nx-list__item:nth-child(2)`;
 
-  const wait1Sec = () => wait(1000);
+  // NOTE: not really 1 sec anymore
+  const wait1Sec = () => wait(1500);
```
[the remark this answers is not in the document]
rpokorny: > I don't think it's that important, but do you think we should update the name because it's no longer 1 sec.

That reminds me, I meant to lower the batch timeout.  I don't think it needs to be a whole second, since it's just the amount of time to wait for stragglers.  All units of work resulting from an initial page render should come in at essentially the same time.  Done in  0ecb3a9dc.  This means that this function name is now truthful again.

entroform (approved):
I think this looks good to me after testing it out in the gallery.

PR #735, opened 2022-06-28
add RequiredFieldNotice to form components - RSC 1083

EmilyBermudez88:
JIRA 
https://issues.sonatype.org/browse/RSC-1083

"Required fields are marked with an asterisk" has been added to discussed components and confirmed with design. Pardon the messiness of the order of commits, as more implementations were found as the ticket progressed.


JasonDSwearingen (approved):

JasonDSwearingen (approved):

gallery/src/components/NxFieldset/NxFieldsetRequiredExample.tsx
```diff
@@ -6,17 +6,22 @@
...
   const [val, setVal] = useState<string | null>(null);
 
   return (
-    <NxFieldset label="Color" isRequired>
-      <NxRadio value="red" isChecked={val === 'red'} name="color" onChange={setVal}>Red</NxRadio>
-      <NxRadio value="green" isChecked={val === 'green'} name="color" onChange={setVal}>Green</NxRadio>
-      <NxRadio value="blue" isChecked={val === 'blue'} name="color" onChange={setVal}>Blue</NxRadio>
-      <NxRadio value="purple" isChecked={val === 'purple'} name="color" onChange={setVal}>Purple</NxRadio>
-    </NxFieldset>
+    <>
```
rpokorny, line 15: I think this one should have been left unchanged.  Only examples that demonstrate a form as a whole need the message - examples of specific form children can stay scoped to those children without worrying about the concerns of a hypothetical wrapping form.

gallery/src/components/NxForm/NxFormExample.tsx
```diff
@@ -117,6 +119,9 @@ export default function NxFormExample() {
                     submitError={submitError}
                     validationErrors={validationErrors}
                     submitMaskState={submitMaskState}>
+      <NxTile.HeaderSubtitle>
```
rpokorny, line 122: This is somewhat problematic, but it's tough to say what the right way forward would be since this example is of a form "in isolation" without any particular wrapping context.  Using `HeaderSubtitle` to wrap the notice would be correct if the form was in a tile (such that it is the entire tile).  That's not the case here.  You might want to ask Design what this message should look like on a form that's "in abstract" as opposed to being in any particular type of container.

gallery/src/components/NxForm/NxFormPage.tsx
```diff
@@ -230,6 +230,17 @@ const NxFormPage = () =>
               function in order to avoid rendering children before loading completes
             </NxTable.Cell>
           </NxTable.Row>
+          <NxTable.Row>
```
rpokorny, line 233: What's this about?  I wouldn't expect a new prop on `NxForm` in this PR, and this one isn't described in a way that makes it sound like it goes on `NxForm` at all

gallery/src/components/NxForm/NxFormPage.tsx
```diff
@@ -292,7 +303,8 @@ const NxFormPage = () =>
                         liveExample={NxFormExample}>
       This example shows an asynchronously loading NxForm. The example is contrived such that the
       form fails to load the first time, but does load (after a brief wait) upon a second attempt.
-      Additionally, the first attempt to submit the form fails, while the retry succeeds.
+      Additionally, the first attempt to submit the form fails, while the retry succeeds. The example also{' '}
+      demonstrates the use of the isRequired flag to remove the "Optional" indicator.
```
rpokorny, line 307: I also don't understand this change.  `isRequired` is used on some of the form children, but not the form itself.  And it doesn't have to do with an "optional" indicator anymore as on this feature branch we've done away with that in favor of the required indicator (the red asterisk).

gallery/src/components/NxFormGroup/NxFormGroupRequiredExample.tsx
```diff
@@ -6,14 +6,19 @@
...
 
 export default function NxFormGroupRequiredExample() {
   const requiredValidator = (val: string) => val ? null : 'Must be non-empty';
 
   return (
-    <NxFormGroup label="Username" isRequired>
-      <NxStatefulTextInput validator={requiredValidator} />
-    </NxFormGroup>
+    <>
+      <NxTile.HeaderSubtitle>
```
rpokorny, line 16: As with `NxFieldset`, I don't think changes were needed here.

gallery/src/components/NxFormSelect/NxFormSelectExample.tsx
```diff
@@ -15,15 +15,20 @@ const NxFormSelectExample = () => {
...
-      <NxFormSelect onChange={onChange} { ...selectState }>
-        <option value="1">Option 1</option>
-        <option value="2">Option 2</option>
-        <option value="3">Option 3</option>
-        <option value="4">Option 4</option>
-        <option value="5">Option 5</option>
-      </NxFormSelect>
-    </NxFormGroup>
+    <>
+      <NxTile.HeaderSubtitle>
```
rpokorny, line 19: Also not needed

gallery/src/components/NxFormSelect/NxFormSelectOverflowExample.tsx
```diff
@@ -5,18 +5,23 @@
...
-    <NxFormSelect>
-      <option value="1">Looooooooooooooooooong Option 1</option>
-      <option value="2">Option 2</option>
-      <option value="3">Option 3</option>
-      <option value="4">Option 4</option>
-      <option value="5">Option 5</option>
-    </NxFormSelect>
-  </NxFormGroup>
+  <>
+    <NxTile.HeaderSubtitle>
```
rpokorny, line 12: Also not needed

gallery/src/components/NxFormSelect/NxFormSelectWidthsExample.tsx
```diff
@@ -21,6 +21,9 @@ const NxFormSelectExample = () => {
 
   return (
     <div className="form-select-width-variants">
+      <NxTile.HeaderSubtitle>
```
rpokorny, line 24: Also not needed

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
```diff
@@ -39,6 +47,9 @@ export default function NxModalFormErrorExample() {
                 <span>Example NxModal header with form content and error styling</span>
               </h2>
             </header>
+            <NxTile.HeaderSubtitle>
```
rpokorny, line 50: This definitely shouldn't use `NxTile.HeaderSubtitle`.  Work with design to determine how the message should be shown in this case.  Updates to https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/nRWjbOl will probably be necessary.

gallery/src/components/NxModal/NxModalFormExample.tsx
```diff
@@ -70,6 +72,9 @@ export default function NxModalFormExample() {
                 <span>NxModal header with form content</span>
               </h2>
             </header>
+            <NxTile.HeaderSubtitle>
```
rpokorny, line 75: Needs to be adjusted along with the other modal example

gallery/src/components/NxRadio/NxRadioPage.tsx
```diff
@@ -107,6 +107,15 @@ const NxRadioPage = () =>
               multiple overlapping tooltips from appearing.
             </NxTable.Cell>
           </NxTable.Row>
+          <NxTable.Row>
```
rpokorny, line 110: `NxRadio` doesn't have this prop.  It's entire radio groups (e.g. `NxFieldset`s) that may be required, not individual radios

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
```diff
@@ -37,6 +39,9 @@ export default function NxFormLayoutExample() {
 
   return (
     <form className="nx-form" onSubmit={onSubmit} aria-label="Horizontal Layout Example">
+      <NxTile.HeaderSubtitle>
```
rpokorny, line 42: This is another case of a form in no specific type of container, same as https://github.com/sonatype/sonatype-react-shared-components/pull/735#discussion_r914795669

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
```diff
@@ -70,6 +72,9 @@ export default function NxFormLayoutExample() {
 
   return (
     <form className="nx-form" onSubmit={onSubmit} aria-label="Default Form Layout Example">
+      <NxTile.HeaderSubtitle>
```
rpokorny, line 75: Another similar to https://github.com/sonatype/sonatype-react-shared-components/pull/735#discussion_r914795669

lib/src/components/NxForm/NxForm.scss
```diff
@@ -0,0 +1,18 @@
...
+    display: block;
+    margin-bottom: var(--nx-spacing-6x);
+}
+
+.nx-modal {
+    .nx-form__required-field-notice {
+        margin-top: var(--nx-spacing-2x);
+        margin-bottom: 0;
+    }
+}
```
rpokorny, line 18: Incorrect indentation

gallery/src/components/NxFieldset/NxFieldsetRequiredExample.tsx
EmilyBermudez88: thanks for this clarification. I misunderstood and thought the requiredFiledNotice was required for all all examples marked with an asterisk. I will remove

EmilyBermudez88: > I think this one should have been left unchanged. Only examples that demonstrate a form as a whole need the message - examples of specific form children can stay scoped to those children without worrying about the concerns of a hypothetical wrapping form.

thanks for the clarification. I misunderstood and thought the requiredFieldNotice was required for all examples marked with an asterisk. I will remove

gallery/src/components/NxForm/NxFormPage.tsx
EmilyBermudez88: Reviewing the documentation on the page, I misunderstood and thought I needed to add the documentation for isRequired. It has been removed.

EmilyBermudez88: Reviewing the documentation on the page, I misunderstood and thought I needed to add the documentation for isRequired. It has been removed.

EmilyBermudez88: > What's this about? I wouldn't expect a new prop on `NxForm` in this PR, and this one isn't described in a way that makes it sound like it goes on `NxForm` at all

Reviewing the documentation on the page, I misunderstood and thought I needed to add the documentation for isRequired. It has been removed.

gallery/src/components/NxFormSelect/NxFormSelectExample.tsx
EmilyBermudez88: > Also not needed

Removed in dd2b57e

gallery/src/components/NxFormSelect/NxFormSelectOverflowExample.tsx
EmilyBermudez88: > Also not needed

Removed in dd2b57e

gallery/src/components/NxFormSelect/NxFormSelectWidthsExample.tsx
EmilyBermudez88: > Also not needed

Removed in dd2b57e

EmilyBermudez88:
> One of the visual test snapshot updates doesn't depict the Required Field Notice and so seems out of scope.

Are you referring to nx-form-select-js-nx-form-select-short-and-long-variants-of-nx-form-select-looks-shorter-and-longer-1-snap.png?  I updated this one as it has the red * now, but am I wrong in doing so?

lib/src/components/NxForm/NxForm.scss
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+.nx-form__required-field-notice {
+    display: block;
```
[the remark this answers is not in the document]
EmilyBermudez88: > The original intent was that this element be an inline element that gets placed into a container that gives it its block layout and also its text styling (e.g. bold, dark). That intent may or may not hold as you work out how this element should be wrapped when not within `NxTile.HeaderSubtitle`, but for now these styles don't seem right

When discussing with Design, we agreed on a margin-bottom of 24px. In order to do so, my solution at the time was to use display:block to add the margin and collapse the margin. I'm trying to find a selector that would target the container element when it holds .nx-form_required-field-notice. I see :has (https://developer.mozilla.org/en-US/docs/Web/CSS/:has),  but it's very unsupported. Do you have any recommendations of how to target the container element to be styled only when it contains .nx-form_required-field-notice, that doesn't involve adding classes?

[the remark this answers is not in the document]
rpokorny: > When discussing with Design, we agreed on a margin-bottom of 24px

That implies that design is abandoning the approach of having the message appear in an existing "slot" such as the tile subheader.  Is that the case? Tile subheaders only have 16px of bottom spacing for instance.

> I'm trying to find a selector that would target the container element when it holds .nx-form_required-field-notice. I see :has (https://developer.mozilla.org/en-US/docs/Web/CSS/:has), but it's very unsupported. Do you have any recommendations of how to target the container element to be styled only when it contains .nx-form_required-field-notice, that doesn't involve adding classes?

No we'll make bigger changes if that's the direction we want to go.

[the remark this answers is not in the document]
EmilyBermudez88: > > When discussing with Design, we agreed on a margin-bottom of 24px
> 
> That implies that design is abandoning the approach of having the message appear in an existing "slot" such as the tile subheader. Is that the case? Tile subheaders only have 16px of bottom spacing for instance.
> 

At the time of discussion, we had liked the bolded version with a 24px bottom and keeping it the same in all instances.I remember we had a discussion to the reasoning about 24px bottom, although what we talked about exactly escapes me now. 
 I did not bring it up that it would mean having to create a new "slot" instead of an existing one, so that is my fault.  Is creating a new "slot" an option, or would you rather we stay within pre-existing elements and styles? (Also, I am trying to locate subheaders, and only see a sublabel. Where is subheader located?)

rpokorny:
> I updated this one as it has the red * now, but am I wrong in doing so?

Yes this ticket isn't about the asterisks themselves, only about the message _about_ the asterisks.

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
EmilyBermudez88: > This definitely shouldn't use `NxTile.HeaderSubtitle`. Work with design to determine how the message should be shown in this case. Updates to https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/nRWjbOl will probably be necessary.

Code updated here : 10fb738 . I will ensure Design updates sketch as well

gallery/src/components/NxModal/NxModalFormExample.tsx
EmilyBermudez88: > Needs to be adjusted along with the other modal example

Updated here: 10fb738

gallery/src/styles/NxFormLayout/NxFormLayoutExample.tsx
EmilyBermudez88: Updated in [cd9d67e](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/cd9d67e06c9a7d3c9eba14c9f8d1c407da96c88c)

gallery/src/styles/NxFormLayout/NxFormInlineLayoutExample.tsx
```diff
@@ -23,6 +24,9 @@ export default function NxFormLayoutExample() {
 
   return (
     <form className="nx-form" onSubmit={onSubmit} aria-label="Inline Form Example">
+      <NxTile.HeaderSubtitle>
```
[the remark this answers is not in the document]
EmilyBermudez88: > Another similar to [#735 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/735#discussion_r914795669)

Updated in [cd9d67e](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/cd9d67e06c9a7d3c9eba14c9f8d1c407da96c88c)

lib/src/components/NxForm/NxForm.scss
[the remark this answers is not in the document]
EmilyBermudez88: > If that latter approach is what we are now going with, then _all_ of the styling of the Required Field Notice is now up to itself. Whereas the current implementation relies on the wrapper (e.g. the Tile Subtitle) to provide the vertical spacing and font weight, in this different approach all of that would be built into the Required Field Notice component. Going along with that, it would be changed from a `span` to a `p`, making it a block element.
> 

After speaking to Jason, they believe that in the given gallery examples (and all scenarios they can think of), leaving the Required Field text as part of the content of the form makes the most sense. However, Jason did say they would like to leave the option for someone to use it inside a header as well, which is why I still wrapped the element in a <p> tag instead of updating and directly styling the <span> element.

gallery/src/components/NxForm/NxFormPage.tsx
rpokorny: > I misunderstood thinking the optional indicator would still be used in addition to the required indicator, and thought about highlighting the face that the required indicator was used in the example. However, as you mention that the flag it is not on the form directly but instead on its children, it has been removed in https://github.com/sonatype/sonatype-react-shared-components/commit/dd2b57e1b851c3604298eb633190bb798bc371b1

That commit doesn't affect this line, I still see the problematic verbiage.  I also see it on the `NxFormSelectPage` btw.

gallery/src/components/NxRadio/NxRadioDisabledExample.tsx
```diff
@@ -12,25 +12,23 @@ export default function NxRadioDisabledExample() {
   const [city, setCity] = useState<string | null>('fulton');
 
   return (
-    <>
```
rpokorny, line 15: Changes to this file are now extraneous to the purpose of this PR, please undo them to keep things relevant

gallery/src/components/NxRadio/NxRadioPage.tsx
```diff
@@ -157,6 +157,7 @@ const NxRadioPage = () =>
                         liveExample={NxRadioDisabledExample}>
       This example shows radios that are disabled.
       Disabled radios can either be checked or unchecked.
+      The example also demonstrates the use of the isRequired flag to remove the "Optional" indicator.
```
rpokorny, line 160: Another instance of this inaccurate sentence

EmilyBermudez88: Removed here: [6ed96b0](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/6ed96b0aa719a655327e16eb493dadc16752ee2e)

lib/src/components/NxForm/NxForm.scss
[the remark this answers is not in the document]
EmilyBermudez88: As discussed, NxForm.RequiredFieldNotice has been updated- the span has been replaced with a paragraph element: [6a6aa83](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/6a6aa8356c243bf8f7437c442254114bbc3209b0)

EmilyBermudez88:
> Missed `NxTileFormErrorExample`. Also `NxFormValidationExample` but that one's being deleted in RSC-1080 so don't worry about it.

Thanks for pointing that out. It has been added here : [6a6aa83](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/6a6aa8356c243bf8f7437c442254114bbc3209b0)

gallery/src/styles/NxFormLayout/NxFormHorizontalLayoutExample.tsx
EmilyBermudez88: Corrected here: [6a6aa83](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/6a6aa8356c243bf8f7437c442254114bbc3209b0)

lib/src/base-styles/_nx-forms.scss
```diff
@@ -14,6 +14,12 @@
     color: var(--nx-swatch-red-40);
   }
 
+  &__required-field-notice {
```
rpokorny, line 17: We're failing to specify the top margin here, and getting the browser default for `<p>` tags.  I think a better approach would be to ensure (in the `RequiredFieldNotice` component) that the element includes both this class and the `nx-p` class (e.g, use `<NxP>` instead of `<p>`) so that we just get the normal `nx-p` margins and don't have to worry about margins at all in _this_ class.

lib/src/base-styles/_nx-forms.scss
```diff
@@ -14,6 +14,12 @@
     color: var(--nx-swatch-red-40);
   }
```
rpokorny, line 16: I'm wondering if `.nx-form` should be using the `container-vertical` mixin.  That would typically remove the top margin from the required field message (or whatever else is first) which is generally what we want.  See http://localhost:4043/#/pages/Container%20Helpers .  I'd definitely say try adding that, and if it breaks any visual tests outside of those intended for this PR, we'll figure out what to do from there.

gallery/src/components/NxModal/NxModalFormErrorExample.tsx
```diff
@@ -6,8 +6,15 @@
  */
 import React, { useState } from 'react';
 
-import { NxModal, NxFontAwesomeIcon, NxTextInput, NxButton, nxTextInputStateHelpers, NxFormGroup, NxStatefulForm }
-  from '@sonatype/react-shared-components';
+import { NxModal,
```
rpokorny, line 9:
```suggestion
import {
  NxModal,
```

gallery/src/components/NxForm/NxFormPage.tsx
EmilyBermudez88: > The row that this thread is about is still present. That commit only affects other files.

Sorry, now removed here: [97b2c40](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/97b2c4096f827cb510a729cd2317fa689a60ab5b)

lib/src/base-styles/_nx-forms.scss
EmilyBermudez88: > I'm wondering if `.nx-form` should be using the `container-vertical` mixin. That would typically remove the top margin from the required field message (or whatever else is first) which is generally what we want. See http://localhost:4043/#/pages/Container%20Helpers . I'd definitely say try adding that, and if it breaks any visual tests outside of those intended for this PR, we'll figure out what to do from there.

Changed here: [5d3df01](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/5d3df01ec33740d86c7ab63f215e25a2fbe3dea8)

After running the visual tests, the test failures came from changes in validation and removing the "optional" tag, which were all there previously. The only one I wanted to confirm is irrelevant is :

nx-formlayout-js-nx-form-nx-fieldset-deprecated-layout-looks-right-1-diff

The margin for the last child appears to be affected in this fieldset, but it seems to be coming from .nx-fieldset<:lastchild(2) vs .nx-fieldset<:last-child

Screen Shot 2022-07-13 at 1 23 43 PM
Screen Shot 2022-07-13 at 1 23 34 PM

rpokorny: > After running the visual tests, the test failures came from changes in validation and removing the "optional" tag, which were all there previously. The only one I wanted to confirm is irrelevant is :

It's irrelevant to this PR yes.  It's a bug that needs to be fixed before the feature branch gets merged to main, and I think I addressed it in the RSC-1080 PR

lib/src/base-styles/_nx-forms.scss
```diff
@@ -10,10 +10,18 @@
...
+  @include nx-container-helpers.container-vertical;
+
   &__required-field-asterisk {
     color: var(--nx-swatch-red-40);
   }
 
+  &__required-field-notice {
+    font-weight: 600;
+    color: var(--nx-color-text-dark);
+    margin-bottom: var(--nx-spacing-6x);
```
rpokorny, line 22: This margin should be redundant now, I'm pretty certain it's what `.nx-p` already provides

rpokorny (approved):
+1 aside from https://github.com/sonatype/sonatype-react-shared-components/pull/735/files#r920382531

lib/src/base-styles/_nx-forms.scss
EmilyBermudez88: I checked to confirm and yes, the margin is the same, so it has been removed here: [4f092cd](https://github.com/sonatype/sonatype-react-shared-components/pull/735/commits/4f092cdf6a2317de6ef8934a45529ed4d7c2b40b)

PR #737, opened 2022-06-30
Updated NxThreatCounter layout - RSC-972

entroform:
https://issues.sonatype.org/browse/RSC-972


entroform:
The updated snapshots look very different from when I run it locally on my machine. I wonder if this is because it's using an much older version of Chrome.

Here's what I see on my local machine:
Screen Shot 2022-06-30 at 10 05 28 AM
Screen Shot 2022-06-30 at 10 05 17 AM
Screen Shot 2022-06-30 at 10 04 58 AM

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -54,16 +54,20 @@
...
 }
 
 .nx-threat-counter-container--column {
-  grid-template-columns: auto;
+  grid-template-columns: min-content;
 }
 
 .nx-threat-counter-container--grid {
-  grid-template-columns: auto auto;
+  grid-template-columns: min-content min-content;
```
rpokorny, line 72: I don't think it was previously believed that the grid mode was buggy.  However, looking closer at it now, I see that it is in that the two columns are supposed to be the same width but they aren't.  But the change you have here does not fix or change that.

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -54,16 +54,20 @@
...
+  display: grid;
   gap: var(--nx-spacing-4x);
-  grid-template-columns: repeat(5, 1fr);
+  grid-template-columns: repeat(auto-fill, minmax($min-width, 1fr));
   margin: var(--nx-spacing-4x) 0 var(--nx-spacing-6x) 0;
 }
 
 .nx-threat-counter-container--column {
-  grid-template-columns: auto;
+  grid-template-columns: min-content;
```
rpokorny, line 68: Why was this changed?

rpokorny:
After looking at this I don't think it's possible with the current capabilities of CSS grid.  If I understand the documentation on MDN correctly, what basically happens is that when you use `auto-fill` with `minmax`, one of the two arguments to `minmax` must be a definite size (e.g. a length or percentage) it performs the layout based on that size and the available space, _not_ on the content of any of the cells.

Additionally, I'm thinking about where this component is used and what this mode of operation would look like in practice, and I'm not sure this really solves anything.  Anywhere that has this component, in its default configuration where all of the indicators are in a row, is probably either some more complex layout that gives the indicators only a certain amount of horizontal and vertical space (i.e. only one row worth of vertical space) or is using it in an inline context were we might want the indicators to wrap to the next line box of the parent, not wrap within an inline rectangular area.  We should get clarity on what we are really trying accomplish here, in terms of the component's surroundings not just the component itself.

entroform:
> I'm afraid this doesn't seem to really be working. If I adjust the data, the counters do not resize accordingly: image

Ah i totally missed the part with the numbers

entroform:
@rpokorny Updated layout here: c7abe5da7a54c76caca051da83b81589dec375ea

lib/src/components/NxThreatCounter/NxThreatCounter.scss
```diff
@@ -54,16 +54,17 @@
 }
 
 .nx-threat-counter-container {
-  display: inline-grid;
+  display: inline-flex;
```
rpokorny, line 57: Didn't the designers agree that it should be fully inline, for instance that here I should be seeing the first two counters or so on the same line as the asdfasfasdf text?
image

entroform: > Didn't the designers agree that it should be fully inline, for instance that here I should be seeing the first two counters or so on the same line as the asdfasfasdf text? image

Hi Ross, after consulting with @JasonDSwearingen and after discussions, they decided that we're changing this to flex: f9f1673b5705fee581e328107afbc544230bbf16

They're working on updating Kaleido and they said they'll approve this ticket on Monday after the update is done 👍

rpokorny:
I see visual test differences in the other modes that weren't supposed to be affected by this ticket.

rpokorny:
We should really have an example and visual test to prove out the new behavior.  How about one where the counters are put inside of a div with an artificially narrow width and a visible border, similar to the "Checkbox label should not wrap" example on the Checkbox page.

entroform:
> We should really have an example and visual test to prove out the new behavior. How about one where the counters are put inside of a div with an artificially narrow width and a visible border, similar to the "Checkbox label should not wrap" example on the Checkbox page?

Added new example and visual test here: 5b33cebbab7a04a5b80eec8842fa7b78536a8b41

entroform:
> I see visual test differences in the other modes that weren't supposed to be affected by this ticket.

Strangely, I compared the "added" snapshots of the other modes that weren't affected by this ticket and it matches what it is currently on my chrome and firefox browser on `main`. 🤔  

For example: before "none" is uniformed sized but when I checked the current live gallery it is non uniform?
Screen Shot 2022-07-26 at 12 34 03 PM

JasonDSwearingen:
+1

rpokorny:
> Strangely, I compared the "added" snapshots of the other modes that weren't affected by this ticket and it matches what it is currently on my chrome and firefox browser on main. thinking
>
> For example: before "none" is uniformed sized but when I checked the current live gallery it is non uniform?

Not sure if you got something mixed up but when I look at that example on gallery.sonatype.com in Chrome I see the Critical and None counters being the same width.

Note that the size of the counters isn't the only difference.  The height of the screenshots themselves has also changed unexpectedly.

entroform:
> > Strangely, I compared the "added" snapshots of the other modes that weren't affected by this ticket and it matches what it is currently on my chrome and firefox browser on main. thinking
> > For example: before "none" is uniformed sized but when I checked the current live gallery it is non uniform?
> 
> Not sure if you got something mixed up but when I look at that example on gallery.sonatype.com in Chrome I see the Critical and None counters being the same width.
> 
> Note that the size of the counters isn't the only difference. The height of the screenshots themselves has also changed unexpectedly.

Fixed the vertical spacing here: b11bda662254c26dbc77e0683f4ea10a9b055922 (Haven't updated the snapshots yet).

I also tested it on gallery.sonatype.com on both Chrome and Firefox: it seems to be non-uniform 🤔
Screen Shot 2022-07-26 at 4 36 39 PM

rpokorny:
> I also tested it on gallery.sonatype.com on both Chrome and Firefox: it seems to be non-uniform thinking

I think we're confusing each other by talking about two different examples and sets of screenshots here.  In the "small row example", the two counters are the same width on main and not in your branch.  Which makes sense given the design decisions made on this ticket.

In the "small grid example", the two counters are different widths but are consistent between main and this branch.  The only difference I see in that screenshot is the height. That said, that does exhibit a bug that was not known prior to this branch, that you had [at one point attempted to fix on this branch](https://github.com/sonatype/sonatype-react-shared-components/pull/737#discussion_r911358995) but which is not currently fixed: that the two columns in that example are supposed to be the same width.  I think a separate ticket can/should be put in to fix that.

entroform:
> at one point attempted to fix on this branch

Ah I see, sorry about that, been all over the place today. Will create a separate ticket to fix the small grid layout.

entroform:
> > I also tested it on gallery.sonatype.com on both Chrome and Firefox: it seems to be non-uniform thinking
> 
> I think we're confusing each other by talking about two different examples and sets of screenshots here. In the "small row example", the two counters are the same width on main and not the same width in your branch. Which makes sense given the design decisions made on this ticket.
> 
> In the "small grid example", the two counters are different widths but are consistent between main and this branch. The only difference I see in that screenshot is the height. That said, that does exhibit a bug that was not known prior to this branch, that you had [at one point attempted to fix on this branch](https://github.com/sonatype/sonatype-react-shared-components/pull/737#discussion_r911358995) but which is not currently fixed: that the two columns in that example are supposed to be the same width. I think a separate ticket can/should be put in to fix that.

Updated snapshots here: bd07eddfc770cb42194ec5296d7864f67a786779
Created ticket to fix the small grid here: https://issues.sonatype.org/browse/RSC-1136

rpokorny:
I'm seeing a bunch of snapshot "updates" that now appear to be identical.  Please revert them back to the originals

rpokorny (approved):
+1 aside from snapshots needing reverting

PR #740, opened 2022-07-06
Zero-styling for NxSmallThreatCounter - RSC-1087

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1087

Update the styling so when an NxSmallThreatIndicator receives a value of 0, it will appear "greyed out" instead of having its visibility set to hidden.


gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "10.2.6",
+  "version": "10.2.8",
```
EmilyBermudez88, line 3: resolved further down. Forgot to merge main, first before version bump

cindyko226:
Look good to me.

rpokorny:
I notice the "infinite max digits" configuration used to shows zeroes in full color.  Did Design consciously decide to change that?  I'm not sure I agree with that change as that variation of the counter is intended to be used in situations where you only have one instance (and not a bunch lined up across different table rows).  In such a situation, you don't have the positional context to know what severity level the zero is referring to, so the color becomes more important IMO.

lib/src/base-styles/_nx-colors.scss
```diff
@@ -40,6 +40,7 @@
   --nx-color-threat-moderate: var(--nx-swatch-yellow-60);
   --nx-color-threat-severe: var(--nx-swatch-orange-50);
   --nx-color-threat-critical: var(--nx-swatch-red-40);
+  --nx-color-threat-zero: var(--nx-swatch-grey-90);
```
rpokorny, line 43: This isn't really a threat color but rather is a special styling treatment specific to `NxSmallThreatCounter`.  So I don't think a CSS variable should be added for it.  In particular it could easily be misconstrued: the other variables here are about different threat severity levels, while the zero-styling could be applied to a counter at any severity level and is about not the level, but the count of violations at that level.

EmilyBermudez88:
> I notice the "infinite max digits" configuration used to shows zeroes in full color. Did Design consciously decide to change that? I'm not sure I agree with that change as that variation of the counter is intended to be used in situations where you only have one instance (and not a bunch lined up across different table rows). In such a situation, you don't have the positional context to know what severity level the zero is referring to, so the color becomes more important IMO.

Having the colour change in the maxDigits example wasn't discussed, so I'll pose it to the channel to get their insight.

lib/src/base-styles/_nx-colors.scss
EmilyBermudez88: > This isn't really a threat color but rather is a special styling treatment specific to `NxSmallThreatCounter`. So I don't think a CSS variable should be added for it. In particular it could easily be misconstrued: the other variables here are about different threat severity levels, while the zero-styling could be applied to a counter at any severity level and is about not the level, but the count of violations at that level.

That makes sense. It has been removed here: [6a16448](https://github.com/sonatype/sonatype-react-shared-components/pull/740/commits/6a16448d850c8714a20c661aa2f7b4b8dde9ce89)

EmilyBermudez88:
> > I notice the "infinite max digits" configuration used to shows zeroes in full color. Did Design consciously decide to change that? I'm not sure I agree with that change as that variation of the counter is intended to be used in situations where you only have one instance (and not a bunch lined up across different table rows). In such a situation, you don't have the positional context to know what severity level the zero is referring to, so the color becomes more important IMO.
> 
> Having the colour change in the maxDigits example wasn't discussed, so I'll pose it to the channel to get their insight.

It has been updated to reflect the original colours when maxDigits is set to infinity : [0735ca9](https://github.com/sonatype/sonatype-react-shared-components/pull/740/commits/0735ca9696ac872fe26e848af94451e9339534d6)

lib/src/components/NxSmallThreatCounter/NxSmallThreatCounter.scss
```diff
@@ -67,9 +68,18 @@
       text-align: center;
       width: 100%;
 
-      &--zero {
-        visibility: visible;
+      @each $category in critical, severe, moderate, low, none, unspecified {
```
rpokorny, line 71: This is repeated code from higher up right? There's probably ways to avoid that.  A `:not` selector if nothing else

EmilyBermudez88: > This is repeated code from higher up right? There's probably ways to avoid that. A `:not` selector if nothing else

In [32c60dc](https://github.com/sonatype/sonatype-react-shared-components/pull/740/commits/32c60dc098f038cdc458457581334b28447078c8) I tried using @if and @ else within a @mixin. However I'm not sure if it should be inside the scss-shared folder since the conditional will just apply to this situation (and I couldn't get an @if and @else statement to work outside of a @mixin properly).

lib/src/components/NxSmallThreatCounter/NxSmallThreatCounter.scss
```diff
@@ -67,9 +72,7 @@
       text-align: center;
       width: 100%;
 
-      &--zero {
-        visibility: visible;
-      }
+      @include counter-colors;
```
rpokorny, line 75: Maybe add a comment that this is to override the zero styling

rpokorny (approved):

lib/src/components/NxSmallThreatCounter/NxSmallThreatCounter.scss
EmilyBermudez88: Added here: [38e9f58](https://github.com/sonatype/sonatype-react-shared-components/pull/740/commits/38e9f5842c881638126d3c12fc863973a388c047)

JasonDSwearingen (approved):

PR #747, opened 2022-07-12
Rename NxCodeSnippet to NxCopyToClipboard - RSC-932

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-932

At this point in time, I have not updated any snapshots with the new NxCopyToClipboard naming. I wanted to confirm the other changes in deprecating NxCodeSnippet and adding NxCopyToClipboard were correct prior to moving forward. I have not touched NXCodeSnippet inside the library, since I assume it will still need to be available in the near future.


EmilyBermudez88:
> > I have not touched NXCodeSnippet inside the library, since I assume it will still need to be available in the near future.
> 
> Observe in `index.ts` how `NxTreeView` is "implemented". That's a very similar situation. We used to have a component called `NxTreeView` but we renamed it to `NxCollapsibleItems`. `NxTreeView` still exists for backwards compatibility but we didn't keep two copies of the code around. Instead, `NxTreeView` is just another export of the same objects under a different name. And then for tests, notice that in `NxCollapsibleItems.test.tsx` there are just a few tests that the old name refers to the same object, rather than fully testing both names. All of this is basically how you should approach `NxCodeSnippet`.

Thanks for the tip. I missed updating the export file in which NxCodeSnippet was coming from here ([322ab51](https://github.com/sonatype/sonatype-react-shared-components/pull/747/commits/322ab51f2bd7f14777280ed15381ba957bcb5bf1)), so that was very helpful. Besides updating snapshots and merging main into RSC-932, I think everything else is done, but not sure.

lib/src/index.ts
```diff
@@ -232,7 +232,12 @@ export { selectableColors, selectableColorClasses, SelectableColor } from './uti
 // deprecated
 export { selectableColors as NX_TAG_COLORS, SelectableColor as NX_TAG_COLORS_TYPE } from './util/selectableColors';
 
-export { default as NxCodeSnippet, Props as NxCodeSnippetProps } from './components/NxCodeSnippet/NxCodeSnippet';
+export { default as NxCopyToClipboard, Props as NxCopyToClipboardProps}
```
rpokorny, line 235:
```suggestion
export { default as NxCopyToClipboard, Props as NxCopyToClipboardProps }
```

rpokorny (approved):

rpokorny (approved):

PR #752, opened 2022-07-26
Update HTML Description List RSC-1077

achen5:
https://issues.sonatype.org/browse/RSC-1077

Turning on Hide Whitespace mode when reviewing the SCSS changes is recommended.


achen5:
Hi @rpokorny,

Can I get your feedback on my `NxDescriptionList.ButtonItem` component? [b78807a](https://github.com/sonatype/sonatype-react-shared-components/pull/752/commits/b78807a10c59ce92e2ede4779acefe5a6cab6060)
To accommodate  `dd` and `dt` limitations, I felt like my JSX has become hacky. Same applies to SCSS as well. I added new SCSS rules to the component's padding so the callback function runs wherever the user clicks on the row.
I want to confirm that I'm on the right track before I commit more time to this ticket. After I get your feedback, I'll build out `NxDescriptionList.LinkItem`, write test cases and update the documentation.

Thanks!

rpokorny:
I don't think you should be reusing `nx-list__link` and `nx-list__btn` for your needs here.  Those two classes replicate `nx-list__item` in terms of their internal structure: they have a grid layout that accomodates the main text area, subtext, action buttons, and special indicators that can all be used within a normal `nx-list` item.  Almost none of that applies within the description list `<dd>` and `<dt>`

lib/src/components/NxList/NxDescriptionListButtonItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+            selected: selectedProp,
+            buttonClassName,
+            buttonAttributes,
+            ...attrs
+          } = props,
+          selected = selectedProp ?? undefined,
+          divClassNames = classnames('nx-list__item', 'nx-list__item--clickable', className),
+          buttonClassNames = classnames('nx-list__btn', buttonClassName, { selected, disabled });
+
+      /* Using children[0].props.children allows for a more natural API for the consuming developer. */
```
rpokorny, line 29: Unfortunately it makes more assumptions about the DOM that the caller is passing in.  For instance it assumes that they're passing in `<dt>My Content</dt>` and not `<MyCustomDt />`.  We try very hard not to make assumptions about JSX children wherever we can avoid it.  In this case I think a better API would be to create `NxDescriptionList.ButtonTerm` and `NxDescriptionList.ButtonDescription` components.  Or an an alternative, have `NxDescriptionList.ButtonItem` take the term and description JSX children as separate props instead of trying to dig them out of its own `children`.

JenniKobylski (approved):
Looks good!

lib/src/components/NxDescriptionList/NxDescriptionListButtonItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    disabled={disabled ? true : false}
+                    onClick={() => onClick()}>
+              <span className="nx-list__text">{term}</span>
+            </button>
+          </dt>
+          <dd className="nx-list__description">
+            {/* This button only exists for the sake of expanding the click target all the way across the
+                row. From a screenreader/keynav perspective only the other button is present. This button's
+                text is attached to the other button via aria-describedby
+            */}
```
EmilyBermudez88, line 53: A clever workaround to resolve expanding the clickable target. Is it possible to wrap a single button around both the `dt` and `dd` elements and then adjust the css accordingly, or is this the cleaner workaround?

gallery/src/components/NxDescriptionList/NxDescriptionListEmptyExample.tsx
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxDescriptionList } from '@sonatype/react-shared-components';
+
+const NxDescriptionListNonInteractiveExample = () => <NxDescriptionList />;
```
EmilyBermudez88, line 10:
```suggestion
const NxDescriptionListEmptyExample = () => <NxDescriptionList />;
```

gallery/src/components/NxDescriptionList/NxDescriptionListLinkExample.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxDescriptionList } from '@sonatype/react-shared-components';
+
+const NxDescriptionListLinkExample = () =>
+  <NxDescriptionList>
+    <NxDescriptionList.LinkItem href="#/"
+                                term="This is the NxDescriptionList.LinkItem component"
+                                description="Clicking this row will give you an alert message." />
```
EmilyBermudez88, line 14: Minor, but this doesn't give you an alert in the Link Example

gallery/src/components/NxDescriptionList/NxDescriptionListPage.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+const NxDescriptionListPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <NxP>
+        <NxCode>NxDescriptionList</NxCode> creates HTML description lists with styling similar
+        to <NxCode>NxList</NxCode>. These lists can support clickable rows implemented as either buttons or links,
+        though in its simplest, noninteractive usage the <NxCode>NxDescriptionList</NxCode> family of components is
+        simply a convenience wrapper for the <NxCode>nx-list--description-list</NxCode> styles documented on the
+        <NxTextLink href="#/pages/List%20(HTML)">List (HTML)</NxTextLink> style page. Creating clickable description
```
EmilyBermudez88, line 29:
```suggestion
       {' '} <NxTextLink href="#/pages/List%20(HTML)">List (HTML)</NxTextLink> style page. Creating clickable description
```

lib/src/components/NxDescriptionList/NxDescriptionListButtonItem.tsx
rpokorny: >  Is it possible to wrap a single button around both the dt and dd elements and then adjust the css accordingly

Not and have it be valid HTML. `<dl>`'s immediate children may only be `<dt>`, `<dd>`, or `<div>` containing only `<dt>` or `<dd>`.

gallery/src/components/NxDescriptionList/NxDescriptionListPage.tsx
rpokorny: [9c04d17](https://github.com/sonatype/sonatype-react-shared-components/pull/752/commits/9c04d17c3808719e30eb0152b0ebc2941bafa45f)

gallery/src/components/NxDescriptionList/NxDescriptionListLinkExample.tsx
rpokorny: [db5788c](https://github.com/sonatype/sonatype-react-shared-components/pull/752/commits/db5788c31295be3cfab9fce5e63880f31876bf4d)

gallery/src/components/NxDescriptionList/NxDescriptionListEmptyExample.tsx
rpokorny: [0ff8daf](https://github.com/sonatype/sonatype-react-shared-components/pull/752/commits/0ff8dafa62e33b8b16debfa7afba61c0abdfda5f)

lib/src/components/NxDescriptionList/NxDescriptionListButtonItem.tsx
EmilyBermudez88: > > Is it possible to wrap a single button around both the dt and dd elements and then adjust the css accordingly
> 
> Not and have it be valid HTML. `<dl>`'s immediate children may only be `<dt>`, `<dd>`, or `<div>` containing only `<dt>` or `<dd>`.

Thanks for the clarification, good to know for the future.

EmilyBermudez88:
+1 approval

entroform:
This is probably not a big deal and unavoidable, but when I use the keyboard nav and spacebar to activate the row only one of the buttons has the active styling.

Screen Shot 2022-08-17 at 10 45 06 AM

EmilyBermudez88 (approved):

rpokorny:
> This is probably not a big deal and unavoidable, but when I use the keyboard nav and spacebar to activate the row only one of the buttons has the active styling.

I wasn't exactly able to fix that but while investigating it I noticed a related problem: that if you click and drag on a row, you'll get the split color background as well.  I've fixed that in 9d3dc5af3, which also affected the way it works for keyboard nav: basically the active state no longer works at all for keyboard nav.  IMO this is a browser bug, as what it comes down to is that the `:active` class should apply to ancestors of activated elements, but in the case of keyboard activation does not.  In the case of mouse activation it does.  Another thing that I noticed is that in Firefox, keyboard activation doesn't even trigger the `:active` class on the activated element (the button) itself, meaning that even before my most recent commit, active styles were not being applied to the rows on keyboard activation, and in fact even our other buttons like `NxButton` get no active styling when used via keyboard in Firefox.

entroform:
> > This is probably not a big deal and unavoidable, but when I use the keyboard nav and spacebar to activate the row only one of the buttons has the active styling.
> 
> I wasn't exactly able to fix that but while investigating it I noticed a related problem: that if you click and drag on a row, you'll get the split color background as well. I've fixed that in [9d3dc5a](https://github.com/sonatype/sonatype-react-shared-components/commit/9d3dc5af3b8aaf874bd796a2a170f7ff40084e48), which also affected the way it works for keyboard nav: basically the active state no longer works at all for keyboard nav. IMO this is a browser bug, as what it comes down to is that the `:active` class should apply to ancestors of activated elements, but in the case of keyboard activation does not. In the case of mouse activation it does. Another thing that I noticed is that in Firefox, keyboard activation doesn't even trigger the `:active` class on the activated element (the button) itself, meaning that even before my most recent commit, active styles were not being applied to the rows on keyboard activation, and in fact even our other buttons like `NxButton` get no active styling when used via keyboard in Firefox.

I've verified that this does fixed it. That is interesting that firefox keyboard navigation doesn't trigger the `:active` class.

entroform:
I noticed that there is a one pixel gap between the focus border. And also the focus border is being overlapped by the selected background (not sure if this one is an issue)

Screen Shot 2022-08-19 at 9 42 41 AM
Screen Shot 2022-08-19 at 9 42 31 AM

entroform (approved):

rpokorny:
> I noticed that there is a one pixel gap between the focus border. And also the focus border is being overlapped by the selected background (not sure if this one is an issue)

[27dfed9](https://github.com/sonatype/sonatype-react-shared-components/pull/752/commits/27dfed95dcf0d6867de3b9e204df04faf1a5d104)

PR #754, opened 2022-07-29
NxFieldset Validation - RSC-1081

rpokorny:
https://issues.sonatype.org/browse/RSC-1081

This PR adds officially adds `isPristine` and `validationErrors` props to `NxFieldset` (they were already there from the spike, but this adds documentation etc) and a number of helper functions to help manage those properties. The helper functions come in two flavors: pure state helpers similar to what we have for `NxTextInput`, intended to be used when state is managed outside of React (e.g. in Redux), and some React hooks which are more convenient for use when managing state within React.


gallery/src/components/NxFieldset/NxFieldsetPage.tsx
```diff
@@ -98,21 +129,242 @@ const NxFieldsetPage = () =>
...
+        <NxTile.SubsectionHeader>
+          <NxH3>State Helpers</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          When an <NxCode>NxFieldset</NxCode> has validation, there are several pieces of state that must be tracked
+          for the fieldset itself in conjunction with the state for the fieldset's contents. Specifically,
+          the values of the <NxCode>isPristine</NxCode> and <NxCode>validationErrors</NxCode> props should be set based
+          on the current values and user interactions within the fieldset. To assist with this, a number of utility
+          function are provided in the <NxCode>nxFieldsetStateHelpers</NxCode> export. These function fall into two
+          categories: pure functions to assist in state managment outside of React (e.g. in Redux), and React hook
```
rjp99, line 142:
```suggestion
          categories: pure functions to assist in state management outside of React (e.g. in Redux), and React hook
```

gallery/src/components/NxFieldset/NxFieldsetPage.tsx
```diff
@@ -98,21 +129,242 @@ const NxFieldsetPage = () =>
...
+                  <NxList.Item>
+                    <NxList.Text>
+                      <NxCode>validator</NxCode> function which will be run on each selected radio value over
+                      time in order to update the <NxCode>validationErrors</NxCode>; optional
+                    </NxList.Text>
+                  </NxList.Item>
+                </NxList>
+              </NxTable.Cell>
+              <NxTable.Cell>
+                Two value tuple where the first value is an bject containing the <NxCode>value</NxCode>,{' '}
```
rjp99, line 297:
```suggestion
                Two value tuple where the first value is an object containing the <NxCode>value</NxCode>,{' '}
```

gallery/src/components/NxFieldset/NxFieldsetPage.tsx
```diff
@@ -98,21 +129,242 @@ const NxFieldsetPage = () =>
...
+              </NxTable.Cell>
+              <NxTable.Cell>
+                Object containing the <NxCode>value</NxCode>, <NxCode>validationErrors</NxCode> as determined by the
+                validator function, and <NxCode>isPristine</NxCode> set to false
+              </NxTable.Cell>
+              <NxTable.Cell>
+                Creates an object representing the state of a group of radio buttons after user input, based on the
+                specified user-selected value, and validated according to the specified validator.
+                This function can be used to update the pieces of state needed for a fieldset containing a radio group
+                upon user input.
```
cindyko226, line 263:
```suggestion
                Creates an object representing the state of a group of checkboxes after user input, based on the
                specified user-selected value, and validated according to the specified validator.
                This function can be used to update the pieces of state needed for a fieldset containing a checkbox group
                upon user input.
```

gallery/visualtests/nxFieldset.js
```diff
@@ -0,0 +1,55 @@
...
+      const [submitBtn, fieldset] = await waitAndGetElements(
+          `${formSelector} .nx-form__submit-btn`,
+          `${formSelector} .nx-fieldset:first-of-type`
+      );
+
+      await submitBtn.click();
+
+      await checkScreenshot(fieldset, undefined, 152);
+    });
+  });
```
cindyko226, line 54: I think we need to add a11yTest as well.

gallery/src/components/NxFieldset/NxFieldsetRequiredPureStateHelperExample.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <>
+      <NxFieldset label="Color"
+                  isRequired
+                  isPristine={color.isPristine}
+                  validationErrors={color.validationErrors}>
+        <NxRadio value="red" isChecked={color.value === 'red'} name="color" onChange={setColor}>Red</NxRadio>
+        <NxRadio value="green" isChecked={color.value === 'green'} name="color" onChange={setColor}>Green</NxRadio>
+        <NxRadio value="blue" isChecked={color.value === 'blue'} name="color" onChange={setColor}>Blue</NxRadio>
+        <NxRadio value="purple" isChecked={color.value === 'purple'} name="color" onChange={setColor}>Purple</NxRadio>
```
cindyko226, line 60: This would need its own unique `name`.

```suggestion
        <NxRadio value="red" isChecked={color.value === 'red'} name="color1" onChange={setColor}>Red</NxRadio>
        <NxRadio value="green" isChecked={color.value === 'green'} name="color1" onChange={setColor}>Green</NxRadio>
        <NxRadio value="blue" isChecked={color.value === 'blue'} name="color1" onChange={setColor}>Blue</NxRadio>
        <NxRadio value="purple" isChecked={color.value === 'purple'} name="color1" onChange={setColor}>Purple</NxRadio>
```

lib/src/components/NxFieldset/types.ts
```diff
@@ -23,3 +23,29 @@ export const propTypes: PropTypes.ValidationMap<Props> = {
...
 };
+
+// The following are for the state helpers
+export type RadioValidator = (value: string | null) => ValidationErrors;
+export type CheckboxValidator = (values: string[]) => ValidationErrors;
+export type RadioSetter = (v: string | null) => void;
+export type CheckboxState = [boolean, () => void];
+export type CheckboxInitValues<K extends string | number> = Record<K, boolean>;
+export type CheckboxStates<K extends string | number> = Record<K, CheckboxState>;
+
```
EmilyBermudez88, line 34: Just for my understanding, is type CheckboxStates essentially an object, in which each of it's keys are type string or number, and the values associated with each key are type checkboxState (and therefore an array containing a boolean and a function- in this case the toggle function)?

gallery/visualtests/nxFieldset.js
rpokorny: I had been relying on it getting tested on the Form Layout Page, but you're right that it's probably better to be clear about it. 606acd7e1

gallery/src/components/NxFieldset/NxFieldsetRequiredPureStateHelperExample.tsx
rpokorny: > This would need its own unique name.

ae1fd03d2

lib/src/components/NxFieldset/types.ts
rpokorny: > Just for my understanding, is type CheckboxStates essentially an object, in which each of it's keys are type string or number, and the values associated with each key are type checkboxState (and therefore an array containing a boolean and a function- in this case the toggle function)?

Yes, exactly.

lib/src/components/NxFieldset/stateHelpers.ts
```diff
@@ -0,0 +1,144 @@
...
+): CheckboxGroupHookReturnValue<K> {
+  type Pair = [K, CheckboxState];
+
+  function wrapState([state, toggler]: CheckboxState): CheckboxState {
+    return [state, () => { setIsPristine(false); toggler(); }];
+  }
+
+  const rawStates = map<CheckboxInitValues<K>, CheckboxStates<K>>(_useToggle, initialValues),
+      [isPristine, setIsPristine] = useState(true),
+      values = map<Pair, K>(head, filter<Pair, Pair[]>(([, [isSet]]) => isSet, toPairs(rawStates)));
```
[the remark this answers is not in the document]
rpokorny: > Can you explain the process of how values is defined here? I understand map/head, but am unsure about the filter method here (what is [, [isSet]], and how is it applied to the rawStates). Thanks!

`toPairs` is a function which turns an object into a list of key-value pairs.  For example,
`toPairs({ a: 1, b: 2 })` results in `[['a', 1], ['b', 2]]`.  But in this case, all of the property values in `rawStates` are themselves lists (2-value tuples to be precise) because they're all return values from `_useToggle`.  So we have a triple-nested list being passed into `filter`.

For each item in that top-level list, `filter` applies the predicate that we've defined here as an arrow function.  That function uses destructing to pull out the only value in that structure that we're interested in, calls it `isSet`, and returns it.  When destructuring, you can ignore the values at list indexes you don't care about by simply writing the comma and no variable name to capture the value.  So for instance destructing with `[, x, , y]` would expect a list of at least four values, and would pull out the second of those values as the local variable `x` and the fourth of those values as the local variable `y`.  In our case, we are interested in the second value of that level of the structure (the filter predicate doesn't care about the name of the state).  

Now for that second value, which we do care about, we don't care about all of it.  This value is the third, deepest level of the triple-nested-list structure.  As I mentioned above, it's a `_useToggle` return value i.e. a `[state, toggleState]` structure.  The filter predicate function doesn't need the toggle function it only needs the state: whether or not this items _is set_.  Hence the name of the variable we destructure into.  So again, we are destructuring a list and only using part of it.  But in this case, the part we're ignoring is at the tail end rather than the front.  We could write this part as `[isSet, ]`, but that's not necessary.  We can just totally ignore that the rest of the list exists and write our destructuring as if that first value is all that's there.

So backing up, we use `toPairs` to turn the object into a list of key-value pairs.  We use filter to filter our the pairs whose state value is not currently `true` (using the arrow function to dig out the state value from each pair).  Then on the remaining pairs, we want their names - the keys.  So we map over them and call `head` on each one.  `head` is a function that retrieves the first value from a list. The first value of each pair is the key so that's what we want.

As I write this, it occurs to me that the filter predicate probably could have been written using ramda's `path` function instead (`path([1, 0])`).  That's certainly shorter; I'm not sure if it's clearer.  If you think it is I'd be happy to make that change.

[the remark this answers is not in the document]
EmilyBermudez88: > As I write this, it occurs to me that the filter predicate probably could have been written using ramda's `path` function instead (`path([1, 0])`). That's certainly shorter; I'm not sure if it's clearer. If you think it is I'd be happy to make that change.

Thanks for such a detailed description, that was a big help. I didn't realize it was destructuring at work, so the code makes a lot more sense with your explanation. I would agree with you, that it may not be any clearer to use path in this case.

cindyko226 (approved):

lib/src/components/NxFieldset/types.ts
rpokorny: > > Just for my understanding, is type CheckboxStates essentially an object, in which each of it's keys are type string or number, and the values associated with each key are type checkboxState (and therefore an array containing a boolean and a function- in this case the toggle function)?
>
> Yes, exactly.

Oh I wanted to elaborate on this a little.  It's not quite that `CheckboxState`'s keys are strings or numbers, it's that they can be specific strings and numbers, known at compile time.  So if it's known at compile time that it's an object like `{ foo: ..., bar: ... }` (say, because it was written as an object literal), then `K` becomes just `'foo' | 'bar'`, not `string | number`.  In this way the compiler can track what keys are involved and what keys will be present in objects derived from this object.

EmilyBermudez88: > Oh I wanted to elaborate on this a little. It's not quite that CheckboxState's keys are strings or numbers, it's that they can be specific strings and numbers, known at compile time. So if it's known at compile time that it's an object like { foo: ..., bar: ... } (say, because it was written as an object literal), then K becomes just 'foo' | 'bar', not string | number. In this way the compiler can track what keys are involved and what keys will be present in objects derived from this object.

Sorry, I never saw this comment. Thanks for expanding on this a bit more.

PR #755, opened 2022-08-15
NxFileUpload - RSC-1061

rpokorny:
https://issues.sonatype.org/browse/RSC-1061

### DIFF
--- a/gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
+++ b/gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
@@ -141,6 +141,55 @@ const NxTextInputPage = () =>
           </NxTable.Body>
         </NxTable>
       </NxTile.Subsection>
+      <NxTile.Subsection>
+        <NxTile.SubsectionHeader>
+          <NxH3>State Helpers</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          Like <NxCode>NxTextInput</NxCode>, <NxCode>NxFileUpload</NxCode> comes with helper functions to more
+          conveniently manage its state. In this case, these helpers only manage the <NxCode>files</NxCode>
+          and <NxCode>isPristine</NxCode> props – validation is not supported other than required value
+          validation, which the component handles internally. The helper functions are availble on
+          the <NxCode>nxFileUploadStateHelpers</NxCode> export. Use of these helpers is not necessary
+          on non-required file uploads, as the <NxCode>isPristine</NxCode> flag is not used there.
+        </NxP>
+        <NxTable>
+          <NxTable.Head>
+            <NxTable.Row>
+              <NxTable.Cell>Function</NxTable.Cell>
+              <NxTable.Cell>Parameters</NxTable.Cell>
+              <NxTable.Cell>Return Value</NxTable.Cell>
+              <NxTable.Cell>Description</NxTable.Cell>
+            </NxTable.Row>
+          </NxTable.Head>
+          <NxTable.Body>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>initialValue</NxCode></NxTable.Cell>
+              <NxTable.Cell>files: <NxCode>FileList | null</NxCode></NxTable.Cell>
+              <NxTable.Cell>
+                Object containing <NxCode>files</NxCode> and <NxCode>isPristine</NxCode> properties
+              </NxTable.Cell>
+              <NxTable.Cell>
+                This function sets up the initial state of a file upload. That is, it returns an object
+                containing the specified file list (usually null) and the <NxCode>isPristine</NxCode> flag set
+                to <NxCode>true</NxCode>.
+              </NxTable.Cell>
+            </NxTable.Row>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>userInput</NxCode></NxTable.Cell>
+              <NxTable.Cell>files: <NxCode>FileList | null</NxCode></NxTable.Cell>
+              <NxTable.Cell>
+                Object containing <NxCode>files</NxCode> and <NxCode>isPristine</NxCode> properties
+              </NxTable.Cell>
+              <NxTable.Cell>
+                This function represents a modification of the state of a file upload. That is, it returns an object
+                containing the specified file list and the <NxCode>isPristine</NxCode> flag set
+                to <NxCode>false</NxCode>.
+              </NxTable.Cell>
+            </NxTable.Row>
+          </NxTable.Body>
+        </NxTable>
+      </NxTile.Subsection>
     </GalleryDescriptionTile>
 
     <GalleryExampleTile title="Simple Example"
--- a/gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
+++ b/gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
@@ -67,7 +67,7 @@ const NxTextInputPage = () =>
                 A callback which is called whenever the user changes their file selection. The new selection,
                 in the form of a <NxCode>FileList</NxCode>, is passed as the parameter. If the user has cleared their
                 selection, the callback will receive <NxCode>null</NxCode>. Otherwise, the <NxCode>FileList</NxCode>
-                will always contain a single <NxCode>File</NxCode>.
+                {' '}will always contain a single <NxCode>File</NxCode>.
               </NxTable.Cell>
             </NxTable.Row>
             <NxTable.Row>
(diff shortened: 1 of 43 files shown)


lib/jest.config.js
```diff
@@ -12,8 +12,10 @@ module.exports = {
     "<rootDir>/src"
   ],
   "transform": {
-    "^.+\\.tsx?$": "ts-jest"
+    "^.+\\.tsx?$": "ts-jest",
+    "^.+\\.jsx?$": "babel-jest"
   },
+  "transformIgnorePatterns": ['/node_modules/(?!(pretty-bytes)/)'],
```
rpokorny, line 18: pretty-bytes uses ES6 exports and so needs to be transpiled.  As third-party code, it isn't by default.

JasonDSwearingen (approved):

gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+            </NxTable.Row>
+            <NxTable.Row>
+              <NxTable.Cell>onChange</NxTable.Cell>
+              <NxTable.Cell>(FileList | null) =&gt; void</NxTable.Cell>
+              <NxTable.Cell>true</NxTable.Cell>
+              <NxTable.Cell/>
+              <NxTable.Cell>
+                A callback which is called whenever the user changes their file selection. The new selection,
+                in the form of a <NxCode>FileList</NxCode>, is passed as the parameter. If the user has cleared their
+                selection, the callback will receive <NxCode>null</NxCode>. Otherwise, the <NxCode>FileList</NxCode>
```
cindyko226, line 69:
```suggestion
                selection, the callback will receive <NxCode>null</NxCode>. Otherwise, the <NxCode>FileList</NxCode>{' '}
```

gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+      </NxTile.Subsection>
+      <NxTile.Subsection>
+        <NxTile.SubsectionHeader>
+          <NxH3>State Helpers</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          Like <NxCode>NxTextInput</NxCode>, <NxCode>NxFileUpload</NxCode> comes with helper functions to more
+          conveniently manage its state. In this case, these helpers only manage the <NxCode>files</NxCode>
+          and <NxCode>isPristine</NxCode> props – validation is not supported other than required value
+          validation, which the component handles internally. The helper functions are availble on
```
cindyko226, line 152:
```suggestion
          validation, which the component handles internally. The helper functions are available on
```

gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+            <NxTable.Row>
+              <NxTable.Cell>onChange</NxTable.Cell>
+              <NxTable.Cell>(FileList | null) =&gt; void</NxTable.Cell>
+              <NxTable.Cell>true</NxTable.Cell>
+              <NxTable.Cell/>
+              <NxTable.Cell>
+                A callback which is called whenever the user changes their file selection. The new selection,
+                in the form of a <NxCode>FileList</NxCode>, is passed as the parameter. If the user has cleared their
+                selection, the callback will receive <NxCode>null</NxCode>. Otherwise, the <NxCode>FileList</NxCode>
+                will always contain a single <NxCode>File</NxCode>.
```
cindyko226, line 70:
```suggestion
                {' '}will always contain a single <NxCode>File</NxCode>.
```

ssr-tests/pages/index.tsx
```diff
@@ -92,6 +92,8 @@ import {
   NxTree,
   NxVulnerabilityDetails,
   NxWarningAlert,
+  NxFileUpload,
+  NxStatefulFileUpload,
```
cindyko226, line 96: Just to confirm, all the new built components should be added to this file, correct?

lib/src/components/NxFileUpload/__tests__/NxFileUpload.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  it('attaches a ref to the top-level element', function() {
+    const ref = React.createRef<HTMLDivElement>(),
+        renderedEl = renderEl({ ref });
+
+    expect(ref.current).toBe(renderedEl);
+  });
+
+  // Everything that involves actually selecting a file cannot be cleanly tested in RTL, will be done in
+  // functional tests
```
cindyko226, line 79: Out of curious, would you please elaborate more why it can't be clean if tested in RTL? Thanks.

rpokorny: > Out of curious, would you please elaborate more why it can't be clean if tested in RTL? Thanks.

The only thing that RTL has for interaction with file upload components is https://testing-library.com/docs/user-event/utility#upload .  That utility is inadequate for our needs for two reasons:
1. It simulates a user clicking on the `<input>` itself.  In `NxFileUpload` the `<input>` is invisible and a real user actually triggers it by clicking the adjacent `<button>`.  So simulating a click on the `<input>` doesn't really verify anything with real-world relevance.
2. The `FileList` object that it results in is fake.  It's a plain JS object with all the the properties that you'd expect from a `FileList` and with its prototype hacked using `Object.setPrototypeOf`.  This is adequate for most cases of application code interacting with the `FileList` but it is _not_ adequate for any case where that object gets reassigned back to the `<input>`'s `value` property – which is exactly what `NxFileUpload` does.  In fact, `<input>`s are necessarily extremely strict about ensuring that it is impossible for a JS-constructed "FileList" to ever be accepted by the `value` property.  This is a security feature: the only way to get a reference to a file on the user's system is by the real user consciously selecting file via their interaction with the file input.  It must remain impossible for some JavaScript code to trigger an upload of, say, `/etc/passwd` without the user's direction.

ssr-tests/pages/index.tsx
rpokorny: > Just to confirm, all the new built components should be added to this file, correct?

Yes. This file is what verifies that the components can be rendered by a server-side rendering framework

lib/src/components/NxFileUpload/types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// Props for NxStatefulFileUpload
+export interface StatefulProps extends Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange' | 'multiple'> {
+  onChange?: ((files: FileList | null) => void) | null;
+  isRequired?: boolean | null;
+}
+
+// Props representing bits of state that need to be managed
+export interface StateProps {
+  files: FileList | null;
```
EmilyBermudez88, line 24: Out of curiosity, why do we use type FileList here instead of type File, if the file upload component only accepts up to one file?

rpokorny: > Out of curiosity, why do we use type FileList here instead of type File, if the file upload component only accepts up to one file?

Because `FileList` is what can be assigned to the `<input>`'s `value` property

ssr-tests/pages/index.tsx
cindyko226: Got it, thanks.

lib/src/components/NxFileUpload/__tests__/NxFileUpload.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+import NxFileUpload from '../NxFileUpload';
+
+describe('NxFileUpload', function() {
+  const minimalProps = {
+        files: null,
+        onChange: () => {}
+      },
+      render = rtlRender(NxFileUpload, minimalProps),
+      renderEl = rtlRenderElement(NxFileUpload, minimalProps),
+      renderInput = pipe(renderEl, el => el?.querySelector('input[type=file]'));
```
EmilyBermudez88, line 20: For clarity, renderInput is a function that takes the returned value of renderEl (which is the render of NxFileUpload, inside of the render() method's container), and then just querySelecting within NxFileUpload to find the input?

lib/src/components/NxFileUpload/__tests__/NxFileUpload.test.tsx
cindyko226: Thanks for the clarification.

lib/src/components/NxFileUpload/__tests__/NxFileUpload.test.tsx
rpokorny: > For clarity, renderInput is a function that takes the returned value of renderEl (which is the render of NxFileUpload, inside of the render() method's container), and then just querySelecting within NxFileUpload to find the input?

Yes. So overall, you pass it any extra props you want, and it returns the `<input>` of the resulting `NxFileUpload`.

lib/src/components/NxFileUpload/types.ts
EmilyBermudez88: Ok, thanks!

lib/src/components/NxFileUpload/__tests__/NxFileUpload.test.tsx
EmilyBermudez88: Great, thanks for confirming that.

EmilyBermudez88:
Screen Shot 2022-08-19 at 11 15 12 AM
Is there a concern if nxOverflowTooltip and nxTooltip overlap when the user hovers over the name of the file input but is focused on the "Dismiss Upload" button?

JasonDSwearingen:
> Screen Shot 2022-08-19 at 11 15 12 AM
> 
> Is there a concern if nxOverflowTooltip and nxTooltip overlap when the user hovers over the name of the file input but is focused on the "Dismiss Upload" button?

I don't think so. I think it's fine as is.

EmilyBermudez88 (approved):

PR #756, opened 2022-08-17
NxSearchDropdown Selection Fix for Safari - RSC-1153

rpokorny:
https://issues.sonatype.org/browse/RSC-1153

This will need to be tested across all supported browser and operating system combinations, especially focusing on Mac.

### DIFF
--- a/lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
+++ b/lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
@@ -28,10 +28,20 @@
     text-align: center;
   }
 
-  &--dropdown-showable:focus-within {
-    .nx-search-dropdown__menu {
-      position: absolute;
-      right: initial;
+  &--dropdown-showable {
+    @mixin show-menu {
+      .nx-search-dropdown__menu {
+        position: absolute;
+        right: initial;
+      }
+    }
+
+    &:focus-within {
+      @include show-menu
+    }
+
+    @supports (overflow: -webkit-paged-x) {
+      @include show-menu;
     }
   }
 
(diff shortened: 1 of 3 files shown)


entroform:
Tested it on Safari, Firefox, and Chrome (Just in case) on Mac Os 12.4 and it works!

entroform (approved):

rpokorny:
Thanks @entroform.  Would you be willing to test whether the current behavior on `main` changes depending on the setting described in https://superuser.com/a/372690 ? To better our understanding of the problem

entroform:
> Thanks @entroform. Would you be willing to test whether the current behavior on `main` changes depending on the setting described in https://superuser.com/a/372690 ? To better our understanding of the problem

Oh it turns out I already have that setting toggled on:
it produced the same result on `main` and Safari for me, whether the setting was toggled on or off (The click didn't work)

lib/src/components/NxSearchDropdown/NxSearchDropdown.scss
```diff
@@ -28,10 +28,26 @@
...
+    &:focus-within {
+      @include show-menu;
+    }
+
+    // Hack to detect Safari. In Safari, focus seems to be intended to strictly be a keyboard nav thing,
+    // and when you click a button, focus does NOT go to that button, it goes to the <body>. This messes up our idea
+    // of only showing the menu when something within the component has focus, because this means that the
+    // component loses focus in the middle of the click, causing the menu to hide while the user is in the middle
+    // of clicking, causing the click not to register at all. As a workaround, in Safari we remove the behavior
+    // where the menu is only visible when the compoent has focus.
```
avnvbhatta, line 48:
```suggestion
    // where the menu is only visible when the component has focus.
```

avnvbhatta (approved):

PR #759, opened 2022-08-23
NxForm RTL tests - RSC-1157

rpokorny:
https://issues.sonatype.org/browse/RSC-1157

### DIFF
--- a/lib/src/components/NxForm/__tests__/NxForm.test.tsx
+++ b/lib/src/components/NxForm/__tests__/NxForm.test.tsx
@@ -5,15 +5,12 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React, { useContext } from 'react';
-import { ShallowWrapper, mount, shallow } from 'enzyme';
+import { within, fireEvent } from '@testing-library/dom';
+import userEvent from '@testing-library/user-event';
 
-import { getShallowComponent, getMountedComponent } from '../../../__testutils__/enzymeUtils';
-import NxForm, { Props } from '../NxForm';
+import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
+import NxForm from '../NxForm';
 import NxButton from '../../NxButton/NxButton';
-import NxLoadError from '../../NxLoadError/NxLoadError';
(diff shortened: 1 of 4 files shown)


lib/src/components/NxForm/__tests__/NxForm.test.tsx
```diff
@@ -22,268 +19,298 @@ describe('NxForm', function() {
...
-      expect(getShallow({ children: () => <div id="foo" /> })).toContainMatchingElement('div#foo');
+      expect(quickRender({ children: () => <div data-testid="foo" /> }).getByTestId('foo')).toBeInTheDocument();
     });
 
     it('provides the showValidationErrors value to the children via FormAriaContext', function() {
       function Fixture() {
         const { showValidationErrors } = useContext(FormAriaContext);
 
-        return <span className="fixture">{showValidationErrors.toString()}</span>;
+        return <span data-testid="foo">{showValidationErrors.toString()}</span>;
```
EmilyBermudez88, line 87: Interesting work-around to follow RTL's 'toHaveTextContent' matcher. For future reference, is this the preferred way (instead of asserting .toBeTrue() ) when moving forward with RTL?

rpokorny: > Interesting work-around to follow RTL's 'toHaveTextContent' matcher. For future reference, is this the preferred way (instead of asserting .toBeTrue() ) when moving forward with RTL?

As far as I can tell `toBeTrue` isn't an option here.  If it was it's what I'd prefer.  But what I'm doing here is creating a child component that uses the context, and who's render output I can use to inspect the context values.  But the render is of course DOM and strings, so I have to assert the value as a string

EmilyBermudez88: > As far as I can tell `toBeTrue` isn't an option here. If it was it's what I'd prefer. But what I'm doing here is creating a child component that uses the context, and who's render output I can use to inspect the context values. But the render is of course DOM and strings, so I have to assert the value as a string

Asserting the value as a string was really clever. Thanks for the explanation

EmilyBermudez88 (approved):

PR #761, opened 2022-08-24
NxFormSelect Support for Validation Errors - RSC-1090

rpokorny:
https://issues.sonatype.org/browse/RSC-1090


lib/src/components/NxFormSelect/NxFormSelect.scss
```diff
@@ -0,0 +1,114 @@
+/*
```
rpokorny, line 1: Note: this file is mostly copied from `_nx-form-select.scss`, but also rearranged as needed for the new more complex structure

lib/src/__testutils__/rtlUtils.tsx
```diff
@@ -9,10 +9,17 @@ import { render, RenderResult } from '@testing-library/react';
...
+import { within } from '@testing-library/dom';
 
 export function rtlRender<P>(Component: ComponentType<P>, minimalProps: P) {
   return function renderWrapper(additionalProps?: Partial<P>): RenderResult {
-    return render(<Component { ...minimalProps } { ...additionalProps } />);
+    const renderResult = render(<Component { ...minimalProps } { ...additionalProps } />),
+        // counterintuitively, the Queries on the return value of `render` are NOT bound to the container
+        // of that render, but rather to the document. For our purposes it is much more helpful to
+        // have them on the render container
+        boundQueries = within(renderResult.container);
```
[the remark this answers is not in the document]
rpokorny: > By adding in boundQueries, we will be able to use querySelector without needing document.querySelector... ?

No, what this does is it means you can use `quickRender(...).getByRole()` and other queries even when using multiple `quickRender`s within the same test.  The default `getByRole` that's present on the RTL `render()` output actually searches the entire `document`, not just the container for that render.  So if you have multiple renders in the same test, their returned query methods can actually find elements from the other renders.  My code here replaces those document-wide query methods with a `getByRole` that only searches that render's container.  If you still want to search the whole document, you can use `screen`.

Generally you shouldn't be using `querySelector` in RTL tests.  The query methods that RTL provides facilitate the query patterns that it endorses.

lib/src/components/NxFormSelect/__tests__/NxFormSelect.test.tsx
```diff
@@ -5,45 +5,490 @@
...
 describe('NxFormSelect', function() {
   const minimalProps = { value: '', onChange: () => {} },
-      getShallow = getShallowComponent<Props>(NxFormSelect, minimalProps);
+      quickRender = rtlRender(NxFormSelect, minimalProps),
+      renderEl = rtlRenderElement(NxFormSelect, minimalProps);
 
-  it('renders a <select> with the nx-form-select class', function() {
-    expect(getShallow()).toMatchSelector('select.nx-form-select');
+  it('renders a <select> ', function() {
+    expect(quickRender().getByRole('combobox').tagName).toBe('SELECT');
```
EmilyBermudez88, line 22: do selects have an inherit role of combobox, or is it specified somewhere in NxFormSelect?

rpokorny: > do selects have an inherit role of combobox, or is it specified somewhere in NxFormSelect?

It's built in to HTML.  See https://www.w3.org/TR/html-aria/#docconformance

lib/src/components/NxFormSelect/__tests__/NxFormSelect.test.tsx
```diff
@@ -5,45 +5,490 @@
...
+
+          it('does not set aria-invalid on the select', function() {
+            expect(quickRender().getByRole('combobox')).not.toHaveAttribute('aria-invalid', 'true');
+          });
+        });
+      });
+    });
+
+    describe('when not pristine', function() {
+      const pristineMinimalProps = { ...nonValidatableMinimalProps, isPristine: true };
```
EmilyBermudez88, line 187: if it's not pristine, should isPristine be set to false?

gallery/src/components/NxFormSelect/NxFormSelectPage.tsx
```diff
@@ -45,11 +49,59 @@ const NxFormSelectPage = () =>
...
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>validatable</NxCode></NxTable.Cell>
+              <NxTable.Cell>boolean</NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell>false</NxTable.Cell>
+              <NxTable.Cell>
+                If true, this <NxCode>NxFormSelect</NxCode> is subject to validation, the result of which should be
+                passed in via the <NxCode>validationErrors</NxCode> prop, resulting in validation CSS classes being
+                applied (see below). If false, the NxTextInput is not considered to be subject to validation, the
+                <NxCode>validationErrors</NxCode> prop is ignored, and validation-related CSS classes
```
EmilyBermudez88, line 64:
```suggestion
               {' '} <NxCode>validationErrors</NxCode> prop is ignored, and validation-related CSS classes
```

lib/src/components/NxFormSelect/__tests__/NxFormSelect.test.tsx
rpokorny: > if it's not pristine, should isPristine be set to false?

Ah yes.  It wasn't failing because 1. this is the "when not validatable" section, and 2. the "when not validatable" props were also wrong. 28ef4ec78

gallery/src/components/NxFormSelect/NxFormSelectPage.tsx
rpokorny: d7be695ee

lib/src/__testutils__/rtlUtils.tsx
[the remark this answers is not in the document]
EmilyBermudez88: > No, what this does is it means you can use `quickRender(...).getByRole()` and other [queries](https://testing-library.com/docs/queries/about/) even when using multiple `quickRender`s within the same test. The default `getByRole` that's present on the RTL `render()` output actually searches the entire `document`, not just the container for that render. So if you have multiple renders in the same test, their returned query methods can actually find elements from each other. My code here replaces those document-wide query methods with query methods that only search that render's container. If you still want to search the whole document, you can use `screen`.
> 
Thanks for the explanation 👍

EmilyBermudez88 (approved):

PR #764, opened 2022-08-31
NxColorPicker validation - RSC-1091

rpokorny:
https://issues.sonatype.org/browse/RSC-1091


lib/src/components/NxColorPicker/__tests__/NxColorPicker.test.tsx
```diff
@@ -4,154 +4,130 @@
...
+  it('does nothing when an input is clicked with no onChange prop', async function() {
+    const user = userEvent.setup(),
+        view = quickRender(),
+        radio = view.getByRole('radio', { name: /turquoise/i });
 
-    component.find('input').filterWhere(input => input.prop('value') === 'turquoise').simulate('change');
-
-    expect(onChange).toHaveBeenCalledWith('turquoise');
+    await user.click(radio);
   });
```
cindyko226, line 85: Is this intended to have no expect?

rpokorny: > Is this intended to have no expect?

Yes, it simply tests that no exception is thrown

cindyko226: Out of curiosity, what might be the exception if there is no expectation?

rpokorny: > Out of curiosity, what might be the exception if there is no expectation?

The purpose of this test is just to verify that `onChange` is in fact optional and that the component doesn't blow up if you don't specify it.

cindyko226: I see, thanks for the explanation.

cindyko226 (approved):

PR #765, opened 2022-09-02
Improvement Day: Efficient Scrolling of Large Lists

rpokorny:
This PR adds a component, `NxScrollRender`, which allows very long lists within small-ish scrolling containers to render efficiently - only the list items that are actually in view at any given time, plus one or two on either side, are actually rendered, while a spacer elements above and below keep the scroll position intact.  Optionally, as the user scrolls, the _same_ DOM nodes can get reused for later items in the list for further efficiency.  Once this is merged, in addition to it being available to RSC consumers, I'll want to create another ticket to use it automatically within `NxTransferListHalf`.

Current Limitations:
* Only supports vertical scrolling
* Only supports DOM structures where the list items are direct children of the scroll container. This means it will not work with tables. Further improvements could be made on this aspect.
* Does not support children with inconsistent heights or heights that change after initial render


lib/src/components/NxScrollRender/NxScrollRender.tsx
```diff
@@ -0,0 +1,137 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { ReactElement, useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import { always, clamp, dec, identity, inc, isNil } from 'ramda';
+import useResizeObserver from '@react-hook/resize-observer';
+import { useThrottleCallback } from '@react-hook/throttle';
+
+import { Props } from './types';
```
cindyko226, line 12: Do we need export Props as well?

lib/src/components/NxScrollRender/NxScrollRender.tsx
```diff
@@ -0,0 +1,137 @@
...
+      parentRef = useRef<HTMLElement>(null),
+      leadingSpacerRef = useRef<HTMLElement | null>(null),
+      leadingSpacerRefFn = (e: HTMLElement | null) => { leadingSpacerRef.current = e; },
+      trailingSpacerRef = useRef<HTMLElement | null>(null),
+      trailingSpacerRefFn = (e: HTMLElement | null) => { trailingSpacerRef.current = e; },
+
+      [parentHeight, setParentHeight] = useState<number | null>(null),
+      [childHeight, setChildHeight] = useState<number | null>(null),
+      normalizedChildHeight = childHeight ?? 0,
+      renderedChildCount = parentHeight == null || childHeight == null ?
```
cindyko226, line 36: Out of curiosity, is there a reason using `==` here instead of `===`?

lib/src/components/NxScrollRender/NxScrollRender.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+      setParentHeight(parentRef.current.clientHeight);
+    }
+  }, []);
+
+  useResizeObserver(parentRef, useCallback(function({ contentBoxSize: [{ blockSize }] }) {
+    setParentHeight(blockSize);
+  }, []));
+
+  useEffect(updateRendering, [updateRendering]);
```
cindyko226, line 134: Just wondering do we need the `useEffect` here? Seems like the `updateRendering` is triggered when scrolling on the parent.

lib/src/components/NxScrollRender/NxScrollRender.tsx
rpokorny: > Do we need export Props as well?

6d3db5159

lib/src/components/NxScrollRender/NxScrollRender.tsx
rpokorny: > Out of curiosity, is there a reason using == here instead of ===?

`== null` is pretty much the only good use case for `==`.  It returns true if the value is `null` or `undefined`, and `false` otherwise.

lib/src/components/NxScrollRender/NxScrollRender.tsx
rpokorny: > Just wondering do we need the useEffect here? Seems like the updateRendering is triggered when scrolling on the parent.

It's needs to run not just during scrolling, but also on initial render in order to get the spacers sized correctly. If I comment out this line, the example starts with the scrollbar the wrong size.

rpokorny: Actually, I've just realized that the spacer heights are derived entirely from other pieces of state and props, and so do not need to be state themselves; they can just be local variables.  Implementing that change has removed the need to call `updateRendering` on first render. 01a02340d

lib/src/components/NxScrollRender/NxScrollRender.tsx
cindyko226: Thanks for the explanation.

lib/src/components/NxScrollRender/NxScrollRender.tsx
cindyko226: > It's needs to run not just during scrolling, but also on initial render in order to get the spacers sized correctly. If I comment out this line, the example starts with the scrollbar the wrong size.

I see, that was for the initial render purpose.

lib/src/components/NxScrollRender/NxScrollRender.tsx
```diff
@@ -0,0 +1,133 @@
...
+      {React.createElement(spacerEl ?? 'div', {
+        ref: trailingSpacerRefFn,
+        role: 'presentation',
+        className: 'nx-scroll-render__spacer',
+        style: { height: trailingSpacerHeight }
+      })}
+    </>
+  ) : null;
+
+  const updateRendering = useThrottleCallback(useCallback(function updateRendering() {
```
EmilyBermudez88, line 82: I'm just trying to wrap my head around this line with its multiple callbacks. Would you be able to clarify for me?

lib/src/components/NxScrollRender/NxScrollRender.tsx
```diff
@@ -0,0 +1,133 @@
...
+        topTooClose = topDifference < normalizedChildHeight,
+        topTooFar = topDifference > normalizedChildHeight * 2,
+        adjust =
+          topTooClose ? dec :
+          topTooFar ? inc :
+          identity,
+        clampFirstRenderedChildIdx =
+            renderedChildCount == null ? always(0) : clamp(0, childCount - renderedChildCount),
+        newFirstRenderedChildIdx = clampFirstRenderedChildIdx(
+            adjust(Math.floor(divOrZero(scrollTop, normalizedChildHeight)) - 2)
```
EmilyBermudez88, line 96: how does the -2 factor in here?

lib/src/components/NxScrollRender/NxScrollRender.tsx
rpokorny: > I'm just trying to wrap my head around this line with its multiple callbacks. Would you be able to clarify for me?

It's probably best explained by the [`useThrottleCallback` documentation](https://www.npmjs.com/package/@react-hook/throttle).  `useThrottleCallback` is described as
> A hook that invokes its callback at most X frames per second.

In the description of its first parameter, it notes
> This is the callback you want to throttle. You need to wrap closures/unstable callbacks in useCallback() so that they are stable, otherwise throttling will break between renders.

So the goal is to create a function that can be called multiple times, potentially across different renders of this component, but which will only execute with a given maximum frequency even if it is called more often than that maximum.  In order for `useThrottleCallback` to always return the same function instance, which is necessary to maintain that call-frequency management, it must always be given the same underlying function instance.  `useCallback` ensures this.  Without `useCallback`, each call of the `NxScrollRender` render function would pass a new function instance to `useThrottleCallback`, and so `useThrottleCallback` would return a new instance, effectively resetting the throttling timers.  `useCallback` ensures that instead of passing a new instance each time, we always pass the same instance, except when the dependencies declared in `useCallback`'s second parameter change.

lib/src/components/NxScrollRender/NxScrollRender.tsx
rpokorny: > how does the -2 factor in here?

Let's just do an example.  Let's say you have children that are 10px high, and you've just (instantaneously, for the sake of simplicity) scrolled from the top of the list down to where the top of the fifth child is exactly aligned with the top of the container.  Also let's say the container is 100px from the top of the viewport, is 70px tall, and there's a total of 20 children.  So here's how I believe our variables would work out:

`scrollTop`: 50
`parentTop`: 100
`leadingSpacerTop`: 50
`topDifference`: 50 (hmm... is this always going to be the same as `scrollTop`? I will double check and optimize if possible)
`topTooClose`: false (50 is not less than 10, so we do not need to add more children at the top)
`topTooFar`: true (50 is greater than 10 times 2, so we current have too many children at the top (and accordingly, not enough at the bottom))
`adjust`: increment function
`clampFirstRenderedChildIdx`: function that clamps between 0 and 13.  9 Because the total child count is 20, and the `renderedChildCount` is the number of visible children plus 2 (notionally, one on each end) which is 7.  See line 39 where `renderedChildCount` is computed.
`newFirstRenderedChildIdx`: first we divide 50 by 10 to get 5.  Then subtract 2 (which is what you're asking about). Then call the `adjust` function which increments to 3. Then that passes the `clamp` and 3 is the final result.  

Remember though, this is an array index, so this 3 refers to the _fourth_ child.  And remember in this example that the fifth child is the first one visible at the new scroll position.  So this result in one child outside of view above the visible area, then the five visible children, and then one child outside of view below the visible area.  The children outside of view on either end are so that as you scroll there's something there to see rather than just immediately exposing the spacer.

TL;DR: I think we subtract 1 because array indexes are zero-based, and 1 more because we want the first rendered child to be the one above the first one you can actually see.

lib/src/components/NxScrollRender/NxScrollRender.tsx
EmilyBermudez88: > So the goal is to create a function that can be called multiple times, potentially across different renders of this component, but which will only execute with a given maximum frequency even if it is called more often than that maximum. In order for `useThrottleCallback` to always return the same function instance, which is necessary to maintain that call-frequency management, it must always be given the same underlying function instance. `useCallback` ensures this. Without `useCallback`, each call of the `NxScrollRender` render function would pass a new function instance to `useThrottleCallback`, and so `useThrottleCallback` would return a new instance, effectively resetting the throttling timers. `useCallback` ensures that instead of passing a new instance each time, we always pass the same instance, except when the dependencies declared in `useCallback`'s second parameter change.

Thanks for this explanation, it helps a lot!

lib/src/components/NxScrollRender/NxScrollRender.tsx
EmilyBermudez88: > > how does the -2 factor in here?
> 
> Let's just do an example. Let's say you have children that are 10px high, and you've just (instantaneously, for the sake of simplicity) scrolled from the top of the list down to where the top of the fifth child is exactly aligned with the top of the container. Also let's say the container is 100px from the top of the viewport, is 70px tall, and there's a total of 20 children. So here's how I believe our variables would work out:
> 
> `scrollTop`: 50 `parentTop`: 100 `leadingSpacerTop`: 50 `topDifference`: 50 (hmm... is this always going to be the same as `scrollTop`? I will double check and optimize if possible) `topTooClose`: false (50 is not less than 10, so we do not need to add more children at the top) `topTooFar`: true (50 is greater than 10 times 2, so we current have too many children at the top (and accordingly, not enough at the bottom)) `adjust`: increment function `clampFirstRenderedChildIdx`: function that clamps between 0 and 11. 11 because the total child count is 20, and the `renderedChildCount` is the number of visible children plus 2 (notionally, one on each end) which is 9. See line 39 where `renderedChildCount` is computed. `newFirstRenderedChildIdx`: first we divide 50 by 10 to get 5. Then subtract 2 (which is what you're asking about). Then call the `adjust` function which increments to 3. Then that passes the `clamp` and 3 is the final result.
> 
> Remember though, this is an array index, so this 3 refers to the _fourth_ child. And remember in this example that the fifth child is the first one visible at the new scroll position. So this result in one child outside of view above the visible area, then the five visible children, and then one child outside of view below the visible area. The children outside of view on either end are so that as you scroll there's something there to see rather than just immediately exposing the spacer.
> 
> TL;DR: I think we subtract 1 because array indexes are zero-based, and 1 more because we want the first rendered child to be the one above the first one you can actually see.

Thanks for taking the time to put in such a detailed response, it's much appreciated. I was thinking about indexing, but my apparently tired brain wasn't counting the child outside of view. The rest of the explanation was also a huge help in solidifying what's going on.

gallery/src/components/NxScrollRender/NxScrollRenderPage.tsx
```diff
@@ -0,0 +1,189 @@
...
+          <NxH3>Caveats</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          There are certain limitations on the conditions in which <NxCode>NxScrollRender</NxCode> may be used,
+          spelled out below:
+        </NxP>
+        <NxList bulleted>
+          <NxList.Item>
+            <NxList.Text>
+              The immediate child of <NxCode>NxScrollRender</NxCode> must render an single container element, with
```
EmilyBermudez88, line 53:
```suggestion
              The immediate child of <NxCode>NxScrollRender</NxCode> must render a single container element, with
```

cindyko226 (approved):
+1

rpokorny:
> I see that there are no unit tests, and wondering if this is the case because NxScrollRender is simply a wrapper for other components with scrolling containers, and its key purpose is about how many children are rendered on the page (which is something that is extensively tested in the visual tests)? Thanks!

There are no unit tests because JSDOM doesn't implement actual DOM layout algorithms, and just returns 0 for any sort of sizes that you try to measure.  That makes it useless for a component like this.

PR #766, opened 2022-09-06
Fix NxToast a11yTest - RSC-1164

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1164

Visual testing was failing on the a11yTest with the following error: “Ensure all page content is contained by landmarks”. 

This was firing because `NxToastContainer` is positioned outside of any landmark elements (outside of header/aside/main/footer etc). It seems as though by supplying roles for all four variations of `NxAlert` (`NxWarningAlert` and `NxInfoAlert` do not have a pre-assigned role), this eliminated the error. There isn't any documentation that explicitly lays out why, but I believe this is because  “Live Regions” (in this case role=alert or role=status) will automatically move the screenreader to the content within the alert. Without roles defined (and outside of a landmark role), the screenreader may not jump to the content inside of the NxAlert.

### DIFF
--- a/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
+++ b/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
@@ -39,15 +39,19 @@ const sidebarLogoPath = require('../../assets/images/logo-plaid-villain-text.png
 
 interface ToastModel {
   id: number;
-  alertComponent: ComponentType;
+  alertComponent: ComponentType<alertComponentProps>;
   message: string;
 }
+
+type alertComponentProps = {
+  role?:string;
+};
 
 export default function NxToastComplexLayoutExample() {
   const [toastIdInc, setToastIdInc] = useState<number>(0);
   const [toasts, setToasts] = useState<ToastModel[]>([]);
 
-  const addToast = (alertComponent: ComponentType, message: string) => {
+  const addToast = (alertComponent: ComponentType<alertComponentProps>, message: string) => {
     const toastId = toastIdInc + 1;
     setToastIdInc(toastId);
     setToasts([
@@ -70,7 +74,11 @@ export default function NxToastComplexLayoutExample() {
             return (
               <NxToast key={id}
                        onClose={()=> removeToast(id)}>
-                <ToastAlert>{message}</ToastAlert>
+                {
+                  ToastAlert === NxWarningAlert || ToastAlert === NxInfoAlert ?
+                    <ToastAlert role= 'status'>{message}</ToastAlert> :
+                    <ToastAlert>{message}</ToastAlert>
+                }
               </NxToast>
             );
           })
--- a/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
+++ b/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
@@ -39,11 +39,11 @@ const sidebarLogoPath = require('../../assets/images/logo-plaid-villain-text.png
 
 interface ToastModel {
   id: number;
-  alertComponent: ComponentType<alertComponentProps>;
+  alertComponent: ComponentType<AlertComponentProps>;
   message: string;
 }
 
-type alertComponentProps = {
+type AlertComponentProps = {
   role?: string;
 };
 
@@ -51,7 +51,7 @@ export default function NxToastComplexLayoutExample() {
   const [toastIdInc, setToastIdInc] = useState<number>(0);
   const [toasts, setToasts] = useState<ToastModel[]>([]);
 
-  const addToast = (alertComponent: ComponentType<alertComponentProps>, message: string) => {
+  const addToast = (alertComponent: ComponentType<AlertComponentProps>, message: string) => {
     const toastId = toastIdInc + 1;
     setToastIdInc(toastId);
     setToasts([
--- a/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
+++ b/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
@@ -75,7 +75,7 @@ export default function NxToastComplexLayoutExample() {
               <NxToast key={id}
                        onClose={()=> removeToast(id)}>
                 {
-                  ToastAlert === NxWarningAlert || ToastAlert === NxInfoAlert ?
+                  ToastAlert === NxInfoAlert ?
                     <ToastAlert role="status">{message}</ToastAlert> :
                     <ToastAlert>{message}</ToastAlert>
                 }
--- a/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
+++ b/gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
@@ -44,7 +44,7 @@ interface ToastModel {
 }
 
 type alertComponentProps = {
-  role?:string;
+  role?: string;
 };
 
 export default function NxToastComplexLayoutExample() {
@@ -76,7 +76,7 @@ export default function NxToastComplexLayoutExample() {
                        onClose={()=> removeToast(id)}>
                 {
                   ToastAlert === NxWarningAlert || ToastAlert === NxInfoAlert ?
-                    <ToastAlert role= 'status'>{message}</ToastAlert> :
+                    <ToastAlert role="status">{message}</ToastAlert> :
                     <ToastAlert>{message}</ToastAlert>
                 }
               </NxToast>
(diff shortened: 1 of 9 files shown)


gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
```diff
@@ -39,15 +39,19 @@ const sidebarLogoPath = require('../../assets/images/logo-plaid-villain-text.png
 
 interface ToastModel {
   id: number;
-  alertComponent: ComponentType;
+  alertComponent: ComponentType<alertComponentProps>;
   message: string;
 }
 
+type alertComponentProps = {
```
rpokorny, line 46: Type names should use PascalCase (e.g. start with an uppercase letter)

rpokorny:
> but I believe this is because “Live Regions” (in this case role=alert or role=status) will automatically move the screenreader to the content within the alert.

I believe you're right that they have to be live regions.  I experimented with changing the role to other live region roles such as `log` and also to other non-live, non-landmark roles such as `generic` and `code`.  The test passed with `log` but failed with `generic` and `code`.

gallery/src/components/NxToast/NxToastPage.tsx
```diff
@@ -108,6 +108,16 @@ const NxToastPage = () =>
...
+                <NxWarningAlert>
+                  <NxCode>Nx*Alert</NxCode>s must be passed a {' '}
+                  <NxTextLink external href="https://www.w3.org/TR/wai-aria/#live_region_roles">
+                    Live Region
+                  </NxTextLink>
+                  {' '}role if not already provided (see
+                  {' '}<NxCode><NxTextLink href="#/pages/Alert">NxAlert</NxTextLink></NxCode>
+                  for more information about default roles). Without a designated role, the content inside
+                  {' '}<NxCode>NxToast</NxCode> will not be immediately read by screenreaders when rendered on the page.
+                </NxWarningAlert>
```
rpokorny, line 120: I'm wondering if alerts in `toasts` should just always use the `alert` role.  Regardless of the level of toast used, that role seems like the best fit for a component that pops up as the result of some event like this.  We could even have `NxToast` pass down the `alert` role automatically unless another role is specified by the caller

EmilyBermudez88: > I'm wondering if alerts in `toasts` should just always use the `alert` role. Regardless of the level of toast used, that role seems like the best fit for a component that pops up as the result of some event like this. We could even have `NxToast` pass down the `alert` role automatically unless another role is specified by the caller

That makes sense to me. With everything set to alert, aria-live will always be set to assertive and the screenreader will interrupt itself to read the toast message. As long as you're good with that, then I'll get started on changing it. So NxToast would pass it down to the alert in the NxToast.tsx file, correct?

rpokorny: > So NxToast would pass it down to the alert in the NxToast.tsx file, correct?

Yes, but it should probably done it in such a way that the if the caller sets a role on the `NxAlert` we keep that role instead

gallery/src/components/NxToast/NxToastComplexLayoutExample.tsx
EmilyBermudez88: > Type names should use PascalCase (e.g. start with an uppercase letter)

Updated here: [b746450](https://github.com/sonatype/sonatype-react-shared-components/pull/766/commits/b7464505042e11a619390940815b23e953f069d8)

gallery/src/components/NxToast/NxToastPage.tsx
EmilyBermudez88: > > So NxToast would pass it down to the alert in the NxToast.tsx file, correct?
> 
> Yes, but it should probably done it in such a way that the if the caller sets a role on the `NxAlert` we keep that role instead

Implemented here: [3d8d907](https://github.com/sonatype/sonatype-react-shared-components/pull/766/commits/3d8d907b09f057b1ac1feb0e3f23346e32a19ace). Since updating the role of NxAlert is not required, I removed AlertComponentProps in the 'Simple' and 'Legacy' examples.

EmilyBermudez88:
> I believe you're right that they have to be live regions. I experimented with changing the role to other live region roles such as `log` and also to other non-live, non-landmark roles such as `generic` and `code`. The test passed with `log` but failed with `generic` and `code`.

I tested marque and timer Live Regions roles (which do not have aria-live set to true)  and they failed, so I specified in the docs that the roles can be updated to "status" and "log" here:  [40b5e47](https://github.com/sonatype/sonatype-react-shared-components/pull/766/commits/40b5e472d8a2dfa33be42b6510b0287cfb54ec1a)

gallery/src/components/NxToast/NxToastPage.tsx
```diff
@@ -100,14 +100,26 @@ const NxToastPage = () =>
...
+                  {' '}<NxCode>Nx*Alert</NxCode> components themselves, but custom wrapping components are also
+                  permitted. Note that the calling code should not specify the <NxCode>onClose</NxCode> prop for the
+                  alert; it will be set up internally by <NxCode>NxToast</NxCode>. Calling code should instead use the
+                  {' '}<NxCode>onClose</NxCode> prop of <NxCode>NxToast</NxCode> itself. Additionally, note that the
+                  alert text content should be brief: only one rendered line of text per toast is supported.
+                </NxP>
+                <NxP>
+                  <NxCode>Nx*Alert</NxCode>s as children of <NxCode>NxToast</NxCode> will be assigned a default role
+                  of <NxCode>alert</NxCode>. The role can be updated to either <NxCode>status</NxCode> or
+                  {' '}<NxCode>log</NxCode>. One of these three {' '}
```
rpokorny, line 116: We shouldn't bring up `log`.  Even though it makes the test pass I can't see it ever making sense as the role for a toast.  At the same time, I'm seeing no documentation in the standards on the rules that we seem to be encountering here, and even the test tool seems to be a bit inconsistent about it.  As you've observed, `log` works while `marquee` does not.  Presumably, the difference is due to the default `aria-live` values that come with those roles.  However if that's the rule - that it must have an effective `aria-live` value other than `off`, then explcitly setting `aria-live` should also allow it to pass, regardless of the `role`.  In my testing however, it does not.  So at this point we well into the weeds of the undocumented and seemingly inconsistent logic of one particular testing tool - which is not something that we should base our documentation on.

I do still think that `alert` makes sense as the default role for all toast alerts. I think we should document that default, and that it can be overridden, and leave it at that.

lib/src/components/NxToast/NxToast.tsx
```diff
@@ -38,7 +38,8 @@ const NxToast = (props: NxToastProps) => {
   };
 
   const validChild = React.Children.only(children),
-      childrenWithProps = React.cloneElement(validChild, { onClose: handleClose });
+      childrenWithProps = React.cloneElement(validChild,
+          { onClose: handleClose, role: children.props.role ? children.props.role : 'alert' });
```
rpokorny, line 42:
```suggestion
          { onClose: handleClose, role: children.props.role ?? 'alert' });
```

Also the more typical formatting would be to break the object declaration into multiple lines like `classes` below

gallery/src/components/NxToast/NxToastPage.tsx
EmilyBermudez88: > I do still think that `alert` makes sense as the default role for all toast alerts. I think we should document that default, and that it can be overridden, and leave it at that.

Sounds good. Updated here: [8cbcf2d](https://github.com/sonatype/sonatype-react-shared-components/pull/766/commits/8cbcf2d165dc3c6dc8c33b35bcb15ae2170f0af9)

lib/src/components/NxToast/NxToast.tsx
EmilyBermudez88: > Also the more typical formatting would be to break the object declaration into multiple lines like `classes` below

Broken into multiple lines here: [3e74381](https://github.com/sonatype/sonatype-react-shared-components/pull/766/commits/3e743819fa73a174bc9e3ebce8a624cd610f9ea8)

lib/src/components/NxToast/NxToast.tsx
```diff
@@ -38,7 +38,10 @@ const NxToast = (props: NxToastProps) => {
   };
 
   const validChild = React.Children.only(children),
-      childrenWithProps = React.cloneElement(validChild, { onClose: handleClose });
+      childrenWithProps = React.cloneElement(validChild, {
+        onClose: handleClose,
+        role: children.props.role ?? 'alert'
```
rpokorny, line 43: Probably safer
```suggestion
        role: validChild.props.role ?? 'alert'
```

rpokorny (approved):

PR #768, opened 2022-09-08
NxColorPicker Visual Updates - RSC-999

rpokorny:
https://issues.sonatype.org/browse/RSC-999

This PR includes a complete change to how the color circles were implemented.  I was finding that the old implementation failed to center the hover/focus circles properly relative to the color circles themselves.  This was the case before but became much more noticeable with the new larger styling on the selected circles.  Switching to SVG resulted in a more precise/correct render.  One cool thing of note is that the RTL-based unit tests required no updates at all to accommodate that change.  this is fitting since the implementation of the circle appearance is an implementation detail.


EmilyBermudez88:
Looks good to me. Great idea to switch to svgs, and nice to see that a change in implementation hasn't affected the RTL testing 👍

EmilyBermudez88 (approved):

PR #769, opened 2022-09-09
Fix a11ytest calls - RSC-1134

rpokorny:
https://issues.sonatype.org/browse/RSC-1134

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.1.3",
+  "version": "11.1.4",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.1.1",
+  "version": "11.1.2",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


EmilyBermudez88:
Looking at all the implementations of a11yTests, I'm wondering  why the second set of parentheses are only needed when the a11yTest is using the async/await syntax?

rpokorny:
> Looking at all the implementations of a11yTests, I'm wondering why the second set of parentheses are only needed when the a11yTest is using the async/await syntax?

That's not exactly what determines it.  `a11yTest` itself, just like `simpleTest`, `hoverTest`, etc, is a function that returns a function.  That _returned_ function is the actual test code.  It's set up this way so you can use the `it('looks right', simpleTest(...))` syntax.  If however you need to do some extra setup where you need to define your own function to be the test itself, but you still want to use `a11yTest` (or `hoverTest` or whichever), then in your test function you need to invoke `a11yTest` to get its test code function, and then _invoke that function_.  Just retrieving the function does nothing.

EmilyBermudez88:
> > Looking at all the implementations of a11yTests, I'm wondering why the second set of parentheses are only needed when the a11yTest is using the async/await syntax?
> 
> That's not exactly what determines it. `a11yTest` itself, just like `simpleTest`, `hoverTest`, etc, is a function that returns a function. That _returned_ function is the actual test code. It's set up this way so you can use the `it('looks right', simpleTest(...))` syntax. If however you need to do some extra setup where you need to define your own function to be the test itself, but you still want to use `a11yTest` (or `hoverTest` or whichever), then in your test function you need to invoke `a11yTest` to get its test code function, and then _invoke that function_. Just retrieving the function does nothing.

Thanks for the clarification on that, it helps a lot

EmilyBermudez88 (approved):

PR #773, opened 2022-09-09
Dark Mode Main Text and Background Colors - RSC-1000 RSC-1009 RSC-1051 RSC-1059

rpokorny:
https://issues.sonatype.org/browse/RSC-1000
https://issues.sonatype.org/browse/RSC-1009
https://issues.sonatype.org/browse/RSC-1051
https://issues.sonatype.org/browse/RSC-1059

To test, either use the `prefers-color-scheme` manual override in the Rendering panel of the Chrome Dev Tools, or use a dark OS theme or browser theme.

### DIFF
--- a/gallery/src/components/NxTooltip/NxTooltipExample.scss
+++ b/gallery/src/components/NxTooltip/NxTooltipExample.scss
@@ -4,8 +4,14 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
+@use '~@sonatype/react-shared-components/scss-shared/nx-dark-mode-helpers';
+
 // Tooltip override classes should always be combined with `.nx-tooltip` to overcome the specificity of
 // the tooltip library's built-in styles
 .nx-tooltip.gallery-tooltip-example {
   color: tan;
+
+  @include nx-dark-mode-helpers.dark-mode {
+    color: sienna;
+  }
 }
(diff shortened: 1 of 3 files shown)


gallery/src/components/NxTooltip/NxTooltipExample.scss
```diff
@@ -4,8 +4,14 @@
...
  */
+@use '~@sonatype/react-shared-components/scss-shared/nx-dark-mode-helpers';
+
 // Tooltip override classes should always be combined with `.nx-tooltip` to overcome the specificity of
 // the tooltip library's built-in styles
 .nx-tooltip.gallery-tooltip-example {
   color: tan;
+
+  @include nx-dark-mode-helpers.dark-mode {
+    color: sienna;
```
EmilyBermudez88, line 15: where did the colour sienna come from? I'm trying to find it in sketch's TypeScale board and don't see it

rpokorny: > where did the colour sienna come from? I'm trying to find it in sketch's TypeScale board and don't see it

The purpose of this example is to show that you can customize the tooltip HTML and CSS.  The color selected here is arbitrary, as is the light-mode one (tan).  Note that these aren't even CSS variables, they are [CSS built-in color names](https://www.w3.org/wiki/CSS/Properties/color/keywords)

EmilyBermudez88: > > where did the colour sienna come from? I'm trying to find it in sketch's TypeScale board and don't see it
> 
> The purpose of this example is to show that you can customize the tooltip HTML and CSS. The color selected here is arbitrary, as is the light-mode one (tan). Note that these aren't even CSS variables, they are [CSS built-in color names](https://www.w3.org/wiki/CSS/Properties/color/keywords)

I figured that the colour was likely arbitrary, but wanted confirmation. Thanks!

EmilyBermudez88 (approved):

JasonDSwearingen:
+1

PR #774, opened 2022-09-09
Adjust NxAlert Height when close Button is not Present - RSC-1114

rpokorny:
https://issues.sonatype.org/browse/RSC-1114

I recommend viewing the `NxLoadError.scss` changes with "Hide whitespace"

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.1.3",
+  "version": "11.1.4",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.1.4",
+  "version": "11.1.5",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.1.5",
+  "version": "11.1.6",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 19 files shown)


EmilyBermudez88:
I'm reading through the ensureElement function and wondering if there are scenarios in which isReactText(content) is false, which would result in the content NOT being wrapped in a <span> element?

EmilyBermudez88 (approved):

rpokorny:
> I'm reading through the ensureElement function and wondering if there are scenarios in which isReactText(content) is false, which would result in the content NOT being wrapped in a span element?

I don't think that's an issue but you did help me realize that `ensureElement` did things other than what its name implied, which are needed in many other areas but which were not what we want here.  So I've renamed the existing `ensureElement ` (and all references to it) and created a new one that's what we need here. 32c8fa226

EmilyBermudez88:
I'm looking through both ensureElement and ensureStartEndElements and now have a follow up question. I threw a console.log in while trying to understand what elements `isReactText()` is pulling. The description for onClose falls within these params, but I'm confused  why it isn't wrapped in a span when it is passed as an argument to `ensureElement()`? It is an array, so wouldn't it fall under the second if statement of `Array.isArray(content)`?
Screen Shot 2022-09-12 at 4 24 37 PM
Screen Shot 2022-09-12 at 4 25 06 PM
Screen Shot 2022-09-12 at 4 25 47 PM

rpokorny:
> I'm looking through both ensureElement and ensureStartEndElements and now have a follow up question. I threw a console.log in while trying to understand what elements isReactText() is pulling. The description for onClose falls within these params, but I'm confused why it isn't wrapped in a span when it is passed as an argument to ensureElement()? It is an array, so wouldn't it fall under the second if statement of Array.isArray(content)?

I assume you are referring to the fact that `NxTableCell` passes its `children` through `ensureStartEndElements`.  While it is true that it does, if you look closely you'll see that the variable that holds (among other things) the result of that operation, `cellSortingContents`, is only used in the return value of `NxTableCell` when the cell is a sortable column header.  In the case that the `onClose` description cell falls into, the `children` are used directly and `cellSortingContents` is ignored.

EmilyBermudez88:
>I assume you are referring to the fact that NxTableCell passes its children through ensureStartEndElements. While it is true that it does, if you look closely you'll see that the variable that holds (among other things) the result of that operation, cellSortingContents, is only used in the return value of NxTableCell when the cell is a sortable column header. In the case that the onClose description cell falls into, the children are used directly and cellSortingContents is ignored.


thanks for the clarification!

EmilyBermudez88 (approved):

PR #775, opened 2022-09-12
Multiple arguments for HasValidationErrors - RSC-1108

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1108

### DIFF
--- a/gallery/src/jsUtilPages/ValidationUtils/ValidationUtilsPage.tsx
+++ b/gallery/src/jsUtilPages/ValidationUtils/ValidationUtilsPage.tsx
@@ -38,7 +38,7 @@ const ValidationUtilsPage = () =>
     <GalleryExampleTile title="hasValidationErrors" codeExamples={hasValidationErrorsExampleCode}>
       <NxCode>hasValidationErrors</NxCode> is a function that takes a series
       of <NxCode>ValidationErrors</NxCode> objects as arguments and returns whether
-      these <NxCode>ValidationErrors</NxCode> represent an error.
+      any of these <NxCode>ValidationErrors</NxCode> represents an error.
     </GalleryExampleTile>
 
     <GalleryExampleTile title="getFirstValidationError" codeExamples={getFirstValidationErrorExampleCode}>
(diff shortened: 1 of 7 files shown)


gallery/src/jsUtilPages/ValidationUtils/ValidationUtilsPage.tsx
```diff
@@ -36,8 +36,9 @@ const ValidationUtilsPage = () =>
     </GalleryExampleTile>
 
     <GalleryExampleTile title="hasValidationErrors" codeExamples={hasValidationErrorsExampleCode}>
-      <NxCode>hasValidationErrors</NxCode> is a function which returns whether
-      the specified <NxCode>ValidationErrors</NxCode> represents an error.
+      <NxCode>hasValidationErrors</NxCode> is a function that takes a series
+      of <NxCode>ValidationErrors</NxCode> objects as arguments and returns whether
+      these <NxCode>ValidationErrors</NxCode> represent an error.
```
rpokorny, line 41:
```suggestion
      any of these <NxCode>ValidationErrors</NxCode> represents an error.
```

rpokorny:
Needs unit tests

EmilyBermudez88:
> Needs unit tests

Added here: [6901b4a](https://github.com/sonatype/sonatype-react-shared-components/pull/775/commits/6901b4a1436b136af5b03b9a5dd8ae16888cb137). I added in a second test for multiple empty arrays, but not sure if it's necessary. I can delete it if you prefer.

rpokorny:
> Added here: [6901b4a](https://github.com/sonatype/sonatype-react-shared-components/pull/775/commits/6901b4a1436b136af5b03b9a5dd8ae16888cb137). I added in a second test for multiple empty arrays, but not sure if it's necessary. I can delete it if you prefer.

I think it needs a wider variety of tests actually.  Test different combinations of strings, arrays (empty and not), and `undefined` parameters.  These shouldn't all be separate `it`s but rather should probably be grouped into a single `it('returns true iff any of its arguments is or contains a string'`.  "iff" meaning "if and only if".

EmilyBermudez88:
> I think it needs a wider variety of tests actually. Test different combinations of strings, arrays (empty and not), and undefined parameters. These shouldn't all be separate its but rather should probably be grouped into a single it('returns true iff any of its arguments is or contains a string'. "iff" meaning "if and only if".


Added in more combinations here: [998dadc](https://github.com/sonatype/sonatype-react-shared-components/pull/775/commits/998dadc08034ff5a14be8adc20f7c0d54dafb1a6)

rpokorny (approved):

PR #776, opened 2022-09-12
Use NxTooltip for NxTransferList Reordering Tooltips - RSC-1113

rpokorny:
https://issues.sonatype.org/browse/RSC-1113

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.2.1",
+  "version": "11.2.2",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.2.0",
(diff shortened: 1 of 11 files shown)


JenniKobylski (approved):
Great!

EmilyBermudez88:
As I'm testing it out, I'm getting this error in the console: 
Screen Shot 2022-09-13 at 11 35 30 AM

rpokorny:
> As I'm testing it out, I'm getting this error in the console:

Yes, I'd noticed that during design PR review but hadn't gotten a chance to look into it yet.  Thanks.  Fixed in 2d885f6cf

EmilyBermudez88:
I noticed this morning that if you have something in focus that has a tooltip and then scroll through the item list, the tooltip doesn't disappear at the top of the `.nx-transfer-list__item-list`, but keeps drifting above it's container. 

Screen Shot 2022-09-14 at 9 06 55 AM

rpokorny:
> It this an edge case we wouldn't need to worry about?

That's normal for MUI tooltips, I don't think we should worry about it

EmilyBermudez88:
> > It this an edge case we wouldn't need to worry about?
> 
> That's normal for MUI tooltips, I don't think we should worry about it

Ok, if that's the case, then everything else looks good!

EmilyBermudez88 (approved):

PR #780, opened 2022-09-13
NxFileUpload and Textarea validation - RSC 1173

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1173

There appeared to be two issues within the formValidationErrors variable. The validationErrors of the textarea was not included, and the string 'A file is required' for NxFileUpload was only being read when isPristine was set to false. However, I assume inside a form, leaving NxFileUpload blank upon form submission should register an error and prevent the form from being submitted, whether isPristine is set to true or false. 

If this is the case, then we may need to reword the gallery docs for isPristine: 
"Only relevant when isRequired is true, this prop sets whether the input is currently pristine (has not yet had its value adjusted by the user), in which case the validation error is not shown."

### DIFF
--- a/gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
+++ b/gallery/src/components/NxFileUpload/NxFileUploadPage.tsx
@@ -85,9 +85,8 @@ const NxTextInputPage = () =>
               <NxTable.Cell>false</NxTable.Cell>
               <NxTable.Cell>false</NxTable.Cell>
               <NxTable.Cell>
-                Only relevant when <NxCode>isRequired</NxCode> is true, this prop sets whether the input is currently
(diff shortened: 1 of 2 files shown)


rpokorny:
> If this is the case, then we may need to reword the gallery docs for isPristine:

Feel free to update the docs in this PR.  I'd use the `isPristine` docs on the other pages that have been updated on this feature branch as a guide

EmilyBermudez88:
> Feel free to update the docs in this PR. I'd use the `isPristine` docs on the other pages that have been updated on this feature branch as a guide

updated here: [5b48a81](https://github.com/sonatype/sonatype-react-shared-components/pull/780/commits/5b48a817a9636e9d3dc6e24352554208c3a6296c)

rpokorny (approved):

PR #782, opened 2022-09-16
Create Dark Mode Modal - RSC-1171

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1171

I currently have two different methods of how to add the various classes to the `html` element.
Although the first option using `classnames()` seems a bit cleaner, I have not yet figured out a good way to prevent the initial classes the `html` element accepts (`nx-html` and `nx-html--page-scrolling`) from being overridden without hardcoding them into classnames. I'm not sure if this is the best idea to override what is currently laid out in the index.html file (even if the classes are the same).
The second / commented-out option has more lines of code, but is simpler by using `classList.add / remove` in a variety of scenarios. 
I can continue fine tuning either option if needed, but wanted an opinion of which approach I should focus first.

### DIFF
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -15,9 +15,11 @@ import {
   useToggle,
   NxFieldset,
   NxRadio,
-  NxP,
-  NxCode
+  NxTooltip,
+  NxFontAwesomeIcon
 } from '@sonatype/react-shared-components';
+
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
 
 const DarkModeModal = () => {
   const [showModal, setShowModal] = useState(false);
@@ -49,17 +51,14 @@ const DarkModeModal = () => {
               </h2>
             </header>
             <div className="nx-modal-content">
-              <NxP>
-                In order to enable dark mode, the class <NxCode>nx-html--enable-color-schemes</NxCode> must first
-                be added to the <NxCode>html</NxCode> element, as demonstrated by the toggle below. Additional
-                classes <NxCode>nx-html--dark-mode</NxCode> or <NxCode>nx-html--light-mode</NxCode> may also be added
-                to force either light or dark mode, as demonstrated by the radio buttons. However, using these
-                simultaneously will result in dark mode and is not recommended. If neither class is added,
-                prefers-color-scheme in the user’s settings will dictate the mode.
-              </NxP>
               <NxFormGroup label="Enable Theme Changes">
                 <NxToggle onChange={toggleEnableModeChange} isChecked={enableModeChange}>
                   Opt-in to Allow Theming
+                  <NxTooltip title="Opting in will allow you to make adjustments to the display theme. Failure
+                             to do so will always result in light mode. This can be done by adding the
+                             'nx-html--enable-color-schemes' class to the HTML element.">
+                    <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+                  </NxTooltip>
                 </NxToggle>
               </NxFormGroup>
               <NxFieldset label="Choose Your Mode">
@@ -69,6 +68,11 @@ const DarkModeModal = () => {
                          isChecked={mode === 'browserChoice'}
                          disabled = {!enableModeChange}>
                   Let Your Browser Color Preference Decide
+                  <NxTooltip title="If theme changes are permitted, the default theme will be dictated by
+                             your browser or OS theme choice. The prefers-color-scheme media query registers
+                             this choice.">
+                    <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+                  </NxTooltip>
                 </NxRadio>
                 <NxRadio name="mode"
                          value="dark"
@@ -76,6 +80,10 @@ const DarkModeModal = () => {
                          isChecked={mode === 'dark'}
                          disabled = {!enableModeChange}>
                   Dark Mode
+                  <NxTooltip title="Override the browser's default display theme to dark mode by adding the class
+                             'nx-html--dark-mode' to the HTML element.">
+                    <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+                  </NxTooltip>
                 </NxRadio>
                 <NxRadio name="mode"
                          value="light"
@@ -83,6 +91,10 @@ const DarkModeModal = () => {
                          isChecked={mode === 'light'}
                          disabled = {!enableModeChange}>
                   Light Mode
+                  <NxTooltip title="Override the browser's default display theme to light mode by adding the class
+                             'nx-html--light-mode' to the HTML element.">
+                    <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+                  </NxTooltip>
                 </NxRadio>
               </NxFieldset>
             </div>
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -28,13 +28,10 @@ const DarkModeModal = () => {
   const [enableModeChange, toggleEnableModeChange] = useToggle(false);
   const [mode, setMode] = useState<string | null>(null);
 
-  const { classList } = document.documentElement;
-
   useEffect(() => {
+    const { classList } = document.documentElement;
+
     classList.toggle('nx-html--enable-color-schemes', enableModeChange);
-  }, [enableModeChange]);
-
-  useEffect(() => {
     classList.toggle('nx-html--dark-mode', mode === 'dark' && enableModeChange);
     classList.toggle('nx-html--light-mode', mode === 'light' && enableModeChange);
   }, [enableModeChange, mode]);
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -15,7 +15,8 @@ import {
   NxRadio,
   NxTooltip,
   NxFontAwesomeIcon,
-  NxCode
+  NxCode,
+  useToggle
 } from '@sonatype/react-shared-components';
 
 import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
@@ -28,7 +29,7 @@ type Props = {
 
 const DarkModeModal = ({ onClose }: Props) => {
 
-  const [themingEnabledState, setThemingEnabledState] = useState(themingEnabled),
+  const [themingEnabledState, toggleThemingEnabledState] = useToggle(themingEnabled),
       [themeOverrideState, setThemeOverrideState] = useState(themeOverride);
 
   useEffect(function() {
@@ -50,7 +51,7 @@ const DarkModeModal = ({ onClose }: Props) => {
       </header>
       <div className="nx-modal-content">
         <NxFormGroup label="Enable Theme Changes">
-          <NxToggle onChange={() => setThemingEnabledState(!themingEnabledState)} isChecked={themingEnabledState}>
+          <NxToggle onChange={toggleThemingEnabledState} isChecked={themingEnabledState}>
             Opt-in to Allow Theming
             <NxTooltip title={
               <>Opting in will allow you to make adjustments to the display theme. Failure
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -4,7 +4,7 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import {
   NxButton,
@@ -30,10 +30,14 @@ const DarkModeModal = () => {
 
   const { classList } = document.documentElement;
 
-  classList.toggle('nx-html--enable-color-schemes', enableModeChange);
+  useEffect(() => {
+    classList.toggle('nx-html--enable-color-schemes', enableModeChange);
+  }, [enableModeChange]);
 
-  classList.toggle('nx-html--dark-mode', mode === 'dark' && enableModeChange);
-  classList.toggle('nx-html--light-mode', mode === 'light' && enableModeChange);
+  useEffect(() => {
+    classList.toggle('nx-html--dark-mode', mode === 'dark' && enableModeChange);
+    classList.toggle('nx-html--light-mode', mode === 'light' && enableModeChange);
+  }, [enableModeChange, mode]);
 
   return (
     <>
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -72,7 +72,7 @@ const DarkModeModal = () => {
 
   return (
     <>
-      <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Light / Dark Mode</NxButton>
+      <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Theme Settings</NxButton>
       { showModal &&
         <NxModal id="nx-modal-dark-mode-example"
                  role="alertdialog"
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -11,7 +11,6 @@ import {
   NxModal,
   NxFormGroup,
   NxToggle,
-  // useToggle,
   NxFieldset,
   NxRadio,
   NxTooltip,
@@ -27,8 +26,6 @@ type Props = {
   modalCloseHandler: () => void;
 };
 
-type themeMode = 'enabled' | 'light' | 'dark'| 'disabled';
-
 const DarkModeModal = (props:Props) => {
   const { modalCloseHandler } = props;
 
@@ -42,39 +39,6 @@ const DarkModeModal = (props:Props) => {
   useEffect(function() {
     setThemingEnabled(themingEnabledState);
   }, [themingEnabled]);
-
-  //const [mode, setMode] = useState<themeMode>('disabled');
-  //const [modeClassnames, setModeClassnames] = useState<string[]>([]);
-
-  //useEffect(() => {
-    //const { classList } = document.documentElement;
-
-    //if (classList.contains('nx-html--enable-color-schemes')) {
-      //setMode('enabled');
-      //if (classList.contains('nx-html--dark-mode')) {
-        //setMode('dark');
-      //}
-      //else if (classList.contains('nx-html--light-mode')) {
-        //setMode('light');
-      //}
-    //}
-    //else {
-      //setMode('disabled');
-    //}
-  //}, []);
-
-  //useEffect(() => {
-
-    //const classesArray = Array.from(classList);
-    //setModeClassnames(classesArray);
-
-  //}, [mode]);
-
-  //const handleThemeChange = () => setMode(mode !== 'disabled' ? 'disabled' : 'enabled');
-
-  //useEffect(() => {
-    //localStorage.setItem('classes', JSON.stringify(modeClassnames));
-  //}, [modeClassnames]);
 
   return (
     <NxModal id="nx-modal-dark-mode-example"
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -58,7 +58,7 @@ const DarkModeModal = ({ onClose }: Props) => {
                 to do so will always result in light mode. This operates by toggling the presence of
                 the <NxCode>nx-html--enable-color-schemes</NxCode> class.
               </>}>
-              <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+              <NxFontAwesomeIcon icon={faInfoCircle}/>
             </NxTooltip>
           </NxToggle>
         </NxFormGroup>
@@ -73,7 +73,7 @@ const DarkModeModal = ({ onClose }: Props) => {
               <>The default theme will be dictated by your browser or OS theme choice, which is
                 communicated to the application via the <NxCode>prefers-color-scheme</NxCode> media query
               </>}>
-              <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+              <NxFontAwesomeIcon icon={faInfoCircle}/>
             </NxTooltip>
           </NxRadio>
           <NxRadio name="mode"
@@ -86,7 +86,7 @@ const DarkModeModal = ({ onClose }: Props) => {
               <>Overrides the browser's default display theme to dark mode by adding the class
                 {' '}<NxCode>nx-html--dark-mode</NxCode> to the HTML element.
               </>}>
-              <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+              <NxFontAwesomeIcon icon={faInfoCircle}/>
             </NxTooltip>
           </NxRadio>
           <NxRadio name="mode"
@@ -99,7 +99,7 @@ const DarkModeModal = ({ onClose }: Props) => {
               <>Overrides the browser's default display theme to dark mode by adding the class
                 {' '}<NxCode>nx-html--light-mode</NxCode> to the HTML element.
               </>}>
-              <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+              <NxFontAwesomeIcon icon={faInfoCircle}/>
             </NxTooltip>
           </NxRadio>
         </NxFieldset>
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -44,7 +44,6 @@ const DarkModeModal = () => {
       <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Theme Settings</NxButton>
       { showModal &&
         <NxModal id="nx-modal-dark-mode-example"
-                 role="dialog"
                  onCancel={modalCloseHandler}
                  aria-labelledby="modal-form-header">
           <NxForm onSubmit={modalCloseHandler}
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -75,9 +75,9 @@ const DarkModeModal = () => {
       <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Theme Settings</NxButton>
       { showModal &&
         <NxModal id="nx-modal-dark-mode-example"
-                 role="alertdialog"
+                 role="dialog"
                  onCancel={modalCloseHandler}
-                 aria-label="NxModal to set light or dark mode">
+                 aria-labelledby="modal-form-header">
           <NxForm onSubmit={modalCloseHandler}
                   submitBtnText= "Save and Close">
             <header className="nx-modal-header">
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -20,7 +20,7 @@ import {
 } from '@sonatype/react-shared-components';
 import classnames from 'classnames';
 
-const DarkModeModel = () => {
+const DarkModeModal = () => {
   const [showModal, setShowModal] = useState(false);
   const modalCloseHandler = () => setShowModal(false);
 
@@ -87,9 +87,9 @@ const DarkModeModel = () => {
             </header>
             <div className="nx-modal-content">
               <NxP>
-                In order to enable dark mode, the class <NxCode>nx-html-enable-color-schemes</NxCode> must first
+                In order to enable dark mode, the class <NxCode>nx-html--enable-color-schemes</NxCode> must first
                 be added to the <NxCode>html</NxCode> element, as demonstrated by the toggle below. Additional
-                classes <NxCode>nx-html–dark-mode</NxCode> or <NxCode>nx-html–light-mode</NxCode> may also be added
+                classes <NxCode>nx-html--dark-mode</NxCode> or <NxCode>nx-html--light-mode</NxCode> may also be added
                 to force either light or dark mode, as demonstrated by the radio buttons. However, using these
                 simultaneously will result in dark mode and is not recommended. If neither class is added,
                 prefers-color-scheme in the user’s settings will dictate the mode.
@@ -130,4 +130,4 @@ const DarkModeModel = () => {
   );
 };
 
-export default DarkModeModel;
+export default DarkModeModal;
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -21,6 +21,8 @@ import {
 
 import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
 
+import { themingEnabled, themeOverride, setThemingEnabled, setThemeOverride } from '../themeClassManager';
+
 type Props = {
   modalCloseHandler: () => void;
 };
@@ -30,43 +32,49 @@ type themeMode = 'enabled' | 'light' | 'dark'| 'disabled';
 const DarkModeModal = (props:Props) => {
   const { modalCloseHandler } = props;
 
-  const [mode, setMode] = useState<themeMode>('disabled');
-  const [modeClassnames, setModeClassnames] = useState<string[]>([]);
+  const [themingEnabledState, setThemingEnabledState] = useState(themingEnabled),
+      [themeOverrideState, setThemeOverrideState] = useState(themeOverride);
 
-  useEffect(() => {
-    const { classList } = document.documentElement;
+  useEffect(function() {
+    setThemingEnabled(themingEnabledState);
+  }, [themingEnabled]);
 
-    if (classList.contains('nx-html--enable-color-schemes')) {
-      setMode('enabled');
-      if (classList.contains('nx-html--dark-mode')) {
-        setMode('dark');
-      }
-      else if (classList.contains('nx-html--light-mode')) {
-        setMode('light');
-      }
-    }
-    else {
-      setMode('disabled');
-    }
-  }, []);
+  useEffect(function() {
+    setThemingEnabled(themingEnabledState);
+  }, [themingEnabled]);
 
-  useEffect(() => {
-    const { classList } = document.documentElement;
+  //const [mode, setMode] = useState<themeMode>('disabled');
+  //const [modeClassnames, setModeClassnames] = useState<string[]>([]);
 
-    classList.toggle('nx-html--enable-color-schemes', mode !== 'disabled');
-    classList.toggle('nx-html--dark-mode', mode === 'dark');
-    classList.toggle('nx-html--light-mode', mode === 'light');
+  //useEffect(() => {
+    //const { classList } = document.documentElement;
 
-    const classesArray = Array.from(classList);
-    setModeClassnames(classesArray);
+    //if (classList.contains('nx-html--enable-color-schemes')) {
+      //setMode('enabled');
+      //if (classList.contains('nx-html--dark-mode')) {
+        //setMode('dark');
+      //}
+      //else if (classList.contains('nx-html--light-mode')) {
+        //setMode('light');
+      //}
+    //}
+    //else {
+      //setMode('disabled');
+    //}
+  //}, []);
 
-  }, [mode]);
+  //useEffect(() => {
 
-  const handleThemeChange = () => setMode(mode !== 'disabled' ? 'disabled' : 'enabled');
+    //const classesArray = Array.from(classList);
+    //setModeClassnames(classesArray);
 
-  useEffect(() => {
-    localStorage.setItem('classes', JSON.stringify(modeClassnames));
-  }, [modeClassnames]);
+  //}, [mode]);
+
+  //const handleThemeChange = () => setMode(mode !== 'disabled' ? 'disabled' : 'enabled');
+
+  //useEffect(() => {
+    //localStorage.setItem('classes', JSON.stringify(modeClassnames));
+  //}, [modeClassnames]);
 
   return (
     <NxModal id="nx-modal-dark-mode-example"
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -79,7 +79,7 @@ const DarkModeModal = () => {
                          isChecked={mode === 'dark'}
                          disabled = {!enableModeChange}>
                   Dark Mode
-                  <NxTooltip title="Override the browser's default display theme to dark mode by adding the class
+                  <NxTooltip title="Overrides the browser's default display theme to dark mode by adding the class
                              'nx-html--dark-mode' to the HTML element.">
                     <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
                   </NxTooltip>
--- a/gallery/src/DarkModeModal/DarkModeModal.tsx
+++ b/gallery/src/DarkModeModal/DarkModeModal.tsx
@@ -18,7 +18,6 @@ import {
   NxP,
   NxCode
 } from '@sonatype/react-shared-components';
-import classnames from 'classnames';
 
 const DarkModeModal = () => {
   const [showModal, setShowModal] = useState(false);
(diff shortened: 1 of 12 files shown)


rpokorny:
Something more along the lines of the second option I think is what we want - it's similar to how we've implemented `SectionScrollingWrapper` (which you'll want to make sure you don't break!).  However I think you could make the code more succinct, especially by using the two-parameter version of the [`classList.toggle`](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle) function

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+        'nx-html--light-mode': mode === 'light' && enableModeChange,
+        'nx-html--dark-mode': mode === 'dark' && enableModeChange
+      });
+
+  useEffect(()=> {
+    htmlElement.className = htmlClasses;
+  }, [enableModeChange, mode]);
+
+  useEffect(()=> {
+    enableModeChange ? setRadioDisabled(false) : setRadioDisabled(true);
```
rpokorny, line 43: Whether the radiogroup is disabled is strictly based on `enableModeChange`, so you don't want `radioDisabled` to be its own piece of state at all – it can just be a local variable

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+  //   }
+  //   else if (!enableModeChange) {
+  //     classList.remove('nx-html--enable-color-schemes', 'nx-html--dark-mode', 'nx-html--light-mode');
+  //     setRadioDisabled(true);
+  //   }
+  // }, [enableModeChange]);
+
+  return (
+    <>
+      <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Light / Dark Mode</NxButton>
```
rpokorny, line 75: How about something like "Theme Settings" for the button label?

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+      { showModal &&
+        <NxModal id="nx-modal-dark-mode-example"
+                 role="alertdialog"
```
rpokorny, line 78: The default `dialog` role seems more correct to this use case

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+                 onCancel={modalCloseHandler}
+                 aria-label="NxModal to set light or dark mode">
```
rpokorny, line 80: Use `aria-labelledby` to reference the `h2` instead of `aria-label`

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <NxForm onSubmit={modalCloseHandler}
+                  submitBtnText= "Save and Close">
+            <header className="nx-modal-header">
+              <h2 className="nx-h2" id="modal-form-header">
+                <span>Setting Preferences for Light or Dark Mode</span>
```
rpokorny, line 85:
```suggestion
                Setting Preferences for Light or Dark Mode
```

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+                In order to enable dark mode, the class <NxCode>nx-html--enable-color-schemes</NxCode> must first
+                be added to the <NxCode>html</NxCode> element, as demonstrated by the toggle below. Additional
+                classes <NxCode>nx-html--dark-mode</NxCode> or <NxCode>nx-html--light-mode</NxCode> may also be added
+                to force either light or dark mode, as demonstrated by the radio buttons. However, using these
+                simultaneously will result in dark mode and is not recommended. If neither class is added,
+                prefers-color-scheme in the user’s settings will dictate the mode.
+              </NxP>
+              <NxFormGroup label="Enable Changes">
+                <NxToggle onChange={toggleEnableModeChange} isChecked={enableModeChange}>
+                  Opt-in to Allow Changes
```
rpokorny, line 99: "Changes" seems a little bit meaningless here, how about "Theming"

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+              </h2>
+            </header>
+            <div className="nx-modal-content">
+              <NxP>
+                In order to enable dark mode, the class <NxCode>nx-html--enable-color-schemes</NxCode> must first
+                be added to the <NxCode>html</NxCode> element, as demonstrated by the toggle below. Additional
+                classes <NxCode>nx-html--dark-mode</NxCode> or <NxCode>nx-html--light-mode</NxCode> may also be added
+                to force either light or dark mode, as demonstrated by the radio buttons. However, using these
+                simultaneously will result in dark mode and is not recommended. If neither class is added,
+                prefers-color-scheme in the user’s settings will dictate the mode.
```
rpokorny, line 95: I'd like the explanation here to be understandable both to technical and non-technical people.  To do that, I think it should focus primarily on the conceptual mechanisms at play (e.g. that theming is opt-in and you always get light mode if you do not opt in, and also that when opted in your browser and/or OS theme choices determine the theme by default) but with brief mention of the technical implementations (classnames and media queries) as well.

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+              </NxP>
```
rpokorny, line 96: One idea I had about presenting the descriptions here was to use "information" icons that produce tooltips on hover for each of the different options, rather than explaining everything in a single paragraph at the top.

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -29,7 +30,9 @@ const PageHeader = () => (
       <span className="gallery-page-header__name">
         React Shared Component Library
       </span>
-
+      <div className="gallery-page-header__display-mode-btn">
```
rpokorny, line 33: Is the extra div needed?

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > How about something like "Theme Settings" for the button label?

Sounds good, replaced here: [2dd372b](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/2dd372bd8036da861ef9ef7d380a830a86e7abf2)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > The default `dialog` role seems more correct to this use case

changed to `dialog` here: [92a7597](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/92a75978913d80a5b8101f5aef71e4c7307c81e7)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > Use `aria-labelledby` to reference the `h2` instead of `aria-label`

changed to `aria-lablledby` here: [92a7597](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/92a75978913d80a5b8101f5aef71e4c7307c81e7)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > Whether the radiogroup is disabled is strictly based on `enableModeChange`, so you don't want `radioDisabled` to be its own piece of state at all – it can just be a local variable

I updated it here : [a91181a](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/a91181a9d5387764cb85f29a9d5bd4cb4b563dde). Since disabled is always the opposite of enableModeChange, I assigned the disabled attribute the opposite value of enableModeChange

EmilyBermudez88:
> Something more along the lines of the second option I think is what we want - it's similar to how we've implemented SectionScrollingWrapper (which you'll want to make sure you don't break!). However I think you could make the code more succinct, especially by using the two-parameter version of the [classList.toggle](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle) function

Thanks for pointing me towards the two-parameter version of classList.toggle, I didn't know this second parameter existed, and has really cleaned things up! I removed the useEffect hooks surrounding the toggles because the dependencies were identical to the conditions of the second parameter in classList.toggle. However, I can re-implement them if you think it's better to leave them. Code updated here: [d42799f](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/d42799fb0d782a63c6d14c3a93afd4d92e768362)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > "Changes" seems a little bit meaningless here, how about "Theming"

Updated here: [ba3a3df](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/ba3a3df0df0eecdf0fdaeb0141e6f465f8a44415)

gallery/src/PageHeader/PageHeader.tsx
EmilyBermudez88: > Is the extra div needed?

Nope, it works fine without. Removed here: [da74491](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/da7449165643e62842d939367922afd0b4c11f25)

gallery/src/DarkModeModal/DarkModeModal.tsx
rpokorny: > changed to aria-lablledby here: 92a7597

The id in question should really be more specific to this modal too

gallery/src/DarkModeModal/DarkModeModal.tsx
rpokorny: > changed to dialog here: [92a7597](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/92a75978913d80a5b8101f5aef71e4c7307c81e7)

My intent was that you remove `role` here entirely, as `dialog` is already the default

rpokorny:
> I removed the useEffect hooks surrounding the toggles because the dependencies were identical to the conditions of the second parameter in classList.toggle. However, I can re-implement them if you think it's better to leave them. Code updated here: [d42799f](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/d42799fb0d782a63c6d14c3a93afd4d92e768362)

It still needs to be in a `useEffect`.  All side-effects should be in `useEffect` as a general rule

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > I'd like the explanation here to be understandable both to technical and non-technical people. To do that, I think it should focus primarily on the conceptual mechanisms at play (e.g. that theming is opt-in and you always get light mode if you do not opt in, and also that when opted in your browser and/or OS theme choices determine the theme by default) but with brief mention of the technical implementations (classnames and media queries) as well.

As discussed in WW, the paragraph has been removed, and explanations implemented in tooltips here: [4e6b7de](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/4e6b7deaa4d5e26ea2c2c25a0722bbe7ee68cbc8)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > One idea I had about presenting the descriptions here was to use "information" icons that produce tooltips on hover for each of the different options, rather than explaining everything in a single paragraph at the top.

Tooltips implemented here: [4e6b7de](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/4e6b7deaa4d5e26ea2c2c25a0722bbe7ee68cbc8)

EmilyBermudez88:
>> I removed the useEffect hooks surrounding the toggles because the dependencies were identical to the conditions of the second parameter in classList.toggle. However, I can re-implement them if you think it's better to leave them. Code updated here: [d42799f](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/d42799fb0d782a63c6d14c3a93afd4d92e768362)

>It still needs to be in a useEffect. All side-effects should be in useEffect as a general rule

Re-implemented here: [afa9fb9](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/afa9fb9ab6107c532fcb23f7ffc5f98cb74e32b7)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > My intent was that you remove `role` here entirely, as `dialog` is already the default

Removed here: [3cfdc2c](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/3cfdc2c2d1a89c6225304e8474917681acdd9c19)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > The id in question should really be more specific to this modal too

Renamed here: [227bab8](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/227bab8fc6853c6e335f2877b23e1b86b482bee5)

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [enableModeChange, toggleEnableModeChange] = useToggle(false);
+  const [mode, setMode] = useState<string | null>(null);
+
+  const { classList } = document.documentElement;
+
+  useEffect(() => {
+    classList.toggle('nx-html--enable-color-schemes', enableModeChange);
+  }, [enableModeChange]);
+
+  useEffect(() => {
```
rpokorny, line 37: I'd just combine these
```suggestion
```

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
+
+const DarkModeModal = () => {
+  const [showModal, setShowModal] = useState(false);
+  const modalCloseHandler = () => setShowModal(false);
+
+  const [enableModeChange, toggleEnableModeChange] = useToggle(false);
+  const [mode, setMode] = useState<string | null>(null);
+
+  const { classList } = document.documentElement;
```
rpokorny, line 31: It would be cleaner with this line moved into the `useEffect` as well

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+                Setting Preferences for Light or Dark Mode
+              </h2>
+            </header>
+            <div className="nx-modal-content">
+              <NxFormGroup label="Enable Theme Changes">
+                <NxToggle onChange={toggleEnableModeChange} isChecked={enableModeChange}>
+                  Opt-in to Allow Theming
+                  <NxTooltip title="Opting in will allow you to make adjustments to the display theme. Failure
+                             to do so will always result in light mode. This can be done by adding the
+                             'nx-html--enable-color-schemes' class to the HTML element.">
```
rpokorny, line 62: I think we could phrase these in a way that makes it more clear that toggling the input will in fact make the change to the classname; the current phrasing almost makes it sound like the user could/should do something _in addition_ to toggling the toggle.  For instance, this one could say "This operates by toggling the presence of the nx-html--enable-color-schemes class"

rpokorny, line 62: Note that tooltip titles can be JSX.  I think it would be worth trying `NxCode` where appropriate within the tooltip to see how it looks.

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+                  </NxTooltip>
+                </NxRadio>
+                <NxRadio name="mode"
+                         value="dark"
+                         onChange={setMode}
+                         isChecked={mode === 'dark'}
+                         disabled = {!enableModeChange}>
+                  Dark Mode
+                  <NxTooltip title="Override the browser's default display theme to dark mode by adding the class
```
rpokorny, line 86:
```suggestion
                  <NxTooltip title="Overrides the browser's default display theme to dark mode by adding the class
```

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
+                  </NxTooltip>
+                </NxRadio>
+                <NxRadio name="mode"
+                         value="light"
+                         onChange={setMode}
+                         isChecked={mode === 'light'}
+                         disabled = {!enableModeChange}>
+                  Light Mode
+                  <NxTooltip title="Override the browser's default display theme to light mode by adding the class
```
rpokorny, line 97:
```suggestion
                  <NxTooltip title="Overrides the browser's default display theme to light mode by adding the class
```

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+              <NxFieldset label="Choose Your Mode">
+                <NxRadio name="mode"
+                         value="browserChoice"
+                         onChange={setMode}
+                         isChecked={mode === 'browserChoice'}
+                         disabled = {!enableModeChange}>
+                  Let Your Browser Color Preference Decide
+                  <NxTooltip title="If theme changes are permitted, the default theme will be dictated by
+                             your browser or OS theme choice. The prefers-color-scheme media query registers
+                             this choice.">
```
rpokorny, line 76: I think we can remove the conditional aspect of this sentence.  The other two radios aren't described conditionally, and IMO the fact that it's conditional is made clear by the radios becoming disabled when the Opt-In toggle is off.

rpokorny, line 76:
```suggestion
                             your browser or OS theme choice, which is communicated to the application via the 'prefers-color-scheme' media query.">
```

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -37,7 +37,7 @@
     min-width: max-content;
   }
 
-  &__version {
+  .nx-btn--tertiary {
```
rpokorny, line 40: In this case I think it would be better to select the button either by an id or a very specific classname: we wouldn't want this style to apply to any other hypothetical tertiary button in the page header.

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }, [enableModeChange]);
+
+  useEffect(() => {
+    classList.toggle('nx-html--dark-mode', mode === 'dark' && enableModeChange);
+    classList.toggle('nx-html--light-mode', mode === 'light' && enableModeChange);
+  }, [enableModeChange, mode]);
+
+  return (
+    <>
+      <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Theme Settings</NxButton>
```
rpokorny, line 44: It's a bit atypical to have the button that opens the modal be in the same component as the modal.  Let's move the button itself and the state it controls to the `PageHeader` component

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  </NxTooltip>
+                </NxToggle>
+              </NxFormGroup>
+              <NxFieldset label="Choose Your Mode">
+                <NxRadio name="mode"
+                         value="browserChoice"
+                         onChange={setMode}
+                         isChecked={mode === 'browserChoice'}
+                         disabled = {!enableModeChange}>
+                  Let Your Browser Color Preference Decide
```
rpokorny, line 73: Wrap the text for each of these in a span, then you won't need the CSS customization.  This is why we have a general rule not to mix bare text nodes and elements as siblings unless the elements are simple inline elements.

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }, [enableModeChange, mode]);
+
+  return (
+    <>
+      <NxButton variant="tertiary" onClick={() => setShowModal(true)}>Theme Settings</NxButton>
+      { showModal &&
+        <NxModal id="nx-modal-dark-mode-example"
+                 onCancel={modalCloseHandler}
+                 aria-labelledby="nx-modal-dark-mode-form-header">
+          <NxForm onSubmit={modalCloseHandler}
```
rpokorny, line 49: The behavior and presentation around when the settings take effect is inconsistent.  The fact that the footer button is a primary button, tied to a form submission, with text "Save and Close" would seem to imply that the settings in this modal do not take effect until you click that button and close the modal.  On the other hand, the use of `NxToggle` implies an immediate change, and in fact you have the logic set up to make the changes immediately.  So we need to rectify that difference, either by making the changes not take effect until the form is submitted (and also by using `NxCheckbox` instead of `NxToggle`), or we need to make it more clear that the settings don't wait for any sort of submission, probably by removing the `NxForm` wrapper entirely and instead creating a simple `NxFooter` with a secondary or tertiary button that's just labelled "Close".

rpokorny:
It'd be really great if this setting was remembered across refreshes.  Look into `localstorage` as a way to accomplish that.

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > I'd just combine these

moved into one here: [803b3a0](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/803b3a0a0d10d335b14f55ba3ce34acb4c51a022)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > It would be cleaner with this line moved into the `useEffect` as well

moved into the useEffect here: [803b3a0](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/803b3a0a0d10d335b14f55ba3ce34acb4c51a022)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > I think we could phrase these in a way that makes it more clear that toggling the input will in fact make the change to the classname; the current phrasing almost makes it sound like the user could/should do something _in addition_ to toggling the toggle. For instance, this one could say "This operates by toggling the presence of the nx-html--enable-color-schemes class"

Thanks for pointing that out, I hadn't thought about it that way. It has been changed here: [e43333b](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/e43333b0e5ea78c9c661c52d2485325d30fdd5c9)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: Browser choice text changed here: [3f5e143](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/3f5e1437e88f0edbdce9ae5bd7708d4676a6d0f4)

EmilyBermudez88: > I think we can remove the conditional aspect of this sentence. The other two radios aren't described conditionally, and IMO the fact that it's conditional is made clear by the radios becoming disabled when the Opt-In toggle is off.

Absolutely, changed here: [3f5e143](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/3f5e1437e88f0edbdce9ae5bd7708d4676a6d0f4)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > Note that tooltip titles can be JSX. I think it would be worth trying `NxCode` where appropriate within the tooltip to see how it looks.

 I did try it initially, but wasn't about it's appearance. Here's a screenshot of what it looks like inside the tooltip:
Screen Shot 2022-09-22 at 8 49 54 AM

rpokorny: > I did try it initially, but wasn't sure about if it looked ok. Here's a screenshot of what it looks like inside the tooltip:

Check with design

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > Wrap the text for each of these in a span, then you won't need the CSS customization. This is why we have a general rule not to mix bare text nodes and elements as siblings unless the elements are simple inline elements.

Spans added here: [033ffd8](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/033ffd88cf7728ebcbd52bb1eb3fddef3e7b99e7)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > The behavior and presentation around when the settings take effect is inconsistent. The fact that the footer button is a primary button, tied to a form submission, with text "Save and Close" would seem to imply that the settings in this modal do not take effect until you click that button and close the modal. On the other hand, the use of `NxToggle` implies an immediate change, and in fact you have the logic set up to make the changes immediately. So we need to rectify that difference, either by making the changes not take effect until the form is submitted (and also by using `NxCheckbox` instead of `NxToggle`), or we need to make it more clear that the settings don't wait for any sort of submission, probably by removing the `NxForm` wrapper entirely and instead creating a simple `NxFooter` with a secondary or tertiary button that's just labelled "Close".

Opted to get rid of the form, which made more sense IMO - [e7e6fad](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/e7e6fad4ba850da86da7c7a127234dab4dced720)

gallery/src/PageHeader/PageHeader.scss
EmilyBermudez88: > In this case I think it would be better to select the button either by an id or a very specific classname: we wouldn't want this style to apply to any other hypothetical tertiary button in the page header.

I added a classname here: [357cbcd](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/357cbcda72bcf2e87d6c013223fa333f5a32ee8e)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > It's a bit atypical to have the button that opens the modal be in the same component as the modal. Let's move the button itself and the state it controls to the `PageHeader` component

Sorry, this one came with a lot of changes. With the button moving to `PageHeader`, the settings inside the modal reset each time it was opened. So there is now an additional useEffect to check if various classnames are present, and set the mode accordingly. I removed state that controlled the toggling of `enableModeChange`, in favour of one `mode` state that has a `disabled` option as well.  [c5cda2c](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/c5cda2ca1dc81c9a3682c94dc0073bf3c1f749e0)

EmilyBermudez88:
> It'd be really great if this setting was remembered across refreshes. Look into `localstorage` as a way to accomplish that.

So I managed to get `localstorage` up and running, which I used to store the classnames to apply to the `html` element. I currently have the stored values being pulled down in Application.tsx so that the settings are applied on each page load. If there's a better place for that, I can move 'localStorage.getItem` elsewhere.
Implemented here:  [f59f81d](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/f59f81d3810be1825b254762db7ffa63077e4e44) and [3f9848d](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/3f9848dbbd57a40cde384ed08e5aa662b3b685e2)

rpokorny:
> So I managed to get localstorage up and running, which I used to store the classnames to apply to the html element. I currently have the stored values being pulled down in Application.tsx so that the settings are applied on each page load. If there's a better place for that, I can move localStorage.getItem elsewhere.
Implemented here: [f59f81d](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/f59f81d3810be1825b254762db7ffa63077e4e44) and [3f9848d](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/3f9848dbbd57a40cde384ed08e5aa662b3b685e2)

I got halfway through some suggested changes to move the localstorage management into its own file, but unfortunately I really have to leave now and will be out next week.  Therefore I've put those changes on this branch: https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-1171_create_dark_mode_toggles...RSC-1171_create_dark_mode_toggles_ross. I suspect you'll be able to finish them up.

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const DarkModeModal = (props:Props) => {
+  const { modalCloseHandler } = props;
+
+  const [themingEnabledState, setThemingEnabledState] = useState(themingEnabled),
+      [themeOverrideState, setThemeOverrideState] = useState(themeOverride);
+
+  useEffect(function() {
+    setThemingEnabled(themingEnabledState);
+  }, [themingEnabled, themingEnabledState]);
```
entroform, line 37:
```suggestion
  }, [themingEnabledState]);
```

EmilyBermudez88:
I noticed that the `nx-html--dark-mode` and `nx-html--light-mode` classes were not being removed when themingEnabled was set to false (aka toggle set to false), so I added in condition for when the classes are applied here : 
[f1e5038](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/f1e5038f715a1204b42e6c0c67055701244249d1)

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
+
+import { themingEnabled, themeOverride, setThemingEnabled, setThemeOverride } from '../themeClassManager';
+
+type Props = {
+  modalCloseHandler: () => void;
+};
+
+const DarkModeModal = (props:Props) => {
+  const { modalCloseHandler } = props;
```
entroform, line 30:
```suggestion
const DarkModeModal = ({ modalCloseHandler }: Props) => {
```

entroform, line 30: Actually, I wonder if it's a good idea to rename this prop to `onClose` so it matches with the other callback names.

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+             onCancel={modalCloseHandler}
+             aria-labelledby="nx-modal-dark-mode-header">
+      <header className="nx-modal-header">
+        <h2 className="nx-h2" id="nx-modal-dark-mode-header">
+          Setting Preferences for Light or Dark Mode
+        </h2>
+      </header>
+      <div className="nx-modal-content">
+        <NxFormGroup label="Enable Theme Changes">
+          <NxToggle onChange={()=> setThemingEnabledState(!themingEnabledState)} isChecked={themingEnabledState}>
```
entroform, line 54:
```suggestion
          <NxToggle onChange={() => setThemingEnabledState(!themingEnabledState)} isChecked={themingEnabledState}>
```

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { themingEnabled, themeOverride, setThemingEnabled, setThemeOverride } from '../themeClassManager';
+
+type Props = {
+  modalCloseHandler: () => void;
+};
+
+const DarkModeModal = (props:Props) => {
+  const { modalCloseHandler } = props;
+
+  const [themingEnabledState, setThemingEnabledState] = useState(themingEnabled),
```
entroform, line 32: I wonder if we could use `useToggle` on the themingEnabledState?

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -4,37 +4,52 @@
...
-
-      <span className="gallery-page-header__name">
-        React Shared Component Library
-      </span>
-
-      <div className="gallery-page-header__version">
-        <HostedVersionsSelect />
+const PageHeader = () => {
+  const [showModal, setShowModal] = useState(false);
+  const modalCloseHandler = () => setShowModal(false);
```
entroform, line 21: I wonder if we could use `useToggle` here too for `setShowModal`, then we can pass in the `toggleDarkModeModal` fn directly into DarkModeModal modalCloseHandler prop below

gallery/src/PageHeader/PageHeader.tsx
```diff
@@ -4,37 +4,52 @@
...
+               alt="" />
+          <img className="gallery-page-header__logotype-image"
+               src={tartanLogotype}
+               alt="Sonatype Tartan" />
+        </a>
+
+        <span className="gallery-page-header__name">
+          React Shared Component Library
+        </span>
+        <NxButton className="nx-btn--open-theme-settings"
```
entroform, line 39: Instead of a modifier `nx-btn--open-theme-settings` which implies changes in behaviour or state or is a variant of, and since this style is being used to adjust spacing within the page header, I wonder if we should change the class name to make it part of the page header:
Maybe something like `gallery-page-header__theme-settings-button` 🤔

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > Actually, I wonder if it's a good idea to rename this prop to `onClose` or `onCancel` so it matches with the other callback names.

I updated the naming here: [0877b05](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/0877b05ef19497dd25364eb2786a988810bf501b). I did keep the name 'onModalCloseHandler' inside pageHeader, since I think it helps distinguish what it does in the context of the page header. If you'd like me to change that one too, I can.

gallery/src/themeClassManager.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+import { includes } from 'ramda';
+
+const validThemeOverrides = [null, 'light', 'dark'] as const;
+type ThemeOverride = (typeof validThemeOverrides)[number];
+
+export let themingEnabled: boolean = localStorage.getItem('theming-enabled') === 'true',
+    themeOverride: ThemeOverride = getThemeOverride();
+
+function getThemeOverride(): ThemeOverride {
```
entroform, line 15: This is a super small nitpick, I wonder if we can reorder these functions so they don't seem arbitrarily ordered: Maybe something like 🤔 :

- getThemeOverride
- setThemingEnabled
- setThemingOverride
- updateHtmlClasses

or

- setThemingEnabled
- setThemingOverride
- getThemeOverride
- updateHtmlClasses

Let me know what you think!

entroform:
I wonder if we should rename `DarkModeModal` to `ThemeSettingsModal` or something like that?

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <div className="nx-modal-content">
+        <NxFormGroup label="Enable Theme Changes">
+          <NxToggle onChange={() => setThemingEnabledState(!themingEnabledState)} isChecked={themingEnabledState}>
+            Opt-in to Allow Theming
+            <NxTooltip title={
+              <>Opting in will allow you to make adjustments to the display theme. Failure
+                to do so will always result in light mode. This operates by toggling the presence of
+                the <NxCode>nx-html--enable-color-schemes</NxCode> class.
+              </>}>
+              <NxFontAwesomeIcon className="dark-mode-modal__info-icon" icon={faInfoCircle}/>
```
entroform, line 60: I wonder if this classname `dark-mode-modal__info-icon` is necessary here and below

gallery/src/PageHeader/PageHeader.tsx
EmilyBermudez88: > I wonder if we could use `useToggle` here too for `setShowModal`, then we can pass in the `toggleDarkModeModal` fn directly into DarkModeModal modalCloseHandler prop below

Makes sense and cleaned things up nicely. Updated here: [861f2c6](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/861f2c675042b530e22789e628967d0fbaa446c8)

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > I wonder if we could use useToggle on the themingEnabledState?

changed to useToggle here: [9e94c49](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/9e94c49f66560c7ce74771f733b5496b3582a7db)

gallery/src/PageHeader/PageHeader.tsx
EmilyBermudez88: > Instead of a modifier `nx-btn--open-theme-settings` which implies changes in behaviour or state or is a variant of, and since this style is being used to adjust spacing within the page header, I wonder if we should change the class name to make it part of the page header: Maybe something like `gallery-page-header__theme-settings-button` 🤔

That's a great point and something I hadn't thought of. Renamed here: [6118e55](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/6118e559989abac6643ed4a9d36d2e06069061bb)

gallery/src/DarkModeModal/DarkModeModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
+
+import { themingEnabled, themeOverride, setThemingEnabled, setThemeOverride } from '../themeClassManager';
+
+type Props = {
+  onClose: () => void;
+};
+
+const DarkModeModal = ({ onClose }: Props) => {
+
```
entroform, line 31:
```suggestion
```

gallery/src/DarkModeModal/DarkModeModal.tsx
EmilyBermudez88: > I wonder if this classname `dark-mode-modal__info-icon` is necessary here and below

Thanks for catching that. I thought I needed it when initially building, and forgot to remove. Removed here: [4780ec8](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/4780ec883c67eeff90b0d1148ab0ff658756fb01)

EmilyBermudez88:
> I wonder if we should rename DarkModeModal to ThemeSettingsModal or something like that, let me know what you think!

Im good with that! Renamed here: 
[afa1076](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/afa1076e5726f9e50cbe2895df313e49deb28ab9)

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+  NxFontAwesomeIcon,
+  NxCode,
+  useToggle
+} from '@sonatype/react-shared-components';
+
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
+
+import { themingEnabled, themeOverride, setThemingEnabled, setThemeOverride } from '../themeClassManager';
+
+type Props = {
```
entroform, line 26: Small thing but we could also move this into its own `types.ts` to follow our pattern

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+  useEffect(function() {
+    setThemingEnabled(themingEnabledState);
+  }, [themingEnabledState]);
+
+  useEffect(function() {
+    setThemeOverride(themeOverrideState);
+  }, [themeOverrideState]);
+
+  return (
+    <NxModal id="nx-modal-dark-mode-example"
```
entroform, line 43: I'm not sure if this id is needed.

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+             onCancel={onClose}
+             aria-labelledby="nx-modal-dark-mode-header">
+      <header className="nx-modal-header">
+        <h2 className="nx-h2" id="nx-modal-dark-mode-header">
```
entroform, line 47: for consistency I wonder if we should also rename this to something like `gallery-theme-settings-modal-header`

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
EmilyBermudez88: > I'm not sure if this id is needed.

My thought was to have an id to make it easier for testing as an easy selector. However, I can remove it if you prefer!

entroform: > > I'm not sure if this id is needed.
> 
> My thought was to have an id to make it easier for testing as an easy and specific selector. However, I can remove it if you prefer!

Oo hmm, I wonder if we could remove this id and add a classname instead, maybe something like `gallery-theme-settings-modal`

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
EmilyBermudez88: > for consistency I wonder if we should also rename this to something like `gallery-theme-settings-modal-header`

Thanks for clarifying about the prefix. Changed here: [af68c48](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/af68c4892214e891dcd143b2b0fa1fc4c71c04e2)

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
EmilyBermudez88: > > > I'm not sure if this id is needed.
> > 
> > 
> > My thought was to have an id to make it easier for testing as an easy and specific selector. However, I can remove it if you prefer!
> 
> Oo hmm, I wonder if we could remove this id and add a classname instead, maybe something like `gallery-theme-settings-modal`

Changed here: [af68c48](https://github.com/sonatype/sonatype-react-shared-components/pull/782/commits/af68c4892214e891dcd143b2b0fa1fc4c71c04e2)

gallery/src/ThemeSettingsModal/ThemeSettingsModal.tsx
entroform: Ops nvm these aren't entirely consistent in the gallery.

gallery/src/themeClassManager.ts
```diff
@@ -0,0 +1,46 @@
...
+  return includes(storedString, validThemeOverrides) ? storedString as ThemeOverride : null;
+}
+
+function updateHtmlClasses() {
+  const { classList } = document.documentElement;
+
+  classList.toggle('nx-html--enable-color-schemes', themingEnabled);
+  classList.toggle('nx-html--dark-mode', themeOverride === 'dark' && themingEnabled);
+  classList.toggle('nx-html--light-mode', themeOverride === 'light' && themingEnabled);
+}
```
entroform, line 45:
```suggestion
}

```

entroform (approved):
Thanks! LGTM 🚀

gallery/src/PageHeader/PageHeader.scss
```diff
@@ -37,7 +37,7 @@
     min-width: max-content;
   }
 
-  &__version {
+  .gallery-page-header__theme-settings-button {
```
entroform, line 40:
```suggestion
  &__theme-settings-button {
```

PR #783, opened 2022-09-19
Add Submit Mask Timing Variable Link in NxForm Page- RSC-1146

rjp99:
https://issues.sonatype.org/browse/RSC-1146

### DIFF
--- a/gallery/src/components/NxForm/NxFormPage.tsx
+++ b/gallery/src/components/NxForm/NxFormPage.tsx
@@ -180,11 +180,11 @@ const NxFormPage = () =>
             <NxTable.Cell>null</NxTable.Cell>
             <NxTable.Cell>
               If null, the submit mask is not shown. If false, the mask is shown in its "submitting" state. If
-              true, the mask is shown in its "success" state. The amount of time the success mask is shown
-              can be changed and is further described in{' '}
+              true, the mask is shown in its "success" state. The amount of time the success mask is shown should be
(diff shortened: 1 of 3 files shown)


gallery/src/components/NxForm/NxFormPage.tsx
```diff
@@ -180,7 +180,11 @@ const NxFormPage = () =>
             <NxTable.Cell>null</NxTable.Cell>
             <NxTable.Cell>
               If null, the submit mask is not shown. If false, the mask is shown in its "submitting" state. If
-              true, the mask is shown in its "success" state.
+              true, the mask is shown in its "success" state. The amount of time the success mask is shown
+              can be changed and is further described in{' '}
```
rpokorny, line 184: We don't want to say it can be changed; the whole idea is for it to be consistent.  I'd say that the amount of time the success mask is shown should be based on the `SUBMIT_MASK_SUCCESS_VISIBLE_TIME_MS` variable as described on the `NxSubmitMask` page.

rjp99: > We don't want to say it can be changed; the whole idea is for it to be consistent. I'd say that the amount of time the success mask is shown should be based on the `SUBMIT_MASK_SUCCESS_VISIBLE_TIME_MS` variable as described on the `NxSubmitMask` page.

Updated here: 1a47cee2a0626388dc151ab7a970348dba56b2ad

PR #787, opened 2022-09-20
Fixed NxDrawer Focus Race Condition - RSC-1176

entroform:
https://issues.sonatype.org/browse/RSC-1176

### DIFF
--- a/gallery/visualtests/nxDrawer.js
+++ b/gallery/visualtests/nxDrawer.js
@@ -17,7 +17,7 @@ describe('NxDrawer', function() {
     await waitAndGetElements(`#${drawerId}`);
 
     // Wait for animation and focus.
-    await wait(300);
+    await wait(500);
   };
 
   describe('NxDrawer with Title, Footer, and non-overflowing Content', function() {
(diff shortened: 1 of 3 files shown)


gallery/visualtests/nxDrawer.js
```diff
@@ -16,8 +16,8 @@ describe('NxDrawer', function() {
 
     await waitAndGetElements(`#${drawerId}`);
 
-    // Wait for animation.
-    await wait(200);
+    // Wait for animation and focus.
+    await wait(300);
```
rpokorny, line 20: I'd rather have a wider margin.  If it failed sometimes at 200, I feel like it might still occasionally fail at 300. Let's bump it up to 500.

entroform: > I'd rather have a wider margin. If it failed sometimes at 200, I feel like it might still occasionally fail at 300. Let's bump it up to 500.

Extended it to 500 here: b9fd128d795458fa31b68ac18d0208c3275ef686

rpokorny (approved):

PR #790, opened 2022-09-21
NxToast with NxDrawer Example - RSC-1166

rjp99:
https://issues.sonatype.org/browse/RSC-1166


gallery/src/components/NxDrawer/NxDrawerPage.tsx
```diff
@@ -328,6 +329,35 @@ export default function NxDrawerPage() {
...
+        <NxTile.Header>
+          <NxTile.HeaderTitle>
+            <NxH2>NxDrawer With NxToast Example</NxH2>
+          </NxTile.HeaderTitle>
+        </NxTile.Header>
+        <NxTile.Content>
+          <NxP>
+            An example of <NxCode>NxDrawer</NxCode> with <NxCode>NxToast</NxCode>. If a toast is already open and
+            then a drawer is opened the toast appears on top of the drawer. This is the same if the toast is opened
+            from a button inside the drawer. Both the drawer and toast will close if a toast is closed.
```
EmilyBermudez88, line 340: Since you mention both scenarios, maybe include a toast button outside of the drawer inside NxPageMain to see both behaviours

EmilyBermudez88:
I would add an example of how the two interact on the Toast page as well. It's positioning above the drawer is noted in Kaleido's Toast docs, so having it accessible on the toast page would be beneficial as well in my opinion.

gallery/visualtests/nxDrawer.js
```diff
@@ -113,6 +113,34 @@ describe('NxDrawer', function() {
...
+    const browserSetup = setupBrowser('#/NxDrawerWithNxToastExample', false);
+    const { getPage, checkFullPageScreenshot, waitAndGetElements, wait } = browserSetup;
+    const openDrawer = openDrawerFromPage(browserSetup);
+
+    const buttonId = 'nx-drawer-with-nx-toast-open-button';
+    const drawerId = 'nx-drawer-with-nx-toast';
+
+    const launchToastsFromPage = async () => {
+
+      const [openToastBtn] = await waitAndGetElements('.nx-btn-bar .nx-btn--secondary');
```
EmilyBermudez88, line 126: you should be able to remove .nx-btn-bar since the openToastBtn is not contained within a bar like the other NxToast examples

gallery/src/components/NxDrawer/NxDrawerPage.tsx
```diff
@@ -326,6 +327,35 @@ export default function NxDrawerPage() {
...
```
rpokorny, line 340: > Both the drawer and toast will close if a toast is closed.

That doesn't sound desirable

gallery/src/components/NxDrawer/NxDrawerPage.tsx
rjp99: > Since you mention both scenarios, maybe include a toast button outside of the drawer inside NxPageMain to see both behaviours

Added it here: abf354e3be917cf7923530d67ed955b645ea326a

gallery/src/components/NxDrawer/NxDrawerPage.tsx
rpokorny: > That doesn't sound desirable

Nor does it seem to be accurate as I try it out

gallery/visualtests/nxDrawer.js
```diff
@@ -113,6 +113,28 @@ describe('NxDrawer', function() {
...
+    const { getPage, checkFullPageScreenshot, waitAndGetElements, wait } = browserSetup;
+    const openDrawer = openDrawerFromPage(browserSetup);
+
+    const buttonId = 'nx-drawer-with-nx-toast-open-button';
+    const drawerId = 'nx-drawer-with-nx-toast';
+
+    it('is positioned correctly', async function() {
+      await getPage().setViewport(viewportSize);
+      const [openSuccessToastButton] = await waitAndGetElements(`#${'nx-toast-success-open-button'}`);
+      const [openErrorToastButton] = await waitAndGetElements(`#${'nx-toast-error-open-button'}`);
```
rpokorny, line 127:
```suggestion
      const [openSuccessToastButton] = await waitAndGetElements('#nx-toast-success-open-button');
      const [openErrorToastButton] = await waitAndGetElements('#nx-toast-error-open-button');
```

gallery/src/components/NxDrawer/NxDrawerPage.tsx
EmilyBermudez88: > > That doesn't sound desirable
> 
> Nor does it seem to be accurate as I try it out

I just tried it out and it does happen for me. If a drawer and toast are both present, when I close a toast, it closes the drawer too. Does that happen for you as well?  I assume this has something to do with the layout of NxDrawer, when you click outside of the drawer, it automatically closes it?

rpokorny: > I just tried it out and it does happen for me. If a drawer and toast are both present, when I close a toast, it closes the drawer too. Does that happen for you as well?

Hmm it does now; I thought it didn't yesterday.  You're right that it's not actually about closing the toast, it's about clicking on the toast at all.  I think that bears checking with Design.  Since the toasts are in front of the drawer, it doesn't feel intuitive that clicking on them would close the drawer.

rjp99:
> I would add an example of how the two interact on the Toast page as well. It's positioning above the drawer is noted in Kaleido's Toast docs, so having it accessible on the toast page would be beneficial as well in my opinion.

I added it here: 59b856f82f35c41b5864c82573cf902f6bdab3ad

EmilyBermudez88:
>>I would add an example of how the two interact on the Toast page as well. It's positioning above the drawer is noted in Kaleido's Toast docs, so having it accessible on the toast page would be beneficial as well in my opinion.
> 59b856f

I know you're still working through everything, just make sure to update the descriptions on both examples

gallery/src/components/NxDrawer/NxDrawerPage.tsx
rjp99: > > I just tried it out and it does happen for me. If a drawer and toast are both present, when I close a toast, it closes the drawer too. Does that happen for you as well?
> 
> Hmm it does now; I thought it didn't yesterday. You're right that it's not actually about closing the toast, it's about clicking on the toast at all. I think that bears checking with Design. Since the toasts are in front of the drawer, it doesn't feel intuitive that clicking on them would close the drawer.

Checked with design and decided to make an exception to not close the drawer when toasts are interacted with: 58b5e5ccbbb3a0f40853646e14d9039260c70626

rjp99:
> > > I would add an example of how the two interact on the Toast page as well. It's positioning above the drawer is noted in Kaleido's Toast docs, so having it accessible on the toast page would be beneficial as well in my opinion.
> > > [59b856f](https://github.com/sonatype/sonatype-react-shared-components/commit/59b856f82f35c41b5864c82573cf902f6bdab3ad)
> 
> I know you're still working through everything, just make sure to update the descriptions on both examples

Thanks for reminding me: c8cbda78d94ccd3cc14745ca6bd146975fa872a2

EmilyBermudez88:
> > I know you're still working through everything, just make sure to update the descriptions on both examples
> 
> Thanks for reminding me: [c8cbda7](https://github.com/sonatype/sonatype-react-shared-components/commit/c8cbda78d94ccd3cc14745ca6bd146975fa872a2)

How about something like "The toast is always positioned above NxDrawer, no matter the order in which they are opened" ? Just to shorten the sentences a bit, and since toasts aren't necessarily opened by buttons outside of the RSC I believe

gallery/visualtests/nxDrawer.js
```diff
@@ -134,18 +134,28 @@ describe('NxDrawer', function() {
...
       await openDrawer(buttonId, drawerId);
+
+      const [openErrorToastButton] = await waitAndGetElements('#nx-toast-error-open-button');
+
       await openErrorToastButton.click();
 
-      const toast = await waitAndGetElements('.nx-toast-container .nx-toast');
+      const [toast] = await waitAndGetElements('.nx-toast');
+
+      await wait(500);
```
EmilyBermudez88, line 158: You may not need this wait(500) at all, since the toast will render immediately, but will slowly animate to its position (I could be wrong!). If you do need it, I'd put it one line above, after openErrorToastButton.click()

gallery/visualtests/nxDrawer.js
```diff
@@ -134,18 +134,28 @@ describe('NxDrawer', function() {
...
 
-    it('it stays open when the toast is closed', async function() {
+    it('stays open when the toast is closed', async function() {
       await getPage().setViewport(viewportSize);
-      const [openErrorToastButton] = await waitAndGetElements('#nx-toast-error-open-button');
-      const drawer = await waitAndGetElements('#nx-drawer-with-nx-toast');
+
+      const [drawer] = await waitAndGetElements('#nx-drawer-with-nx-toast');
 
       await openDrawer(buttonId, drawerId);
```
EmilyBermudez88, line 152: Just my preference, but maybe put this above the waitAndGetElements for `nx-drawer-with-nx-toast`. IMO it makes more sense for readability to open the drawer, then get the element. But works either way

gallery/visualtests/nxDrawer.js
rjp99: > 

I tried running it without the wait, but the the expect for if the toast isVisible fails. I moved that line up here: 95752e9a5cf87abd4d87c552048814cf10dc28f3

gallery/visualtests/nxDrawer.js
rjp99: > Just my preference, but maybe put this above the waitAndGetElements for `nx-drawer-with-nx-toast`. IMO it makes more sense for readability to open the drawer, then get the element. But works either way

That makes sense: 95752e9a5cf87abd4d87c552048814cf10dc28f3

EmilyBermudez88:
> > How about something like "The toast is always positioned above NxDrawer, no matter the order in which they are opened" ? Just to shorten the sentences a bit, and since toasts aren't necessarily opened by buttons outside of the RSC I believe
> 
> That makes much more sense: [b237ae0](https://github.com/sonatype/sonatype-react-shared-components/commit/b237ae0344c14e8b01655003e9fbdb21e4538ebd)

Sorry, if you'd like you can keep the drawer not closing when a toast is closed (I was thinking about the two sentences about positioning). Or you could just add it back in for NxDrawer since that behaviour is important to know? Sorry for the confusion there.

rjp99:
> > > How about something like "The toast is always positioned above NxDrawer, no matter the order in which they are opened" ? Just to shorten the sentences a bit, and since toasts aren't necessarily opened by buttons outside of the RSC I believe
> > 
> > 
> > That makes much more sense: [b237ae0](https://github.com/sonatype/sonatype-react-shared-components/commit/b237ae0344c14e8b01655003e9fbdb21e4538ebd)
> 
> Sorry, if you'd like you can keep the drawer not closing when a toast is closed (I was thinking about the two sentences about positioning). Or you could just add it back in for NxDrawer since that behaviour is important to know? Sorry for the confusion there.

Yeah, I think it would be useful to keep that part for the drawer page: 000a0fe3b2a4d07bd0e54f0201877896be27c9b5

gallery/src/components/NxToast/NxToastPage.tsx
```diff
@@ -252,6 +253,34 @@ const NxToastPage = () =>
...
+          </NxTextLink>
+        </NxP>
+      </NxTile.Content>
+      <NxTile.Content className= "nx-tile-content--accordion-container">
+        <NxStatefulAccordion>
+          <NxAccordion.Header>
+            <NxAccordion.Title>Example Code</NxAccordion.Title>
+          </NxAccordion.Header>
+          <CodeExample content={NxDrawerWithNxToastExampleCode} />
+          <GalleryTileFooter clipboardContent={NxDrawerWithNxToastExampleCode}/>
```
EmilyBermudez88, line 280: Because the same NxDrawerWithNxToastExample is being used for Toast and Drawer, the back button brings you back to the NxDrawer page, no matter if you are looking at the home page from NxToast or NxDrawer

rjp99: > Because the same NxDrawerWithNxToastExample is being used for Toast and Drawer, the back button brings you back to the NxDrawer page, no matter if you are looking at the home page from NxToast or NxDrawer

I made a new example for the toast page: 7266925a146d039a3b8a1bfdaa6411ddf6021aff

gallery/visualtests/nxDrawer.js
```diff
@@ -125,6 +125,48 @@ describe('NxDrawer', function() {
...
+    const { getPage, checkFullPageScreenshot, waitAndGetElements, wait, isVisible } = browserSetup;
+    const openDrawer = openDrawerFromPage(browserSetup);
+
+    const buttonId = 'nx-drawer-with-nx-toast-open-button';
+    const drawerId = 'nx-drawer-with-nx-toast';
+
+    it('is positioned correctly', async function() {
+      await getPage().setViewport(viewportSize);
+      const [openSuccessToastButton] = await waitAndGetElements('#nx-toast-success-open-button');
+      const [openErrorToastButton] = await waitAndGetElements('#nx-toast-error-open-button');
```
rpokorny, line 139: The reason `waitAndGetElements` returns an array is because it can take multiple arguments and fetch multiple `ElementHandle`s at the same time.  This reduces the round trips between the test and the browser (which is a separate process) and speeds up the tests

gallery/src/components/NxToast/NxToastWithNxDrawerExample.tsx
```diff
@@ -0,0 +1,132 @@
...
+  NxSuccessAlert
+} from '@sonatype/react-shared-components';
+
+interface ToastModel {
+  id: number;
+  alertComponent: ComponentType;
+  message: string;
+}
+
+export default function NxDrawerWithNxToastExample() {
```
rpokorny, line 32: Is there any meaningful difference between the "toast with drawer" example and the "drawer with toast" example? I don't think we need both.  Personally I would have only put the example on one of the two documentation pages, but even if you have it on both, they could both link to the same actual example page.

rjp99: The only difference is so the 'Back to Documentation' button links back to the drawer or toast page depending on where it was accessed from. I initially had it as just one example, but the Back to Documentation link from the toast page would go back to the Drawer page instead, but if that isn't a problem I can just link them to the same example.

rpokorny: > The only difference is so the 'Back to Documentation' button links back to the drawer or toast page depending on where it was accessed from. I initially had it as just one example, but the Back to Documentation link from the toast page would go back to the Drawer page instead, but if that isn't a problem I can just link them to the same example.

Oh that's a good point.  Just leave it like this then.  There are ways to address that that could result in less code duplication, but at the expense of making the example code more complicated which is undesirable since they are examples

gallery/visualtests/nxDrawer.js
rjp99: > The reason `waitAndGetElements` returns an array is because it can take multiple arguments and fetch multiple `ElementHandle`s at the same time. This reduces the round trips between the test and the browser (which is a separate process) and speeds up the tests

Thanks for letting me know that can be done! : ec1c68211bc63b6f64180fee56d855cd79f3042f

gallery/visualtests/nxDrawer.js
```diff
@@ -125,6 +125,49 @@ describe('NxDrawer', function() {
...
+      await wait(500);
+      await checkFullPageScreenshot();
+    });
+
+    it('stays open when the toast is closed', async function() {
+      await getPage().setViewport(viewportSize);
+      await openDrawer(buttonId, drawerId);
+      const [drawer] = await waitAndGetElements('#nx-drawer-with-nx-toast');
+
+      const [openErrorToastButton] = await waitAndGetElements('#nx-toast-error-open-button');
```
rpokorny, line 155:
```suggestion
      const [drawer, openErrorToastButton] = await waitAndGetElements(
        '#nx-drawer-with-nx-toast',
        '#nx-toast-error-open-button'
      );
```

rpokorny (approved):

PR #799, opened 2022-09-26
Support disabling NxIndeterminatePagination buttons - RSC-1172

entroform:
https://issues.sonatype.org/browse/RSC-1172


cindyko226:
Hi Andrew, based on the mockups https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/EL9382q, the disabled buttons are hidden from users, maybe there should be some CSS changes.

EmilyBermudez88:
In the brief description of indeterminate pagination, the docs read "A pagination control for use in cases where the current page number and total number of pages is indeterminate." Now that we know when we are on the first or last page, perhaps update this

entroform:
> Hi Andrew, based on the mockups https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/EL9382q, the disabled buttons are hidden from users, maybe there should be some CSS changes.

Thanks Cindy! I'll be reaching out to the designers to find out what happens if both are hidden.

entroform:
> Hi Andrew, based on the mockups https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/EL9382q, the disabled buttons are hidden from users, maybe there should be some CSS changes.

Hi Cindy! After discussions with design we've confirmed the implementation so the changes are made here:
382d570ffd1dba0624d1b27ae1e013c7c7f83619

I'm torn between adding a classname to toggle visibility, but it may not be worth it for now since it's only one property. maybe we should create a global utility class modifier for things like hidden down the road 🤔

entroform:
> In the brief description of indeterminate pagination, the docs read "A pagination control for use in cases where the current page number and total number of pages is indeterminate." Now that we know when we are on the first or last page, perhaps update this

Thanks for noticing this @EmilyBermudez88! I updated the description in this commit:
382d570ffd1dba0624d1b27ae1e013c7c7f83619

JasonDSwearingen (approved):

gallery/src/components/NxIndeterminatePagination/NxIndeterminatePaginationPage.tsx
```diff
@@ -72,6 +96,22 @@ const NxIndeterminatePaginationPage = () =>
                         codeExamples={nxIndeterminatePaginationCode}>
       An <NxCode>NxIndeterminatePagination</NxCode> component.
     </GalleryExampleTile>
+
+    <GalleryExampleTile title="First Page Example"
+                        id="nx-indeterminate-pagination-first-page-example"
+                        liveExample={NxIndeterminatePaginationFirstPageExample}
+                        codeExamples={nxIndeterminatePaginationFirstPageCode}>
+      An<NxCode>NxIndeterminatePagination</NxCode> first page example.{' '}
```
cindyko226, line 104:
```suggestion
      An <NxCode>NxIndeterminatePagination</NxCode> first page example.{' '}
```

cindyko226:
> I'm torn between adding a classname to toggle visibility, but it may not be worth it for now since it's only one property. maybe we should create a global utility class modifier for things like hidden down the road 🤔

I am thinking maybe we should add `hidden` in the section of Utility classes in https://gallery.sonatype.dev/#/pages/Styling%20Components since it is related to component's visual state, what do you think?

entroform:
> > I'm torn between adding a classname to toggle visibility, but it may not be worth it for now since it's only one property. maybe we should create a global utility class modifier for things like hidden down the road 🤔
> 
> I am thinking maybe we should add `hidden` in the section of Utility classes in https://gallery.sonatype.dev/#/pages/Styling%20Components since it is related to component's visual state, what do you think?

Thanks Cindy, yeah I think that's cleaner. I've added it here: 6129c6f393516623ac60a173c299a3f2a1b67786

cindyko226 (approved):

PR #802, opened 2022-09-27
Add Support for Wrapping NxTooling around NxTag - RSC-1182 - RSC-1191

entroform:
https://issues.sonatype.org/browse/RSC-1182
https://issues.sonatype.org/browse/RSC-1191

### DIFF
--- a/gallery/src/components/NxTag/NxTagPage.tsx
+++ b/gallery/src/components/NxTag/NxTagPage.tsx
@@ -139,7 +139,7 @@ const NxTagPage = () =>
                         liveExample={NxTagWithTooltipExample}
                         codeExamples={NxTagWithTooltipExampleCode}>
       An example of <NxCode>NxTag</NxCode> wrapped inside <NxCode>NxTooltip</NxCode>.{' '}
-      Notice that <NxCode>NxTooltip</NxCode> title overrides the <NxCode>NxOverflowTooltip</NxCode>{' '}
+      Notice that the wrapping <NxCode>NxTooltip</NxCode> title overrides the <NxCode>NxOverflowTooltip</NxCode>{' '}
       generated when the label is truncated.
     </GalleryExampleTile>
 
@@ -148,7 +148,7 @@ const NxTagPage = () =>
                         liveExample={NxSelectableTagWithTooltipExample}
                         codeExamples={NxSelectableTagWithTooltipExampleCode}>
       An example of <NxCode>NxSelectableTag</NxCode> wrapped inside <NxCode>NxTooltip</NxCode>.{' '}
-      Notice that <NxCode>NxTooltip</NxCode> title overrides the <NxCode>NxOverflowTooltip</NxCode>{' '}
+      Notice that the wrapping <NxCode>NxTooltip</NxCode> title overrides the <NxCode>NxOverflowTooltip</NxCode>{' '}
       generated when the label is truncated.
     </GalleryExampleTile>
   </>;
--- a/gallery/src/components/NxTag/NxTagPage.tsx
+++ b/gallery/src/components/NxTag/NxTagPage.tsx
@@ -139,7 +139,8 @@ const NxTagPage = () =>
                         liveExample={NxTagWithTooltipExample}
                         codeExamples={NxTagWithTooltipExampleCode}>
       An example of <NxCode>NxTag</NxCode> wrapped inside <NxCode>NxTooltip</NxCode>.{' '}
-      Notice that <NxCode>NxTooltip</NxCode> title overrides the tooltip generated when the text is truncated.
+      Notice that <NxCode>NxTooltip</NxCode> title overrides the <NxCode>NxOverflowTooltip</NxCode>{' '}
+      generated when the label is truncated.
     </GalleryExampleTile>
 
     <GalleryExampleTile title="NxSelectableTag With Tooltip"
@@ -147,7 +148,8 @@ const NxTagPage = () =>
                         liveExample={NxSelectableTagWithTooltipExample}
                         codeExamples={NxSelectableTagWithTooltipExampleCode}>
       An example of <NxCode>NxSelectableTag</NxCode> wrapped inside <NxCode>NxTooltip</NxCode>.{' '}
-      Notice that <NxCode>NxTooltip</NxCode> title overrides the tooltip generated when the text is truncated.
+      Notice that <NxCode>NxTooltip</NxCode> title overrides the <NxCode>NxOverflowTooltip</NxCode>{' '}
+      generated when the label is truncated.
     </GalleryExampleTile>
   </>;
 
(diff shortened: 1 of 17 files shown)


entroform:
> Hi Andrew, there is another type of `NxTag`, `NxSelectableTag`, do we want to forward ref on this as well so it can be wrapped inside `NxTooltip`?

Hi Cindy! Yeah I think so: I've added it here: 45bdede425fb7e8aee4187a0c85118ba237ed4c2

gallery/src/components/NxTag/NxTagPage.tsx
```diff
@@ -130,6 +134,21 @@ const NxTagPage = () =>
       Light grey border added for clarity.
     </GalleryExampleTile>
 
+    <GalleryExampleTile title="NxTag With Tooltip"
+                        id="nx-tag-with-tooltip-example"
+                        liveExample={NxTagWithTooltipExample}
+                        codeExamples={NxTagWithTooltipExampleCode}>
+      An example of <NxCode>NxTag</NxCode> wrapped inside <NxCode>NxTooltip</NxCode>.{' '}
+      Notice that <NxCode>NxTooltip</NxCode> title overrides the tooltip generated when the text is truncated.
```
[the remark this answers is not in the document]
cindyko226: > I think the NxTooltip title overrides both whether the text is truncated, right?

Ah I think I know what you mean here, it overrides overflowing tooltip.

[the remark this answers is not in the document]
entroform: Yup! Oo I wonder if I should phrase it differently so it's more clear 🤔 I'll try to think of a better way to word it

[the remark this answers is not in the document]
entroform: > > I think the NxTooltip title overrides both whether the text is truncated, right?
> 
> Ah I think I know what you mean here, it overrides overflowing tooltip.

I updated it here:
65b9f6875e9ebcc26539754a81dde17a5499de62
bf62e6d178a8a0a7407b97c9153ac961a7b1761e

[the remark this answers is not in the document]
cindyko226: Further on this, I noticed in NxTag With Tooltip Example, if you remove the `title` attribute from `NxTooltip` overflowing tag, the `NxOverflowTooltip` is not showing. Is this expected?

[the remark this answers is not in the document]
entroform: > Further on this, I noticed in NxTag With Tooltip Example, if you remove the `title` attribute from `NxTooltip` overflowing tag, the `NxOverflowTooltip` is not showing. Is this expected?

I believe so, because the `title` prop is required for NxTooltip

gallery/visualtests/nxTag.js
```diff
@@ -59,5 +59,73 @@ describe('NxTag', function() {
...
+
+    it('shows custom tooltip', async function() {
+      const [exampleElement, tooltipTagElement] = await waitAndGetElements(selector, tagSelector);
+
+      // hover the tag to activate its tooltip.
+      await scrollIntoView(exampleElement);
+
+      await tooltipTagElement.hover();
+      await wait(1500);
+      const { x, y, height, width } = await exampleElement.boundingBox();
```
cindyko226, line 78:
```suggestion
      await scrollIntoView(exampleElement);
      await tooltipTagElement.hover();
      await wait(1500);
      const { x, y, height, width } = await exampleElement.boundingBox();
```

gallery/visualtests/nxTag.js
```diff
@@ -59,5 +59,73 @@ describe('NxTag', function() {
...
```
cindyko226, line 112:
```suggestion
      await scrollIntoView(exampleElement);
      await tooltipTagElement.hover();
      await wait(1500);
      const { x, y, height, width } = await exampleElement.boundingBox();
```

entroform:
Cindy found a bug when the truncated selected tag is selected the text seems to shift left only on Chrome (works fine on Firefox and Safari). It turns out this bug is also in the current live version. I've created a ticket to address it here: https://issues.sonatype.org/browse/RSC-1191

entroform:
@cindyko226 Hi Cindy! I fixed the Chrome bug here: 63f7fae81ea62cc211892fcc54738e34868f7b39
Let me know if you could verify it! Thanks!

cindyko226:
> Let me know if you could verify it! Thanks!

Hi Andrew, I have verified the truncated text is not shifted to the left when selected.
However, I noticed another behavior that if you adjust the viewport width to have the horizontal scrollbar shown, then click the truncated selected tag, the scrollbar will scroll to the right, is it expected?

lib/src/components/NxTag/NxTag.scss
```diff
@@ -64,6 +64,10 @@
 
 .nx-tag__input {
   @include nx-hide-checkbox-helper.hide-checkbox;
+  // This fixed the checkbox shifting the truncated text
+  // to the left when focused on Chrome.
+  position: absolute;
+  right: 0;
```
rpokorny, line 70: Weird, I wonder if if this a regression in more recent versions of Chrome.  I doubt we wouldn't have noticed this issue when we first created `NxTag`.  IMO we should have a visual test that would have caught this

lib/src/components/NxTag/NxTag.tsx
```diff
@@ -15,29 +15,28 @@ import NxOverflowTooltip from '../NxTooltip/NxOverflowTooltip';
...
-      );
-    };
+  const label = (
+    <label className={tagClasses} title={title} ref={ref} {...attrs}>
+      <span className="nx-tag__text">{children}</span>
+      {selectedIcons}
+    </label>
+  );
+
+  return typeof title === 'undefined' ? <NxOverflowTooltip>{label}</NxOverflowTooltip> : label;
```
rpokorny, line 32: We have a more robust mechanism for components to detect when they are wrapped in a tooltip: the `TooltipContext` defined in `NxTooltip`.  That should be used instead of checking for `title`.

rpokorny:
Add `ref` unit tests

entroform:
> Add `ref` unit tests

Fixed the ref issue here: 29574cd860b314ebd2fb244df133bf3e9cf1bb11
Currently looking into adding test for the NxOverflowTooltip forwardRef

lib/src/components/NxTag/NxTag.tsx
entroform: > We have a more robust mechanism for components to detect when they are wrapped in a tooltip: the `TooltipContext` defined in `NxTooltip`. That should be used instead of checking for `title`.

Updated here! 59ced2571cd1548f718660958d93212af59f7860

lib/src/components/NxTag/NxTag.scss
entroform: > Weird, I wonder if if this a regression in more recent versions of Chrome. I doubt we wouldn't have noticed this issue when we first created `NxTag`. IMO we should have a visual test that would have caught this

Added an extra visual test  here: f907f61fec9666bba6ca301b7c1036a8c000096c

entroform:
@rpokorny Thanks for the suggestions. I've updated the implementation here:
6c0185cd7462f4b4d53d04aed9f315a3eca8f375

rpokorny (approved):
+1 aside from integrating https://github.com/sonatype/sonatype-react-shared-components/commit/692013be28a33015c67b8c10f46fdd416d595c37

PR #803, opened 2022-09-27
NxCode Stretched Vertically - RSC-1174

rjp99:
https://issues.sonatype.org/browse/RSC-1174

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
 #
 
-FROM node:16-bullseye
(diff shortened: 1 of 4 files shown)


entroform (commented):
Looks good to me! I think it's worth getting design approval to make sure they're aware of this scenario of NxCode inside list term.

I found this issue where NxCode overflows, not sure if we want to implement something like `word-break: break-all;`, it could be something you could demo to the designers too.

nvm I'll create a new ticket, this is probably out of scope. 👍 

Screen Shot 2022-09-28 at 10 14 57 AM

entroform (approved):

PR #804, opened 2022-09-28
Colorful Icon Utility Class - RSC-1119

rjp99:
https://issues.sonatype.org/browse/RSC-1119


gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -260,15 +260,31 @@ const NxDropdownPage = () =>
...
             <NxTable.Row>
-              <NxTable.Cell><NxCode>nx-dropdown-icon</NxCode></NxTable.Cell>
               <NxTable.Cell>
-                Set custom icons to the correct color.
+                <NxCode><NxTextLink href="#/pages/Icon">nx-icon--colorful</NxTextLink></NxCode>
               </NxTable.Cell>
               <NxTable.Cell>
-                This class should be applied to custom icons that do not have built-in color{' '}
-                so they are set to the correct <NxCode>NxDropdown</NxCode> icon color.
+                <NxCode>NxIcon</NxCode> inside <NxCode>NxDropdown</NxCode>
```
rpokorny, line 285: Probably meant `NxFontAwesomeIcon`; there's no such think as `NxIcon`

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -13,10 +13,10 @@
   display: inline-block;
   position: relative;
   width: nx-dropdown-variables.$dropdown-width;
-}
 
-.nx-dropdown-icon {
-  color: var(--nx-swatch-grey-05);
+  .nx-icon:not(.nx-icon--colorful) {
```
rpokorny, line 17: :thinking: is this technically a breaking change? If a user in the past wanted to include an icon with a specific color in and `NxDropdown`, all they had to do is _not_ provide it with the `nx-dropdown-icon` class.  With this PR, that will no longer work, they must instead provide the `.nx-icon--colorful` class. I think we need to deprecate but still support the previous way.

gallery/src/components/NxDropdown/NxDropdownPage.tsx
rjp99: > Probably meant `NxFontAwesomeIcon`; there's no such think as `NxIcon`

Updated here: 51f0453ac0b7f4def7648ce7d3277d2c447267b2

gallery/src/styles/NxIcon/NxIconPage.tsx
```diff
@@ -24,6 +24,16 @@ const NxIconPage = () =>
...
         the <NxCode>viewBox</NxCode>, while for IE, both the width and the height are necessary.
       </NxP>
+      <NxTile.Subsection>
+        <NxTile.SubsectionHeader>
+          <NxH3>Modifiers</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          The <NxCode>nx-icon--colorful</NxCode> modifier class can be applied to icons that have context
+          sensitive color. Using this modifier will prevent the icon's color from changing even if a componentx
+          tries to override the color.
```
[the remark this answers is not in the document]
rjp99: Couldn't commit the suggestion for some reason, updated here: 2caae7288c76c2ccff4df4edf2975f897d9f8250

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: > 🤔 is this technically a breaking change? If a user in the past wanted to include an icon with a specific color in and `NxDropdown`, all they had to do is _not_ provide it with the `.nx-dropdown-icon` class. With this PR, that will no longer work, they must instead provide the `.nx-icon--colorful` class. I think we need to deprecate but still support the previous way.

Updated here: 48a6e95bce989f25086e29f9dc8caccb6653530b

rpokorny: > Updated here: https://github.com/sonatype/sonatype-react-shared-components/commit/48a6e95bce989f25086e29f9dc8caccb6653530b

Actually the more I think about it the more I'm not sure it's possible for this to not be a breaking change.  Previously, if the caller wanted the icon to keep its original color, they would _not_ specify a class.  And if they wanted it to get the dark color, they'd specify `.nx-dropdown-icon`.  Now, if they want it to keep its original color, they specify `.nx-icon--colorful`.  But what do they do if they want it to get the dark color? The only answer that makes sense is that they don't specify any class.  But that conflicts with the previous behavior, so I think it has to be a breaking change.  Luckily, we are close to being able to release RSC 12.0.0 so this could just be lumped in with that.

gallery/src/components/NxDropdown/NxDropdownLinksExample.tsx
```diff
@@ -19,15 +19,15 @@ function NxDropdownActionsExample() {
                 onToggleCollapse={onToggleCollapse}>
       <a href="https://www.google.com/" className="nx-dropdown-link">
         <span>Go somewhere</span>
-        <NxFontAwesomeIcon icon={faExternalLinkAlt}/>
+        <NxFontAwesomeIcon className="nx-icon--colorful" icon={faExternalLinkAlt}/>
```
rpokorny, line 22: We should probably build this into the `.nx-dropdown-link` styling rather than requiring the user to specify it.  That is, when an icon is within `.nx-dropdown-link` and _does not_ have `.nx-icon--colorful`, we should set its color to the link blue.

rjp99: > We should probably build this into the `.nx-dropdown-link` styling rather than requiring the user to specify it. That is, when an icon is within `.nx-dropdown-link` and _does not_ have `.nx-icon--colorful`, we should set its color to the link blue.

Fixed here: f300e51084a77e4f23ff4be98780714f48b13e5b

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -13,6 +13,10 @@
   display: inline-block;
   position: relative;
   width: nx-dropdown-variables.$dropdown-width;
+
+  .nx-icon:not(.nx-icon--colorful):not(.nx-dropdown-link .nx-icon):not(.nx-dropdown .disabled .nx-icon) {
+    color: var(--nx-swatch-grey-05);
```
rpokorny, line 18:
```suggestion
    color: var(--nx-swatch-grey-05);
```

gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -260,15 +260,35 @@ const NxDropdownPage = () =>
                 any <NxCode>.nx-dropdown-right-button</NxCode>s.
               </NxTable.Cell>
             </NxTable.Row>
+          </NxTable.Body>
+        </NxTable>
+        <NxWarningAlert>
+          The <NxCode>nx-dropdown-icon</NxCode> class name has been deprecated.
```
rpokorny, line 266: Since we're considering this PR to be a breaking change, I don't think we need to mention `nx-dropdown-icon` anymore at all.  Especially since icons which previously used that class no longer have to use any class at all to achieve the same effect (though they are still safe to use it, it just doesn't do anything anymore)

lib/src/components/NxIconDropdown/NxIconDropdown.tsx
```diff
@@ -47,7 +47,7 @@ const NxIconDropdown: FunctionComponent<Props> = function NxIconDropdown(props)
               aria-haspopup="true"
               aria-expanded={isOpen}
               title={title}>
-      <NxFontAwesomeIcon icon={icon || faEllipsisV}/>
+      <NxFontAwesomeIcon className="nx-icon--colorful" icon={icon || faEllipsisV}/>
```
rpokorny, line 50: This has no effect

lib/src/components/NxDropdown/NxDropdown.scss
rpokorny, line 18: Semantically speaking, is it supposed to be strictly grey-05, or is it supposed to match the dark (stark) text color? I suspect it might be the latter which will make things smoother as we add dark mode support.  Check with design.

lib/src/components/NxTextLink/NxTextLink.tsx
```diff
@@ -47,7 +47,7 @@ const NxTextLink = forwardRef<HTMLAnchorElement, Props>(
           target = specifiedTarget || (newTab || (external && newTab !== false) ? '_blank' : ''),
           wrappedChildren = truncate ?
             <span className="nx-text-link__truncate-text">{children}</span> : ensureStartEndElements(children),
-          icon = <NxFontAwesomeIcon icon={faExternalLinkAlt} size="sm" />;
+          icon = <NxFontAwesomeIcon className="nx-icon--colorful" icon={faExternalLinkAlt} size="sm" />;
```
rpokorny, line 50: What's the purpose of this change?  I guess it would affect external links within `NxTree`s, but I'm wondering if there's something else that caused you to add this.

rpokorny:
The `NxTree` Utility Class Example code and description needs to be updated.  The description still refers to `nx-tree__colored-icon` rather than `nx-icon--colorful`.  Additionally, since the existing colored icon within the example is an `NxThreatIndicator`, it actually _doesn't_ demonstrate adding that class manually, as `NxThreatIndicator` now takes care of that internally.  IMO the example should be updated to include both the `NxThreatIndicator` and some other custom icon with a custom color, and the description should be updated to explain that `NxThreatIndicator` sets `nx-icon--colorful` internally, so the caller doesn't have to specify it for that one.

lib/src/icons/Close.tsx
```diff
@@ -9,7 +9,7 @@ import React from 'react';
 import './Close.scss';
 
 const Close = () =>
-  <svg viewBox="0 0 16 16" className="nx-icon nx-icon--close">
+  <svg viewBox="0 0 16 16" className="nx-icon nx-icon--close nx-icon--colorful">
```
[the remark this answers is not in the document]
rjp99: > > We want this here?
> 
> I removed it and added it to the css: [6d951eb](https://github.com/sonatype/sonatype-react-shared-components/commit/6d951ebe770da58579e20756049192a0d09ff770)

Made it more organized: 00938f26f901043a30f94c35c896d247d2a7c099

lib/src/components/NxIconDropdown/NxIconDropdown.tsx
rjp99: > This has no effect

Removed here: 542e7e67935e50ab1f2f4d7efd8e4971846e2ad3

lib/src/components/NxTextLink/NxTextLink.tsx
rjp99: > What's the purpose of this change? I guess it would affect external links within `NxTree`s, but I'm wondering if there's something else that caused you to add this.

It was also for the `NxIconDropdown` external links, but it isn't needed anymore so I removed it here: 21c1ad2f8281c16939fc5afecf905965acb90e69

gallery/src/components/NxDropdown/NxDropdownPage.tsx
rjp99: > Since we're considering this PR to be a breaking change, I don't think we need to mention `nx-dropdown-icon` anymore at all. Especially since icons which previously used that class no longer have to use any class at all to achieve the same effect (though they are still safe to use it, it just doesn't do anything anymore)

Removed the deprecated warning: 723a56df94ba202e58d35fe44c909e0c581f8657

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: > Semantically speaking, is it supposed to be strictly grey-05, or is it supposed to match the dark (stark) text color (which happens to be grey-05)? I suspect it might be the latter which will make things smoother as we add dark mode support. Check with design.

I asked Jenni about it, and she said all the light mode icons should be grey-05 unless in an interaction state. Since they haven't started on the icon collection for dark mode it is not sure if it will match the text color.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -141,6 +145,19 @@
   }
 }
 
+.nx-dropdown-link .nx-icon {
+  &:hover {
+    color: var(--nx-color-link-hover);
+    text-decoration: underline;
```
rpokorny, line 151: In the dropdown mocks, the external link icons are not underlined.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -141,6 +145,19 @@
...
+  }
+
+  &.disabled {
```
rpokorny, line 154: The icon isn't going to have the `disabled` class, so I doubt this block is doing anything

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -141,6 +145,19 @@
   }
 }
 
+.nx-dropdown-link .nx-icon {
+  &:hover {
+    color: var(--nx-color-link-hover);
```
rpokorny, line 150: Is this needed? Even without it I'm seeing the link icons change color when the link is hovered

rjp99: > Is this needed? Even without it I'm seeing the link icons change color when the link is hovered

It's for the close icon next to the reset button under the filter dropdown.

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: It's for the close icon next to the reset button under the filter dropdown. When the filter doesn't have anything selected the reset button and the icon are both disabled.

rjp99:
> The `NxTree` Utility Class Example code and description needs to be updated. The description still refers to `nx-tree__colored-icon` rather than `nx-icon--colorful`. Additionally, since the existing colored icon within the example is an `NxThreatIndicator`, it actually _doesn't_ demonstrate adding that class manually, as `NxThreatIndicator` now takes care of that internally. IMO the example should be updated to include both the `NxThreatIndicator` and some other custom icon with a custom color, and the description should be updated to explain that `NxThreatIndicator` sets `nx-icon--colorful` internally, so the caller doesn't have to specify it for that one.

I added an example for that here: 51ba5b839a85ef91a7792edd85601a651d6c9602

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: Oh yeah, removed it here: 602eb123d4617a696be4b955b4cdb781fd26ed1b

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: > In the dropdown mocks, the external link icons are not underlined.

Removed: 602eb123d4617a696be4b955b4cdb781fd26ed1b

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: > Is this needed? Even without it I'm seeing the link icons change color when the link is hovered

Yeah, it isn't needed anymore: 602eb123d4617a696be4b955b4cdb781fd26ed1b

lib/src/components/NxTree/NxTree.scss
```diff
@@ -142,10 +142,14 @@
     margin-left: $item-label-left-margin;
     margin-right: var(--nx-spacing-base);
 
-    .nx-icon:not(.nx-tree__colored-icon) {
+    .nx-icon:not(.nx-tree__colored-icon, .nx-icon--colorful) {
       color: var(--nx-swatch-grey-05);
     }
 
+    .nx-tree-key-icon {
+      color: purple;
```
rpokorny, line 150: This is part of the one gallery example, and needs to be moved to a scss file for just that example.  See other examples that include scss, such as `WithClassExample.tsx`.  The classname should also use a `gallery-` prefix rather than an `nx-` prefix.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -13,6 +13,10 @@
   display: inline-block;
   position: relative;
   width: nx-dropdown-variables.$dropdown-width;
+
+  .nx-icon:not(.nx-icon--colorful):not(.nx-dropdown-link .nx-icon):not(.nx-dropdown .disabled .nx-icon) {
```
rpokorny, line 17: This can be made more compact with commas instead of successive `:not`s
```suggestion
  .nx-icon:not(.nx-icon--colorful, .nx-dropdown-link .nx-icon, .nx-dropdown .disabled .nx-icon) {
```

But since we're handling disabled here, we should probably handle disabling by attribute as well
```suggestion
  .nx-icon:not(.nx-icon--colorful, .nx-dropdown-link .nx-icon, .nx-dropdown :is(.disabled, [disabled]) .nx-icon) {
```

Note that I've never used `:is` before, and am suggesting this without testing it first :slightly_smiling_face:.  Let me know if it doesn't work.

lib/src/components/NxTextLink/NxTextLink.tsx
rjp99: > > It was also for the NxIconDropdown external links, but it isn't needed anymore so I removed it here: [21c1ad2](https://github.com/sonatype/sonatype-react-shared-components/commit/21c1ad2f8281c16939fc5afecf905965acb90e69)
> 
> Well I do think it was a desirable change, for external links in trees. In fact lets go ahead and add such a thing to one of the tree examples to confirm whether it would be needed for that to work correctly.

I added the example here: d4c760e166fb8d517b09f2a8e2940685fe1e95b5
There was some css that was already setting the color of all icons for links, but it was not using the `nx-icon--colorful` class so I removed it and added the className to the icon.

lib/src/components/NxTree/NxTree.scss
rjp99: > This is part of the one gallery example, and needs to be moved to a scss file for just that example. See other examples that include scss, such as `WithClassExample.tsx`. The classname should also use a `gallery-` prefix rather than an `nx-` prefix.

Updated here: 1e0619f0c14757945f2a0141c630eeee2972e602

lib/src/components/NxDropdown/NxDropdown.scss
rjp99: It worked, thanks!

lib/src/components/NxTree/NxTree.scss
rpokorny: > Updated here: https://github.com/sonatype/sonatype-react-shared-components/commit/1e0619f0c14757945f2a0141c630eeee2972e602

Also include the scss as part of the readable example code.  Again, see `WithClassPage.tsx` to see how.

gallery/src/components/NxTree/NxTreeExample.scss
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ .gallery-example-custom-key-icon {
```
rpokorny, line 7:
```suggestion
.gallery-example-custom-key-icon {
```

gallery/src/components/NxTree/NxTreeNoGutterExample.tsx
```diff
@@ -105,6 +105,15 @@ export default function NxTreeCollapsibleExample() {
                                 <span>cucumber.webm</span>
                               </NxTree.ItemLabel>
                             </NxTree.Item>
+                            <NxTree.Item>
+                              <NxTree.ItemLabel>
+                                <NxTextLink external
+                                            className="nx-dropdown-link"
```
rpokorny, line 111: This classname shouldn't be here.  This isn't a dropdown, and the styling effects that the class has are incorrect here.

gallery/src/components/NxTree/NxTreeNoGutterExample.tsx
```diff
@@ -105,6 +105,15 @@ export default function NxTreeCollapsibleExample() {
...
+                                            href="http://www.sonatype.com">
```
rpokorny, line 112:
```suggestion
                                            href="https://www.sonatype.com">
```

gallery/src/components/NxTree/NxTreeNoGutterExample.tsx
rjp99: > This classname shouldn't be here. This isn't a dropdown, and the styling effects that the class has are incorrect here.

I copied the code from the dropdown page but forgot to remove the class, removed here: 0aac6314d6ad47a8895ab77c4b87d5b57339c786

rjp99: Updated visual tests: 69e9d24ed7701ed9fdcc973182e46474a1f64e2d

gallery/src/components/NxTree/NxTreeNoGutterExample.tsx
```diff
@@ -105,6 +105,14 @@ export default function NxTreeCollapsibleExample() {
                                 <span>cucumber.webm</span>
                               </NxTree.ItemLabel>
                             </NxTree.Item>
+                            <NxTree.Item>
+                              <NxTree.ItemLabel>
+                                <NxTextLink external
+                                            href="https://www.sonatype.com">
```
rpokorny, line 111:
```suggestion
                                <NxTextLink external href="https://www.sonatype.com">
```

rpokorny (approved):

PR #808, opened 2022-09-30
Setup dark mode visual testing - RSC-1170

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1170

Visual tests are set up to run jest twice, once in its default (light), then in a dark mode environment. Visual tests have also been added for the `themeSettingsModal`, which will not run in dark mode (since these tests are testing the functionality of the modal, and not how the modal appears in light and dark mode). 

To define dark mode, `process.env.THEME` is used, but throws a 'process is not defined' lintr error.  It seems that a possible solution could be to implement `"process": true` under `globals` or `"node":true` under `env` in the eslintrc file
https://stackoverflow.com/questions/50894000/eslint-process-is-not-defined

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -11,9 +11,10 @@
     "lint-tests": "eslint --no-eslintrc -c .eslintrc-tests \"visualtests/*.js\"",
     "webpack": "webpack --env production --mode=production",
     "webpack-dev-server": "webpack serve",
-    "test": "npm-run-all check-install lint-tests build jest",
+    "test": "npm-run-all check-install lint-tests build jest jest-dark",
     "check-install": "yarn install --check-files --frozen-lockfile",
     "jest": "jest --testTimeout=120000",
+    "jest-dark": "THEME=DARK jest --testTimeout=120000",
     "clean": "rimraf dist webpack-modules test-results"
   },
   "repository": {
--- a/gallery/setupTests.js
+++ b/gallery/setupTests.js
@@ -12,7 +12,8 @@ jest.retryTimes(3);
 
 const toMatchImageSnapshot = configureToMatchImageSnapshot({
   customDiffDir: path.join(__dirname, 'test-results'),
-  customSnapshotIdentifier: ({ defaultIdentifier }) => defaultIdentifier
+  // eslint-disable-next-line no-undef
+  customSnapshotIdentifier: ({ defaultIdentifier }) => defaultIdentifier + (process.env.THEME === 'DARK' ? '-dark' : '')
 });
 
 expect.extend({ toMatchImageSnapshot });
--- a/gallery/visualtests/testUtils.js
+++ b/gallery/visualtests/testUtils.js
@@ -56,6 +56,12 @@ module.exports = {
 
     beforeEach(async function() {
       page = await browser.newPage();
+
+      // eslint-disable-next-line no-undef
+      if (process.env.THEME === 'DARK') {
+        await page.emulateMediaFeatures([{name: 'prefers-color-scheme', value: 'dark'}]);
+      }
+
       await page.goto(pageUrl + pageFragmentIdentifier);
 
       if (ignoreVersionNumber) {
--- a/gallery/visualtests/testUtils.js
+++ b/gallery/visualtests/testUtils.js
@@ -272,5 +272,11 @@ module.exports = {
         };
       }
     };
+  },
+
+  describeIf(envCondition) {
+    return (
+      envCondition ? describe : describe.skip
+    );
   }
 };
--- a/README.md
+++ b/README.md
@@ -231,6 +231,19 @@ You can execute the tests within the Docker container based on that image by run
 ```
 docker run --rm -it -w /home/jenkins/gallery -v $PWD:/home/jenkins rsc-visualtesting yarn test
 ```
+
+Note that by running this command, all visual tests will run in both light (default) and dark modes. To run tests for light mode only, run:
+
+```
+docker run --rm -it -w /home/jenkins/gallery -v $PWD:/home/jenkins rsc-visualtesting yarn jest test
+```
+
+To run tests for dark mode only, run: 
+
+```
+docker run --rm -it -w /home/jenkins/gallery -v $PWD:/home/jenkins rsc-visualtesting yarn jest-dark test
+```
+
 On some computers, visual testing may take more than 20 minutes, so let the tests run in the background. Note that the
 `yarn test` command will (re-)install the gallery dependencies, ensuring that puppeteer downloads the correct version
 of Chromium before running the gallery build and tests.
@@ -245,9 +258,14 @@ specifics of your system. If seemingly spurious diffs are detected, try running 
 ### Updating Visual Test Screenshots
 
 If running the tests results in differences that are expected/intended based on new changes to the library, then the
-baseline screenshots may be updated by running the following command:
+baseline screenshots may be updated. To update light mode screenshots, run the following command:
 ```
 docker run --rm -it -w /home/jenkins/gallery -v $PWD:/home/jenkins rsc-visualtesting yarn jest -u
+```
+
+To update dark mode screenshots, run the following command:
+```
+docker run --rm -it -w /home/jenkins/gallery -v $PWD:/home/jenkins rsc-visualtesting yarn jest-dark -u
 ```
 
 The command above ensures you update the screenshots within the Docker container. The updates need be run within the
--- a/gallery/visualtests/themeSettingsModal.js
+++ b/gallery/visualtests/themeSettingsModal.js
@@ -5,9 +5,13 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 
-const { setupBrowser } = require('./testUtils');
+const { setupBrowser, describeIf } = require('./testUtils');
 
-describe('ThemeSettingsModal', function() {
+// Tests for ThemeSettingsModal are only testing the functionality of the modal itself, so
+// it is unnecessary to run these tests in both light and dark modes
+
+// eslint-disable-next-line no-undef
+describeIf(process.env.THEME !== 'DARK')('ThemeSettingsModal', function() {
   const { waitAndGetElements, wait, getPage, checkFullPageScreenshot } = setupBrowser('#/');
 
   const openModalBtn = '.gallery-page-header__theme-settings-button',
@@ -60,5 +64,4 @@ describe('ThemeSettingsModal', function() {
   it('changes display theme to light when light mode radio is selected', async function() {
     await selectModeChoice(lightModeRadio);
   });
-
 });
--- a/gallery/visualtests/themeSettingsModal.js
+++ b/gallery/visualtests/themeSettingsModal.js
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
+const { setupBrowser } = require('./testUtils');
+
+describe('ThemeSettingsModal', function() {
+  const { waitAndGetElements, wait, getPage, checkFullPageScreenshot } = setupBrowser('#/');
+
+  const openModalBtn = '.gallery-page-header__theme-settings-button',
+      themeSettingsModal = '.gallery-theme-settings-modal',
+      enableChangesToggle = `${themeSettingsModal} .nx-form-group label:nth-of-type(2)`,
+      darkModeRadio = `${themeSettingsModal} .nx-radio:nth-of-type(2)`,
+      lightModeRadio = `${themeSettingsModal} .nx-radio:nth-of-type(3)`;
+
+  async function openThemeSettingsModal() {
+    const [targetElement] = await waitAndGetElements(openModalBtn);
+    await targetElement.click();
+    await waitAndGetElements(themeSettingsModal);
+  }
+
+  async function selectModeChoice(radioElement) {
+    await openThemeSettingsModal();
+
+    const [targetElement] = await waitAndGetElements(radioElement);
+    targetElement.click();
+    await wait(400);
+
+    await checkFullPageScreenshot();
+  }
+
+  beforeEach(async function() {
+    await getPage().setViewport({ width: 1366, height: 1000 });
+  });
+
+  it('disallows any theme changes by default', async function() {
+    await openThemeSettingsModal();
+    await checkFullPageScreenshot();
+  });
+
+  it('enables radios when opted in to theme changes',
+      async function() {
+        await openThemeSettingsModal();
+
+        const [toggle] = await waitAndGetElements(enableChangesToggle);
+        toggle.click();
+        await wait(400);
+
+        await checkFullPageScreenshot();
+      }
+  );
+
+  it('changes display theme to dark when dark mode radio is selected', async function() {
+    await selectModeChoice(darkModeRadio);
+  });
+
+  it('changes display theme to light when light mode radio is selected', async function() {
+    await selectModeChoice(lightModeRadio);
+  });
+
+});


rpokorny:
>  it seems as though https://issues.sonatype.org/browse/RSC-1118 may have dealt with this already, so merging it into FEATURE_RSC-928_dark_mode may remedy this problem.

I have just merged `main` into `FEATURE_RSC-928_dark_mode`.  I tend to doubt that will fix the problem but it will give us the proper basis from which to work on it

gallery/setupTests.js
```diff
@@ -12,7 +12,10 @@ jest.retryTimes(3);
 
 const toMatchImageSnapshot = configureToMatchImageSnapshot({
   customDiffDir: path.join(__dirname, 'test-results'),
-  customSnapshotIdentifier: ({ defaultIdentifier }) => defaultIdentifier
+  // eslint-disable-next-line no-undef
+  customSnapshotsDir: process.env.THEME === 'DARK' && path.join(__dirname, 'visualtests', '__image_snapshots_dark__'),
+  // eslint-disable-next-line no-undef
+  customSnapshotIdentifier: ({ defaultIdentifier }) => defaultIdentifier + (process.env.THEME === 'DARK' ? '-dark' : '')
```
rpokorny, line 18: Once the lastest on the feature branch (which includes the latest from `main`) is merged into this branch, I think the eslint config will need `setupTests.js` added to the list of file patterns in the `overrides` config in the gallery's eslintrc

rpokorny:
> will not run in jest-dark (since these tests are testing the functionality of the modal, and not how the modal appears in light and dark mode).

Actually I think they should be run in both modes to verify how they work in the presence of both browser defaults (e.g. that if the browser is in dark mode, you get dark mode when Enable Theme Changes is off, and get light mode when it's on and Light Mode is selected, etc).  I also think those tests should be renamed to be not about the modal itself but about something more abstract like "dark mode settings".  The fact that we're doing it through the modal is incidental, what we're really trying to test in these tests is that the logic in `_nx-dark-mode-helpers.scss` works correctly.

rpokorny:
Rather than committing a million screenshots in this PR, many of which are for tickets that have not yet been worked and which are therefore probably not really correct, let's leave most of them uncommitted, and have the relevant ticket for each one be responsible for committing it.  In this PR, just commit the ones for dark mode components that have already been worked.

Further, I'm seeing that there's a `--ci` flag for jest that will cause it to fail if snapshots are missing.  I'm wondering we should add that to the commands ran on CI but not locally (I know, explosion of different command variants, sorry).  That way, when we are finally done dark mode and looking to merge the feature branch, we will know we didn't miss any snapshots because the build will finally be passing.  The downside would be that in the meantime, dark mode ticket branches will never have passing CI builds.

rpokorny:
> Actually I think they should be run in both modes to verify how they work in the presence of both browser defaults (e.g. that if the browser is in dark mode, you get dark mode when Enable Theme Changes is off, and get light mode when it's on and Light Mode is selected, etc)

When I made that comment I hadn't yet looked at how you were setting up dark mode for the various tests.  I think it's important that we test at least the dark mode settings themselves using _the actual browser-level preference_ (the media query) rather than the overrides in the modal, as that's the only way to fully verify that the overrides work correctly in all cases.  See https://github.com/sonatype/sonatype-react-shared-components/compare/FEATURE_RSC-928_dark_mode...RSC-1170_dark_mode_browser_emulation for a demonstration of how to control the browser-level setting as well as some changes to how we can control the gallery's theme settings - just setting the `localStorage` values in `testUtils` and then refreshing wasn't working for me, and refreshing isn't ideal for test performance anyway.

gallery/package.json
```diff
@@ -14,9 +14,11 @@
     "webpack": "webpack --env production --mode=production",
     "webpack-dev-server": "webpack serve",
     "test": "npm-run-all check-install lint-tests build jest",
+    "test-dark": "npm-run-all check-install lint-tests build jest-dark",
     "ci-test": "npm-run-all check-install ci-lint-tests build jest",
     "check-install": "yarn install --check-files --frozen-lockfile",
     "jest": "jest --testTimeout=120000",
+    "jest-dark": "THEME=DARK jest --testTimeout=120000",
```
rpokorny, line 21: I'd rather use something slightly more specific as the environment variable name, maybe `RSC_GALLERY_THEME`

gallery/setupTests.js
rpokorny, line 18: Do we need both of these, the custom directly and custom snapshot naming?

EmilyBermudez88: > Do we need both of these, the custom directory and custom snapshot naming?

I thought it may be a good idea that every snapshot have a unique name to avoid any confusion searching for snapshots etc, but I can remove the -dark suffix if you prefer

EmilyBermudez88:
> When I made that comment I hadn't yet looked at how you were setting up dark mode for the various tests. I think it's important that we test at least the dark mode settings themselves using the actual browser-level preference (the media query) rather than the overrides in the modal, as that's the only way to fully verify that the overrides work correctly in all cases. See https://github.com/sonatype/sonatype-react-shared-components/compare/FEATURE_RSC-928_dark_mode...RSC-1170_dark_mode_browser_emulation for a demonstration of how to control the browser-level setting as well as some changes to how we can control the gallery's theme settings - just setting the localStorage values in testUtils and then refreshing wasn't working for me, and refreshing isn't ideal for test performance anyway.

I updated the testing here:
merge RSC-1170_dark_mode_browser_emulation here: [e3aab4c](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/e3aab4c3f60a23529ea662f4f0926b628773b2a4)
Update it to only apply the media query when theme is set to dark: [c8216e7](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/c8216e7cf609ddb2d0e7a74586128bc6e1c6e09b)
Update the visual tests to use the setOverrideTheme and setThemingEnabled functions: [56274cb](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/56274cb48a80a60a1a5f4d5e4ce44d7075428434)

EmilyBermudez88:
> Rather than committing a million screenshots in this PR, many of which are for tickets that have not yet been worked and which are therefore probably not really correct, let's leave most of them uncommitted, and have the relevant ticket for each one be responsible for committing it. In this PR, just commit the ones for dark mode components that have already been worked.

Unnecessary snapshots removed here: [24a004c](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/24a004c312727a073540af33bf489657032c6fa5)

gallery/package.json
EmilyBermudez88: > I'd rather use something slightly more specific as the environment variable name, maybe `RSC_GALLERY_THEME`

Absolutely, I changed it here: [48b2b2b](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/48b2b2bd6d3120f10387e0f2e1cda2078519e9b7)

EmilyBermudez88:
>>it seems as though https://issues.sonatype.org/browse/RSC-1118 may have dealt with this already, so merging it into FEATURE_RSC-928_dark_mode may remedy this problem.

>I have just merged main into FEATURE_RSC-928_dark_mode. I tend to doubt that will fix the problem but it will give us the proper basis from which to work on it

So with the merge of RSC-1118 (https://issues.sonatype.org/browse/RSC-1118), it set the eslintrc file to ignore `visualtests/*.js`, which removes the lintr error of `process not defined` inside of testUtils.js. If we add `setupTests.js` to the list of file paths (suggested here https://github.com/sonatype/sonatype-react-shared-components/pull/808#discussion_r1003251307), I assume it should remove all the lintr errors regarding `process not defined`. Perhaps this issue can be considered closed then (unless we want to make a separate ticket to deal with other instances where process.env is used - like `telemetry.js` and `HostedVersionsSelect.tsx`)?

EmilyBermudez88:
> Further, I'm seeing that there's a `--ci` flag for jest that will cause it to fail if snapshots are missing. I'm wondering we should add that to the commands ran on CI but not locally (I know, explosion of different command variants, sorry). That way, when we are finally done dark mode and looking to merge the feature branch, we will know we didn't miss any snapshots because the build will finally be passing. The downside would be that in the meantime, dark mode ticket branches will never have passing CI builds.

Added `"ci-jest"` with the `--ci flag` to the `"ci-test"` commands in: [c203d87](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/c203d87e9735f1d0ff33546968576e98e8ddd7e2)

gallery/package.json
```diff
@@ -14,9 +14,12 @@
     "webpack": "webpack --env production --mode=production",
     "webpack-dev-server": "webpack serve",
     "test": "npm-run-all check-install lint-tests build jest",
-    "ci-test": "npm-run-all check-install ci-lint-tests build jest",
+    "test-dark": "npm-run-all check-install lint-tests build jest-dark",
+    "ci-test": "npm-run-all check-install ci-lint-tests build ci-jest",
     "check-install": "yarn install --check-files --frozen-lockfile",
+    "ci-jest": "jest --ci --testTimeout=120000",
     "jest": "jest --testTimeout=120000",
+    "jest-dark": "RSC_GALLERY_THEME=DARK jest --testTimeout=120000",
```
rpokorny, line 22: I believe we need a `ci-jest-dark`, and `ci-test` needs to run both of them.  `test` should probably also run both `jest` and `jest-dark` rather than having a separate `test-dark`.

gallery/setupTests.js
```diff
@@ -12,7 +12,12 @@ jest.retryTimes(3);
 
 const toMatchImageSnapshot = configureToMatchImageSnapshot({
   customDiffDir: path.join(__dirname, 'test-results'),
-  customSnapshotIdentifier: ({ defaultIdentifier }) => defaultIdentifier
+  // eslint-disable-next-line no-undef
+  customSnapshotsDir: process.env.RSC_GALLERY_THEME === 'DARK'
+  && path.join(__dirname, 'visualtests', '__image_snapshots_dark__'),
```
rpokorny, line 17:
```suggestion
  customSnapshotsDir: process.env.RSC_GALLERY_THEME === 'DARK' && 
      path.join(__dirname, 'visualtests', '__image_snapshots_dark__'),
```

gallery/setupTests.js
```diff
@@ -12,7 +12,12 @@ jest.retryTimes(3);
...
+  customSnapshotIdentifier: ({ defaultIdentifier }) =>
+    // eslint-disable-next-line no-undef
```
rpokorny, line 19: These eslint disables don't see necessary on my system

gallery/setupTests.js
```diff
@@ -12,7 +12,12 @@ jest.retryTimes(3);
 
 const toMatchImageSnapshot = configureToMatchImageSnapshot({
   customDiffDir: path.join(__dirname, 'test-results'),
-  customSnapshotIdentifier: ({ defaultIdentifier }) => defaultIdentifier
+  // eslint-disable-next-line no-undef
+  customSnapshotsDir: process.env.RSC_GALLERY_THEME === 'DARK'
```
rpokorny, line 16: Rather than doing this comparison twice, I'd say move it out to a local variable above.

gallery/visualtests/darkModeSettings.js
```diff
@@ -0,0 +1,52 @@
...
+    getPage,
+    checkFullPageScreenshot,
+    overrideDisplayTheme,
+    removeThemeEnabling
+  } = setupBrowser('#/');
+
+  const openModalBtn = '.gallery-page-header__theme-settings-button',
+      themeSettingsModal = '.gallery-theme-settings-modal';
+
+  async function openThemeSettingsModal() {
```
rpokorny, line 21: Since we're no longer using the modal to drive the settings, there's really no reason to have it open for these tests

gallery/visualtests/darkModeSettings.js
```diff
@@ -0,0 +1,52 @@
...
+    await overrideDisplayTheme('dark');
+    await openThemeSettingsModal();
+  });
+
+  it('sets display theme to light when themeOverride is set to light', async function() {
+    await overrideDisplayTheme('light');
+    await openThemeSettingsModal();
+  });
+
+  it('sets display theme to light mode when theme changes are disabled', async function() {
```
rpokorny, line 48: IMO to be really thorough we should test all three of the `overrideDisplayTheme` settings both with and without theming enabled

gallery/visualtests/testUtils.js
```diff
@@ -189,6 +211,8 @@ module.exports = {
       dismissResultingDialog,
       disableLoadingSpinnerAnimation,
       scrollIntoView,
+      overrideDisplayTheme,
+      removeThemeEnabling,
```
rpokorny, line 215: Let's keep the names of both of these consistent with the gallery-side functions that they invoke

gallery/visualtests/testUtils.js
```diff
@@ -56,15 +56,37 @@ module.exports = {
...
+        await page.emulateMediaFeatures([{ name: 'prefers-color-scheme', value: 'dark' }]);
+      }
+
       if (ignoreVersionNumber) {
         await hideVersionNumber();
       }
 
+      await page.evaluate(() => {
+        window.setThemingEnabled(true);
+      });
```
rpokorny, line 73: Related to https://github.com/sonatype/sonatype-react-shared-components/pull/808/files#r1004805779 , once `removeThemeEnabling` becomes `setThemeEnabling` you can call it here instead of dealing with `page.evaluate` again

gallery/setupTests.js
rpokorny: > These eslint disables don't see necessary on my system

Actually I don't think this file is getting linted at all.  The various lint scripts in `package.json` are all constrained to files in `src` and `visualtests`, and this file is in neither of those directories.  It should get included in the test linting.

gallery/package.json
EmilyBermudez88: > I believe we need a `ci-jest-dark`, and `ci-test` needs to run both of them. `test` should probably also run both `jest` and `jest-dark` rather than having a separate `test-dark`.

I had taken a look at trying to get `jest` and` jest-dark` both to run under a single `test` command, but if there are any snapshots that fail in the initial `jest` command, yarn exits with `"error Command failed with exit code 1"` and `jest-dark` is not tested. Is there a way to override that exit command in this instance where multiple jest commands are run in one `test` command?

rpokorny: > I had taken a look at trying to get jest and jest-dark both to run under a single test command, but if there are any snapshots that fail in the initial jest command, yarn exits with "error Command failed with exit code 1" and jest-dark is not tested. Is there a way to override that exit command in this instance where multiple jest commands are run in one test command?

That's not particularly a problem.  All of our build is set up so that a failure at one step prevents later steps from executing.  In some ways it could be advantageous to try to run later steps after earlier failures, but that's quite a bit harder to set up and has some disadvantages too, such as the reason for a failure not necessarily being near the bottom of the build log and thus being harder to find.

gallery/setupTests.js
EmilyBermudez88: > > These eslint disables don't see necessary on my system
> 
> Actually I don't think this file is getting linted at all. The various lint scripts in `package.json` are all constrained to files in `src` and `visualtests`, and this file is in neither of those directories. It should get included in the test linting.

This had been a lintr error prior to merging RSC-1118 into FEAUTRE_RSC-928, but now that all visual test files are overridden in the gallery's eslintrc file, it's no longer needed

rpokorny:
We should create a dark mode flavor of the checkered background that can appear behind examples.  That would effect some of the snapshots being added in this branch, though I suppose its OK for this branch to move forward without waiting for that.
image

gallery/setupTests.js
EmilyBermudez88: Screen Shot 2022-10-25 at 2 16 17 PM
Unfortunately, I am still getting it on my end, and checked to see that my branch was currently up to date. You suggested in https://github.com/sonatype/sonatype-react-shared-components/pull/808#discussion_r1003251307 to override the lintr errors in setupTest.js. Would this override cancel out the benefits of adding `setupTest.js` to the lint scripts in package.json ?

gallery/package.json
EmilyBermudez88: > That's not particularly a problem. All of our build is set up so that a failure at one step prevents later steps from executing. In some ways it could be advantageous to try to run later steps after earlier failures, but that's quite a bit harder to set up and has some disadvantages too, such as the reason for a failure not necessarily being near the bottom of the build log and thus being harder to find.
Ok that's good to know. So I removed the unnecessary `test-dark` script and added in `ci-jest-dark` here: [7b218cc](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/7b218cce14b47564c98ebeeaa0c24b5e1b0fc58f)

gallery/setupTests.js
EmilyBermudez88: > Rather than doing this comparison twice, I'd say move it out to a local variable above.

Absolutely. Stored in variable and removed eslint warning here: [3d57fe2](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/3d57fe24579829f1c6cfaf4ffe28d1d6838a7d68)

gallery/visualtests/testUtils.js
EmilyBermudez88: > Let's keep the names of both of these consistent with the gallery-side functions that they invoke

Updated here: [e3c5036](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/e3c5036f7782f983fb251bbb0fa140836f3600d9)

gallery/package.json
```diff
@@ -13,10 +13,13 @@
     "ci-lint-tests": "eslint \"visualtests/*.js\" --max-warnings=0",
     "webpack": "webpack --env production --mode=production",
     "webpack-dev-server": "webpack serve",
-    "test": "npm-run-all check-install lint-tests build jest",
-    "ci-test": "npm-run-all check-install ci-lint-tests build jest",
+    "test": "npm-run-all check-install lint-tests build jest jest-dark",
+    "ci-test": "npm-run-all check-install ci-lint-tests build ci-jest ci-jest-dark",
     "check-install": "yarn install --check-files --frozen-lockfile",
+    "ci-jest": "jest --ci --testTimeout=120000",
+    "ci-jest-dark": "jest-dark --ci --testTimeout=120000",
```
rpokorny, line 20:
```suggestion
    "ci-jest-dark": "RSC_GALLERY_THEME=DARK jest --ci --testTimeout=120000",
```

gallery/visualtests/testUtils.js
```diff
@@ -75,13 +75,13 @@ module.exports = {
...
 
-    async function overrideDisplayTheme(theme) {
+    async function setThemeOverride(theme) {
       await page.evaluate((t) => {
         window.setThemeOverride(t);
       }, theme);
     }
 
-    async function removeThemeEnabling() {
+    async function setThemingEnabled() {
```
rpokorny, line 84: This needed to not just be renamed but also to take the boolean parameter

EmilyBermudez88: > This needed to not just be renamed but also to take the boolean parameter

Sorry, you beat me to the push of the updated function: [5740744](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/574074445b615d7b483810cd4cb99bfa87a28506)

gallery/visualtests/testUtils.js
EmilyBermudez88: > Related to https://github.com/sonatype/sonatype-react-shared-components/pull/808/files#r1004805779 , once `removeThemeEnabling` becomes `setThemeEnabling` you can call it here instead of dealing with `page.evaluate` again

Updated here: [5740744](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/574074445b615d7b483810cd4cb99bfa87a28506)

EmilyBermudez88:
> We should create a dark mode flavor of the checkered background that can appear behind examples. That would effect some of the snapshots being added in this branch, though I suppose its OK for this branch to move forward without waiting for that. image

I'll reach out to design to get this started

gallery/setupTests.js
EmilyBermudez88: > Once the lastest on the feature branch (which includes the latest from `main`) is merged into this branch, I think the eslint config will need `setupTests.js` added to the list of file patterns in the `overrides` config in the gallery's eslintrc

added to `overrides` here" [21ea274](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/21ea2743176a4beb332f83f3c2b032a38d07021b)

rpokorny:
> Should we consider this issue closed then (unless we want to make a separate ticket to deal with other instances where process.env is used - like telemetry.js and HostedVersionsSelect.tsx)?

As of now I think the linting questions in this ticket are resolved.  `telemetry.js` and `HostedVersionsSelect.tsx` are a different matter, because they actually _aren't_ running in node and so it would be inappropriate to tell eslint that they are.  They only have access to `process.env` because we have the webpack `EnvironmentPlugin` set up.  That plugin, which does run in node, basically forwards the node environment variables to the bundle code.  Just using `eslint-disable` for the places that we use that is adequate I think.

gallery/visualtests/darkModeSettings.js
rpokorny: > I went back on forth on whether including the modal was helpful or not. Makes sense to leave it out, so openThemeSettingsModal() removed here: [b01b922](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/b01b922d9a59522d5dce573c40b883865447131d)

You'll need to update the screenshots.

gallery/visualtests/darkModeSettings.js
```diff
@@ -0,0 +1,59 @@
...
+const { setupBrowser } = require('./testUtils');
+
+describe('Dark Mode Settings', function() {
+  const { getPage,
+    checkFullPageScreenshot,
+    setThemeOverride,
+    setThemingEnabled
+  } = setupBrowser('#/');
+
+  async function themeOverrideTest(theme) {
```
rpokorny, line 17: To be more succinct and more consistent with our utilities like `simpleTest`, `themeOverrideTest` should be a non-async function which returns an anonymous async function which runs the actual test logic.  Then the `it` statements would just be `it('...', themeOverrideTest(...));`

gallery/visualtests/testUtils.js
```diff
@@ -56,15 +56,35 @@ module.exports = {
 
     beforeEach(async function() {
       page = await browser.newPage();
+
       await page.goto(pageUrl + pageFragmentIdentifier);
 
+      // eslint-disable-next-line no-undef
```
rpokorny, line 62: This shouldn't be needed now
```suggestion
```

EmilyBermudez88: > This shouldn't be needed now

thanks for catching that, updated here [fac266c](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/fac266cd71d7dacb045c29ca4b52629201bf25db)

EmilyBermudez88:
> As of now I think the linting questions in this ticket are resolved. `telemetry.js` and `HostedVersionsSelect.tsx` are a different matter, because they actually _aren't_ running in node and so it would be inappropriate to tell eslint that they are. They only have access to `process.env` because we have the webpack `EnvironmentPlugin` set up. That plugin, which does run in node, basically forwards the node environment variables to the bundle code. Just using `eslint-disable` for the places that we use that is adequate I think.

Thanks for clarification on this!

gallery/visualtests/darkModeSettings.js
EmilyBermudez88: Screenshots updated here: [3d2e65f](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/3d2e65f70b29ea49997bd5433be89411631ca285)

gallery/visualtests/darkModeSettings.js
EmilyBermudez88: > To be more succinct and more consistent with our utilities like `simpleTest`, `themeOverrideTest` should be a non-async function which returns an anonymous async function which runs the actual test logic. Then the `it` statements would just be `it('...', themeOverrideTest(...));`

Sounds great, updated here : [667c925](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/667c9254abec3c6cccb7e54e9204c424e012c988)

EmilyBermudez88:
I added in then increased a wait() command for themeOverrideTest() here:[631218a](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/631218a270a5e25a4d1ebed9dc9f4e88d7c31426). The carets in the Gallery's navigation were not changing between light mode and dark mode colours quickly enough, and causing tests to fail. I've run tests multiple times after the last snapshot was updated, and everything appears to be passing correctly now.

gallery/visualtests/darkModeSettings.js
```diff
@@ -0,0 +1,51 @@
...
+  describe('when theme changes are disabled', function() {
+    beforeEach(async function() {
+      await setThemingEnabled(false);
+    });
+
+    it('sets display theme to light mode', async function() {
+      await checkFullPageScreenshot();
+    });
+
+    it('disallows browser preference', themeOverrideTest(null));
```
rpokorny, line 45: Aren't these two the same?

EmilyBermudez88: > Aren't these two the same?

yes actually, it hadn't crossed my mind. Which 'it' statement would you prefer to keep?

rpokorny:
> I added in then increased a wait() command for themeOverrideTest() here:[631218a](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/631218a270a5e25a4d1ebed9dc9f4e88d7c31426). The carets in the Gallery's navigation were not changing between light mode and dark mode colours quickly enough, and causing tests to fail. I've run tests multiple times after the last snapshot was updated, and everything appears to be passing correctly now.

I think that's because the CSS `transition` on those carets was written too broadly; it uses `all` instead of actually naming the property that it should transition.  I think in the past there were multiple properties being transitioned, which is probably why it's like that.  But now it's only `transform` that gets transitioned and it should say that clearly so that it switches to dark mode instantly like the rest of the application.  That can be fixed in this PR.

gallery/visualtests/darkModeSettings.js
rpokorny: > yes actually, it hadn't crossed my mind. Which 'it' statement would you prefer to keep?

The second one, it's most similar to the others.

EmilyBermudez88: > The second one, it's most similar to the others.

Removed here: [caa023f](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/caa023f2a4a2a08d59bdf2bb36444a12682ba8cc)

EmilyBermudez88:
> I think that's because the CSS transition on those carets was written too broadly; it uses all instead of actually naming the property that it should transition. I think in the past there were multiple properties being transitioned, which is probably why it's like that. But now it's only transform that gets transitioned and it should say that clearly so that it switches to dark mode instantly like the rest of the application. That can be fixed in this PR.

You were right in your thinking. I changed the transition to specify transform only, and removed the wait() command inside of themeOverrideTest, and the tests passed multiple times. [0ce0397](https://github.com/sonatype/sonatype-react-shared-components/pull/808/commits/0ce0397dcba16e908a62e902d6aef37bb1477787)

rpokorny (approved):

PR #809, opened 2022-10-03
Reset Scroll Position on Gallery Links - RSC-1089

cindyko226:
https://issues.sonatype.org/browse/RSC-1089


rpokorny (approved):
+1.  This seems to be the crucial sentence from the react-router docs:
> A location object is never mutated so you can use it in the lifecycle hooks to determine when navigation happens, this is really useful for data fetching and animation.

https://v5.reactrouter.com/web/api/location

cindyko226:
> A location object is never mutated so you can use it in the lifecycle hooks to determine when navigation happens, this is really useful for data fetching and animation. 

Thanks for the info.

PR #811, opened 2022-10-04
Fix eslint Indent Conflict RSC-1162 

rpokorny:
https://issues.sonatype.org/browse/RSC-1162

There were two issues, one of which still exists but is very minor.

1. JSX that started with a fragment (`<>`) was not being ignored by the typescript indentation plugin.  It turns out these nodes have a unique identity in the AST that needed to be added to the `ignoredNodes` list.  This is what this PR fixes.
2. The react indentation rules do not appear to be able to handle the case where you have a series of variable assignments in a single expression, one of which is a function call that receives multi-line JSX as its argument.  For instance the following:

```
const foo = render(
      <div>
        <span>asdfasdf</span>
      </div>
    ),
    bar = 1;
 ```
 
 It doesn't understand the expectation that everything beyond the first line have an extra two levels of indentation.  So as a workaround, we must continue to break up such declarations into multiple statements, avoiding the double indentation.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.5.14",
+  "version": "11.5.15",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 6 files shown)


entroform (approved):
👍

PR #813, opened 2022-10-05
NxLoadError to type submit - RSC-1192

entroform:
https://issues.sonatype.org/browse/RSC-1192

### DIFF
--- a/gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
+++ b/gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
@@ -56,7 +56,17 @@ const NxLoadErrorPage = () =>
             <NxTable.Cell>Function</NxTable.Cell>
             <NxTable.Cell>No</NxTable.Cell>
             <NxTable.Cell>
-              If this is defined, a Retry button will be rendered which executes this function when clicked
+              If this is defined, a Retry button will be rendered with type "button" or "submit" if
+              <NxCode>useSubmitRetry</NxCode> is also set to true,
+              which executes this function when clicked.
+            </NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell>useSubmitRetry</NxTable.Cell>
+            <NxTable.Cell>boolean</NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>
+              If this is set to true, a Retry button will be rendered with type "submit".
             </NxTable.Cell>
           </NxTable.Row>
           <NxTable.Row>
--- a/gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
+++ b/gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
@@ -56,17 +56,18 @@ const NxLoadErrorPage = () =>
             <NxTable.Cell>Function</NxTable.Cell>
             <NxTable.Cell>No</NxTable.Cell>
             <NxTable.Cell>
-              If this is defined, a Retry button will be rendered with type "button" or "submit" if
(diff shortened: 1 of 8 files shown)


rpokorny:
Does this cause the `onSubmit` logic in an `NxForm` to execute twice when the Retry button is clicked - once for the button click handler and once for the form submit handler?

entroform:
> Does this cause the `onSubmit` logic in an `NxForm` to execute twice when the Retry button is clicked - once for the button click handler and once for the form submit handler?

So interestingly, from testing, it called `onSubmitProp` only once, when the retry button is clicked or enter is pressed on all browsers even with `retryHandler` still being attached. 

I did make `retryHandler` optional when `useSubmitRetry` true, let me know your thought son this!

lib/src/components/NxLoadError/NxLoadError.tsx
```diff
@@ -36,8 +39,11 @@ const NxLoadError = forwardRef<HTMLDivElement, Props>(
               {' '}
               { error }
             </span>
-            { retryHandler &&
-              <NxButton type="button" variant="error" onClick={retryHandler} className="nx-load-error__retry">
+            { (retryHandler || useSubmitRetry) &&
+              <NxButton type={useSubmitRetry ? 'submit' : 'button'}
+                        variant="error"
+                        {...(retryHandler ? { onClick: retryHandler } : {})}
```
rpokorny, line 45:
```suggestion
                        onClick={retryHandler ?? undefined}
```

gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
```diff
@@ -56,7 +56,17 @@ const NxLoadErrorPage = () =>
...
             <NxTable.Cell>No</NxTable.Cell>
             <NxTable.Cell>
-              If this is defined, a Retry button will be rendered which executes this function when clicked
+              If this is defined, a Retry button will be rendered with type "button" or "submit" if
+              <NxCode>useSubmitRetry</NxCode> is also set to true,
+              which executes this function when clicked.
+            </NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell>useSubmitRetry</NxTable.Cell>
```
rpokorny, line 65: How would you feel about renaming this to `submitRetry`, `submitOnRetry`, or something else?  `use...` typically makes people think of hooks

gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
```diff
@@ -56,7 +56,17 @@ const NxLoadErrorPage = () =>
             <NxTable.Cell>Function</NxTable.Cell>
             <NxTable.Cell>No</NxTable.Cell>
             <NxTable.Cell>
-              If this is defined, a Retry button will be rendered which executes this function when clicked
+              If this is defined, a Retry button will be rendered with type "button" or "submit" if
```
rpokorny, line 59: I think the phrasing here would be cleaner if the mention of the `type` was removed.  That can be fully explained in the `useSubmitRetry` description instead.  Though you might also want to mention here that this prop should typically not be used if `useSubmitRetry` is true, since you'll typically use the form's `onSubmit` instead.

gallery/src/components/NxLoadError/NxLoadErrorPage.tsx
entroform: > How would you feel about renaming this to `submitRetry`, `submitOnRetry`, or something else? `use...` typically makes people think of hooks

Thanks yeah I was also trying to think of a better name. Thanks for the suggestions, I went with `submitOnRetry`:
2b1a9ca065d55b324f748a2d776bfe929efc9d53

rpokorny (approved):

PR #814, opened 2022-10-05
Made the Drawer Close Button Hip To Be Square - RSC-1193

entroform:
https://issues.sonatype.org/browse/RSC-1193

### DIFF
--- a/lib/src/components/NxDrawer/NxDrawer.scss
+++ b/lib/src/components/NxDrawer/NxDrawer.scss
@@ -125,6 +125,7 @@
 
   &__close-button {
     grid-area: close-button;
+    height: min-content;
   }
 
   &__title {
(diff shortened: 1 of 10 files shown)


lib/src/components/NxDrawer/NxDrawer.scss
```diff
@@ -125,6 +125,7 @@
 
   &__close-button {
     grid-area: close-button;
+    height: min-content;
```
rpokorny, line 128: Feels like it might be cleaner to use `align-content: top` on the header element

entroform: > Feels like it might be cleaner to use `align-content: top` on the header element

Good call, I've updated it here: 0e555bb70000ba0a3eedea4f432131f169b26c66

lib/src/components/NxDrawer/NxDrawer.scss
```diff
@@ -117,6 +117,7 @@
   grid-template: "title close-button"
                  "subtitle subtitle"
                  "description description" / auto min-content;
+  align-items: start;
```
rpokorny, line 120: We do try to alphabetize our CSS properties

rpokorny (approved):

PR #815, opened 2022-10-06
Avoid need for downstream changes to Alerts - RSC-1201

rpokorny:
https://issues.sonatype.org/browse/RSC-1201

Screenshot of IQ at commit 171249e8f3 with these changes:
image

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "11.5.17",
+  "version": "11.5.18",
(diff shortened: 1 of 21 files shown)


rpokorny:
Most of the visual test differences are caused by an increase in the space above the first line of text in a multi-line alert without a close button.  I believe the new spacing more correctly matches kaleido.  There are also a few snapshots which had minor pixel changes in the focus glow around some elements, presumably due to changes in the size of elements higher up on the page.  These are fine.

avnvbhatta (approved):

PR #816, opened 2022-10-06
Fix Dropdown Links Are Missing Underline When Not Hovered - RSC-1198

entroform:
https://issues.sonatype.org/browse/RSC-1198


cindyko226:
Hi Andrew, should we have underline on disabled links as well? 
Seems like it is depicted in Sketch https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/9Pm3pGz.

entroform:
> Hi Andrew, should we have underline on disabled links as well? Seems like it is depicted in Sketch https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/9Pm3pGz.

Thanks Cindy! Totally missed that.
I updated it here: 2df4fd1ffc19b5a0214dc4cc1693e4cd230652a8
Will update the snapshots momentarily.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -126,17 +126,13 @@
...
-    text-decoration: underline;
-  }
+  text-decoration: underline;
 
   &.disabled {
     &, &:hover, &:focus {
       color: var(--nx-color-link-disabled);
       cursor: default;
-      text-decoration: none;
+      text-decoration: underline;
```
cindyko226, line 135: Do we need this? I think it would inherit the text-decoration defined on line 129.

entroform: > Do we need this? I think it would inherit the text-decoration defined on line 129.

Thanks! yeah we don't need it: 0937f12d715455aa795db5220d4c36c687140d46

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -126,17 +126,12 @@
 
 .nx-dropdown-link {
   color: var(--nx-color-link);
-
-  &:hover {
-    color: var(--nx-color-link-hover);
```
cindyko226, line 131: The text link has darker blue when hover, should we add it back?

entroform: > The text link has darker blue when hover, should we add it back?

Yup! I don't know why i removed that 🤔. Thanks for catching that here's the commit: aa26b63eef59d56acde5231d24b62d69b9eea3e7

cindyko226 (approved):

PR #817, opened 2022-10-06
Fix NxCollapsibleItems Roles - RSC-1197

cindyko226:
https://issues.sonatype.org/browse/RSC-1197


rpokorny:
We should document that the role goes to the children element now and not to the top-level element rendered by this component.  I think we can get away with not calling this a breaking change, if only because the way we were using the role made little sense anyway, so there's not much a consuming project could have been doing with it that would be less correct now than before.

cindyko226:
> We should document that the role goes to the children element now and not to the top-level element rendered by this component. I think we can get away with not calling this a breaking change, if only because the way we were using the role made little sense anyway, so there's not much a consuming project could have been doing with it that would be less correct now than before.

Updated here 45c606c0ff78e7d36a22bc2f9b434567110c8250. Thanks.

gallery/src/components/NxCollapsibleItems/NxCollapsibleItemsPage.tsx
```diff
@@ -137,6 +137,16 @@ const NxCollapsibleItemsPage = () =>
...
         </ul>
       </NxTile.Subsection>
+      <NxTile.Subsection>
+        <NxTile.SubsectionHeader>
+          <NxH3>Accessibility</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          The <NxCode>NxCollapsibleItems</NxCode> role attribute will be applied to the children element instead of
+          the top-level element. By default the role for the children element is set to <NxCode>list</NxCode> and
+          the child element is set to <NxCode>listitem</NxCode> unless otherwise specified.
```
rpokorny, line 147:
```suggestion
          the roles for the individual child elements are set to <NxCode>listitem</NxCode> unless otherwise specified.  The component as a whole has a hardcoded role of <NxCode>group</NxCode>.
```

gallery/src/components/NxCollapsibleItems/NxCollapsibleItemsPage.tsx
```diff
@@ -137,6 +137,16 @@ const NxCollapsibleItemsPage = () =>
           </li>
         </ul>
       </NxTile.Subsection>
+      <NxTile.Subsection>
+        <NxTile.SubsectionHeader>
+          <NxH3>Accessibility</NxH3>
+        </NxTile.SubsectionHeader>
+        <NxP>
+          The <NxCode>NxCollapsibleItems</NxCode> role attribute will be applied to the children element instead of
```
rpokorny, line 145:
```suggestion
          The <NxCode>NxCollapsibleItems</NxCode> role prop will be applied to the children element instead of
```

rpokorny (approved):

PR #818, opened 2022-10-07
Fix nx-read-only Label Colour - RSC-1148

entroform:
https://issues.sonatype.org/browse/RSC-1148


lib/src/base-styles/_nx-read-only.scss
```diff
@@ -29,6 +29,7 @@
 
 .nx-read-only__label {
   @include nx-text-helpers.bold;
+  color: var(--nx-swatch-grey-05);
```
rpokorny, line 32:
```suggestion
  color: var(--nx-color-text-dark);
```

rpokorny:
What's different in the `nx-card` snapshot?

rpokorny (approved):

PR #820, opened 2022-10-11
Fix ComboBox Stuck Open in Safari - RSC-1204

entroform:
https://issues.sonatype.org/browse/RSC-1204


rpokorny:
I think you need to do something similar on the error retry button as well.  In other browsers, clicking that button results in another data load, with the dropdown staying open.  In Safari, clicking that button closes the dropdown

rpokorny:
I'm really surprised that this works.  I would have thought calling `preventDefault` on `onMouseDown` would prevent the `onClick` from ever firing.  I want to do a little bit of research before approving but I'm not coming across any issues in manual testing

rpokorny (approved):
Reading the specs, it seems that the default action for `mousedown` is not actually tied to click, but rather to things like initiating drag and drop.  So this change does in fact seem safe. https://w3c.github.io/uievents/#event-type-mousedown

entroform:
> I'm really surprised that this works. I would have thought calling `preventDefault` on `onMouseDown` would prevent the `onClick` from ever firing. I want to do a little bit of research before approving but I'm not coming across any issues in manual testing

I was surprised by that too!

> I think you need to do something similar on the error retry button as well. In other browsers, clicking that button results in another data load, with the dropdown staying open. In Safari, clicking that button closes the dropdown

Thanks! I fixed that in this commit: b7aadd7b121686294a45fc3eb73ccfcf93e2746d

lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -344,7 +344,18 @@ function NxComboboxRender<T extends string | number = string>(
...
+                      // In Safari, focus seems to be intended to strictly be a keyboard nav thing,
+                      // and when you click a button, focus does NOT go to that button, it goes to the <body>.
+                      // This messes up our idea of only showing the menu when something within the component has focus,
+                      // because this means that the component loses focus in the middle of the click,
+                      // causing the menu to hide while the user is in the middle of clicking,
+                      // causing the click not to register at all.
+                      // Using preventDefault on onMouseDown seem to have prevent the focus from going to the <body>.
+                      onMouseDown={(event) => {
+                        event.preventDefault();
+                        event.currentTarget.focus();
```
rpokorny, line 357: I'm pretty sure calling focus here removes the need for the preventDefault

rpokorny: > I'm pretty sure calling focus here removes the need for the preventDefault

Nevermind it's not working now, I must've been running out of date code when I thought it was

rpokorny (approved):

PR #821, opened 2022-10-11
Eslint Jest/Global variable errors - RSC-1118

rjp99:
https://issues.sonatype.org/browse/RSC-1118

### DIFF
--- a/lib/.eslintrc
+++ b/lib/.eslintrc
@@ -1,6 +1,12 @@
 {
   "extends": ["./.eslintrc-base", "plugin:jsx-a11y/recommended"],
   "plugins": ["jsx-a11y"],
+  "globals": {
+    "require": true
+  },
+  "env": {
+    "jest": true
+  },
   "rules": {
     "jsx-a11y/label-has-associated-control": [ 2, {
       "controlComponents": ["NxTextInput", "NxStatefulTextInput", "NxFilterInput"]


lib/.eslintrc
```diff
@@ -1,6 +1,12 @@
 {
   "extends": ["./.eslintrc-base", "plugin:jsx-a11y/recommended"],
   "plugins": ["jsx-a11y"],
+  "globals": {
+    "require": true
+  },
+  "env": {
+    "jest": true
```
rpokorny, line 8: We have separate eslint configs for our tests.  In fact, this file is _not_ used for our tests.  I suspect VSCode might not actually be smart enough to use the correct eslint config file for a given source file.  In that case I think we could do one of two things:

* Leave this issue unaddressed, VSC users would just have to live with it
* Move this file to another name such as `.eslintrc-impl` and have a new `.eslintrc` which isn't actually used by the build but which is only for the benefit of editors and things, which includes all of the necessary settings (maybe by extending both `.eslintrc-test` and `.eslintrc-impl`)

rjp99: > We have separate eslint configs for our tests. In fact, this file is _not_ used for our tests. I suspect VSCode might not actually be smart enough to use the correct eslint config file for a given source file. In that case I think we could do one of two things:
> 
> * Leave this issue unaddressed, VSC users would just have to live with it
> * Move this file to another name such as `.eslintrc-impl` and have a new `.eslintrc` which isn't actually used by the build but which is only for the benefit of editors and things, which includes all of the necessary settings (maybe by extending both `.eslintrc-test` and `.eslintrc-impl`)

Okay, I will make another file like you suggested and try it out, but I did notice that just adding those configs made all the errors I had in the test files go away.

rpokorny: > but I did notice that just adding those configs made all the errors I had in the test files go away.

Sure, but it will also cause eslint to *not* give errors if someone for some reason tried to use `jest` in the implementation code

rjp99: > > but I did notice that just adding those configs made all the errors I had in the test files go away.
> 
> Sure, but it will also cause eslint to _not_ give errors if someone for some reason tried to use `jest` in the implementation code

Ohh, that makes sense. I also saw overrides on here: https://eslint.org/docs/latest/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns
so I was wondering if adding something like this would work?
```
"overrides": [
    {
      "files": ["*.test.tsx"],
      "globals": {
        "require": true
      },
      "env": {
        "jest": true
      }
    }
  ]

```

rpokorny: > so I was wondering if adding something like this would work?

oohh yeah potenttially.  Try that out and if it works that would help is simplify our overall build in addition to fixing the VSC problem

rjp99: > > so I was wondering if adding something like this would work?
> 
> oohh yeah potentially. Try that out and if it works that would help is simplify our overall build in addition to fixing the VSC problem

It works, when I add that to the `lib/.eslintrc` file then the errors in the test files go away but if I add a describe in a file like NxDrawer.tsx then the error is there.

rpokorny: > I think it does work, when I add that to the lib/.eslintrc file then the errors in the test files go away but if I add a describe in a file like NxDrawer.tsx then the error is there.
https://github.com/sonatype/sonatype-react-shared-components/commit/316d39b3a4e40c7ea4515c527dd1d5402d30d3a6

If I'm not mistaken, this allows us to completely clean up the multi-config-file approach that we've been using for eslint, and to just have one `.eslintrc` each in `lib` and `gallery`.  I'd want to see that cleanup implemented in this PR

gallery/.eslintrc
```diff
@@ -10,5 +10,14 @@
   "rules": {
     "@typescript-eslint/no-var-requires": "off",
     "react/prop-types": "off"
-  }
+  },
+  "overrides": [
+    {
+      "files": ["*.tsx", "*.js", "*.ts"],
```
rpokorny, line 16: This pattern doesn't seem right, shouldn't this be constrained to `js` files in the `visualtests` directory?

rpokorny (approved):

gallery/.eslintrc
rjp99: > This pattern doesn't seem right, shouldn't this be constrained to `js` files in the `visualtests` directory?

Yeah, that makes sense: 2cf88baf5f34897fd2122a48e2d00899b5d835bd

PR #822, opened 2022-10-11
Fix nx-page-content in page-scrolling mode cutting off dropdowns - RSC-1194

entroform:
https://issues.sonatype.org/browse/RSC-1194


lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -50,7 +50,6 @@
   // Cover all columns and the third through last rows of the nx-page grid template
   grid-area: 3 / 1 / -1 / -1;
   max-width: var(--nx-width-page-max);
-  overflow-y: hidden;
```
entroform, line 53: I'm not entirely sure this is the right solution.
I tried looking for things that might break removing this, but couldn't find anything.
Not entirely sure the purpose behind this, let me know your thoughts!

With this removed the dropdown extends the scrollbar down when opened:

Screen Shot 2022-10-11 at 3 26 23 PM

rpokorny: > I'm not entirely sure this is the right solution.
I tried looking for things that might break removing this, but couldn't find anything.
Not entirely sure the purpose behind this, let me know your thoughts!

Do section-scrolling layouts still work?

My assumption had been that we would override this back to visible in `.nx-html--page-scrolling .nx-page-content`, but I too am skeptical that it's safe to remove in when section-scrolling.

entroform: Hmm I remember it did work before and testing it again and it looks like section-scrolling layouts still works 🤔.
I've updated the implementation a bit based on your comment here: cf28564974bf32e1d942a8082ad1dffa987e6148

Going to look into it a bit more!

rpokorny (approved):

PR #823, opened 2022-10-11
Add Clear Functionality to Filter Input - RSC-1184

cindyko226:
https://issues.sonatype.org/browse/RSC-1184


rpokorny:
Doesn't look good on my system with the datalist example: 
image

Using a datalist here is probably now redundant with `NxCombobox`. I wonder if we should deprecate it.  Either way though we can't just drop it, so this would need to be fixed

gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
```diff
@@ -25,7 +25,13 @@ const NxFilterInputPage = () =>
   <>
     <GalleryDescriptionTile>
       <NxP>
-        An input to be used for filtering purposes
+        An input to be used for filtering purposes. After a user has text within the input field, a
+        clear filter button appears within the component and is only visible/usable when there is text within
+        the input field.
+      </NxP>
+      <NxP>
+        Note: This button is for mouse use only and is therefore not focusable but still hoverable and
```
rpokorny, line 33: IMO this should be in the same paragraph as the previous sentence

rpokorny:
It doesn't seem like design intended/is aware that this affect `NxSearchDropdown`, as the mockups for that component do not have this button.  Check with them whether it should (as it does in this PR).  If so, the mockups should be updated.

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -24,4 +24,18 @@
     margin-left: 0;
     margin-right: 8px;
   }
+
+  .nx-btn--clear {
+    visibility: visible;
```
rpokorny, line 29: Redundant
```suggestion
```

rpokorny:
Text should not be visible behind the button.  Check with design on exactly where text should truncate.
image

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -23,16 +26,43 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
           cleanedProps = omit(['validatable', 'validationErrors', 'type'], otherProps),
           filterIcon = searchIcon ? faSearch : faFilter,
-          prefixContent = <NxFontAwesomeIcon icon={filterIcon} className="nx-icon--filter-icons" />;
+          prefixContent = <NxFontAwesomeIcon icon={filterIcon} className="nx-icon--filter-icons" />,
+          suffixContent =
+            <NxButton className={btnClassName}
+                      variant="icon-only"
+                      title="clear filter"
+                      onClick={clearFilterInputText}
+                      tabIndex={-1}><Close/>
```
rpokorny, line 42:
```suggestion
                      tabIndex={-1}>
              <Close/>
```

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -23,16 +26,43 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
+            </NxButton>;
+
+      function clearFilterInputText() {
```
rpokorny, line 45: This seems way over-complicated.  Is there a reason you can't just call the `onChange` prop with `''`?

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -12,10 +12,13 @@ import { faFilter, faSearch } from '@fortawesome/free-solid-svg-icons';
...
 import { PrivateNxTextInput } from '../../NxTextInput/NxTextInput';
 import NxFontAwesomeIcon from '../../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import NxButton from '../../NxButton/NxButton';
+import Close from '../../../icons/Close';
 
 describe('NxFilterInput', function() {
   const minimalProps = { value: '' },
-      shallowComponent = enzymeUtils.getShallowComponent<Props>(NxFilterInput, minimalProps);
+      shallowComponent = enzymeUtils.getShallowComponent<Props>(NxFilterInput, minimalProps),
+      MountedComponent = enzymeUtils.getMountedComponent<Props>(NxFilterInput, minimalProps);
```
rpokorny, line 21:
```suggestion
      mountedComponent = enzymeUtils.getMountedComponent<Props>(NxFilterInput, minimalProps);
```

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -96,6 +99,38 @@ describe('NxFilterInput', function() {
...
+  it('clears input text when the Escape key is pressed', function() {
+    const onKeyDown = jest.fn(),
+        component = shallowComponent({ onKeyDown, value: 'a' });
+
+    component.simulate('keyDown', { key: 'Escape' });
+
+    expect(component).toHaveText('');
+  });
+
+  it('clears input text when the clear filter button is clicked', function() {
```
rpokorny, line 123: This is a stateless component.  The effect of the button/ESC shouldn't really be to clear the text, it should be to fire `onChange` with the empty string.  It's up to typical calling code to respond to that by passing `''` as the new `value`, causing the component text to actually become empty.

gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
cindyko226: Updated here be92dbb5730ca5e08c7792134dfb29890331f319. Thanks.

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -24,4 +24,18 @@
...
+    margin: 0;
+    padding-top: calc(var(--nx-spacing-base) - 1px);
+    padding-bottom: calc(var(--nx-spacing-base) - 1px);
+    position: absolute;
+    right: var(--nx-spacing-2x);
+    width: calc(1em * var(--nx-line-height) + 2 * var(--nx-spacing-base));
```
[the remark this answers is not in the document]
cindyko226: Updated here 3f2ace0faaf74836ac3fc7e0235375dfd8afc734, thanks.

lib/src/components/NxFilterInput/NxFilterInput.tsx
rpokorny: > But then I found a way to avoid the type warning https://github.com/sonatype/sonatype-react-shared-components/commit/4c7a67337521345304121c8fc8bb88c051e2bd83, not sure if this is okay, please let me know your opinions, thanks.

Using `partial` there to just break the typechecking isn't good, a more straightforward way to do the same thing would be to just cast to `any`.  But of course we'd rather not break the typechecking.  It's interesting that the type for `onChange` here doesn't match what's described in the documentation.  The type has it taking two parameters, the string and the event, while the gallery docs make no mention of the event.  Since the docs are silent about it anyway, I think we can change the type to indicate that the second parameter is optional.

cindyko226: Thanks for the opinions and suggestions! Here is the updates 06843871f8a0937f619868f15845bf540ccd0893.

cindyko226:
> Doesn't look good on my system with the datalist example

Seems like it only shows in Chrome, I update the CSS to hide the dropdown arrow, f97f056ff4c7d092422bcf6b7e40373ab6a50907.


> Using a datalist here is probably now redundant with `NxCombobox`. I wonder if we should deprecate it. 

I am not 100% sure if it should be deprecated. In terms of filter/search function, it is kinda redundant, but it supports icon and clear button whereas combobox does not.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Update here cc5699c70171de5574aae0a2e70aa4bbe6cad3b5.
The RTL unit tests will be in another PR.

lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -19,9 +20,25 @@
     }
   }
 
+  .nx-text-input__input::-webkit-calendar-picker-indicator {
+      display: none !important;
```
rpokorny, line 24: If you use `visibility: hidden` you don't need the `!important`

rpokorny:
> I am not 100% sure if it should be deprecated. In terms of filter/search function, it is kinda redundant, but it supports icon and clear button whereas combobox does not.

I'd check with design about guidance for what we want to push people towards going forward.  It doesn't seem good to me to have several components that do almost the same thing but with slight differences in capabilities that might force a downstream app to compromise if no single option does everything they want.

I also think design should be asked about the `datalist` dropdown arrow, whether hiding it was the right move.

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -23,16 +26,39 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
+        if (props.onChange) {
+          props.onChange('');
+        }
+      }
+
+      const handleKeyDown: KeyboardEventHandler<HTMLInputElement | HTMLTextAreaElement> = (e) => {
+        if (e.key === 'Escape') {
+          clearFilterInputText();
```
rpokorny, line 54: Should this also do a `preventDefault`? I'm imagining if we have something like a filter input inside of a modal, we don't want Escape here to close the modal. `NxDropdown` has similar logic.

rpokorny: > Should this also do a preventDefault? I'm imagining if we have something like a filter input inside of a modal, we don't want Escape here to close the modal. NxDropdown has similar logic.

In fact let's add one of these to the modal form example and (once we've decided how it should behave) test that it behaves correctly there.

lib/src/components/NxFilterInput/NxFilterInput.scss
cindyko226: With `visibility: hidden`, it takes up space in the input field, but I will double check with design if it should be hidden.

lib/src/components/NxFilterInput/NxFilterInput.tsx
cindyko226: I test it in the modal form example and it closes the modal when Escape is pressed, also notice it is mentioned on the modal page doc that any components within the modal that have their own ESC handling should call `preventDefault` on any ESC keydowns that they handle in order to prevent the modal from also handling them. 
Updated here 22e4fcdbfb6fdfc0dfbfc3c5cb87df9bd5c49cd2. Thanks for catching this.

cindyko226:
> Text should not be visible behind the button. Check with design on exactly where text should truncate.

Confirmed with design @JenniKobylski that there should be 24px between icon only button and the end of text 3662430b99da0f8ff0ae9c1a3e5708c09c00d2af.

Updated mockups: https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/3OJr0zm.

cindyko226:
> It doesn't seem like design intended/is aware that this affect `NxSearchDropdown`, as the mockups for that component do not have this button. Check with them whether it should (as it does in this PR). If so, the mockups should be updated. Also check about the filter boxes within `NxTransferListHalf`

Updated mockups:
Transfer List - https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/m1095Gz
Search - https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/R13r33j

cindyko226:
> I'd check with design about guidance for what we want to push people towards going forward. It doesn't seem good to me to have several components that do almost the same thing but with slight differences in capabilities that might force a downstream app to compromise if no single option does everything they want.

As discussion with design, they agree the usage with `datalist` should be deprecated and use Combobox instead. Updated desc here 56a6b584c300643b60efb45b23cfced0f07d4ca5.

> I also think design should be asked about the `datalist` dropdown arrow, whether hiding it was the right move.

It should be hidden.

rpokorny:
> As discussion with design, they agree the usage with datalist should be deprecated and use Combobox instead. Updated desc here https://github.com/sonatype/sonatype-react-shared-components/commit/56a6b584c300643b60efb45b23cfced0f07d4ca5.

In that case should we create a ticket to add the option for filter and magnifying glass icons to `NxCombobox`? What about the clear button like we're adding here to `NxFilterInput`.  In other words, should `NxCombobox` be able to do everything this component can do plus more?

gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
```diff
@@ -25,7 +25,10 @@ const NxFilterInputPage = () =>
   <>
     <GalleryDescriptionTile>
       <NxP>
-        An input to be used for filtering purposes
+        An input to be used for filtering purposes. After a user has text within the input field, a
+        clear filter button appears within the component and is only visible/usable when there is text within
+        the input field. Note: This button is for mouse use only and is therefore not focusable but still
+        hoverable and clickable.
```
rpokorny, line 31: Maybe add a mention of the clear-on-escape functionality here too

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -23,16 +26,40 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
           // just in case these props get passed in, avoid passing them to NxTextInput as they would cause
           // malfunction
           cleanedProps = omit(['validatable', 'validationErrors', 'type'], otherProps),
           filterIcon = searchIcon ? faSearch : faFilter,
-          prefixContent = <NxFontAwesomeIcon icon={filterIcon} className="nx-icon--filter-icons" />;
+          prefixContent = <NxFontAwesomeIcon icon={filterIcon} className="nx-icon--filter-icons" />,
+          suffixContent =
+            <NxButton className={btnClassName}
+                      variant="icon-only"
+                      title="clear filter"
```
rpokorny, line 40: Should the title be "clear search" when `searchIcon` is true?

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -23,16 +26,40 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
...
+                      onClick={clearFilterInputText}
+                      tabIndex={-1}>
+              <Close/>
+            </NxButton>;
+
+      function clearFilterInputText() {
+        if (props.onChange) {
+          props.onChange('');
+        }
```
rpokorny, line 49:
```suggestion
        props.onChange?.('');
```

lib/src/components/NxFilterInput/NxFilterInput.tsx
```diff
@@ -23,16 +26,40 @@ const NxFilterInput = forwardRef<HTMLDivElement, Props>(
           className = classnames('nx-filter-input', classNameProp, {
             'nx-filter-input--empty': isEmpty
           }),
-
+          btnClassName = classnames('nx-btn--clear', { 'hidden': isEmpty }),
+          fieldRef = useRef<HTMLDivElement>(null),
```
rpokorny, line 30: Not needed anymore, which means the `mergedRef` isn't either

cindyko226:
> In that case should we create a ticket to add the option for filter and magnifying glass icons to NxCombobox? What about the clear button like we're adding here to NxFilterInput. In other words, should NxCombobox be able to do everything this component can do plus more?

I think it is a good idea, will check with design about this!

lib/src/components/NxFilterInput/NxFilterInput.tsx
cindyko226: It makes sense to me but will confirm with design.

gallery/src/components/NxFilterInput/NxFilterInputPage.tsx
cindyko226: Updated here 447daff63725ceb4aa77ad580798f2adac4d1a05. Thanks.

lib/src/components/NxFilterInput/NxFilterInput.tsx
cindyko226: Updated here 80b272ee24ecab0f0a8279530895d1bbb9098bcb. Thanks.

lib/src/components/NxFilterInput/NxFilterInput.tsx
cindyko226: Confirmed with designer @JasonDSwearingen that it should read "Clear search" when `searchIcon` is true. 
Updated mockups https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/g0GK9mj.
Updated title 7037e5e100f7b9e7829500a1b141c7b9a7ded913
Updated unit tests 6db707ba13ca7b6bbaba1769b0d74d0551a74628
Updated text with sentence case to match final mockups 506f08770df82ddf592cfc324fa6aa904516f5d0

JasonDSwearingen (approved):

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -96,6 +99,61 @@ describe('NxFilterInput', function() {
...
+        },
+        btn = shallow(<BtnFixture />);
+
+    expect(btn).toExist();
+    expect(btn).toMatchSelector(NxButton);
+    expect(btn.children()).toMatchSelector(Close);
+    expect(btn).toHaveClassName('nx-btn--clear');
+  });
+
+  it('sets the title as "Clear filter" to the Nxbutton when searchIcon is undefined', function() {
```
rpokorny, line 114: Also test `false` and `null` in this test.

rpokorny (approved):

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Updated here e4cd31e9323b6c71452478b70739de92d89bd170. Thanks.

PR #826, opened 2022-10-12
Change NxToast Animation - RSC-1200

rjp99:
https://issues.sonatype.org/browse/RSC-1200


rpokorny:
The toast animation description in sketch no longer makes any mention of the fade-in nor of the movements of other toasts up and down to make room for a new toast or to remove the space for a closing toast.  As I think about that, I think it calls into question this whole ticket - the animations for the two components are fundamentally different.  Drawers _just_ slide in from the side while toasts do that plus other things some of which happen before or after the slide.  So matching the total durationss of the animations doesn't necessarily make sense; it's not like it results in the durations of the slides themselves matching.  I think this needs to go back to design

lib/src/components/NxToast/NxToast.scss
```diff
@@ -39,12 +39,12 @@
...
   opacity: 1;
   margin-top: var(--nx-spacing-2x);
-  animation: animationIn 0.4s linear;
+  animation: animationIn 300ms ease-out forwards;
   overflow: hidden;
   pointer-events: auto;
 
   &.nx-toast--closing {
-    animation-name: animationOut;
+    animation: animationOut 250ms ease-in forwards;
```
rpokorny, line 47: Specifying the `animation-fill-mode` (the `forwards`) doesn't seem necessary to me, I think it's working without it.

rjp99: > Specifying the `animation-fill-mode` (the `forwards`) doesn't seem necessary to me, I think it's working without it.

I removed it here: cbfa9f3d3854e60cc99e25d18e159c831eb5e65e

rpokorny (approved):

PR #827, opened 2022-10-14
NxBreadcrumb - RSC-1156

rpokorny:
https://issues.sonatype.org/browse/RSC-1156


lib/src/components/NxBreadcrumb/NxBreadcrumb.tsx
```diff
@@ -0,0 +1,99 @@
...
+        crumbsToBecomeLinks = [crumbBeforeDropdown, ...init(crumbsAfterDropdown)],
+
+        // if we're in this block, the list was at least two items long so last on the "after" crumbs will
+        // always have a value
+        /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+        currentCrumb = last(crumbsAfterDropdown)!,
+        currentCrumbRender = <CurrentBreadcrumb key={currentCrumb.href} { ...currentCrumb } />,
+        rowLinks = map(c => <BreadcrumbLink key={c.href} { ...c } />, crumbsToBecomeLinks),
+        dropdown = dropdownCrumbs.length ?
+          <BreadcrumbDropdown key="%%nx-breadcrumb-dropdown"
```
cindyko226, line 76: Is the "%%" here on purpose?

gallery/src/components/NxBreadcrumb/NxBreadcrumbPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+            Click here to navigate to the live example.
+          </NxTextLink>
+        </NxP>
+        <NxStatefulAccordion>
+          <NxAccordion.Header>
+            <NxAccordion.Title>Example Code</NxAccordion.Title>
+          </NxAccordion.Header>
+          <CodeExample content={globalheaderExampleCode} />
+          <GalleryTileFooter clipboardContent={globalheaderExampleCode}/>
+        </NxStatefulAccordion>
```
cindyko226, line 212:
```suggestion
      </NxTile.Content>
      <NxTile.Content className= "nx-tile-content--accordion-container">
        <NxStatefulAccordion>
          <NxAccordion.Header>
            <NxAccordion.Title>Example Code</NxAccordion.Title>
          </NxAccordion.Header>
          <CodeExample content={globalheaderExampleCode} />
          <GalleryTileFooter clipboardContent={globalheaderExampleCode}/>
        </NxStatefulAccordion>
```

gallery/src/pageConfig.ts
```diff
@@ -178,6 +180,8 @@ const pageConfig: PageConfig = {
   },
   'Buttons and Dropdowns': {
     'Back Button': { content: NxBackButtonPage, type: 'react' },
+    'Breadcrumb': { content: NxBreadcrumbPage, type: 'react' },
+    'StatefulBreadcrumb': { content: NxStatefulBreadcrumbPage, type: 'react' },
```
cindyko226, line 184:
```suggestion
    'Stateful Breadcrumb': { content: NxStatefulBreadcrumbPage, type: 'react' },
```

gallery/visualtests/nxBreadcrumb.js
```diff
@@ -0,0 +1,95 @@
...
+
+  describe('in NxGlobalHeader', function() {
+    const { checkFullPageScreenshot, a11yTest, getPage } = setupBrowser('#/NxBreadcrumbGlobalHeaderExample', false);
+
+    it('looks right', async function() {
+      await getPage().setViewport({ width: 1366, height: 800 });
+      await checkFullPageScreenshot();
+    });
+
+    it('passes a11y checks', a11yTest(null, true));
```
cindyko226, line 91: Out of curiosity, is the `a11Test` testing the `NxBreadcrumbGlobalHeaderExample` page without calling `getPage().setViewport({ width: 1366, height: 800 });` here?

gallery/visualtests/nxBreadcrumb.js
```diff
@@ -0,0 +1,95 @@
...
+        longSegmentsExample,
+        `${longSegmentsExample} a`
+    );
+
+    await link.hover();
+    await wait(500);
+
+    const { x, y, width, height } = await example.boundingBox();
+
+    await checkScreenshotCoordinates(x, y - 21, width, height + 21);
```
cindyko226, line 65: Is there a reason to specify the coordinates here instead of using `checkScreenshot`?

lib/src/components/NxBreadcrumb/__tests__/NxBreadcrumb.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    expect(link2).toHaveAccessibleName('B');
+    expect(link2).toHaveTextContent('B');
+    expect(link2).toHaveAttribute('href', 'b');
+
+    expect(within(finalItem).queryByRole('link')).not.toBeInTheDocument();
+    expect(finalItem).toHaveTextContent('C');
+  });
+
+  it('renders four crumbs in a list with the first two being links', function() {
```
cindyko226, line 102:
```suggestion
  it('renders four crumbs in a list with the first three being links', function() {
```

lib/src/components/NxBreadcrumb/stateful/__tests__/NxStatefulBreadcrumb.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+    expect(link2).toHaveAccessibleName('B');
+    expect(link2).toHaveTextContent('B');
+    expect(link2).toHaveAttribute('href', 'b');
+
+    expect(within(finalItem).queryByRole('link')).not.toBeInTheDocument();
+    expect(finalItem).toHaveTextContent('C');
+  });
+
+  it('renders four crumbs in a list with the first two being links', function() {
```
cindyko226, line 100:
```suggestion
  it('renders four crumbs in a list with the first three being links', function() {
```

cindyko226:
Do we need the screenshot of tooltip on dropdown button?

lib/src/components/NxBreadcrumb/NxBreadcrumb.tsx
rpokorny: > Is the "%%" here on purpose?

Yes; all of the other items in this list use the `href` values (which should be URLs) as their keys.  Keys need to be unique of course, so the way to ensure that this one is unique is to make it a string that cannot be a valid `href`.  This was the first way I thought of to do so - `%` is the escape character in URLs and should always (in URLs) be followed by two hex digits, so following it with another `%` instead is one way to construct a string that is not a valid URL or URL part.

gallery/visualtests/nxBreadcrumb.js
rpokorny: > Out of curiosity, is the a11Test testing the NxBreadcrumbGlobalHeaderExample page without calling getPage().setViewport({ width: 1366, height: 800 }); here?

Yes, the custom viewport size on the visual test was just to avoid being wasteful with a needlessly humongous image.  Our full-page screenshots are quite tall by default now

rpokorny:
> Do we need the screenshot of tooltip on dropdown button?

Doesn't hurt. a8972d104

lib/src/components/NxBreadcrumb/NxBreadcrumb.tsx
cindyko226: Thanks for the explanation.

gallery/visualtests/nxBreadcrumb.js
cindyko226: Got it, yes the default viewport hight is 3000 by default.

lib/src/components/NxBreadcrumb/types.ts
```diff
@@ -0,0 +1,41 @@
...
+  crumbs: Crumb[];
+}
+
+export interface Props extends StatefulProps {
+  isDropdownOpen: boolean;
+  onToggleDropdown: () => void;
+}
+
+export const statefulPropTypes: PropTypes.ValidationMap<StatefulProps> = {
+  crumbs: PropTypes.arrayOf(PropTypes.shape({
```
EmilyBermudez88, line 31: Just for clarification here, is using `PropTypes.shape` a form of validation, to make sure that each crumb contains the two specified props/keys (name + href)?

lib/src/components/NxBreadcrumb/types.ts
```diff
@@ -0,0 +1,41 @@
...
+    name: PropTypes.string.isRequired,
+    href: PropTypes.string.isRequired
+  }).isRequired).isRequired
```
EmilyBermudez88, line 34: And for additional clarification, the first `.isRequired` on line 34 refers to the shape of each crumb, and the second `.isRequired` refers to the array of crumbs being a required prop?

gallery/src/components/NxBreadcrumb/NxBreadcrumbPage.tsx
```diff
@@ -0,0 +1,219 @@
...
+      proportionally such that the component maintains its default maximum width.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Example with Constrained Width"
+                        id="nx-breadcrumb-constrained-width-example"
+                        liveExample={NxBreadcrumbConstrainedWidthExample}
+                        codeExamples={constrainedWidthExampleCode}>
+      An example of <NxCode>NxBreadcrumb</NxCode> within a container that is narrower
+      than <NxCode>NxBreadcrumb</NxCode>'s maximum width. <NxCode>NxBreadcrumb</NxCode> shrinks to fit within the
+      container, and the truncation of the longs adjusts accordingly.
```
EmilyBermudez88, line 148:
```suggestion
      container, and the truncation of the long names adjust accordingly.
```

JenniKobylski (approved):
Excellent work!

lib/src/components/NxBreadcrumb/types.ts
rpokorny: > And for additional clarification, the first .isRequired on line 34 refers to the shape of each crumb, and the second .isRequired refers to the array of crumbs being a required prop?

Yeah all propTypes validator accept `null` and `undefined` by default in addition to what they normally do, and you have to use the variation attached to their `isRequired` property to avoid that.  So here the inner `isRequired` effectively means that the array children cannot be `null` or `undefined`, and the outer one means that the array itself cannot be `null` or `undefined`

gallery/src/components/NxBreadcrumb/NxBreadcrumbPage.tsx
rpokorny: @JenniKobylski just had me change how the truncation works. I fixed this typo as part of the corresponding doc update. 90768ca35

cindyko226:
Do we need more padding on the right of each dropdown link as mentioned in the mockups?
Mockups: 
Screen Shot - mockups

Current implementation:
Screen Shot - dropdown

lib/src/components/NxBreadcrumb/types.ts
EmilyBermudez88: excellent, thanks!

lib/src/components/NxBreadcrumb/types.ts
EmilyBermudez88: thanks for the confirmation

rpokorny:
> Do we need more padding on the right of each dropdown link as mentioned in the mockups?

Yes, I missed that. e5606bb65

cindyko226 (approved):
+1

PR #829, opened 2022-10-14
Allow NxStatefulForm to work even without submit mask usage - RSC-1209

rpokorny:
https://issues.sonatype.org/browse/RSC-1209


EmilyBermudez88:
I am not sure if this is the exact purview of this ticket, but when testing the horizontal form layout example under 'form layouts', I see that the `NxFormSelect` needs the validatable prop. Currently, the form can still be submitted if it is left with the default 'select an option'

rpokorny:
> I am not sure if this is the exact purview of this ticket, but when testing the horizontal form layout example under 'form layouts', I see that the NxFormSelect needs the validatable prop. Currently, the form can still be submitted if it is left with the default 'select an option'

That's the other one, RSC-1208

EmilyBermudez88:
> > I am not sure if this is the exact purview of this ticket, but when testing the horizontal form layout example under 'form layouts', I see that the NxFormSelect needs the validatable prop. Currently, the form can still be submitted if it is left with the default 'select an option'
> 
> That's the other one, RSC-1208

Ah ok, I was starting from the top down and should have reviewed in the other direction!

lib/src/components/NxForm/stateful/__tests__/NxStatefulForm.test.tsx
```diff
@@ -350,77 +354,28 @@ describe('NxStatefulForm', function() {
 
     await user.click(screen.getByRole('button', { name: 'Submit' }));
```
cindyko226, line 355: Could you please explain why we have `await user.click on Submit button` here, is it to test that clicking on Submit button will not set `showValidationErrors` to true when there is no validationErrors? 
Seems like the line# 357 will still pass without this.

lib/src/components/NxForm/stateful/__tests__/NxStatefulForm.test.tsx
```diff
@@ -441,12 +396,16 @@ describe('NxStatefulForm', function() {
     // only changing back to not having validationErrors counts
```
EmilyBermudez88, line 396: does this test need to have a user.click event? Or is it considered addressed on line 365?

lib/src/components/NxForm/stateful/__tests__/NxStatefulForm.test.tsx
rpokorny: > Seems like the line# 357 will still pass without this.

Yes, but we want to make sure it passes _even with_ this.  As the two tests above this one establish, `showValidationErrors` always starts false and only gets set to true when the submit button is clicked and the `validationErrors` prop has certain types of values.  In this test here, I basically am testing a wide variety of transitions between different kinds of `validationErrors` values. For each transition where `showValidationErrors` was already `false`, I have to click the submit button to find out whether the transition did anything, because even if that transition would change `showValidationErrors`, it won't do so until the button is clicked.

lib/src/components/NxForm/stateful/__tests__/NxStatefulForm.test.tsx
rpokorny: See https://github.com/sonatype/sonatype-react-shared-components/pull/829/files/b444d99fb300a2f0c8894ee54a14abee26b45cbb#r1002072022 .  I add the click to each transition where the previous `showValidationErrors` value was `false`.

lib/src/components/NxForm/stateful/__tests__/NxStatefulForm.test.tsx
cindyko226: Thanks for the explanation.

lib/src/components/NxForm/stateful/__tests__/NxStatefulForm.test.tsx
EmilyBermudez88: > See https://github.com/sonatype/sonatype-react-shared-components/pull/829/files/b444d99fb300a2f0c8894ee54a14abee26b45cbb#r1002072022 . I add the click to each transition where the previous `showValidationErrors` value was `false`.

Ah I see, thanks for linking the explanation!

PR #831, opened 2022-10-25
Convert NxFilterInput unit tests to RTL - RSC-1212

cindyko226:
https://issues.sonatype.org/browse/RSC-1212

### DIFF
--- a/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
+++ b/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
@@ -85,9 +85,8 @@ describe('NxFilterInput', function() {
 
   it('fires onChange with the empty string when the Escape key is pressed', async function() {
     const user = userEvent.setup(),
-        onKeyDown = jest.fn(),
         onChange = jest.fn(),
-        el = quickRender({ onKeyDown, value: 'a', onChange }),
+        el = quickRender({ value: 'a', onChange }),
         input = el.getByRole('textbox');
 
     expect(onChange).not.toHaveBeenCalled();
--- a/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
+++ b/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
@@ -58,12 +58,21 @@ describe('NxFilterInput', function() {
     expect(ref.current).toBe(el);
   });
 
-  it('renders a button with an accessible name of "Clear filter" when searchIcon is undefined', async function() {
-    const clearBtn = quickRender().getByRole('button');
+  it('renders a button with an accessible name of "Clear filter" when searchIcon is undefined, false or null',
+      async function() {
+        const clearBtn = quickRender().getByRole('button'),
+            clearBtnWithSearchIconFalse = quickRender({ searchIcon: false }).getByRole('button'),
+            clearBtnWithSearchIconNull = quickRender({ searchIcon: null }).getByRole('button');
 
-    expect(clearBtn).toBeInTheDocument();
-    await waitFor(() => expect(clearBtn).toHaveAccessibleName('Clear filter'));
-  });
+        expect(clearBtn).toBeInTheDocument();
+        await waitFor(() => expect(clearBtn).toHaveAccessibleName('Clear filter'));
+
+        expect(clearBtnWithSearchIconFalse).toBeInTheDocument();
+        await waitFor(() => expect(clearBtnWithSearchIconFalse).toHaveAccessibleName('Clear filter'));
+
+        expect(clearBtnWithSearchIconNull).toBeInTheDocument();
+        await waitFor(() => expect(clearBtnWithSearchIconNull).toHaveAccessibleName('Clear filter'));
+      });
 
   it('renders a button with an accessible name of "Clear search" when searchIcon is true', async function() {
     const clearBtn = quickRender({ searchIcon: true }).getByRole('button');
--- a/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
+++ b/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
@@ -12,9 +12,9 @@ import userEvent from '@testing-library/user-event';
 import NxFilterInput, { Props } from '../NxFilterInput';
 
 describe('NxFilterInput', function() {
-  const minimalProps = { value: '' },
-      quickRender = rtlRender<Props>(NxFilterInput, minimalProps),
-      renderEl = rtlRenderElement<Props>(NxFilterInput, minimalProps);
+  const minimalProps: Props = { value: '' },
+      quickRender = rtlRender(NxFilterInput, minimalProps),
+      renderEl = rtlRenderElement(NxFilterInput, minimalProps);
 
   it('renders an Input', function() {
     expect(quickRender().getByRole('textbox').tagName).toBe('INPUT');
--- a/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
+++ b/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
@@ -69,11 +69,18 @@ describe('NxFilterInput', function() {
     expect(icons[0]).toHaveAttribute('data-icon', 'search');
   });
 
-  it('renders a button with an accessible name of "clear filter"', async function() {
+  it('renders a button with an accessible name of "Clear filter" when searchIcon is undefined', async function() {
     const clearBtn = quickRender().getByRole('button');
 
     expect(clearBtn).toBeInTheDocument();
-    await waitFor(() => expect(clearBtn).toHaveAccessibleName('clear filter'));
+    await waitFor(() => expect(clearBtn).toHaveAccessibleName('Clear filter'));
+  });
+
+  it('renders a button with an accessible name of "Clear search" when searchIcon is true', async function() {
+    const clearBtn = quickRender({ searchIcon: true }).getByRole('button');
+
+    expect(clearBtn).toBeInTheDocument();
+    await waitFor(() => expect(clearBtn).toHaveAccessibleName('Clear search'));
   });
 
   it('fires onChange with the empty string when the Escape key is pressed', async function() {
--- a/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
+++ b/lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
@@ -5,7 +5,7 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React, { RefAttributes } from 'react';
-import { waitFor } from '@testing-library/react';
+import { waitFor, screen } from '@testing-library/react';
 import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
 import userEvent from '@testing-library/user-event';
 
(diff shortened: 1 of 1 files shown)


lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,109 @@
...
       type: 'textarea'
-    } as Partial<Props>);
-
-    expect(component).not.toHaveProp('validatable');
-    expect(component).not.toHaveProp('validationErrors');
-    expect(component).not.toHaveProp('type');
-  });
+    } as Partial<Props>).getByRole('textbox');
 
-  it('passes isPristine = false to the PrivateNxTextInput', function() {
```
cindyko226, line 63: The `isPristine=false` is passed down to `PrivateNxTextInput` and is used to check whether to have `showValidationErrors` and add `pristine` classname in the `NxTextInput` component, for this it assertion, should we just test if the input does not have the `pristine` classname? Please kindly let me know your opinions, thanks.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,109 @@
...
-    expect(btn).toHaveProp('title', 'Clear search');
+    expect(clearBtn).toBeInTheDocument();
+    await waitFor(() => expect(clearBtn).toHaveAccessibleName('Clear search'));
   });
 
-  it('fires onChange with the empty string when the Escape key is pressed', function() {
-    const onKeyDown = jest.fn(),
+  it('fires onChange with the empty string when the Escape key is pressed', async function() {
+    const user = userEvent.setup(),
+        onKeyDown = jest.fn(),
```
EmilyBermudez88, line 88: Do you need `onKeyDown` here?

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,109 @@
...
-          onChange,
-          onKeyPress,
-          id: 'bar'
-        });
-
-    expect(component).toHaveProp('value', 'foo');
-    expect(component).toHaveProp('onChange', onChange);
-    expect(component).toHaveProp('onKeyPress', onKeyPress);
-    expect(component).toHaveProp('id', 'bar');
+  it('passes additional props to the Input', function() {
```
EmilyBermudez88, line 34:
```suggestion
  it('passes additional attrs to the Input', function() {
```

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,109 @@
...
+import userEvent from '@testing-library/user-event';
 
 import NxFilterInput, { Props } from '../NxFilterInput';
-import { PrivateNxTextInput } from '../../NxTextInput/NxTextInput';
-import NxFontAwesomeIcon from '../../NxFontAwesomeIcon/NxFontAwesomeIcon';
-import NxButton from '../../NxButton/NxButton';
-import Close from '../../../icons/Close';
 
 describe('NxFilterInput', function() {
   const minimalProps = { value: '' },
```
EmilyBermudez88, line 15: If you assign minimalProps as type Props here, you would be able to get rid of <Props> in quickRender and renderEl below. Just a thought, and probably personal preference!

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: No, thanks for catching it, updated here c801738d2fa4bbfeb10c78ace8bf91b46a0ee65b.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Thanks, it is cleaner, updated here c79b2d1ca6a07085d952e9a2ff1a27c46df4219d.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,108 @@
...
+  it('passes additional attrs to the Input', function() {
+    const input = quickRender({ id: 'foo', lang: 'en-US' }).getByRole('textbox');
+
+    expect(input).toHaveAttribute('id', 'foo');
+    expect(input).toHaveAttribute('lang', 'en-US');
   });
 
-  it('does not pass validatable, validationErrors, or type props to the PrivateNxTextInput', function() {
-    const component = shallowComponent({
+  it('does not pass validatable, validationErrors, or type props to the Input', function() {
```
rpokorny, line 41: Testing that `type` is ignored in this way is valid, but for the others they wouldn't have become attributes on the `<input>`  anyway, so I think a different approach is needed.  Maybe check that `aria-invalid` doesn't get set to true, that `aria-errormessage` doesn't get set, and that the `validationErrors` text doesn't show up anywhere in the rendered DOM.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,108 @@
...
-    expect(shallowComponent()).toHaveProp('isPristine', false);
+    expect(input).not.toHaveAttribute('validatable', true);
+    expect(input).not.toHaveAttribute('validationErrors', 'It\'s all wrong');
+    expect(input).not.toHaveAttribute('type', 'textarea');
   });
 
-  it('puts the ref on the PrivateNxTextInput', function() {
+  it('sets ref on the Input', function() {
     const ref = React.createRef<HTMLDivElement>(),
+        el = renderEl({ ref } as Partial<Props>);
```
rpokorny, line 55: If you change the type of `minimalProps` to `Props & RefAttributes<HTMLDivElement>` then you don't need this cast.  `RefAttributes` is an export from `react`.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,108 @@
...
+
+    expect(icons[0]).toHaveAttribute('data-icon', 'filter');
   });
 
-  it('passes an NxButton with the Close icon and nx-btn--clear classname as the suffixContent', function() {
-    const BtnFixture = function() {
-          return shallowComponent().prop('suffixContent');
-        },
-        btn = shallow(<BtnFixture />);
+  it('renders a search icon when searchIcon is true', function() {
```
rpokorny, line 66: As we convert to RTL some things that were unit tests will need to become visual tests.  Choice of icon is one of those things.  I think these two should be removed.  The corresponding visual tests already exist.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
```diff
@@ -5,160 +5,108 @@
...
-    expect(btn).toHaveClassName('nx-btn--clear');
+    expect(icons[0]).toHaveAttribute('data-icon', 'search');
   });
 
-  it('sets the title as "Clear filter" to the Nxbutton when searchIcon is undefined', function() {
-    const BtnFixture = function() {
-          return shallowComponent().prop('suffixContent');
-        },
-        btn = shallow(<BtnFixture />);
+  it('renders a button with an accessible name of "Clear filter" when searchIcon is undefined', async function() {
```
rpokorny, line 72: As in the other PR, also test `false` and `null`.

rpokorny:
I'm realizing that this is one that has a dependency on another - `NxTextInput`.  Ideally, we would have done `NxTextInput`'s RTL conversion first, and then copied the tests for the common behavior (basic text input behavior) into the `NxFilterInput` tests in this PR.  Since we failed to notice that before you did this work, I'd say that from here what you can do is either go ahead and port `NxTextInput` in this PR as well, and copy the relevant tests into `NxFilterInput` as well, or test write tests for those common aspects directly on `NxFilterInput` and then we'll copy them the opposite direction once we get to `NxTextInput` in the future.  But either way, with RTL `NxFilterInput` will need direct tests of all of its functionality, including the functionality that it inherits from `NxTextInput`.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Updated here 85138bcbbccb76c45f2c9c54cc9b8a3ab70a7c4b.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Removed here 022713cb070daad192ed4d837776f74f260c2fef.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Thanks for the info. Updated here 7cd3180bedc9a6842ee950fc523e59059ac33e60.

lib/src/components/NxFilterInput/__tests__/NxFilterInput.test.tsx
cindyko226: Updated here 67f47c9c38f4b270233459595b587866d52e5173. Thanks.
For the `aria-errormessage`, it is set with a unique ID if not disabled in `NxTextInput`. 
So there is `aria-errormessage` when valid, seems like it is correct as long as the element referenced by the `aria-message` is hidden according to the WAI-ARIA doc https://www.w3.org/TR/wai-aria-1.2/#aria-errormessage 

> When an object is in a valid state, it has either "aria-invalid" set to false or it does not have the "aria-invalid" attribute. Authors MAY use aria-errormessage on an object that is currently valid, but only if the element referenced by aria-errormessage is "hidden", because the message it contains is not pertinent.

rpokorny: > For the aria-errormessage, it is set with a unique ID if not disabled in NxTextInput.
So there is aria-errormessage when valid, seems like it is correct as long as the element referenced by the aria-message is hidden according to the WAI-ARIA doc https://www.w3.org/TR/wai-aria-1.2/#aria-errormessage

OK, looking closer what happens is that the `aria-errormessage` attribute is set but the element that it refers to is not present in the DOM.  I'm not totally sure that's valid, but our a11y tests seem to be happy with it so I guess it is.  At any rate, the jest RTL matchers include a [`toHaveErrorMessage`](https://github.com/testing-library/jest-dom#tohaveerrormessage) matcher which I think would be good to use here `(with a `not`)

cindyko226: > OK, looking closer what happens is that the aria-errormessage attribute is set but the element that it refers to is not present in the DOM. I'm not totally sure that's valid, but our a11y tests seem to be happy with it so I guess it is.

This is out of scope of this ticket, but do you think it might be much more clear if we apply the `aria-errormessage` attr to the input only when it's invalid?

> At any rate, the jest RTL matchers include a [toHaveErrorMessage](https://github.com/testing-library/jest-dom#tohaveerrormessage) matcher which I think would be good to use here (with a not`)

Thanks for the suggestion, with this matcher I think I don't need the query of alert and error text, updated here 914b2f7062be0607d25692deaa454124621921a1.

rpokorny: > but do you think it might be much more clear if we apply the aria-errormessage attr to the input only when it's invalid?

A little bit maybe, but what we have currently isn't really wrong according to the a11y tests.

cindyko226:
> or test write tests for those common aspects directly on NxFilterInput and then we'll copy them the opposite direction once we get to NxTextInput in the future. But either way, with RTL NxFilterInput will need direct tests of all of its functionality, including the functionality that it inherits from NxTextInput.

Updated here
d0614577ae581d68bb15c4eb3ddace6b64dd4b8c
edfcec5d4e41a3b07e18dee3a617035bde410b82 I removed `onKeyPress` here and think it should be tested within `NxTextInput`, but I might be wrong, please let me know your opinions. Thanks.

rpokorny:
> https://github.com/sonatype/sonatype-react-shared-components/commit/edfcec5d4e41a3b07e18dee3a617035bde410b82 I removed onKeyPress here and think it should be tested within NxTextInput, but I might be wrong, please let me know your opinions. Thanks.

I think you misunderstand.  Everything about `NxTextInput` that is relevant to `NxFilterInput` must also be tested on `NxFilterInput` - the fact that `NxFilterInput` uses `NxTextInput` internally is an implementation detail.  So I would expect to see tests for:
1. `onKeyPress`
2. `onChange`
3. `disabled`

cindyko226:
> I think you misunderstand. Everything about NxTextInput that is relevant to NxFilterInput must also be tested on NxFilterInput - the fact that NxFilterInput uses NxTextInput internally is an implementation detail. So I would expect to see tests for:
`onKeyPress`
`onChange`
`disabled`

Thanks for the explanation. Updated here 1725850516f62c3d71af6d44bdf66b6b8d723dfd.

rpokorny (approved):
+1 aside from https://github.com/sonatype/sonatype-react-shared-components/commit/1725850516f62c3d71af6d44bdf66b6b8d723dfd#r88640817

PR #832, opened 2022-10-25
Collapsible Items NxIconOnlyButton and NxIconDropdown Support - RSC-1185 

entroform:
https://issues.sonatype.org/browse/RSC-1185


JasonDSwearingen (approved):

gallery/visualtests/nxCollapsibleItems.js
```diff
@@ -106,6 +107,30 @@ describe('NxCollapsibleItems', function() {
     it('looks right', simpleTest(disabledTreeViewSelector));
   });
 
+  describe('NxCollapsibleItems with action content', function() {
```
rpokorny, line 110: Also test focus of the collapse trigger, since we want to make sure that it does not overlap the action button

gallery/visualtests/nxCollapsibleItems.js
```diff
@@ -106,6 +107,30 @@ describe('NxCollapsibleItems', function() {
...
+        await waitAndGetElements(firstTreeSelector, secondTreeSelector, iconDropdownToggleSelector);
+
+      await firstTree.click();
+      await secondTree.click();
+
+      await wait(350);
+
+      await iconDropdownToggle.click();
+
+      await simpleTest(actionContentCollapsibleItemsSelector)();
```
rpokorny, line 130: I'd actually rather test this with the first collapse section closed, in order to prove that the dropdown menu doesn't get cut off when it overflows out the bottom of the `NxCollapsibleItems`

rpokorny: > I'd actually rather test this with the first collapse section closed, in order to prove that the dropdown menu doesn't get cut off when it overflows out the bottom of the NxCollapsibleItems

At the same time though, I think it's important to test that the two click targets don't interfere with each other: that clicking the collapse trigger doesn't open or close the dropdown, and that clicking the action button doesn't collapse or expand the items.  That should probably just be a separate test.

lib/src/components/NxCollapsibleItems/NxCollapsibleItems.scss
```diff
@@ -28,7 +29,7 @@ $-form-child-left-padding: calc(#{$-trigger-side-padding} + #{$-triangle-box-wid
   }
 }
 
-.nx-collapsible-items--expanded {
+.nx-collapsible-items--expanded .nx-collapsible-items__header {
```
rpokorny, line 32: We use `>` where it makes sense because it makes the selector slightly more efficient.

```suggestion
.nx-collapsible-items--expanded > .nx-collapsible-items__header {
```

rpokorny:
How should the `disabled` prop affect these buttons?  Either way, that needs an example and a mockup

lib/src/components/NxCollapsibleItems/types.ts
```diff
@@ -15,6 +15,7 @@ export interface Props extends HTMLAttributes<HTMLDivElement> {
   disabled?: boolean | null;
   triggerContent: ReactNode;
   triggerTooltip?: TooltipConfigProps | string | null;
+  actionContent?: ReactNode | null;
```
rpokorny, line 18:
```suggestion
  actionContent?: ReactElement | null;
```

lib/src/components/NxCollapsibleItems/types.ts
```diff
@@ -31,6 +32,10 @@ export const propTypes: PropTypes.ValidationMap<Props> = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node
   ]),
+  actionContent: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node
+  ]),
```
rpokorny, line 38: I'm pretty sure that `PropTypes.node` can match an array anyway, it's basically equivalent to `ReactNode`.  But anyway, just like the `ReactNode` above should be `ReactElement`, this should be `PropTypes.element`.

entroform: > I'm pretty sure that `PropTypes.node` can match an array anyway, it's basically equivalent to `ReactNode`. But anyway, just like the `ReactNode` above should be `ReactElement`, this should be `PropTypes.element`.

Thanks, I updated that here: ddfde84270f8f18a0e362d2d12f8dc9f40f0d774

gallery/visualtests/nxCollapsibleItems.js
entroform: > > I'd actually rather test this with the first collapse section closed, in order to prove that the dropdown menu doesn't get cut off when it overflows out the bottom of the NxCollapsibleItems
> 
> At the same time though, I think it's important to test that the two click targets don't interfere with each other: that clicking the collapse trigger doesn't open or close the dropdown, and that clicking the action button doesn't collapse or expand the items. That should probably just be a separate test.

Thanks for the feedback! I updated the tests here: 6366984bd3762ab77d0ab0c90fda012da1e53315

gallery/visualtests/nxCollapsibleItems.js
entroform: > Also test focus of the collapse trigger, since we want to make sure that it does not overlap the action button

Good call, I added the focus test here:
aeafd77f1ecd1a29cea7b911e95387b3f1bcecac

plus the snapshots: c12e595a8a433cdf9b3b5368cf6c5001d955c4b2

rpokorny (approved):

entroform:
> How should the `disabled` prop affect these buttons? Either way, that needs an example and a mockup

Hi @rpokorny, I added an example here: 2901cac26652f6aa337c4032d9d4844d5c973043 and @JenniKobylski added a mockup for it. Let me know your thoughts!

rpokorny:
I'm seeing a real difference in gallery/visualtests/__image_snapshots__/nx-collapsible-items-js-nx-collapsible-items-nx-collapsible-items-with-clickable-children-has-items-that-look-right-on-hover-1-snap.png and at least one of the other snapshots. It looks like the focus glow on the collapse button has changed.  That shouldn't be the case right?

entroform:
> I'm seeing a real difference in gallery/visualtests/**image_snapshots**/nx-collapsible-items-js-nx-collapsible-items-nx-collapsible-items-with-clickable-children-has-items-that-look-right-on-hover-1-snap.png and at least one of the other snapshots. It looks like the focus glow on the collapse button has changed. That shouldn't be the case right?

It looks like the focus glow was how it was set initially but the stacking order changed so now the glow is visible.
I brought this to the awareness of the designer and they are okay with this subtle change. Let me know what you think.

rpokorny (approved):

PR #833, opened 2022-10-26
Allow custom Tooltips on Transfer List Items - RSC-1217

rpokorny:
https://issues.sonatype.org/browse/RSC-1217

Note that I didn't update the unit tests.  We've discussed that we should generally go ahead and convert unit tests to RTL when we do a ticket that would result in non-trivial changes to the unit tests, but in this ticket's case I don't think that's a good idea for several reasons:
1. The downstream team that wants this changes wants it fast
2. This change affects several components
3. The components that this changes affects are especially complex, and built on top of multiple other components for which we do not yet have RTL unit tests.

I think testing the new behavior using the visual tests, as this PR does, is adequate for now.  The new prop should however get unit tests when these components' tests are converted to RTL in the future.


gallery/src/components/NxSearchTransferList/NxSearchTransferListPage.tsx
```diff
@@ -189,6 +196,62 @@ const NxSearchTransferListPage = () =>
...
+              <NxTable.Cell>
+                Normally – when this property is <em>not</em> specified – the display text of a selected data item is
+                wrapped in an <NxCode>NxOverflowTooltip</NxCode>, such that if its <NxCode>displayName</NxCode>{' '}
+                overflows, it gets a tooltip displaying the full <NxCode>displayName</NxCode>. If
+                this <NxCode>tooltip</NxCode> property is specified however, instead of
+                an <NxCode>NxOverflowTooltip</NxCode> it gets a regular <NxCode>NxTooltip</NxCode> configured with the
+                specified props or, if the value of this property is a string, that string as the tooltip title. The
+                effect being that the properties specified here are used to construct a tooltip for the data item which
+                is active (on hover) regardless of whether the text is overflowing. Note that this replaces the default
+                overflow tooltip, and so must account the usability that would have been provided by that tooltip as
```
EmilyBermudez88, line 247:
```suggestion
                overflow tooltip, and so must account for the usability that would have been provided by that tooltip as
```
small fix on each of the pages

EmilyBermudez88 (approved):
Look great, minus the one minor fix!

gallery/src/components/NxSearchTransferList/NxSearchTransferListPage.tsx
rpokorny: > small fix on each of the gallery pages

[99b5b73](https://github.com/sonatype/sonatype-react-shared-components/pull/833/commits/99b5b730fd1e14fffb295385f3742c5a76f7e794)

PR #834, opened 2022-10-26
Gallery Example Deep Linking - RSC-1179

rpokorny:
https://issues.sonatype.org/browse/RSC-1179

Note that not all gallery examples currently have ids and that the ones that don't do not generate clickable links. There should be a followup ticket to add ids where they don't currently exist.


gallery/src/main.scss
```diff
@@ -46,6 +46,24 @@
...
+.gallery-example > .nx-tile-header > .nx-tile-header__title {
+  align-items: baseline;
+  display: flex;
+  overflow: visible;
+  gap: var(--nx-spacing-2x);
+  min-width: 0;
+
+  > .nx-h2 {
+    min-width: 0;
+  }
```
cindyko226, line 58: Do we need this?

gallery/src/gallery-components/GalleryTiles.tsx
```diff
@@ -49,13 +54,21 @@ interface GalleryExampleTileProps extends GalleryBaseProps {
...
-    const tileClasses = classnames('nx-tile', className);
+    const tileClasses = classnames('nx-tile', className),
+        routeParams = useParams<{ pageName: string }>();
 
     return (
       <div id={id} className={tileClasses}>
         <div className="nx-tile-header">
           <div className="nx-tile-header__title">
             <h2 className="nx-h2">{title}</h2>
+            { id && (
```
EmilyBermudez88, line 65: are the smooth brackets surrounding NxTooltip necessary?

EmilyBermudez88:
>Note that not all gallery examples currently have ids and that the ones that don't do not generate clickable links. There should be a followup ticket to add ids where they don't currently exist.

I know there are locations in the gallery that mention looking for a specific use case of a component in an example on another page (eg. Font Awesome Icon Page referring to an example in the NxButton page). Is it worth making a ticket to link these properly as well (or is the effort more than the benefit)?

gallery/src/gallery-components/GalleryTiles.tsx
rpokorny: > are the smooth brackets surrounding NxTooltip necessary?

No but they match the parentheses that we typically put around multi-line return statements (like the one starting on line 61 that this line is part of)

rpokorny:
> I know there are locations in the gallery that mention looking for a specific use case of a component in an example on another page (eg. Font Awesome Icon Page referring to an example in the NxButton page). Is it worth making a ticket to link these properly as well (or is the effort more than the benefit)?

I'd say that's pretty optional.  I suspect most of the cross-page links that we have really are intended to link to whole pages and not individual examples.  Finding the few that would be best linked to specific examples would probably be more trouble than its worth.

EmilyBermudez88:
I tested opening up a new tab and pasting in the example's localhost URL, and it seems as though this changes the positioning of the viewport slightly. In localhost, if you paste the URL in while already on Tartan it works as expected. 

Screen Shot 2022-11-01 at 11 25 56 AM
Screen Shot 2022-11-01 at 11 26 15 AM

rpokorny:
> I tested opening up a new tab and pasting in the example's localhost URL, and it seems as though this changes the positioning of the viewport slightly (first image). In localhost, if you paste the URL in while already on Tartan it works as expected (second image).

I believe the problem was that the scroll was happening before fonts loaded, and once they did load, the lengths of paragraphs changed and pushed things down. Fixed in 1e57f1855

EmilyBermudez88:
> I believe the problem was that the scroll was happening before fonts loaded, and once they did load, the lengths of paragraphs changed and pushed things down. Fixed in [1e57f18](https://github.com/sonatype/sonatype-react-shared-components/commit/1e57f1855431fbdcff3a189b8e0ca9857a72a011)

Great thought, looks good

EmilyBermudez88 (approved):

PR #837, opened 2022-10-31
Fix NxOverflowTooltip in Firefox - RSC-1216

rpokorny:
https://issues.sonatype.org/browse/RSC-1216

While this issue was discovered on `NxBreadcrumb`, it is not specific to it.  The issue is that previously, in Firefox, any content which overflowed the content box of an `NxOverflowTooltip`-enabled container, but not the padding box, would not get a tooltip even though it would get overflow truncation.  Investigation deeply into the [relevant specs](https://www.w3.org/TR/cssom-view-1/#dom-element-scrollwidth), Firefox's definitions of the relevant properties (namely `scrollWidth`) were actually the ones that are correct to the spec.  In such a standards compliant browser, it seems that it is actually impossible to detect text truncation using `scrollWidth`.  This PR takes a different approach by comparing bounding client rectangles of the container and all of its text node children.


lib/src/components/NxTooltip/NxOverflowTooltip.tsx
```diff
@@ -12,15 +12,72 @@ import { textContent } from '../../util/childUtil';
...
+    if (parsedSizes[0] == null) {
+      console.warn('Got non-pixel computed value for padding-left, assuming 0');
+    }
+    if (parsedSizes[1] == null) {
+      console.warn('Got non-pixel computed value for padding-right, assuming 0');
+    }
+    if (parsedSizes[2] == null) {
+      console.warn('Got non-pixel computed value for border-left-width, assuming 0');
+    }
+    if (parsedSizes[2] == null) {
```
cindyko226, line 44:
```suggestion
    if (parsedSizes[3] == null) {
```

cindyko226:
This might be the edge case when the element contains icons and the text, I notice that the tooltip is not showing while testing it locally in the `NxTransferListHalf` minimal example, if you change the displayName of the first item to `Loooooooooooooooooooooooooo`, the text is truncated but not triggering the tooltip when hovered. Maybe we should consider the elements rendered in front of the text when computing the `getTextBoundingRectWidth`?

rpokorny:
> This might be the edge case when the element contains icons and the text, I notice that the tooltip is not showing while testing it locally in the NxTransferListHalf minimal example, if you change the displayName of the first item to Loooooooooooooooooooooooooo, the text is truncated but not triggering the tooltip when hovered. Maybe we should consider the elements rendered in front of the text when computing the getTextBoundingRectWidth?

Great point.  Thinking about it more I think it's more correct to compare only the right edges of the container and children, not the widths.  Fixed in 062668223

cindyko226 (approved):

PR #840, opened 2022-11-04
Disable Browser autocomplete in NxCombobox - RSC-1221

cindyko226:
https://issues.sonatype.org/browse/RSC-1221

### DIFF
--- a/lib/src/components/NxCombobox/NxCombobox.tsx
+++ b/lib/src/components/NxCombobox/NxCombobox.tsx
@@ -315,7 +315,9 @@ function NxComboboxRender<T extends string | number = string>(
                    aria-activedescendant={focusableBtnId}
                    aria-required={ariaRequired}
                    aria-describedby={inputDescribedby}
-                   aria-label={ariaLabel}/>
+                   aria-label={ariaLabel}
+                   // disable browser autocomplete
+                   autoComplete="off"/>
       { isAlert ?
         <div id={alertDropdownId}
              role="alert"
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.3.7",
+  "version": "12.3.8",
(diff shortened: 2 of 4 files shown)


lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -315,7 +315,9 @@ function NxComboboxRender<T extends string | number = string>(
                    aria-activedescendant={focusableBtnId}
                    aria-required={ariaRequired}
                    aria-describedby={inputDescribedby}
-                   aria-label={ariaLabel}/>
+                   aria-label={ariaLabel}
+                   // disable browser autocomplete
+                   autoComplete="off"/>
```
rpokorny, line 320: Were you able to create a test case that had the browser auto-complete showing up, and then prove that this fixes it? I ask because we never saw the browser autocomplete during the development of `NxCombobox`, but Inception gets it consistently as the work on the Select Contact modal.  There must be something about the way that modal is set up that triggers it, and we want to make sure that whatever change we make in this PR will fix that.

cindyko226: Below are the steps to manually test it, the browser auto-complete is triggered when rendered within the Form.

- Go to Form Layout Examples Page - General Example
- Fill out `User Name` text input and `State` combobox
- Click `Submit` button
- Go back to `User Name` text input, delete the text, click outside of the text input then click the text input again, the browser auto-complete is showing up
- Repeat the previous step on the combobox won't trigger the browser auto-complete when `autoComplete` is set to "off"

Base on the testing, I think the change in this PR fix it.

rpokorny: > Base on the testing, I think the change in this PR fix it.

Nice, please add those steps to the JIRA ticket too

rpokorny (approved):

lib/src/components/NxCombobox/NxCombobox.tsx
cindyko226: Thanks, will do.

PR #841, opened 2022-11-04
Fix NxOverflowTooltip at Non-100% Zoom Levels - RSC-1240

rpokorny:
https://issues.sonatype.org/browse/RSC-1240

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.4.2",
+  "version": "12.4.3",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.3.7",
+  "version": "12.3.8",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "12.3.7",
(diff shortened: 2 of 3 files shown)


rpokorny:
> Testing in Chrome, I still found a bug on the Breadcrumb page with NxOverflowTooltip at 80%-110% when 'Penitente Canyon' is in the dropdown, and in the Five Examples San Luis triggers the tooltip unnecessarily just at 90%. However, this seems to be the only issue I could find testing in different browsers at different zoom levels.

Does this fix them all for you? ef38c258e

EmilyBermudez88:
> > Testing in Chrome, I still found a bug on the Breadcrumb page with NxOverflowTooltip at 80%-110% when 'Penitente Canyon' is in the dropdown, and in the Five Examples San Luis triggers the tooltip unnecessarily just at 90%. However, this seems to be the only issue I could find testing in different browsers at different zoom levels.
> 
> Does this fix them all for you? [ef38c25](https://github.com/sonatype/sonatype-react-shared-components/commit/ef38c258e3b1c5510042dfbd97018e2abc9f6c39)

So now at 110% 'Earth' it triggering a tooltip on the `Example with many Segments`. Sorry, this seems to be quite the whack-a-mole situation.

rpokorny:
> So now at 110% 'Earth' it triggering a tooltip on the Example with many Segments. Sorry, this seems to be quite the whack-a-mole situation.

The fact of the matter is that without knowing exactly what logic the browser uses internally (which might vary between browsers anyway), we can't know how far we have to round, or if we should maybe be using `floor` or `ceil` instead of rounding.  This is all just our best guess of how far we have to fudge things to get around the unknown logic driving the rounding errors in the numbers that the browser gives us for the bounding boxes.  So there's always going to be certain risks of false positives (activating the tooltip on non-truncated text) and false negatives (failing to activate the tooltip on truncated text).  The further we round, the lower the risk of false positives but the higher the risk of false negatives, I think.  I think we should be erring on the side of false positives; it's better to activate the tooltip when you don't need it than to not activate it when you do need it.  For that reason I'm inclined not to round things further than we now have and to accept the current behavior.

EmilyBermudez88 (approved):

PR #842, opened 2022-11-07
Configurable Tooltips on NxCombobox Items - RSC-1243

rpokorny:
https://issues.sonatype.org/browse/RSC-1243


lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -375,6 +378,11 @@ function NxComboboxRender<T extends string | number = string>(
   );
 }
 
-const NxCombobox = Object.assign(forwardRef(NxComboboxRender), { propTypes });
+const NxCombobox = Object.assign(forwardRef(NxComboboxRender), {
+  // propTypes types can actually effect the overall type of NxCombobox in ways we don't want, the cast
+  // to `any` prevents that
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  propTypes: propTypes as any
```
cindyko226, line 385: > propTypes types can actually effect the overall type of NxCombobox in ways we don't want

Out of curiosity, is it because the conditional type is added for NxCombobox?

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -296,6 +295,33 @@ describe('NxCombobox', function() {
...
+    const user = userEvent.setup(),
+        view = quickRender({
+          matches: [{ id: '1', displayName: 'Foo' }]
+        }),
+        btn = view.getByRole('option');
+
+    await user.hover(btn);
+    const tooltipPromise = screen.findByRole('tooltip');
+
+    await expect(tooltipPromise).rejects.toBeTruthy();
```
cindyko226, line 322: I might think it too much..but is it possible that more than one tooltip is found which will make the `tooltipPromise` rejected, and it still passed on line 322?

lib/src/components/NxCombobox/NxCombobox.tsx
rpokorny: > Out of curiosity, is it because the conditional type is added for NxCombobox?

That might be what's tripping it up but it's something from the React typings that I don't fully understand.  I just know I've seen times where attaching the `propTypes` to a component changes the TS type for that component in a way that doesn't always work if the props or component have generic type parameters.  We do want to ensure that our PropTypes match our TS interfaces (which is why we use the `ValidationMap` type where the `propTypes` are defined), but we aren't relying on PropTypes _in lieu_ of TS interfaces, so we really don't want/need them to affect the component type like that.

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
rpokorny: > I might think it too much..but is it possible that more than one tooltip is found which will make the tooltipPromise rejected, and it still passed on line 322?

Nice catch. 00f39d3c5

lib/src/components/NxCombobox/NxCombobox.tsx
cindyko226: Thanks for the explanations, it is helpful.

cindyko226 (approved):

PR #844, opened 2022-11-08
Relocate Dropdown Menu Item Styles - RSC-1241

cindyko226:
https://issues.sonatype.org/browse/RSC-1241


rpokorny:
In the JIRA ticket it mentions the base styles as another candidate for where these styles could be relocated.  I think the deciding factor there is whether there are an places that we support or demonstrate using these tiles outside of React components built on `NxDropdownMenu`.  If so, they need to be in the base styles.  I tend to think we don't, and the `NxDropdownMenu` is in fact the right place, but we should double check.

rpokorny (approved):

PR #845, opened 2022-11-08
Added managed keyboard nav and roles to NxDropdown - RSC-989

entroform:
https://issues.sonatype.org/browse/RSC-989

### DIFF
--- a/gallery/src/components/NxDropdown/NxDropdownCloseHandlerExample.tsx
+++ b/gallery/src/components/NxDropdown/NxDropdownCloseHandlerExample.tsx
@@ -18,10 +18,10 @@ function NxDropdownCloseHandlerExample() {
                 onToggleCollapse={onToggleCollapse}
                 onCloseClick={(evt: MouseEvent) => evt.preventDefault()}
                 onCloseKeyDown={(evt: KeyboardEvent) => evt.preventDefault()}>
-      <a onClick={onClick} href="#/pages/Dropdown" className="nx-dropdown-button">
+      <a onClick={onClick} href="#/pages/Dropdown" role="menuitem" className="nx-dropdown-button">
         Link
       </a>
-      <button onClick={onClick} className="nx-dropdown-button">
+      <button role="menuitem" onClick={onClick} className="nx-dropdown-button">
         Button
       </button>
     </NxDropdown>
--- a/gallery/src/components/NxDropdown/NxDropdownRightButtonsExample.tsx
+++ b/gallery/src/components/NxDropdown/NxDropdownRightButtonsExample.tsx
@@ -14,10 +14,10 @@ function NxDropdownRightButtonsExample() {
 
   return (
     <NxDropdown label="Navigation" isOpen={isOpen} onToggleCollapse={onToggleCollapse}>
-      <a href="#/" className="nx-dropdown-button">
+      <a href="#/" className="nx-dropdown-button" role="menuitem">
         <span className="nx-dropdown-button-content">Text Link</span>
       </a>
-      <a href="#/" className="nx-dropdown-button">
+      <a href="#/" className="nx-dropdown-button" role="menuitem">
         <span className="nx-dropdown-button-content">
           Text Link - this link should trigger truncation
         </span>
@@ -25,19 +25,21 @@ function NxDropdownRightButtonsExample() {
       <NxButton onClick={() => alert('icon click')}
                 className="nx-dropdown-right-button"
                 variant="icon-only"
-                title="Delete Button Link2">
+                title="Delete Button Link2"
+                role="menuitem">
         <NxFontAwesomeIcon icon={faTrash}/>
       </NxButton>
-      <a href="#/" className="nx-dropdown-button">
+      <a href="#/" className="nx-dropdown-button" role="menuitem">
         <span className="nx-dropdown-button-content">Button Link2</span>
       </a>
       <NxButton onClick={() => alert('icon click')}
                 className="nx-dropdown-right-button"
                 variant="icon-only"
-                title="Delete Text Link3">
+                title="Delete Text Link3"
+                role="menuitem">
         <NxFontAwesomeIcon icon={faTrash}/>
       </NxButton>
-      <a href="#/" className="nx-dropdown-button">
+      <a href="#/" className="nx-dropdown-button" role="menuitem">
         <span className="nx-dropdown-button-content">Text Link3 - this link should trigger truncation</span>
       </a>
     </NxDropdown>
--- a/gallery/visualtests/nxDropdown.js
+++ b/gallery/visualtests/nxDropdown.js
@@ -16,7 +16,8 @@ describe('NxDropdown', function() {
     waitAndGetElements,
     moveMouseAway,
     checkScreenshot,
-    a11yTest
+    a11yTest,
+    getPage
   } = setupBrowser('#/pages/Dropdown');
 
   const defaultSelector = '#nx-dropdown-scrolling-example .nx-dropdown',
@@ -83,6 +84,16 @@ describe('NxDropdown', function() {
       const [targetElement] = await waitAndGetElements(selector);
 
       await moveMouseAway();
+
+      await checkScreenshot(targetElement, 251, 218);
+    });
+
+    it('looks right when focused via keyboard', async function() {
+      const [targetElement] = await waitAndGetElements(selector);
+      const page = getPage();
+
+      await moveMouseAway();
+      await page.keyboard.press('ArrowLeft');
 
       await checkScreenshot(targetElement, 251, 218);
     });
--- a/lib/src/components/NxCheckbox/NxCheckbox.tsx
+++ b/lib/src/components/NxCheckbox/NxCheckbox.tsx
@@ -35,6 +35,7 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
             checkboxId,
             overflowTooltip,
             children,
+            checkboxRole,
             inputAttributes = {},
             ...otherProps
           } = props,
@@ -63,6 +64,7 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
                  disabled={!!disabled}
                  checked={isChecked}
                  readOnly={!onChange}
+                 role={checkboxRole}
                  onChange={onChange || undefined}
                  { ...otherInputAttributes } />
           <span className="nx-radio-checkbox__control nx-checkbox__box">
--- a/lib/src/components/NxDropdown/NxDropdown.tsx
+++ b/lib/src/components/NxDropdown/NxDropdown.tsx
@@ -15,6 +15,7 @@ import NxFontAwesomeIcon from '../NxFontAwesomeIcon/NxFontAwesomeIcon';
 import { wrapTooltipProps } from '../../util/tooltipUtils';
 import './NxDropdown.scss';
 import NxOverflowTooltip from '../NxTooltip/NxOverflowTooltip';
+import { useUniqueId } from '../../util/idUtil';
 
 import AbstractDropdown, { AbstractDropdownRenderToggleElement } from './AbstractDropdown';
 import withClass from '../../util/withClass';
@@ -28,6 +29,7 @@ const _NxDropdown = forwardRef<HTMLDivElement, Props>(function NxDropdown(props,
     label,
     toggleTooltip,
     variant,
+    id: idProp,
     ...otherProps
   } = props;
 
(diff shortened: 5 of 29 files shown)


entroform:
According to the pattern, when the dropdown menu is opened, it first focuses on the menu as shown here in the snapshot.

nx-dropdown-js-nx-dropdown-nx-dropdown-links-looks-right-when-focused-via-keyboard-1-snap

Will be reaching out to the designer about how it should look like when focused.

entroform:
Curiously, in the example here, https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-links.html, it focuses on the menuitem when you hover over it with mouse. Something we could consider implementing in another ticket?

rpokorny:
> Curiously, in the example here, https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-links.html, it focuses on the menuitem when you hover over it with mouse. Something we could consider implementing in another ticket?

That's pretty weird.  I've never seen any accessibility guidance or other examples that have focus following the mouse, and I don't see that called out as desired/required on the [docs for the dropdown pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubutton/) or the [docs for the relevant roles](https://www.w3.org/TR/wai-aria-1.2/#menu).  I'd ignore it.

gallery/src/components/NxDropdown/NxDropdownCloseHandlerExample.tsx
```diff
@@ -18,10 +18,10 @@ function NxDropdownCloseHandlerExample() {
                 onToggleCollapse={onToggleCollapse}
                 onCloseClick={(evt: MouseEvent) => evt.preventDefault()}
                 onCloseKeyDown={(evt: KeyboardEvent) => evt.preventDefault()}>
-      <a onClick={onClick} href="#/pages/Dropdown" className="nx-dropdown-button">
+      <a onClick={onClick} href="#/pages/Dropdown" role="menuitem" className="nx-dropdown-button">
         Link
       </a>
-      <button onClick={onClick} className="nx-dropdown-button">
+      <button role="menuitem" onClick={onClick} className="nx-dropdown-button">
```
rpokorny, line 24: I think we should create convenience components that encapsulate the needed classes and roles

rpokorny: > I think we should create convenience components that encapsulate the needed classes and roles

The tooltip logic will need to be adjusted to accomodate this as well

rpokorny:
> According to the pattern, when the dropdown menu is opened, it first focuses on the menu as shown here in the snapshot

That's not what I'm seeing in the ARIA patterns docs and examples.  Everything I'm seeing says to focus the first item in the dropdown, except if it was opened by the user pressing `UP`, in which case the last item should be focused.

gallery/src/components/NxDropdown/NxDropdownRightButtonsExample.tsx
```diff
@@ -14,30 +14,32 @@ function NxDropdownRightButtonsExample() {
...
         <span className="nx-dropdown-button-content">
           Text Link - this link should trigger truncation
         </span>
       </a>
       <NxButton onClick={() => alert('icon click')}
                 className="nx-dropdown-right-button"
                 variant="icon-only"
-                title="Delete Button Link2">
+                title="Delete Button Link2"
+                role="menuitem">
```
rpokorny, line 29: Check with design about how these buttons should be added into the recommended patterns.  These are pretty nonstandard, so I think we're on our own in terms of how they behave.  Having them just be part of the up/down key cycling doesn't necessarily seem right though.

gallery/visualtests/nxDropdown.js
```diff
@@ -86,6 +87,16 @@ describe('NxDropdown', function() {
 
       await checkScreenshot(targetElement, 251, 218);
     });
+
+    it('looks right when focused via keyboard', async function() {
+      const [targetElement] = await waitAndGetElements(selector);
+      const page = getPage();
+
+      await moveMouseAway();
+      await page.keyboard.press('ArrowLeft');
```
rpokorny, line 96: I don't see any behavior for the Left arrow key defined in the ARIA docs; what's this for?

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -56,6 +56,26 @@ describe('NxDropdown', () => {
...
+    const menuId = 'foo';
+    const component = getMountedComponent({ isOpen: true, menuId });
+    const dropdown = component.find(AbstractDropdown);
+    const button = dropdown.find(NxButton);
+    const menu = component.find(NxDropdownMenu);
+
+    expect(button).toHaveProp('aria-controls', menuId);
+    expect(menu).toHaveProp('id', menuId);
+  });
+
```
rpokorny, line 78: `aria-expanded` should be tested too

lib/src/components/NxDropdown/types.ts
```diff
@@ -36,6 +37,7 @@ export type Props = Omit<HTMLAttributes<HTMLDivElement>, 'className'> & {
   onCloseClick?: ((e: MouseEvent) => void) | null;
   toggleTooltip?: TooltipConfigProps | string | null;
   menuRef?: Ref<HTMLDivElement>;
+  menuId?: string;
```
rpokorny, line 40: To be compatible with `propTypes` optional properties should allow both `undefined` and `null`

lib/src/components/NxDropdownMenu/NxDropdownMenu.tsx
```diff
@@ -4,28 +4,108 @@
...
 
 import { Props, propTypes } from './types';
 
 import './NxDropdownMenu.scss';
 
 export { Props };
 
+const FOCUSABLE_MENU_ITEMS_SELECTOR = 'a:not([disabled]):not(.disabled), '
++ 'button:not([disabled]):not(.disabled), '
++ 'input:not([disabled]):not(.disabled)';
```
rpokorny, line 19: I think this is the same
```suggestion
const FOCUSABLE_MENU_ITEMS_SELECTOR = ':is(a, button, input):not([disabled], .disabled])';
```

rpokorny:
A key aspect of compound ARIA widgets like this is that the whole thing is one tabstop.  That is, the user can tab to the dropdown, and regardless of whether it is open or not or how many menu items it has, they can tab away from it with a single press of the Tab key.  Moving focus within the component should not be possible with Tab, but only with the arrow keys.  This is typically accomplished by setting `tabIndex` to `-1` on all of the menu items and using either `aria-activedesendant` or programmatic focus to keep track of which menu item is currently focused (I recommend the latter as it makes activation of the menu items more natural).

lib/src/components/NxDropdownMenu/NxDropdownMenu.tsx
rpokorny: Might not hurt to also prefix this with `:scope > `

lib/src/components/NxDropdownMenu/NxDropdownMenu.tsx
```diff
@@ -4,28 +4,108 @@
...
   useLayoutEffect(() => onClosing, []);
 
-  return <div ref={ref} { ...{ className, ...attrs } } />;
+  useEffect(() => {
+    if (!disableMenuKeyNav && menuRef.current) {
+      menuRef.current.focus();
+    }
+  }, [disableMenuKeyNav]);
+
+  function setFocusToMenuItem(direction: -1 | 1) {
```
rpokorny, line 43: Both `NxSearchDropdown` and `NxComboxbox` have also had to implement logic to do what this is doing, which they did with a function called `adjustBtnFocus`.  As this is the third time we've needed similar logic, I think it probably makes sense to see if we can extract that out into a utility and reuse it here too.

lib/src/components/NxFilterDropdown/NxFilterDropdown.tsx
```diff
@@ -68,12 +68,16 @@ function NxFilterDropdownRender<T extends string | number = string>(props: Props
 
   function onResetClick() {
     onChange(new Set());
+    if (menuRef.current) {
+      menuRef.current.focus();
```
rpokorny, line 72: As I mentioned in another comment, I'm not seeing anything indicating that focusing the menu itself is truly a part of the pattern.  If we get rid of that, what you could do here instead is focus the dropdown toggle.

lib/src/components/NxDropdownMenu/NxDropdownMenu.tsx
entroform: > Might not hurt to also prefix this with `:scope > `

Ooo I didn't know :is is supported with major browsers now, I think multiple argument :not is not supported with the current safari though, not 100% sure, going to double check.

rpokorny: > Ooo I didn't know :is is supported with major browsers now, I think multiple argument :not is not supported with the current safari though, not 100% sure, going to double check.

MDN seems to say it's been supported since safari 9

gallery/src/components/NxDropdown/NxDropdownCloseHandlerExample.tsx
entroform: > > I think we should create convenience components that encapsulate the needed classes and roles
> 
> The tooltip logic will need to be adjusted to accomodate this as well

This is still in progress, but I wanted to get your thoughts on the addition of adding wrapper prop to withClass here: 3353a939799a2e11b4606976a08be82e902252d9

lib/src/components/NxCheckbox/NxCheckbox.tsx
```diff
@@ -35,6 +35,7 @@ const NxCheckbox = forwardRef<HTMLLabelElement, Props>(
             checkboxId,
             overflowTooltip,
             children,
+            checkboxRole,
```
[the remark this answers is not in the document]
entroform: > This doesn't seem needed when `inputAttributes` exists

Ops, totally missed `inputAttributes`, fixed here: 1ee7a6f8f1584d3335d97084cd92a59e3e878d2a

lib/src/components/NxDropdown/types.ts
entroform: > To be compatible with `propTypes` optional properties should allow both `undefined` and `null`

Fixed: d9fb143850b74df0bc73bc5fb767cc6b9f607bdd

gallery/src/components/NxDropdown/NxDropdownCloseHandlerExample.tsx
rpokorny: > This is still in progress, but I wanted to get your thoughts on the addition of adding wrapper prop to withClass here: https://github.com/sonatype/sonatype-react-shared-components/commit/3353a939799a2e11b4606976a08be82e902252d9

Hmm is there any other kind of wrapper that would make sense here besides `NxOverflowTooltip`? I wonder if we should just add a boolean flag for that possibility only, and leave anything more obscure for custom component implementations.

entroform: > > This is still in progress, but I wanted to get your thoughts on the addition of adding wrapper prop to withClass here: [3353a93](https://github.com/sonatype/sonatype-react-shared-components/commit/3353a939799a2e11b4606976a08be82e902252d9)
> 
> Hmm is there any other kind of wrapper that would make sense here besides `NxOverflowTooltip`? I wonder if we should just add a boolean flag for that possibility only, and leave anything more obscure for custom component implementations.

I can't think of anything else that would be wrapper. I added a boolean flag for `NxOverflowTooltip` here: 641905f546159b0d24ca397f0c2308a99990992d

lib/src/components/NxDropdown/NxDropdown.tsx
```diff
@@ -77,7 +83,10 @@ const _NxDropdown = forwardRef<HTMLDivElement, Props>(function NxDropdown(props,
 _NxDropdown.propTypes = propTypes;
 
 const NxDropdown = Object.assign(_NxDropdown, {
-  Divider: withClass('hr', 'nx-dropdown__divider')
+  Divider: withClass('hr', 'nx-dropdown__divider', 'separator'),
```
rpokorny, line 86: It's bad practice to re-specify roles that are already the default on a given type of HTML element. `separator` is in fact the default role for `<hr>`.

entroform: > It's bad practice to re-specify roles that are already the default on a given type of HTML element. `separator` is in fact the default role for `<hr>`.

Ops, yes I removed it here: 103669afeb61fd3dfd3ac5172c4e6d3a1951aa6e

PR #846, opened 2022-11-08
Prevent Shrinking on Inner Sidebar at Narrow Screen Size - RSC-1245

rpokorny:
https://issues.sonatype.org/browse/RSC-1245

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.4.3",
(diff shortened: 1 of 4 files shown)


rpokorny:
Note: the `NxCollapsibleItems` screenshot changes here are basically incidental.  The examples that those screenshots come from use `.nx-page-sidebar` in a simple div container rather than in `.nx-page` or `.nx-page-content` which are the only places it's truly supported.  As such the width that those examples "should" be is somewhat undefined.

PR #847, opened 2022-11-09
RTL Conversion for Links

rjp99:
https://issues.sonatype.org/browse/RSC-1229


lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
-    expect(getShallow()).toHaveProp('aria-disabled', false);
-    expect(getShallow({ disabled: true })).toHaveProp('aria-disabled', true);
+    expect(renderEl()).toHaveAttribute('aria-disabled', 'false');
+    expect(renderEl({ disabled: true })).toHaveAttribute('aria-disabled', 'true');
   });
 
   it('sets link role on the text link when the disabled prop is set', function() {
-    expect(getShallow()).not.toHaveProp('role', 'link');
-    expect(getShallow({ disabled: true })).toHaveProp('role', 'link');
+    expect(renderEl()).not.toHaveAttribute('role', 'link');
```
EmilyBermudez88, line 71: Not sure if this is relevant or not, but the test will pass with any kind of role attached to NxTextLink as long as it's not 'role'. If there isn't supposed to be a role attached to the link unless it's disabled, perhaps it could be `expect(renderEl()).not.toHaveAttribute('role')`. As I said though, not sure if testing whether ANY role is applied prior to disabling NxTextLink is part of testing

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
 
   it('adds an icon after the children if external is true', function() {
-    expect(getShallow({ children: <span/> })).not.toContainMatchingElement(NxFontAwesomeIcon);
-    expect(getShallow({ children: <span/>, external: true })).toContainMatchingElement(NxFontAwesomeIcon);
-  });
+    const { container } = quickRender({ children: 'foo', external: true });
 
-  it('wraps non-element children in an element sibling to the icon', function() {
-    expect(getShallow({ children: 'foo', external: true })).toMatchElement(
-      <a>
```
EmilyBermudez88, line 105: This may still be useful to test, to check that a <span> automatically renders when text is provided (and that when the text is wrapped in something, it doesn't create a span) . I'm sure there's a cleaner way, but could be something like:    

```
const el = renderEl({children: 'foo'});
    expect(el?.children[0].tagName).toBe('SPAN');
```

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
 import NxTextLink from '../NxTextLink';
-import NxFontAwesomeIcon from '../../NxFontAwesomeIcon/NxFontAwesomeIcon';
-import { IconProp } from '@fortawesome/fontawesome-svg-core';
 
 describe('NxTextLink', function() {
-  const getShallow = getShallowComponent(NxTextLink, {});
+  const quickRender = rtlRender(NxTextLink, {});
+  const renderEl = rtlRenderElement(NxTextLink, {});
 
   it('renders an <a> with the nx-text-link class', function() {
```
rpokorny, line 16: The class name is an implementation detail as far as RTL is concerned

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
-    const component = getShallow({ id: 'foo', lang: 'en_US' });
+    const el = renderEl({ id: 'foo', lang: 'en_US' });
 
-    expect(component).toHaveProp('id', 'foo');
-    expect(component).toHaveProp('lang', 'en_US');
+    expect(el).toHaveAttribute('id', 'foo');
+    expect(el).toHaveAttribute('lang', 'en_US');
   });
 
   it('adds the nx-text-link--external class if external is true', function() {
```
rpokorny, line 37: Implementation detail.  The relevant external behavior is the presence of the external link icon, which is tested by visual tests

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
   });
 
   it('sets aria-disabled on the text link when the disabled prop is set', function() {
-    expect(getShallow()).toHaveProp('aria-disabled', false);
-    expect(getShallow({ disabled: true })).toHaveProp('aria-disabled', true);
+    expect(renderEl()).toHaveAttribute('aria-disabled', 'false');
+    expect(renderEl({ disabled: true })).toHaveAttribute('aria-disabled', 'true');
   });
 
   it('sets link role on the text link when the disabled prop is set', function() {
```
rpokorny, line 70: The setting of the attribute itself isn't important.  What's important is what the actual computed role of the element is: it should always be `link` regardless of whether the link is disabled.  You can use `getByRole` to check this.

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
-    expect(getShallow()).not.toHaveProp('role', 'link');
-    expect(getShallow({ disabled: true })).toHaveProp('role', 'link');
+    expect(renderEl()).not.toHaveAttribute('role', 'link');
+    expect(renderEl({ disabled: true })).toHaveAttribute('role', 'link');
   });
 
   it('does not pass href value on the text link when the disabled prop is set', function() {
```
rpokorny, line 75: Implementation detail.  What we really want to check is whether clicking the link does anything when it is/isn't disabled

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
+    expect(renderEl({ external: true, newTab: true })).toHaveAttribute('target', '_blank');
+    expect(renderEl({ external: true, newTab: false })).toHaveAttribute('target', '');
 
-    expect(getShallow({ external: true, target: 'asdf' })).toHaveProp('target', 'asdf');
+    expect(renderEl({ external: true, noReferrer: true, target: 'asdf' })).toHaveAttribute('target', 'asdf');
+  });
 
-    expect(getShallow({ external: true, newTab: true })).toHaveProp('target', '_blank');
-    expect(getShallow({ external: true, newTab: false })).toHaveProp('target', '');
+  it('does not add an icon after the children if external is false', function() {
```
rpokorny, line 99: As I said above the icon presentation is tested via visual test and testing it at this level is an implementation detail.

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > The class name is an implementation detail as far as RTL is concerned

Removed here: 4e7e494279e16de3a908ade9be773ce6d2bd9065

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > Implementation detail. The relevant external behavior is the presence of the external link icon, which is tested by visual tests

Removed here: 4c42970c97bdf179a2b81e369280dce5396bc4de

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > The setting of the attribute itself isn't important. What's important is what the actual computed role of the element is: it should always be `link` regardless of whether the link is disabled. You can use `getByRole` to check this.

Updated here: 03d46791ff22a5aff9a1b1c96a074e5d957adac8

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > Implementation detail. What we really want to check is whether clicking the link does anything when it is/isn't disabled

Something like this? b72d4c0c3aaf75b02b9195688d8c0b38d231ad7f

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > As I said above the icon presentation is tested via visual test and testing it at this level is an implementation detail.

Removed here: 70a65dd9c54784f2922ba72a48d7db238e0ecc4e

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -5,114 +5,106 @@
...
 
   it('adds additional specified classes', function() {
-    const component = getShallow({ className: 'foo' });
+    const el = renderEl({ className: 'foo' });
 
-    expect(component).toHaveClassName('foo');
-    expect(component).toHaveClassName('nx-text-link');
+    expect(el).toHaveClass('foo');
+    expect(el).toHaveClass('nx-text-link');
   });
```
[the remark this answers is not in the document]
rjp99: > For this test, you could do something like this (for example in Drawer, ColorPicker etc ):

Updated here: 2404df74b517db0b783ddec5bd265266bf40df6c

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > Not sure if this is relevant or not, but the test will pass with any kind of role attached to NxTextLink as long as it's not 'role'. If there isn't supposed to be a role attached to the link unless it's disabled, perhaps it could be `expect(renderEl()).not.toHaveAttribute('role')`. As I said though, not sure if testing whether ANY role is applied prior to disabling NxTextLink is part of testing

Updated here: 7910971123c15e7e2be37ae5dd3a497e0ffa19ec

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > This may still be useful to test, to check that a automatically renders when text is provided (and that when the text is wrapped in something, it doesn't create a span) . I'm sure there's a cleaner way, but could be something like:
> 
> ```
> const el = renderEl({children: 'foo'});
>     expect(el?.children[0].tagName).toBe('SPAN');
> ```

Added it here: 69ec81407e68f1a5c3c53a15bd104a400005753d

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rpokorny: > Removed here: https://github.com/sonatype/sonatype-react-shared-components/commit/4e7e494279e16de3a908ade9be773ce6d2bd9065

IMO it was still worthwhile to test that it was an `<a>`, since `<a>` tags have special accessibility behaviors in the browser such as showing the hyperlink target in the bottom of the viewport when you hover the link.

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rpokorny: > Updated here: https://github.com/sonatype/sonatype-react-shared-components/commit/03d46791ff22a5aff9a1b1c96a074e5d957adac8

`toBe` doesn't work that way.  Use `.toBeInTheDocument()`.  (Make sure you put the `()` on in order to actually invoke the assertion).

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rpokorny: > Something like this? https://github.com/sonatype/sonatype-react-shared-components/commit/b72d4c0c3aaf75b02b9195688d8c0b38d231ad7f

Nevermind, after a little bit of research it doesn't look like JSDOM really implements enough to verify that links can truly be followed.  Checking the `href` attribute seems like the best approach afterall.  What you had originally was good.

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > > Removed here: [4e7e494](https://github.com/sonatype/sonatype-react-shared-components/commit/4e7e494279e16de3a908ade9be773ce6d2bd9065)
> 
> IMO it was still worthwhile to test that it was an `<a>`, since `<a>` tags have special accessibility behaviors in the browser such as showing the hyperlink target in the bottom of the viewport when you hover the link.

Added that part back: a5cc4005f6e6e5133b043cff46e94438a409293e

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > > Something like this? [b72d4c0](https://github.com/sonatype/sonatype-react-shared-components/commit/b72d4c0c3aaf75b02b9195688d8c0b38d231ad7f)
> 
> Nevermind, after a little bit of research it doesn't look like JSDOM really implements enough to verify that links can truly be followed. Checking the `href` attribute seems like the best approach afterall. What you had originally was good.

Changed it: 6748453f2635cc4c13e839c3db1dac9ea6ed782c

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -4,115 +4,96 @@
...
-import { IconProp } from '@fortawesome/fontawesome-svg-core';
 
 describe('NxTextLink', function() {
-  const getShallow = getShallowComponent(NxTextLink, {});
+  const quickRender = rtlRender(NxTextLink, {});
+  const renderEl = rtlRenderElement(NxTextLink, {});
 
-  it('renders an <a> with the nx-text-link class', function() {
-    expect(getShallow()).toMatchSelector('a.nx-text-link');
+  it('renders an <a> with the nx-text-link', function() {
```
rpokorny, line 15:
```suggestion
  it('renders an <a>', function() {
```

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
```diff
@@ -4,115 +4,96 @@
...
-    expect(getShallow({ external: true, noReferrer: true, target: 'asdf' })).toHaveProp('target', 'asdf');
-  });
-
-  it('adds an icon after the children if external is true', function() {
-    expect(getShallow({ children: <span/> })).not.toContainMatchingElement(NxFontAwesomeIcon);
-    expect(getShallow({ children: <span/>, external: true })).toContainMatchingElement(NxFontAwesomeIcon);
+    expect(renderEl({ external: true, noReferrer: true, target: 'asdf' })).toHaveAttribute('target', 'asdf');
   });
 
   it('wraps non-element children in an element sibling to the icon', function() {
```
rpokorny, line 94: This test is probably also an implementation detail.  The span wrapper is just there so the styling works correctly, which is tested in the visual tests.

rpokorny (approved):

lib/src/components/NxTextLink/__tests__/NxTextLink.test.tsx
rjp99: > This test is probably also an implementation detail. The span wrapper is just there so the styling works correctly, which is tested in the visual tests.

Removed here: f1666b672214f5311a9852cb1c13ecf1de9419dd

PR #849, opened 2022-11-10
RTL Conversion for NxTextInput - RSC-1232

cindyko226:
https://issues.sonatype.org/browse/RSC-1232


lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -4,258 +4,518 @@
...
+import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
+import userEvent from '@testing-library/user-event';
 
-import NxTextInput, { PrivateNxTextInput, Props } from '../NxTextInput';
-import { mount } from 'enzyme';
-import { FormAriaContext } from '../../NxForm/context';
+import NxTextInput, { PublicProps } from '../NxTextInput';
+import NxForm from '../../NxForm/NxForm';
 
 describe('PrivateNxTextInput', function() {
```
rpokorny, line 15: `PrivateNxTextInput` isn't what's being tested here, and doesn't need to be directly tested since it's not exposed outside of the RSC library.
```suggestion
describe('NxTextInput', function() {
```

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -4,258 +4,518 @@
...
 
-  it('contains an nx-text-input__box containing the input and validation icons', function() {
-    expect(getShallowComponent()).toContainMatchingElement('.nx-text-input__box');
-    expect(getShallowComponent()).toContainMatchingElement('.nx-text-input__box input.nx-text-input__input');
-    expect(getShallowComponent()).toContainMatchingElement('.nx-text-input__box .nx-icon--valid');
-    expect(getShallowComponent()).toContainMatchingElement('.nx-text-input__box .nx-icon--invalid');
-  });
+  it('renders a password input if type is "password"', function() {
+    // input type="password" don't have a role
+    const inputEl = quickRender({ type: 'password' }).container.querySelector('input');
```
rpokorny, line 30: Why not use `getByRole` here?

rpokorny: > Why not use getByRole here?

Ah I see, password inputs have no role.

lib/src/components/NxTextInput/__tests__/NxTextInput.test.tsx
```diff
@@ -4,258 +4,518 @@
...
-  it('renders a password input if type is "password"', function() {
-    expect(getShallowComponent({ type: 'password' }).find('input')).toHaveProp('type', 'password');
+    expect(input).toHaveAttribute('id', 'foo');
+    expect(input).toHaveAttribute('lang', 'en-US');
   });
 
-  it('renders a date input if type is "date"', function() {
-    expect(getShallowComponent({ type: 'date' }).find('input')).toHaveProp('type', 'date');
+  it('sets the value as specified', function() {
+    expect(quickRender({ value: 'boo' }).getByRole('textbox')).toHaveAttribute('value', 'boo');
```
rpokorny, line 66: The attribute isn't what matters to us and is not kept in sync with the actual `value` DOM property
```suggestion
    expect(quickRender({ value: 'boo' }).getByRole('textbox')).toHaveValue('boo');
```

lib/src/components/NxTextInput/stateful/NxStatefulTextInput.tsx
```diff
@@ -10,7 +10,7 @@ import { initialState, userInput } from '../stateHelpers';
 import { PrivateNxTextInput } from '../NxTextInput';
 
 import { Props, PublicProps, propTypes } from './types';
-export { Props, propTypes } from './types';
+export { Props, PublicProps, propTypes } from './types';
```
rpokorny, line 13: This is also missing/incorrect in `index.ts` I guess

lib/src/components/NxTextInput/stateful/__tests__/NxStatefulTextInput.test.tsx
```diff
@@ -4,148 +4,288 @@
...
+        input = quickRender({ onKeyPress }).getByRole('textbox');
 
     expect(onKeyPress).not.toHaveBeenCalled();
 
-    component.prop('onKeyPress')('q');
+    await user.type(input, 'a');
 
-    expect(onKeyPress).toHaveBeenCalledWith('q');
+    expect(onKeyPress).toHaveBeenCalledWith('a');
   });
```
rpokorny, line 91: Unless I missed it this seems like exactly the same tests that `NxTextInput` has.  The stateful component should have tests for its state handling - that when the user types in the input its `value` and validation change accordingly.

lib/src/components/NxTextInput/stateful/NxStatefulTextInput.tsx
cindyko226: Thanks for catching it, updated here 285bbd59a7ee49a0a3e81ae5b1b202191d6b005a.

lib/src/components/NxTextInput/stateful/__tests__/NxStatefulTextInput.test.tsx
cindyko226: Updated here 8fae95e4f6c234f4b80c51c634b3ee8b4ef02e95.

rpokorny (approved):

cindyko226:
> for `NxTextInput.test.tsx`, Should there be a test confirming no validation errors appear when isPristine: true, validationErrors: 'foo' , validatable: true, BUT input is disabled?

Add the validatable part of validation 78093b770682aae201ff6940bf3ac30db943bff8.

PR #851, opened 2022-11-15
Radio RTL Conversion - RSC-1236

rjp99:
https://issues.sonatype.org/browse/RSC-1236


lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+      expect(container.querySelector('span.nx-radio__content .bar')).toBe;
     });
 
     it('does not render the .nx-radio__content element if there are no children', function() {
-      expect(getShallow()).not.toContainMatchingElement('.nx-radio__content');
+      const el = renderEl()!;
+      expect(el).not.toHaveClass('nx-radio__content');
     });
-
-    it('wraps the .nx-radio__content element in an NxOverflowTooltip unless overflowTooltip is set to false',
```
rjp99, line 116: Added a visual test for the tooltip. Andrew and I were talking about it, and it was getting complicated to convert to RTL so we thought a functional test would be better.

gallery/visualtests/nxRadio.js
```diff
@@ -76,5 +77,17 @@ describe('NxRadio', function() {
     it('looks disabled when hovered', hoverTest(disabledCheckedSelector));
   });
 
+  describe('Tooltip for NxRadio', function() {
+    it('has a tooltip on hover when set to true', async function() {
```
rpokorny, line 81: It seems like this could have been done with `hoverTest()`.  Notice its optional third parameter which is for this case.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+  const quickRender = rtlRender<Props>(NxRadio, minimalProps);
 
   it('renders a <label> containing a radio <input> and .nx-radio__circle and .nx-radio__outer-circle  elements',
       function() {
-        const shallowRender = getShallow();
+        const el = renderEl()!;
 
-        expect(shallowRender).toMatchSelector('label.nx-radio');
+        expect(el.tagName).toBe('LABEL');
+        expect(el).toHaveClass('nx-radio');
```
rpokorny, line 33: The class is an implementation detail

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
 
-        expect(shallowRender.find('input')).toHaveProp('name', 'color');
-        expect(shallowRender.find('input')).toHaveProp('type', 'radio');
-        expect(shallowRender.find('input')).toHaveProp('disabled', false);
-        expect(shallowRender.find('input')).toHaveProp('checked', false);
+        expect(el.querySelector('input')).toHaveAttribute('name', 'color');
+        expect(el.querySelector('input')).toHaveAttribute('type', 'radio');
+        expect(el.querySelector('input')).not.toHaveAttribute('disabled');
+        expect(el.querySelector('input')).not.toHaveAttribute('checked');
```
rpokorny, line 38: Use `toBeDisabled` and `toBeChecked` instead

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
 
-        expect(shallowRender.find('input')).toHaveClassName('nx-radio__input');
+        expect(el.querySelector('input')).toHaveClass('nx-radio__input');
```
rpokorny, line 40: Implementation detail

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
 
-        expect(shallowRender.find('svg.nx-radio__circle')).toExist();
-        expect(shallowRender.find('.nx-radio__circle .nx-radio__outer-circle')).toExist();
+        expect(el.querySelector('svg')).toHaveClass('nx-radio__circle');
+        expect(el.querySelector('circle')).toHaveClass('nx-radio__outer-circle');
```
rpokorny, line 43: Implementation detail, the visuals of the circle are covered by visual tests

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
   );
 
   it('adds classes specified with the className prop', function() {
-    const component = getShallow({ className: 'foo' });
+    const component = renderEl({ className: 'foo' });
 
-    expect(component).toHaveClassName('foo');
-    expect(component).toHaveClassName('nx-radio');
+    expect(component).toHaveClass('foo');
+    expect(component).toHaveClass('nx-radio');
```
rpokorny, line 51: This should use the `for` loop pattern we've established in other RTL tests

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
   });
 
-  it('renders .nx-radio__inner-circle iff it is checked', function() {
-    expect(getShallow()).not.toContainMatchingElement('.nx-radio__inner-circle');
-    expect(getShallow({ isChecked: true })).toContainMatchingElement('.nx-radio__circle .nx-radio__inner-circle');
-    expect(getShallow({ isChecked: true, disabled: true }))
-        .toContainMatchingElement('.nx-radio__circle .nx-radio__inner-circle');
+  it('renders .nx-radio__inner-circle if it is checked', function() {
```
rpokorny, line 54: Covered by visual tests

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+    expect(renderEl()?.querySelector('circle')).not.toHaveClass('nx-radio__inner-circle');
+
+    expect(renderEl({ isChecked: true })?.querySelector('circle')).toHaveClass('nx-radio__inner-circle');
+    expect(renderEl({ isChecked: true, disabled: true })?.querySelector('circle'))
+        .toHaveClass('nx-radio__inner-circle');
   });
 
-  it('uses null as the value passed to onChange if the supplied value is null', function() {
+  it('uses null as the value passed to onChange', async function() {
```
rpokorny, line 62: Why'd you get rid of the second half of the test name here? It's important

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+    const el = renderEl({ id: 'foo', htmlFor: 'baz' });
 
-    expect(component).toHaveProp('id', 'foo');
-    expect(component).toHaveProp('htmlFor', 'baz');
+    expect(el).toHaveAttribute('id', 'foo');
+    expect(el).toHaveAttribute('for', 'baz');
   });
 
   describe('when disabled prop is true', function () {
     it('adds the nx-radio-checkbox--disabled class', function() {
```
rpokorny, line 80: Impl detail

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+      expect(renderEl({ disabled: true })).toHaveClass('nx-radio-checkbox--disabled');
     });
 
     it('disables the input', function() {
-      expect(getShallow().find('input')).toHaveProp('disabled', false);
-      expect(getShallow({ disabled: false }).find('input')).toHaveProp('disabled', false);
-      expect(getShallow({ disabled: true }).find('input')).toHaveProp('disabled', true);
+      expect(renderEl()?.querySelector('input')).not.toHaveAttribute('disabled');
+      expect(renderEl({ disabled: false })?.querySelector('input')).not.toHaveAttribute('disabled');
+      expect(renderEl({ disabled: true })?.querySelector('input')).toHaveAttribute('disabled');
```
rpokorny, line 88: Use `toBeDisabled`

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
     });
   });
 
   describe('isChecked prop', function () {
     it('sets the input to checked per the value of isChecked', function() {
-      expect(getShallow({ isChecked: false }).find('input')).toHaveProp('checked', false);
-      expect(getShallow({ isChecked: true }).find('input')).toHaveProp('checked', true);
+      expect(renderEl({ isChecked: false })?.querySelector('input')).not.toHaveAttribute('checked');
+      expect(renderEl({ isChecked: true })?.querySelector('input')).toHaveAttribute('checked');
```
rpokorny, line 95: Query by role, use `toBeChecked`

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
 
-      expect(getShallow({ isChecked: true })).not.toHaveClassName('tm-unchecked');
-      expect(getShallow({ isChecked: true })).toHaveClassName('tm-checked');
+      expect(renderEl({ isChecked: true })).not.toHaveClass('tm-unchecked');
+      expect(renderEl({ isChecked: true })).toHaveClass('tm-checked');
     });
   });
 
   describe('children prop', function () {
     it('renders children nodes within an nx-radio__content <span>', function() {
```
rpokorny, line 108: The `span.nx-radio__content` itself is an implementation detail, this test should be simplified to just check that `.bar` is among the descendants.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+      const { container, rerender } = quickRender();
+
+      rerender(
+        <NxRadio { ...minimalProps }>
           <div className="bar"></div>
         </NxRadio>
       );
 
-      expect(render).toContainMatchingElement('label span.nx-radio__content .bar');
+      expect(container.querySelector('span.nx-radio__content .bar')).toBe;
```
rpokorny, line 117: `toBe` doesn't do anything until you actually call it.  You probably want `.toBeInTheDocument()` here.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
     });
 
     it('does not render the .nx-radio__content element if there are no children', function() {
```
rpokorny, line 120: Impl detail

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+
     expect(onChange).toHaveBeenCalledWith('red');
   });
 
   it('sets the input as readonly if there is no onChange handler', function() {
-    expect(getShallow().find('input')).toHaveProp('readOnly', false);
-    expect(getShallow({ onChange: undefined }).find('input')).toHaveProp('readOnly', true);
-    expect(getShallow({ onChange: null }).find('input')).toHaveProp('readOnly', true);
+    expect(renderEl()?.querySelector('input')).not.toHaveAttribute('readOnly');
+    expect(renderEl({ onChange: undefined })?.querySelector('input')).toHaveAttribute('readOnly');
```
rpokorny, line 139: At the HTML level it's `readonly` (all lowercase)

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
+    expect(renderEl({ onChange: null })?.querySelector('input')).toHaveAttribute('readOnly');
   });
 
   it('adds id attribute to radio input when radioId prop is provided', function () {
-    const component = getShallow({ radioId: 'color-red'});
-    expect(component.find('input')).toHaveProp('id', 'color-red');
+    const component = renderEl({ radioId: 'color-red'});
+    expect(component?.querySelector('input')).toHaveAttribute('id', 'color-red');
```
rpokorny, line 145: Query by role

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -175,11 +161,11 @@ describe('NxRadio', function() {
       } as Props['inputAttributes']
     });
 
-    expect(component.find('input')).toHaveProp('id', 'garfield');
-    expect(component.find('input')).toHaveProp('name', 'garfield');
-    expect(component.find('input')).toHaveProp('disabled', true);
-    expect(component.find('input')).toHaveClassName('input-classname');
-    expect(component.find('input')).not.toHaveClassName('label-classname');
-    expect(component.find('input')).toHaveProp('checked', true);
+    expect(component?.querySelector('input')).toHaveAttribute('id', 'garfield');
```
rpokorny, line 164: Query by role

gallery/visualtests/nxRadio.js
rjp99: > It seems like this could have been done with `hoverTest()`. Notice its optional third parameter which is for this case.

I didn't know that I could have two selectors and wait for the tooltip, thanks! 28e27ee9fdb3d586de9a5536e59d0f52643e0883

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > The class is an implementation detail

Removed here: 7bc38eee5e88ff8e0e4785a09cdb57813fbaa056

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Use `toBeDisabled` and `toBeChecked` instead

Updated here: 94e7448e1893bf604efaafb4810a0ed864ed8f13

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,131 @@ describe('NxRadio', function() {
...
-        expect(shallowRender).toMatchSelector('label.nx-radio');
+        expect(el.tagName).toBe('LABEL');
+        expect(el).toHaveClass('nx-radio');
 
-        expect(shallowRender.find('input')).toHaveProp('name', 'color');
-        expect(shallowRender.find('input')).toHaveProp('type', 'radio');
-        expect(shallowRender.find('input')).toHaveProp('disabled', false);
-        expect(shallowRender.find('input')).toHaveProp('checked', false);
+        expect(el.querySelector('input')).toHaveAttribute('name', 'color');
+        expect(el.querySelector('input')).toHaveAttribute('type', 'radio');
```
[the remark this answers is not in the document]
rjp99: > Probably better to query by role rather than by element. I also don't think there's a need to check the `type` if you query by role.

Like this?: 5cab60ae3b1ac98a479a240989a9c696753669d3

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Implementation detail

Removed here: 5a92d00f1f77978ce7e5f510f4c26d5f13c7ac74

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > This should use the `for` loop pattern we've established in other RTL tests

Updated her: 984bdacd64b861440c86635dd61de604207c7a83

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Covered by visual tests

Removed here: 342dd2aab0bf0bd2a6dc0df26e0672fb04ff9dfe

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Why'd you get rid of the second half of the test name here? It's important

Don't know why it was changed, must've done it accidentally: bfcfcb45e3e795890140811d7307ef4fb3153403

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Impl detail

removed here: 919e7787714691360b37f39cbde36fbf9f3b79f6

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Use `toBeDisabled`, and query by role

Updated here: 967874349164ef10c678e7a76bea5e78d2d0bee8

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Query by role, use `toBeChecked`

Updated here: d8ebbecdb39597bd0cb70a70212b791520355608

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > The `span.nx-radio__content` itself is an implementation detail, this test should be simplified to just check that `.bar` is among the descendants.

Updated here: 2c79beb07760dd08ceb80320a591cf9689044f2d

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > `toBe` doesn't do anything until you actually call it. You probably want `.toBeInTheDocument()` here.

Updated here: 2c79beb07760dd08ceb80320a591cf9689044f2d

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Impl detail

Removed here: 26b725ed6a5ca9ecd47ef2f738f59e983444a6c6

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > At the HTML level it's `readonly` (all lowercase)

Updated here: d90926f741e3ce45781229b29afce572b46bf788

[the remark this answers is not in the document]
rjp99: > Query by role

Updated here: b78163223ca7c1cb98381cfbf3839f0fdc8c58b6

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Query by role

Updated here: 8312e31a17f55c28d146b06c236638c847499b6d

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Query by role

Updated here: 1d352481a17440bd55150969e9a540e74ef2c4fe

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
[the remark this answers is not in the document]
rpokorny: > Like this?: https://github.com/sonatype/sonatype-react-shared-components/commit/5cab60ae3b1ac98a479a240989a9c696753669d3

On the lines below that too though

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,109 @@ describe('NxRadio', function() {
...
-        expect(shallowRender.find('input')).toHaveProp('type', 'radio');
-        expect(shallowRender.find('input')).toHaveProp('disabled', false);
-        expect(shallowRender.find('input')).toHaveProp('checked', false);
+        expect(el.tagName).toBe('LABEL');
 
-        expect(shallowRender.find('input')).toHaveClassName('nx-radio__input');
-
-        expect(shallowRender.find('svg.nx-radio__circle')).toExist();
-        expect(shallowRender.find('.nx-radio__circle .nx-radio__outer-circle')).toExist();
+        expect(quickRender().queryByRole('radio')).toHaveAttribute('name', 'color');
```
rpokorny, line 34: Sorry, everywhere I said "query by role" I didn't mean the literal `queryByRole` function, I meant any appopriate [RTL query function](https://testing-library.com/docs/queries/about/) by role.  `queryByRole` itself is actually rarely used.  The only difference between it and `getByRole` is that if the element isn't found, `getByRole` will throw an exception (with a very elaborate and helpful error message) while `queryByRole` will just return `null`.  As such, the documented guidance is to always use `getByRole` unless you are trying to prove that a particular element is _not_ present.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -175,11 +139,9 @@ describe('NxRadio', function() {
...
-    expect(component.find('input')).toHaveProp('id', 'garfield');
-    expect(component.find('input')).toHaveProp('name', 'garfield');
-    expect(component.find('input')).toHaveProp('disabled', true);
-    expect(component.find('input')).toHaveClassName('input-classname');
-    expect(component.find('input')).not.toHaveClassName('label-classname');
-    expect(component.find('input')).toHaveProp('checked', true);
+    expect(component.queryByRole('radio')).toHaveAttribute('id', 'garfield');
+    expect(component.queryByRole('radio')).toHaveAttribute('name', 'garfield');
+    expect(component.queryByRole('radio')).toBeDisabled();
+    expect(component.queryByRole('radio')).toBeChecked();
```
rpokorny, line 145: Custom classnames here were not implementation details.  We want to make sure that `inputAttributes.className` works and that it doesn't get mixed up with the top-level `className`.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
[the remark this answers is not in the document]
rjp99: > > Like this?: [5cab60a](https://github.com/sonatype/sonatype-react-shared-components/commit/5cab60ae3b1ac98a479a240989a9c696753669d3)
> 
> On the lines below that too though

I thought I got them all: f587d8d25edfbd1c7cac6f46abd13da5f6e45941

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > > Added a visual test for the tooltip. Andrew and I were talking about it, and it was getting complicated to convert to RTL so we thought a functional test would be better.
> 
> Hmm, the visual test doesn't cover the `overflowTooltip` prop though, does it? It's true that `NxOverflowTooltip` can't meaningfully be tested in unit tests though, because the DOM layout logic that it relies on is not implemented in JSDOM. So testing it in functional tests is the right way to go, but I suspect the current test doesn't fully test the prop.

What would be a good way to test it, can I maybe add an expect there to check if the radio has NxOverflowTooltip as an attribute or would that not work?

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Sorry, everywhere I said "query by role" I didn't mean the literal `queryByRole` function, I meant any appopriate [RTL query function](https://testing-library.com/docs/queries/about/) by role. `queryByRole` itself is actually rarely used. The only difference between it and `getByRole` is that if the element isn't found, `getByRole` will throw an exception (with a very elaborate and helpful error message) while `queryByRole` will just return `null`. As such, the documented guidance is to always use `getByRole` unless you are trying to prove that a particular element is _not_ present.

Ohh okay, thanks for explaining the difference. Changed here: 00092cc4c5329310821b561d1677cdbf73832497

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: > Custom classnames here were not implementation details. We want to make sure that `inputAttributes.className` works and that it doesn't get mixed up with the top-level `className`.

Added them back: 395719f19c7504346ded37566cbc6ed07d72f0f2

The test for the class for `isChecked`, `'tm-checked'`, would be the same thing?

rpokorny: > The test for the class for isChecked, 'tm-checked', would be the same thing?

The tm classes are a bit of a special case.  at least one of the frontend user analytics tools that Sonatype products use or have used can't pick up the actual checked state of the checkboxes, but can pick up classnames.  So as a workaround we have our checkboxes and radios expose their checked state as a classname.  Since that is an externally-relevant feature, yes it should still be tested

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rpokorny: > What would be a good way to test it, can I maybe add an expect there to check if the radio has NxOverflowTooltip as an attribute or would that not work?

In my experience the `userEvent.hover` function does work for triggering tooltips in unit tests.  You can use that on an element and then do `screen.getByRole('tooltip')` to detect whether the tooltip appeared.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,107 @@ describe('NxRadio', function() {
...
-
-        expect(shallowRender).toMatchSelector('label.nx-radio');
-
-        expect(shallowRender.find('input')).toHaveProp('name', 'color');
-        expect(shallowRender.find('input')).toHaveProp('type', 'radio');
-        expect(shallowRender.find('input')).toHaveProp('disabled', false);
-        expect(shallowRender.find('input')).toHaveProp('checked', false);
-
-        expect(shallowRender.find('input')).toHaveClassName('nx-radio__input');
+        expect(renderEl()!.tagName).toBe('LABEL');
```
rpokorny, line 30: Noticed when reviewing the `NxCheckbox` ticket, but it applies here as well:

> I tend to think the label element is an implementation detail. We do want to check that it gets connected as the accessible name of the input though.

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
```diff
@@ -21,146 +22,113 @@ describe('NxRadio', function() {
     children: undefined
   };
 
-  const getShallow = getShallowComponent<Props>(NxRadio, simpleProps);
+  const renderEl = rtlRenderElement(NxRadio, minimalProps);
+  const quickRender = rtlRender(NxRadio, minimalProps);
 
-  it('renders a <label> containing a radio <input> and .nx-radio__circle and .nx-radio__outer-circle  elements',
+  it('renders a <label> containing a radio <input> and disabled and checked props',
```
rpokorny, line 28:
```suggestion
  it('renders radio <input> with an accessible name specified by the children,and disabled and checked props',
```

lib/src/components/NxRadio/__tests__/NxRadio.test.tsx
rjp99: I added the visual tests for the Radio label does not wrap example here: ec817cb345ce1fc17da4ea313ca1bf66bc744166

gallery/visualtests/nxRadio.js
```diff
@@ -76,5 +76,19 @@ describe('NxRadio', function() {
     it('looks disabled when hovered', hoverTest(disabledCheckedSelector));
   });
 
+  describe('Tooltip for NxRadio', function() {
+    const selector = '#nx-radio-example .gallery-example-live',
+        hoverSelector = '#nx-radio-example .gallery-example-live label:nth-of-type(3)';
+    it('has a tooltip on hover when set to true', hoverTest(selector, hoverSelector, true));
```
rpokorny, line 82: Let's have this one test for the visuals of it and then use actual assertions on the DOM in the two tests below, expecting the `.nx-tooltip` to either be or not be present.  Using more screenshots to verify that is a bit wasteful.

rjp99: > Let's have this one test for the visuals of it and then use actual assertions on the DOM in the two tests below, expecting the `.nx-tooltip` to either be or not be present. Using more screenshots to verify that is a bit wasteful.

Updated here: 89466db8325d301cbfe8cc61516a92123789f831

rpokorny: > Updated here: https://github.com/sonatype/sonatype-react-shared-components/commit/89466db8325d301cbfe8cc61516a92123789f831

That lost the test for the one where the prop is set to false

rjp99: > > Updated here: [89466db](https://github.com/sonatype/sonatype-react-shared-components/commit/89466db8325d301cbfe8cc61516a92123789f831)
> 
> That lost the test for the one where the prop is set to false

Added here: abeaf022ac8711ce20477aca59311b5cee67bd01

gallery/visualtests/nxRadio.js
```diff
@@ -76,5 +78,31 @@ describe('NxRadio', function() {
...
+
+    it('does have a tooltip on hover only when set to true', async function() {
```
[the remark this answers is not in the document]
rjp99: > Three minor improvements:
> 
> 1. the prop defaults to true, so the way you named the test wasn't quite accurate
> 2. mentioning which prop we are talking about is a good idea
> 3. The test should probably also check that hovering a radio that doesn't set `overflowTooltip` to false, but which is also not actually overflowing, does not get a tooltip.

Added the test here: 7bf4bcc925783a695b7e672c41dca6b130a63aa2

rpokorny (approved):

PR #857, opened 2022-11-23
Allow to Disable Moving Feature in NxTransferListHalf - RSC-1254

cindyko226:
https://issues.sonatype.org/browse/RSC-1254

### DIFF
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
@@ -23,6 +23,6 @@ export default function NxTransferListHalfDisableTransferExample() {
   return <NxTransferListHalf label="Example Items"
                              filterValue={filter}
                              onFilterChange={setFilter}
-                             items={Items}
-                             footerContent={`${Items.length} items`} />;
+                             items={items}
+                             footerContent={`${items.length} items`} />;
 }
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
@@ -9,7 +9,7 @@ import { DataItem, NxFontAwesomeIcon, NxTransferListHalf } from '@sonatype/react
 import { map, range, prepend } from 'ramda';
 import { faArrowsAltH } from '@fortawesome/free-solid-svg-icons';
 
-const Items: DataItem<number>[] = prepend(
+const items: DataItem<number>[] = prepend(
     {
       id: 0,
       displayName: <><NxFontAwesomeIcon icon={faArrowsAltH} /><span>Loooooooooooooooooooooooooong Name</span></>
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
@@ -24,6 +24,5 @@ export default function NxTransferListHalfDisableTransferExample() {
                              filterValue={filter}
                              onFilterChange={setFilter}
                              items={Items}
-                             isSelected={true}
                              footerContent={`${Items.length} items`} />;
 }
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
@@ -23,8 +23,6 @@ export default function NxTransferListHalfDisableTransferExample() {
   return <NxTransferListHalf label="Example Items"
                              filterValue={filter}
                              onFilterChange={setFilter}
-                             showMoveAll={false}
-                             onMoveAll={() => {}}
                              items={Items}
                              isSelected={true}
                              footerContent={`${Items.length} items`} />;
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
@@ -23,11 +23,9 @@ export default function NxTransferListHalfDisableTransferExample() {
   return <NxTransferListHalf label="Example Items"
                              filterValue={filter}
                              onFilterChange={setFilter}
-                             disableItemMove={true}
                              showMoveAll={false}
                              onMoveAll={() => {}}
                              items={Items}
                              isSelected={true}
-                             onItemChange={() => {}}
                              footerContent={`${Items.length} items`} />;
 }
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { useState } from 'react';
+import { DataItem, NxFontAwesomeIcon, NxTransferListHalf } from '@sonatype/react-shared-components';
+import { map, range, prepend } from 'ramda';
+import { faArrowsAltH } from '@fortawesome/free-solid-svg-icons';
+
+const Items: DataItem<number>[] = prepend(
+    {
+      id: 0,
+      displayName: <><NxFontAwesomeIcon icon={faArrowsAltH} /><span>Loooooooooooooooooooooooooong Name</span></>
+    },
+    map<number, DataItem<number>>(i => ({ id: i, displayName: `Item ${i}` }), range(1, 101))
+);
+
+export default function NxTransferListHalfDisableTransferExample() {
+  const [filter, setFilter] = useState('');
+
+  return <NxTransferListHalf label="Example Items"
+                             filterValue={filter}
+                             onFilterChange={setFilter}
+                             disableItemMove={true}
+                             showMoveAll={false}
+                             onMoveAll={() => {}}
+                             items={Items}
+                             isSelected={true}
+                             onItemChange={() => {}}
+                             footerContent={`${Items.length} items`} />;
+}
--- a/gallery/src/components/NxTransferListHalf/NxTransferListHalfPage.tsx
+++ b/gallery/src/components/NxTransferListHalf/NxTransferListHalfPage.tsx
@@ -13,6 +13,8 @@ import NxTransferListHalfExample from './NxTransferListHalfExample';
 import NxTransferListHalfCustomTooltipExample from './NxTransferListHalfCustomTooltipExample';
 import NxTransferListHalfOrderingExample from './NxTransferListHalfOrderingExample';
 import NxTransferListHalfDisableTransferExample from './NxTransferListHalfDisableTransferExample';
+import NxTransferListHalfOrderingWithDisableTransferExample
+  from './NxTransferListHalfOrderingWithDisableTransferExample';
 
 import '../NxTooltip/NxTooltipExample.scss';
 
@@ -20,7 +22,9 @@ const nxTransferListHalfExample = require('./NxTransferListHalfExample?raw'),
     nxTransferListHalfCustomTooltipExample = require('./NxTransferListHalfCustomTooltipExample?raw'),
     nxTransferListHalfOrderingExample = require('./NxTransferListHalfOrderingExample?raw'),
     tooltipScss = require('../NxTooltip/NxTooltipExample.scss?raw'),
-    nxTransferListHalfDisableTransferExample = require('./NxTransferListHalfDisableTransferExample?raw');
+    nxTransferListHalfDisableTransferExample = require('./NxTransferListHalfDisableTransferExample?raw'),
+    nxTransferListHalfOrderingWithDisableTransferExample =
+      require('./NxTransferListHalfOrderingWithDisableTransferExample?raw');
 
 const tooltipExampleCode = [nxTransferListHalfCustomTooltipExample, { language: 'scss', content: tooltipScss }];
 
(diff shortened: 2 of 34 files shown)


EmilyBermudez88:
I think it's a good idea to add an example of the new disabledItemMove prop in the gallery

gallery/src/components/NxTransferListHalf/NxTransferListHalfPage.tsx
```diff
@@ -108,6 +108,15 @@ const NxTransferListPage = () =>
                 the <NxCode>isSelected</NxCode> prop) and the id of the item.
               </NxTable.Cell>
             </NxTable.Row>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>disableItemMove</NxCode></NxTable.Cell>
+              <NxTable.Cell>boolean</NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell></NxTable.Cell>
+              <NxTable.Cell>
+                Set to true to disable moving item feature and hide the move icon on each item.
```
EmilyBermudez88, line 117: Maybe something like : 'When true it disable’s the ability to transfer or remove items by hiding the item’s icon.'

rpokorny:
Rather than introducing a new prop, would it make sense to just make `onItemChange` optional, and to not show the icons when it is not set?

cindyko226:
> I think it's a good idea to add an example of the new disabledItemMove prop in the gallery

Sure. Added here 
914f60cb038492809ddb496129f90f9b0036bca8
aaf191ce1d866ecba96aa6ffa3c95f93c7bb490e

cindyko226:
> Rather than introducing a new prop, would it make sense to just make `onItemChange` optional, and to not show the icons when it is not set?

Yes, it makes sense to me since the `onItemChange` prop is not necessary when disabling the transfer feature. 
Updated here aaf191ce1d866ecba96aa6ffa3c95f93c7bb490e, thanks.

gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
```diff
@@ -0,0 +1,31 @@
...
+);
+
+export default function NxTransferListHalfDisableTransferExample() {
+  const [filter, setFilter] = useState('');
+
+  return <NxTransferListHalf label="Example Items"
+                             filterValue={filter}
+                             onFilterChange={setFilter}
+                             showMoveAll={false}
+                             onMoveAll={() => {}}
```
rpokorny, line 27: It seems a little goofy that `showMoveAll` and `onMoveAll` need to be provided in this case.  For that matter, it seems a little goofy that `onMoveAll` needs to be provided any time `showMoveAll` is false.  IMO those two props should also be made optional and excluded from this example.

rpokorny:
The items still have hover and click styles when moving is disabled, even though clicking them doesn't do anything.  Should those styles be disabled?

rpokorny:
Needs a visual test.  Also I think the original use case was to have reordering but not removal.  So we should have an example (and visual test) for that too.

gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
cindyko226: Updated here 8a6881fef17ec3292383fa277f21d76019b90b93.
Further on this, since there won't be selected/unselected items in this case, should `isSelected` prop be optional as well?

rpokorny: > Further on this, since there won't be selected/unselected items in this case, should isSelected prop be optional as well?

Sure

cindyko226: Updated here 8b6823bcb2af5335fecb102363cf56f47854aed8

cindyko226:
> The items still have hover and click styles when moving is disabled, even though clicking them doesn't do anything. Should those styles be disabled?

Checked with design @JasonDSwearingen, in this case the item should not be clickable or have interaction states while tooltip still shows on hover when display name is truncated.
Updated sketch mockups: https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/m1095Gz

Those styles should be disabled, updated here 9370018954ef06e3f5b78538ed371cab13bea0e5

gallery/src/components/NxTransferListHalf/NxTransferListHalfDisableTransferExample.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React, { useState } from 'react';
+import { DataItem, NxFontAwesomeIcon, NxTransferListHalf } from '@sonatype/react-shared-components';
+import { map, range, prepend } from 'ramda';
+import { faArrowsAltH } from '@fortawesome/free-solid-svg-icons';
+
+const Items: DataItem<number>[] = prepend(
```
rpokorny, line 12: There's no reason this should start uppercase

```suggestion
const items: DataItem<number>[] = prepend(
```

gallery/src/components/NxTransferListHalf/NxTransferListHalfPage.tsx
```diff
@@ -69,16 +71,19 @@ const NxTransferListPage = () =>
             <NxTable.Row>
               <NxTable.Cell><NxCode>showMoveAll</NxCode></NxTable.Cell>
               <NxTable.Cell>boolean</NxTable.Cell>
-              <NxTable.Cell>Yes</NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
               <NxTable.Cell></NxTable.Cell>
```
rpokorny, line 75: Optional props need to have their default value specified when applicable

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
```diff
@@ -40,14 +40,17 @@ function _TransferListItem<T extends string | number = string>(props: TransferLi
...
+    onChangeProp?.(evt.currentTarget.checked, id);
   }
 
   const classes = classnames(
       'nx-transfer-list__item',
       {
         'nx-transfer-list__item--with-reordering': !!showReorderingButtons
       },
+      {
+        'nx-transfer-list__moving--disabled': !onChangeProp
```
rpokorny, line 52: This isn't correct BEM.  I'd call it `nx-transfer-list__item--moving-disabled`.  Or put a `.nx-transfer-list__half--moving-disabled` class on the `.nx-transfer-list__half`

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -171,9 +168,23 @@ describe('NxTransferListHalf', function() {
...
+    expect(select.find(NxFontAwesomeIcon)).not.toExist();
+    expect(select.find('input.nx-transfer-list__checkbox')).not.toExist();
+  });
+
   it('sets the item checked prop to isSelected', function() {
-    const withSelected = getMounted({ items: [{ id: 1, displayName: 'foo' }], isSelected: true }),
-        withoutSelected = getMounted({ items: [{ id: 1, displayName: 'foo' }] });
+    const onItemChange = jest.fn(),
+        withSelected = getMounted({ items: [{ id: 1, displayName: 'foo' }], isSelected: true, onItemChange }),
+        withoutSelected = getMounted({ items: [{ id: 1, displayName: 'foo' }], onItemChange });
```
rpokorny, line 187: Should test `isSelected = false` as well

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
rpokorny: > This isn't correct BEM. I'd call it nx-transfer-list__item--moving-disabled. Or put a .nx-transfer-list__half--moving-disabled class on the .nx-transfer-list__half

Better yet make a class for the opposite case: a `.nx-transfer-list__item--movable` or something.  Writing the CSS for that should be easier

gallery/src/components/NxTransferListHalf/NxTransferListHalfPage.tsx
cindyko226: Updated here 9cf90c48d024ff1921bec3ca8088ae74e46aa821

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
cindyko226: Ah I miss the item.
Thanks for the suggestions, updated here 6f2d800213933eb20a58d91c8b7f579d8ef176b8.

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
cindyko226: Updated here 4dc735f5e5008c35a9245561f423c697ce343183.

cindyko226:
> Needs a visual test. 

Added here 
d658e734d93d6adc1165298a9550120c4eaa62ed
a73bfb86162ef154bfc35f817c42ee5a6e804235.

> Also I think the original use case was to have reordering but not removal. So we should have an example (and visual test) for that too.

Added here 
8802b01b53ca01d321a4347d69d1d2755b8003fc
This is the combination of disable transfer and ordering example, not sure if it is necessary to also have the visual tests for interaction states on each item and move up/down buttons, let me know your opinions, thanks.

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
```diff
@@ -148,7 +148,7 @@ export default function NxTransferListHalf<T extends string | number = string>(p
             <span>{isSelected ? 'Remove' : 'Transfer'} All</span>
           </button>
         }
-        <div className="nx-transfer-list__item-list">
+        <div className="nx-transfer-list__item-list" tabIndex={onItemChange || allowReordering ? undefined : 0}>
```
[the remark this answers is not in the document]
rpokorny: > Add the tabIndex here to meet the a11y requirement that scrollable region must have keyboard access

Great.  I think that's worth adding a code comment

gallery/src/components/NxTransferListHalf/NxTransferListHalfPage.tsx
```diff
@@ -90,8 +99,8 @@ const NxTransferListPage = () =>
             <NxTable.Row>
               <NxTable.Cell><NxCode>isSelected</NxCode></NxTable.Cell>
               <NxTable.Cell>boolean</NxTable.Cell>
-              <NxTable.Cell>Yes</NxTable.Cell>
-              <NxTable.Cell></NxTable.Cell>
+              <NxTable.Cell>No</NxTable.Cell>
+              <NxTable.Cell>false</NxTable.Cell>
```
rpokorny, line 103: IMO we should actually make this default to true.  In cases where you have just a single `NxTransferListHalf`, it would presumably be holding things that are selected.

lib/src/components/NxTransferListHalf/NxTransferListHalf.scss
```diff
@@ -104,22 +104,29 @@ $-visible-item-count: 10;
 
   border-radius: var(--nx-border-radius);
   border: var(--nx-border-default);
-  cursor: pointer;
   flex: auto;
   padding: var(--nx-spacing-2x);
+  padding-left: var(--nx-spacing-4x);
```
rpokorny, line 109: Why is this 4x? Shouldn't it be 2x whether the icon is there or not?

rpokorny: > Why is this 4x? Shouldn't it be 2x whether the icon is there or not?

I see that this does match the mockup.  However I recall a discussion with the designers where I believe they confirmed the spacing would be the same with and without the icon.  So unless you've doubled checked this with them already, I'm wondering if it's a mistake in the mockup.

gallery/visualtests/nxTransferListHalf.js
```diff
@@ -111,5 +116,21 @@ describe('NxTransferListHalf', function() {
...
   });
 
+  describe('disable transfer', function() {
+    it('looks right', simpleTest(disableTransferListSelector));
+    it('does not put a dark border on hovered items',
+        hoverTest(disableTransferListSelector, secondDisableTransferItemSelector));
+    it('does not put a blue border and glow on focused items',
+        focusTest(disableTransferListSelector, secondDisableTransferItemSelector));
+    it('does not put a blue border and blue glow on focused+hovered items',
+        focusAndHoverTest(disableTransferListSelector, secondDisableTransferItemSelector));
```
rpokorny, line 126: Given that the user can't actually focus the items I don't think these two tests are necessary

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
```diff
@@ -40,14 +40,17 @@ function _TransferListItem<T extends string | number = string>(props: TransferLi
...
 
   const classes = classnames(
       'nx-transfer-list__item',
       {
         'nx-transfer-list__item--with-reordering': !!showReorderingButtons
       },
+      {
+        'nx-transfer-list__item--movable': !!onChangeProp
+      }
   );
```
rpokorny, line 54:
```suggestion
  const classes = classnames('nx-transfer-list__item', {
    'nx-transfer-list__item--with-reordering': !!showReorderingButtons
    'nx-transfer-list__item--movable': !!onChangeProp
  });
```

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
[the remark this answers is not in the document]
cindyko226: Updated here. 0211efb136eb72fcc99e81fdd4fd32ad193237cd

lib/src/components/NxTransferListHalf/NxTransferListHalf.scss
cindyko226: I confirmed with designs @JasonDSwearingen, the padding would be 4x when rendered without the icon.

JasonDSwearingen (approved):

lib/src/components/NxTransferListHalf/NxTransferListHalf.tsx
```diff
@@ -106,7 +108,7 @@ export default function NxTransferListHalf<T extends string | number = string>(p
         onFilterChange,
         showMoveAll,
         onMoveAll,
-        isSelected,
+        isSelected = true,
```
rpokorny, line 111: Because we also allow null on optional props, this syntax is unfortunately not usually useful for us - it only catches `undefined`.  You should add unit tests for when `isSelected` is `null`, and you'll see that they don't pass.

cindyko226: Thanks for catching it, updated here 15fa580e13a126a1104f8bf8e856d24e4e1ed3df.

PR #860, opened 2022-11-25
Remove suffixContent from NxTextInput's PublicProps - RSC-1255

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1255

### DIFF
--- a/lib/src/components/NxTextInput/types.ts
+++ b/lib/src/components/NxTextInput/types.ts
@@ -52,7 +52,7 @@ export type Props = Omit<StateProps, 'trimmedValue'> & HTMLProps & {
   suffixContent?: ReactNode | null;
 };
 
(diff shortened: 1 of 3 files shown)


lib/src/components/NxTextInput/types.ts
```diff
@@ -52,7 +52,7 @@ export type Props = Omit<StateProps, 'trimmedValue'> & HTMLProps & {
   suffixContent?: ReactNode | null;
 };
 
-export interface PublicProps extends Omit<Props, 'prefixContent'> {
+export interface PublicProps extends Omit<Props, 'prefixContent'|'suffixContent'> {
```
rpokorny, line 55:
```suggestion
export interface PublicProps extends Omit<Props, 'prefixContent' | 'suffixContent'> {
```

PR #861, opened 2022-11-25
Fixed nx-counter inside NxCollapsibleMultiSelect not aligned to the right when disabled - RSC-1262

entroform:
https://issues.sonatype.org/browse/RSC-1262

### DIFF
--- a/gallery/visualtests/nxCollapsibleMultiSelect.js
+++ b/gallery/visualtests/nxCollapsibleMultiSelect.js
@@ -10,8 +10,14 @@ describe('NxCollapsibleMultiSelect', function() {
   const { focusTest, simpleTest, a11yTest } = setupBrowser('#/pages/Collapsible Multi-Select');
 
   const selector = '#nx-collapsible-multi-select-example .nx-collapsible-items--select';
+  const disabledWithTooltipSelector
+    = '#nx-collapsible-multi-select-disabled-with-tooltip-example .gallery-example-live';
 
   it('looks right', simpleTest(selector));
+
+  describe('NxCollapsibleMultiSelect Disabled With Tooltip', function() {
+    it('looks right', simpleTest(disabledWithTooltipSelector));
+  });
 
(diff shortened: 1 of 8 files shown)


lib/src/components/NxCollapsibleItems/NxCollapsibleItems.scss
```diff
@@ -74,6 +74,10 @@ $-form-child-left-padding: calc(#{$-trigger-side-padding} + #{$-triangle-box-wid
   }
 }
 
+.nx-collapsible-items__tooltip-wrapper {
+  display: contents;
```
rpokorny, line 78: This breaks the tooltip placement unfortunately

rpokorny: > This breaks the tooltip placement unfortunately

I think the better solution is to add the `.nx-collapsible-items__header` styles to this element as well.

gallery/visualtests/nxCollapsibleMultiSelect.js
```diff
@@ -10,9 +10,15 @@ describe('NxCollapsibleMultiSelect', function() {
   const { focusTest, simpleTest, a11yTest } = setupBrowser('#/pages/Collapsible Multi-Select');
 
   const selector = '#nx-collapsible-multi-select-example .nx-collapsible-items--select';
+  const disabledWithTooltipSelector
+    = '#nx-collapsible-multi-select-disabled-with-tooltip-example .gallery-example-live';
 
   it('looks right', simpleTest(selector));
 
+  describe('NxCollapsibleMultiSelect Disabled With Tooltip', function() {
+    it('looks right', simpleTest(disabledWithTooltipSelector));
```
rpokorny, line 19: Should probably have a hover+tooltip test as well given that the initial fix in this PR didn't address the tooltip correctly.

rpokorny (approved):

PR #865, opened 2022-11-30
Accordion RTL Conversion - RSC-1222

rjp99:
https://issues.sonatype.org/browse/RSC-1222


lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
+      <NxAccordion>
+        <NxAccordion.Header>
+          <span className='foo'>Foo</span>
+        </NxAccordion.Header>
+        <span className='bar'>Bar</span>
+      </NxAccordion>
+    );
+
+    expect(container.querySelector('.nx-accordion__header .foo')).toBeInTheDocument();
+    expect(container.querySelector('.nx-accordion__content .bar')).toBeInTheDocument();
```
rpokorny, line 61: Rather than testing that it's in `.nx-accordion__content`, I'd test that it's in the accordion in general, and that it's not in the `<summary>` element

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
-          }),
-          wrapper = component.find('summary').children();
-
-      expect(wrapper.children().first()).toMatchSelector(NxFontAwesomeIcon);
-      expect(wrapper.children().at(1)).toMatchSelector('span');
-      expect(wrapper.children().last()).toMatchSelector('div');
-
-      expect(wrapper.find(NxFontAwesomeIcon)).toHaveClassName('nx-accordion__chevron');
+      expect(container.querySelector('#headerId')).toBeInTheDocument();
+      expect(container.querySelector('.nx-accordion__summary-wrapper')).toBeInTheDocument();
```
rpokorny, line 75: This class is an impl detail

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
-            children: (
-              <NxAccordion.Header>
-                <span>Foo</span>
-              </NxAccordion.Header>
-            )
-          }),
-          explicitOpenHeader = explicitOpenComponent.find('summary');
-
-      expect(explicitOpenHeader.find(NxFontAwesomeIcon)).toHaveProp('icon', faChevronCircleDown);
+    it('sets the nx-accordion__header class along with any provided className on the header', function() {
```
rpokorny, line 78: `.nx-accordion__header` is an impl detail.  I'd replace this with our usual style of test for custom classnames that checks that all classes present without a custom classname are also still present with a custom classname.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
-            children: (
-              <NxAccordion.Header>
-                <span>Foo</span>
-              </NxAccordion.Header>
-            )
-          }),
-          header = component.find('summary');
-
-      expect(header.find(NxFontAwesomeIcon)).toHaveProp('icon', faChevronCircleUp);
+    it('renders an icon with nx-accordion__chevron class as the first child of the header wrapper', function() {
```
rpokorny, line 93: Impl/visual detail

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
+        const { container } = render(
+          <NxAccordion onToggle={onToggle} open={false}>
+            <NxAccordion.Header>
+              <span>Foo</span>
+            </NxAccordion.Header>
+          </NxAccordion>,
+          { baseElement: divContainer as HTMLElement }
+        );
+
+        const header = container.querySelector('summary') as HTMLElement;
```
rpokorny, line 148:
```suggestion
        const header = container.querySelector<HTMLElement>('summary');
```

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
+    rerender(
+      <NxAccordion>
+        <NxAccordion.Header>
+          <span className='foo'>Foo</span>
+        </NxAccordion.Header>
+        <span className='bar'>Bar</span>
+      </NxAccordion>
+    );
+
+    expect(container.querySelector('.nx-accordion__header .foo')).toBeInTheDocument();
```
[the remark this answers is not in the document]
rjp99: > Content in the header isn't part of this particular test

Removed here: 26910c530b16aefd690112d7d281c3efa3a0b3b0

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Rather than testing that it's in `.nx-accordion__content`, I'd test that it's in the accordion in general, and that it's not in the `<summary>` element

Like this? ad3b1d55d33707c24cabbaad0ac7cddeae469eab

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > This class is an impl detail

Removed here: 7907f12daefbcc73441d1fc90f29256059f91c55

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,204 @@
...
+  it('sets the provided className', function() {
+    const el = renderEl()!;
+    const customEl = renderEl({ className: 'foo' })!;
+
+    expect(customEl).toHaveClass('foo');
+
+    for (const cls of Array.from(el.classList)) {
+      expect(customEl).toHaveClass(cls);
+    }
   });
```
rjp99, line 36: @rpokorny Didn't know how to link this in your other comment, but is this what you meant by _'I'd replace this with our usual style of test for custom classnames that checks that all classes present without a custom classname are also still present with a custom classname.'_?

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Impl/visual detail

Removed here: e1c2b242c83f8888f450bbca8a272c9a7a237698

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
+          </NxAccordion.Header>
+        ) });
+
+      expect(container.querySelector('.nx-accordion__header')).toHaveAttribute('aria-controls', 'foo');
     });
   });
 
   describe('onToggle', function() {
-    let mountContainer: HTMLElement | null = null;
+    let divContainer: HTMLElement | null = null;
```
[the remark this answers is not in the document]
rjp99: > Creating a distinct container element for each render is actually what RTL does by default, so we don't need to do this manually anymore

Removed it here: ba71c22ad1b1609a2f2b213b8903e8a94169d4ca

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,277 +5,221 @@
...
+          { baseElement: divContainer as HTMLElement }
+        );
+
+        const header = container.querySelector('summary') as HTMLElement;
 
         expect(onToggle).not.toHaveBeenCalled();
 
-        act(function() {
-          header.getDOMNode().dispatchEvent(createClickEvent());
+        await act(async function() {
```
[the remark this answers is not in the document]
rjp99: > Using `act` is almost never necessary with RTL. Was it actually necessary here for some reason?

I just assumed it was needed, but worked without it: d651816435232a9bab4a73d5061c504318928c13

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > Like this? https://github.com/sonatype/sonatype-react-shared-components/commit/ad3b1d55d33707c24cabbaad0ac7cddeae469eab

Close but not quite.  `.nx-accordion` is also an implementation detail.  I'd just check that `.bar` is present in the rendered output at all (in addition to checking that `summary .bar` isn't)

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > Removed here: https://github.com/sonatype/sonatype-react-shared-components/commit/7907f12daefbcc73441d1fc90f29256059f91c55

The test name indicates that it checks whether the header renders a `<summary>`.  I don't see that happening.  In addition, the test name still talks about `.nx-accordion__summary-wrapper`.  That part should be removed.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > I think the custom classnames test is already added:

If I understand what you're linking to correctly, that's the classname test for `NxAccordion`.  You also need one for `NxAccordion.Header`.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > > Removed here: [7907f12](https://github.com/sonatype/sonatype-react-shared-components/commit/7907f12daefbcc73441d1fc90f29256059f91c55)
> 
> The test name indicates that it checks whether the header renders a `<summary>`. I don't see that happening. In addition, the test name still talks about `.nx-accordion__summary-wrapper`. That part should be removed.

Added here: 735be7af5df2ea43ca7885546fb8c2a3593882e0

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > Added here: https://github.com/sonatype/sonatype-react-shared-components/commit/6bab09f43a65a7fe9dee14d24d331842795c88d2

The test needs to add the `foo` class to the header, not the accordion.  And the comparison with the default classes needs to be done across the header elements, not the accordion elements.

rjp99: > > Added here: [6bab09f](https://github.com/sonatype/sonatype-react-shared-components/commit/6bab09f43a65a7fe9dee14d24d331842795c88d2)
> 
> The test needs to add the `foo` class to the header, not the accordion. And the comparison with the default classes needs to be done across the header elements, not the accordion elements.

Updated here: 254a5eecf167c8f59dd9b9a0d73a5aa91cc15a31

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,210 @@
...
+import NxButton from '../../NxButton/NxButton';
+import { render } from '@testing-library/react';
 
 describe('NxAccordion', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxAccordion, {}),
-      getMountedComponent = enzymeUtils.getMountedComponent<Props>(NxAccordion, {});
+  const quickRender = rtlRender(NxAccordion, {});
+  const renderEl = rtlRenderElement(NxAccordion, {});
 
   it('renders a <details> element with the provided props', function() {
```
rpokorny, line 20: This isn't currently testing that it's a `<details>` element like the test name claims.  Given that `<details>` and `<summary>` are kinda specific elements, and I don't really think the `group`/`button` roles do them justice, I'm inclined to say that the actual tag names are more than impl details here and should in fact be tested.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,210 @@
...
-    const component = getShallowComponent({ id: 'foo', open: true, title: 'test title' });
-
-    expect(component).toMatchSelector('details#foo');
-    expect(component).toHaveProp('open', true);
-    expect(component).toHaveProp('title', 'test title');
+    expect(quickRender({ id: 'foo' }).getByRole('group')).toHaveAttribute('id', 'foo');
```
rpokorny, line 21: We don't just want to test that some child is a `group` with the expected id, we want to test that the _top level element_ is a `group` with the expected id.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,210 @@
...
+    expect(customEl).toHaveClass('foo');
+
+    for (const cls of Array.from(el.classList)) {
+      expect(customEl).toHaveClass(cls);
+    }
   });
 
   it('sets an id if none is specified', function() {
-    expect(getShallowComponent().prop('id')).toBeTruthy();
+    expect(quickRender().getByRole('group')).toHaveAttribute('id');
```
rpokorny, line 38: Check the top-level element (`renderEl` would be the most convenient way)

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,210 @@
...
-        ),
-        childrenWrapper = component.find('.nx-accordion__content');
-
-    expect(childrenWrapper).toExist();
-    expect(childrenWrapper).toContainReact(<span>Bar</span>);
-    expect(childrenWrapper).not.toContainReact(<span>Foo</span>);
+  it('renders non-header children in content wrapper', function() {
+    const { rerender, container } = quickRender();
+
+    rerender(
```
rpokorny, line 50: Why `rerender` as opposed to setting the desired children in `quickRender`

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,210 @@
...
-              <NxAccordion.Header>
-                <span>Foo</span>
-              </NxAccordion.Header>
-            )
-          }),
-          header = component.find('summary');
-
-      expect(header.find(NxFontAwesomeIcon)).toHaveProp('icon', faChevronCircleUp);
+    it('renders the NxAccordion.Header as a <summary>', function() {
+      const { container } = quickRender({ children:
```
rpokorny, line 78: Check the top element specifically, using `renderEl`

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > We don't just want to test that some child is a `group` with the expected id, we want to test that the _top level element_ is a `group` with the expected id.

Like this? 2caf1192854b872e28771e153d04634e4cf24e1e

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > This isn't currently testing that it's a `<details>` element like the test name claims. Given that `<details>` and `<summary>` are kinda specific elements, and I don't really think the `group`/`button` roles do them justice, I'm inclined to say that the actual tag names are more than impl details here and should in fact be tested.

Updated here: 61fc75e0ae0e643c61c58c1a4d8f12224dc4e646

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > Like this? https://github.com/sonatype/sonatype-react-shared-components/commit/2caf1192854b872e28771e153d04634e4cf24e1e

That's one way.  I would have just used `renderEl` instead of `quickRender` though.  I mispoke slightly above as (related to some of my other comments) the `group` role isn't that important here and I'd rather we check the element tag name instead.  So in summary, I think this test should use `renderEl` to get the top-level element, and then test that that element is a `details` and that its id is `foo`.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Check the top-level element (`renderEl` would be the most convenient way)

Does this just test the top-level element? 94c174e048866249ef04e2a58a2b2e43860c8cc4

rpokorny: > Does this just test the top-level element? https://github.com/sonatype/sonatype-react-shared-components/commit/94c174e048866249ef04e2a58a2b2e43860c8cc4

Yes.  `renderEl(...)` is equivalent to `quickRender(...).container.firstElementChild`.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > using

Not sure if I did it right, but I tried something like this for the first child but it didn't work with renderEl
`expect(el.firstChild!).tagName.toBe('SUMMARY');`
b77c174cead31780dc234f12837628a25254fe80

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Why `rerender` as opposed to setting the desired children in `quickRender`?

Updated here: f681812b1f772f51d259d2abc4ec028c24d8a3be

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > > Like this? [2caf119](https://github.com/sonatype/sonatype-react-shared-components/commit/2caf1192854b872e28771e153d04634e4cf24e1e)
> 
> That's one way. I would have just used `renderEl` instead of `quickRender` though. I mispoke slightly above as (related to some of my other comments) the `group` role isn't that important here and I'd rather we check the element tag name instead. So in summary, I think this test should use `renderEl` to get the top-level element, and then test that that element is a `details` and that its id is `foo`.

Updated here: 24a336d3c82d2224871268ce37740dc8275c6565

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,212 @@
...
+    expect(el.getByRole('group')).toBe(el.container.firstChild);
+    expect(el.getByRole('group')).toHaveAttribute('id', 'foo');
 
-    expect(component).toMatchSelector('details#foo');
-    expect(component).toHaveProp('open', true);
-    expect(component).toHaveProp('title', 'test title');
+    expect(renderEl({ open: true })).toHaveAttribute('open');
+    expect(renderEl({ title: 'test title' })).toHaveAttribute('title', 'test title');
+
+    expect(quickRender({ open: true }).getByRole('group').tagName).toBe('DETAILS');
```
rpokorny: This is over-complicated.  This is what I've been trying to suggest:
```
  it('renders a <details> element with the provided props', function() {
    const el = renderEl({ id: 'foo', open: true })!;
    expect(el.tagName).toBe('DETAILS');
    expect(el).toHaveAttribute('id', 'foo');
    expect(el).toHaveAttribute('open', 'true');
  });

```

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,212 @@
...
-              <NxAccordion.Header>
-                <span>Foo</span>
-              </NxAccordion.Header>
-            )
-          }),
-          header = component.find('summary');
-
-      expect(header.find(NxFontAwesomeIcon)).toHaveProp('icon', faChevronCircleUp);
+    it('renders the NxAccordion.Header as a <summary>', function() {
+      const el = renderEl({ children:
```
rpokorny, line 77: Don't forget you redefined `renderEl` in the Header `describe` block.  So by specifying an `NxAccordion.Header` in the `children` here, you are rendering a header nested inside another header.  Same in the next test below

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,212 @@
...
+        (
+          <NxAccordion.Header id="headerId">
+            <span>Foo</span>
+          </NxAccordion.Header>
+        )})!;
+      expect(el.querySelector('#headerId')!.tagName).toBe('SUMMARY');
+      expect(el.querySelector('summary')).toBeInTheDocument();
```
rpokorny, line 84: This is overcomplicated/imprecise similarly to the "renders a <details> element with the provided props" test above.  I recommend fixing it similarly to what I recommended in https://github.com/sonatype/sonatype-react-shared-components/pull/865/files#r1048612547

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > This is over-complicated. This is what I've been trying to suggest:
> 
> ```
>   it('renders a <details> element with the provided props', function() {
>     const el = renderEl({ id: 'foo', open: true })!;
>     expect(el.tagName).toBe('DETAILS');
>     expect(el).toHaveAttribute('id', 'foo');
>     expect(el).toHaveAttribute('open', 'true');
>   });
> ```

Okay, that makes sense. d1ec88c521ffb315c0e189bb34198c0fee01e927

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Don't forget you redefined `renderEl` in the Header `describe` block. So by specifying an `NxAccordion.Header` in the `children` here, you are rendering a header nested inside another header. Same in the next test below

Updated the renderEl here: 0176eb8acf4d984deb4cb1a18659eb0341fe65b6

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > This is overcomplicated/imprecise similarly to the "renders a `<details>` element with the provided props" test above. I recommend fixing it similarly to what I recommended in https://github.com/sonatype/sonatype-react-shared-components/pull/865/files#r1048612547

Updated here: 0176eb8acf4d984deb4cb1a18659eb0341fe65b6

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,202 @@
...
   it('renders a <details> element with the provided props', function() {
-    const component = getShallowComponent({ id: 'foo', open: true, title: 'test title' });
-
-    expect(component).toMatchSelector('details#foo');
-    expect(component).toHaveProp('open', true);
-    expect(component).toHaveProp('title', 'test title');
+    const el = renderEl({ id: 'foo', open: true })!;
+    expect(el.tagName).toBe('DETAILS');
+    expect(el).toHaveAttribute('id', 'foo');
+    expect(el).toHaveAttribute('open');
```
rpokorny, line 24: Since the `open` attr is especially crucial, I think we should have a separate test testing that it is set correctly for all possible values of the corresponding prop (`true`, `false`, and `undefined`).

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,202 @@
...
+    const customEl = renderEl({ className: 'foo' })!;
+
+    expect(customEl).toHaveClass('foo');
+
+    for (const cls of Array.from(el.classList)) {
+      expect(customEl).toHaveClass(cls);
+    }
   });
 
   it('sets an id if none is specified', function() {
```
rpokorny, line 38: Rather than this test, I think what we really need is additional testing of the `aria-controls` on the header, for the case where an explicit id was not used.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,202 @@
...
-    expect(childrenWrapper).toContainReact(<span>Bar</span>);
-    expect(childrenWrapper).not.toContainReact(<span>Foo</span>);
+  it('renders non-header children in content wrapper', function() {
+    const { container } = quickRender({ children: (
+      <NxAccordion.Header>
+        <span>Foo</span>
+      </NxAccordion.Header>
+
+    ),
+    className: 'bar' });
```
rpokorny, line 55: Something got messed up here.  The `bar` class needs to be applied to non-header children, not to the accordion itself

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
```diff
@@ -5,313 +5,202 @@
...
-          autoId = autoIdAccordion.prop('id');
-
-      expect(explicitIdAccordion.find('summary')).toHaveProp('aria-controls', 'foo');
-      expect(autoIdAccordion.find('summary')).toHaveProp('aria-controls', autoId);
+      const { container } = quickRender({ id: 'foo',
+        children: (
+          <NxAccordion.Header>
+            <span>Foo</span>
+          </NxAccordion.Header>
+        ) });
```
rpokorny, line 87:
```suggestion
      const { container } = quickRender({ 
        id: 'foo',
        children: (
          <NxAccordion.Header>
            <span>Foo</span>
          </NxAccordion.Header>
        ) 
      });
```

rpokorny:
It's not clear in the ticket, but I think this JIRA ticket was intended to convert the `NxStatefulAccordion` tests too.  Since this PR is already so far along let's create a subtask for that and do it separately.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Since the `open` attr is especially crucial, I think we should have a separate test testing that it is set correctly for all possible values of the corresponding prop (`true`, `false`, and `undefined`).

Added the test: e2508b1fddf6f6ae84c2ee5bb5ae1eecf2a4cd60

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > 

Added it here: 7bd9b2c0b3e19a89c13646166fef04205e303e4a

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > Something got messed up here. The `bar` class needs to be applied to non-header children, not to the accordion itself

Fixed here: 323786008e9a71801d1165fd650fa9b0c4cc49b5

rpokorny: > Fixed here: https://github.com/sonatype/sonatype-react-shared-components/commit/323786008e9a71801d1165fd650fa9b0c4cc49b5

Why'd you switch it away from `quickRender`?

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > Added it here: https://github.com/sonatype/sonatype-react-shared-components/commit/7bd9b2c0b3e19a89c13646166fef04205e303e4a

The ids are randomly/dynamically generated.  Hardcoding the value it might be is definitely not going to work.  The test should just test that the value of `aria-controls` and the `id` of the accordion match and are truthy, not that they are any specific value.

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > > Fixed here: [3237860](https://github.com/sonatype/sonatype-react-shared-components/commit/323786008e9a71801d1165fd650fa9b0c4cc49b5)
> 
> Why'd you switch it away from `quickRender`?

Because when I quickRender and then specify an Accordion within it would that be rendering an accordion within an accordion similar to what I was doing with the headers before?

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rjp99: > > Added it here: [7bd9b2c](https://github.com/sonatype/sonatype-react-shared-components/commit/7bd9b2c0b3e19a89c13646166fef04205e303e4a)
> 
> The ids are randomly/dynamically generated. Hardcoding the value it might be is definitely not going to work. The test should just test that the value of `aria-controls` and the `id` of the accordion match and are truthy, not that they are any specific value.

Updated: 91d01901daf3ebdebbac2d98fb6a2d7c5cf78b5f

lib/src/components/NxAccordion/__tests__/NxAccordion.test.tsx
rpokorny: > Because when I quickRender and then specify an Accordion within it would that be rendering an accordion within an accordion similar to what I was doing with the headers before?

So don't add an extra layer of accordions.  Do it like you had at the beginning of this thread, but remove the `className: 'bar'` from the `quickRender` properties and add an element with `className="bar"` as a sibling to the `NxAccordion.Header`.

Oh, you'll need a fragment to wrap the header and sibling.  Is that what was tripping you up?

rjp99: > > Because when I quickRender and then specify an Accordion within it would that be rendering an accordion within an accordion similar to what I was doing with the headers before?
> 
> So don't add an extra layer of accordions. Do it like you had at the beginning of this thread, but remove the `className: 'bar'` from the `quickRender` properties and add an element with `className="bar"` as a sibling to the `NxAccordion.Header`.
> 
> Oh, you'll need a fragment to wrap the header and sibling. Is that what was tripping you up?

Yeah when I do this: 
`const { container } = quickRender(
      <>
        <NxAccordion.Header>
          <span>Foo</span>
        </NxAccordion.Header>
        <span className='bar'>Bar</span>
      </>
    );`

I get that it has a value of null

rpokorny: > Yeah when I do this:
const { container } = quickRender( <> <NxAccordion.Header> <span>Foo</span> </NxAccordion.Header> <span className='bar'>Bar</span> </> );
>
> I get that it has a value of null

Don't forget that the parameter to `quickRender` should be the entire props object, not just the children.  I have pushed a commit that addresses this.

rpokorny (approved):

PR #867, opened 2022-12-02
NxAlert RTL Tests - RSC-1223

rpokorny:
https://issues.sonatype.org/browse/RSC-1223

### DIFF
--- a/lib/src/components/NxAlert/__tests__/NxAlert.test.tsx
+++ b/lib/src/components/NxAlert/__tests__/NxAlert.test.tsx
@@ -111,7 +111,7 @@ describe('NxAlert', function() {
         expect(view.container.firstElementChild).toBe(view.getByRole('alert'));
       });
 
-      it('renders an img with an accessible nameof "Error"', function() {
+      it('renders an img with an accessible name of "Error"', function() {
         expect(quickRender().getByRole('img')).toHaveAccessibleName('Error');
       });
     });
@@ -125,7 +125,7 @@ describe('NxAlert', function() {
         expect(renderEl()).not.toHaveAttribute('role');
       });
 
-      it('renders an img with an accessible nameof "Info"', function() {
+      it('renders an img with an accessible name of "Info"', function() {
         expect(quickRender().getByRole('img')).toHaveAccessibleName('Info');
       });
     });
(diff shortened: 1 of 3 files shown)


lib/src/components/NxAlert/__tests__/NxAlert.test.tsx
```diff
@@ -4,263 +4,158 @@
...
-        component = getNxAlert({ onClose });
+  const minimalProps = {
+        children: 'A message to show in an alert',
+        icon: faBiohazard
+      },
+      quickRender = rtlRender(NxAlert, minimalProps);
 
-    expect(component).toContainMatchingElement(NxCloseButton);
-    expect(onClose).not.toHaveBeenCalled();
+  baseTests(NxAlert, minimalProps);
```
EmilyBermudez88, line 80: does the generic NxAlert need to have tests for role assignment as well, or is it considered sufficient when its tested in all the variations of NxAlert?

lib/src/components/NxAlert/__tests__/NxAlert.test.tsx
```diff
@@ -4,263 +4,158 @@
...
+        expect(viewWithRole.container.firstElementChild).toHaveAttribute('role', 'foo');
       });
 
-      it('passes any other props to the NxAlert', function() {
-        const onClose = jest.fn(),
-            component = getNxInfoAlert({ id: 'foo', title: 'baz', onClose });
-        expect(component).toHaveProp('id', 'foo');
-        expect(component).toHaveProp('title', 'baz');
-        expect(component).toHaveProp('onClose', onClose);
+      it('renders an img with an accessible nameof "Warning"', function() {
```
EmilyBermudez88, line 141:
```suggestion
      it('renders an img with an accessible name of "Warning"', function() {
```

EmilyBermudez88: (in all `it('renders an img with an accessible nameof...)`

lib/src/components/NxAlert/__tests__/NxAlert.test.tsx
rpokorny: > does the generic NxAlert need to have tests for role assignment as well, or is it considered sufficient when its tested in all the variations of NxAlert?

This made me realize there was room for several improvements around role testing.  All of the different alert variations _can_ receive a role, and the ones that don't provide a role by default should probably have tests proving that.  Addressed in 72da7ca06

lib/src/components/NxAlert/__tests__/NxAlert.test.tsx
rpokorny: 3c5b1c0a2

EmilyBermudez88 (approved):

PR #868, opened 2022-12-06
NxTooltip RTL Tests - RSC-1227

rpokorny:
https://issues.sonatype.org/browse/RSC-1227

### DIFF
--- a/lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
+++ b/lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
@@ -4,113 +4,278 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
+import { within, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
-import { Tooltip } from '@material-ui/core';
+import { rtlRender } from '../../../__testutils__/rtlUtils';
 
-import { getMountedComponent, getShallowComponent } from '../../../__testutils__/enzymeUtils';
-import NxTooltip, { TooltipContext, Props } from '../NxTooltip';
-import { act } from 'react-dom/test-utils';
+import NxTooltip from '../NxTooltip';
 
 describe('NxTooltip', function() {
   const minimalProps = {
-        children: <div id="foo" />,
+        children: <div data-testid="foo" />,
         title: 'tip'
       },
-      getMounted = getMountedComponent(NxTooltip, minimalProps),
-      getShallow = getShallowComponent(NxTooltip, minimalProps),
-      getTooltip = (extraProps?: Partial<Props>) => getMounted(extraProps).children();
+      quickRender = rtlRender(NxTooltip, minimalProps);
 
   beforeEach(function() {
     jest.useFakeTimers();
   });
 
-  it('creates a TooltipContext.Provider that provides a value of true', function() {
-    expect(getShallow()).toMatchSelector(TooltipContext.Provider);
-    expect(getShallow()).toHaveProp('value', true);
+  // an element containing a11y description text to be added to the child in some tests
+  let descriptionEl: HTMLElement;
+
+  beforeEach(function() {
+    descriptionEl = document.createElement('div');
+    descriptionEl.id = 'descriptionId';
+    descriptionEl.textContent = 'Description Text';
+
+    document.body.append(descriptionEl);
   });
 
-  it('creates a MUI tooltip as the context providers child', function() {
-    expect(getTooltip()).toMatchSelector(Tooltip);
+  afterEach(function() {
+    descriptionEl.remove();
   });
 
-  it('passes through all props except placement to MUI Tooltip', function() {
-    const onOpen = () => {},
-        onClose = () => {},
-        component = getTooltip({ open: false, onOpen, onClose, placement: 'left' });
+  it('adds specified classes to an element within the tooltip and not to the child', async function() {
+    const user = userEvent.setup(),
+        view = quickRender({ className: 'foo' }),
+        defaultView = quickRender(),
+        child = view.getByTestId('foo'),
+        defaultChild = defaultView.getByTestId('foo');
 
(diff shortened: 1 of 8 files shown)


lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -4,113 +4,278 @@
...
 
-    expect(component.children()).toHaveProp('title', title);
+    expect(tooltip).toBeInTheDocument();
+    expect(tooltip).toHaveTextContent('tip');
   });
 
-  it('passes "nx-tooltip" as the `tooltip` property on the `classes` prop of the Tooltip', function() {
-    expect(getTooltip().prop('classes').tooltip).toBe('nx-tooltip');
-  });
+  it('calls onOpen when the tooltip renders by hover but not due to the open prop', async function() {
```
EmilyBermudez88, line 73: When I tested the second quickRender function with different props (ie. `quickRender({ onOpen, className: 'foo', placement: 'bottom'})` ), and used `await user.hover(child)` for the second time, the test still came back as having `onOpen` being called one time. Perhaps the `open` prop set to `true` should be tested first to show `onOpen` wasn't called, and then use `quickRender({onOpen})` & `await user.hover(child)` to assert `onOpen` has indeed been called.

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -4,113 +4,278 @@
...
+      expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+
+      await user.hover(child);
+
+      await expect(screen.findByRole('tooltip')).rejects.toBeTruthy();
+    });
+
+    it('does not affect the accessible name or description of the child', function() {
+      const children = <button aria-describedby="descriptionId">Foo</button>,
+          view = quickRender({ children }),
```
EmilyBermudez88, line 122:
```suggestion
          view = quickRender({ children, title: '' }),
```

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -4,113 +4,278 @@
...
+          await user.hover(button);
+          await waitFor(() => expect(button).toHaveAccessibleDescription('tiptip'));
+          expect(button).not.toHaveAttribute('title');
+        }
+    );
+
+    it('does not override the accessible description if it is already set',
+        async function() {
+          const user = userEvent.setup(),
+              childrenWithOwnDescription = <button aria-describedby="descriptionId">Foo</button>,
```
EmilyBermudez88, line 271: need the JSX title prop added here

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
rpokorny: ea70cf4a2

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
rpokorny: > If it's under the 'when title is JSX' describe, I think you need the JSX title prop added here

ea70cf4a2

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
EmilyBermudez88: >When I change the code as follows, it passes

Thanks for writing out a solution. I called `hover` and `waitFor`, but completely forgot to assign a variable to the second `quickRender` (just called `quickRender({ onOpen, placement: 'bottom'}), which of course caused things to fail. Thanks!

EmilyBermudez88 (approved):

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -4,113 +4,317 @@
...
+              viewWithOwnDescription = quickRender({ children: childrenWithOwnDescription }),
+              buttonWithOwnDescription = viewWithOwnDescription.getByRole('button');
+
+          await user.hover(buttonWithOwnDescription);
+          expect(buttonWithOwnDescription).toHaveAccessibleDescription('Description Text');
+          expect(buttonWithOwnDescription).not.toHaveAttribute('title');
+
+          // Ensure that nothing changes after waiting for tooltip to initialize
+          await expect(waitFor(() => expect(buttonWithOwnDescription).not.toHaveAccessibleDescription('Description Text')))
```
EmilyBermudez88, line 215: lintr error here with too many characters on line 215

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -117,13 +119,19 @@ describe('NxTooltip', function() {
...
+    it('does not affect the accessible name or description of the child', async function() {
       const children = <button aria-describedby="descriptionId">Foo</button>,
           view = quickRender({ title: '', children }),
           button = view.getByRole('button');
 
       expect(button).toHaveAccessibleName('Foo');
       expect(button).toHaveAccessibleDescription('Description Text');
+
+      // Ensure that nothing changes after waiting for tooltip to initialize
+      await expect(waitFor(() => expect(button).not.toHaveAccessibleName('Foo')))
```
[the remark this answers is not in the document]
rpokorny: > For clarification, why do we need the test here? If the title is an empty string, wouldn't the tooltip fail to render on the page?

The issue is that the tooltips initialize asynchronously, so any test that's trying to test that they _did not_ change something really needs to ensure not just that nothing changed immediately, but that nothing changed after waiting a bit for the tooltip to initialize as well.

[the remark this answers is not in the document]
EmilyBermudez88: > The issue is that the tooltips initialize asynchronously, so any test that's trying to test that they _did not_ change something really needs to ensure not just that nothing changed immediately, but that nothing changed after waiting a bit for the tooltip to initialize as well.

Makes sense for the situation here, thanks!

EmilyBermudez88 (approved):
Once the eslint errors are fixed, looks good!

PR #870, opened 2022-12-12
SPIKE: Skeleton Loader - RSC-1196

rpokorny:
https://issues.sonatype.org/browse/RSC-1196

### DIFF
--- a/gallery/src/components/NxSkeletonLoader/NxSkeletonLoaderPage.tsx
+++ b/gallery/src/components/NxSkeletonLoader/NxSkeletonLoaderPage.tsx
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import { NxSkeletonLoader } from '@sonatype/react-shared-components';
+import React from 'react';
+import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
+
+export default function NxSkeletonLoaderPage() {
+  return (
+    <>
+      <GalleryDescriptionTile>
+        <NxSkeletonLoader />
+      </GalleryDescriptionTile>
+    </>
+  );
+}
--- a/gallery/src/components/NxSkeletonLoader/NxSkeletonLoaderPage.tsx
+++ b/gallery/src/components/NxSkeletonLoader/NxSkeletonLoaderPage.tsx
@@ -6,17 +6,25 @@
  */
 import { NxH2, NxP, NxSkeletonLoader, NxSpan, NxTable, NxThreatCounter, NxThreatIndicator, NxTile }
   from '@sonatype/react-shared-components';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
 
 export default function NxSkeletonLoaderPage() {
+  const [showSecondTile, setShowSecondTile] = useState(false);
+
+  useEffect(function() {
+    setTimeout(() => { setShowSecondTile(true); }, 800);
+  }, []);
+
   return (
     <>
       <GalleryDescriptionTile>
+        {/*
         <NxSkeletonLoader.Block style={{ height: '100px', width: '100px' }}/>
         <NxSkeletonLoader.Block style={{ marginLeft: '300px', height: '100px', width: '100px' }}/>
         <NxSkeletonLoader.Block style={{ height: '100px', width: '400px' }}/>
         <NxSkeletonLoader.Block style={{ height: '100px', width: '3000px' }}/>
+        */}
         <div className="gallery-example--checkered-background">
           <div className="gallery-example-live">
             <NxSkeletonLoader>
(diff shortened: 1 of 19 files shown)


gallery/src/components/NxSkeletonLoader/NxSkeletonLoaderPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+export default function NxSkeletonLoaderPage() {
+  const [showSecondTile, setShowSecondTile] = useState(false);
+
+  useEffect(function() {
+    setTimeout(() => { setShowSecondTile(true); }, 800);
+  }, []);
+
+  return (
+    <>
+      <GalleryDescriptionTile>
```
rpokorny, line 22: Since this is a spike, the gallery page isn't properly built and is just a dumping ground for experiments and examples

gallery/src/components/NxSkeletonLoader/NxSkeletonLoaderPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { NxFontAwesomeIcon, NxH2, NxP, NxSkeletonLoader, NxSpan, NxTable, NxThreatCounter, NxThreatIndicator, NxTile }
+  from '@sonatype/react-shared-components';
+import React, { useEffect, useState } from 'react';
+import { GalleryDescriptionTile } from '../../gallery-components/GalleryTiles';
+
+export default function NxSkeletonLoaderPage() {
+  const [showSecondTile, setShowSecondTile] = useState(false);
+
+  useEffect(function() {
+    setTimeout(() => { setShowSecondTile(true); }, 800);
```
rpokorny, line 17: The purpose of the second tile is to prove that the skeleton animations are synchronized even for elements that render at different times

lib/src/components/NxSkeletonLoader/NxSkeletonLoader.tsx
```diff
@@ -0,0 +1,110 @@
...
+      activeSkeletonLoaders.delete(instance);
+      if (!activeSkeletonLoaders.size) {
+        animationStartTime = null;
+        window.removeEventListener('animationstart', handleAnimationStart);
+      }
+    };
+  }, []);
+
+  return (
+    <div className="nx-skeleton-loader" aria-busy="true">
```
rpokorny, line 75: In a real impl, `NxSkeletonLoader` would function more like `NxLoadWrapper`.  It would receive a `loading` boolean prop and would only enable its loading state if that prop is true.  i.e., `aria-busy` here would be set conditionally, as would `aria-hidden` below an the context provider value.  The `.nx-skeleton-loader__status` would also be conditionally hidden from the screenreader.

rpokorny: > In a real impl, NxSkeletonLoader would function more like NxLoadWrapper. It would receive a loading boolean prop and would only enable its loading state if that prop is true. i.e., aria-busy here would be set conditionally, as would aria-hidden below an the context provider value. The .nx-skeleton-loader__status would also be conditionally hidden from the screenreader. For the spike I left that out and just have it always active

Also like `NxLoadWrapper`, this component will need to handle load errors, which it should do by rendering an error alert instead of content.

lib/src/components/NxSkeletonLoader/NxSkeletonLoader.tsx
```diff
@@ -0,0 +1,111 @@
...
+          }
+        }
+      }
+    }
+
+    if (!activeSkeletonLoaders.size) {
+      window.addEventListener('animationstart', handleAnimationStart);
+    }
+
+    activeSkeletonLoaders.add(instance);
```
EmilyBermudez88, line 64: I'm having a bit of trouble trying to understand the use of `instance`- why are we passing an empty object to the activeSkeletonLoaders Set instead of the target or animation?

lib/src/components/NxTable/NxTableBody.tsx
```diff
@@ -62,11 +64,24 @@ const NxTableBody = function NxTableBody(props: NxTableBodyProps) {
     </NxTableRow>
   );
 
+  function makeSkeletonChildren() {
+    const firstChild = React.Children.toArray(children)[0] as ReactElement;
+
+    return (
+      <>
```
EmilyBermudez88, line 71: Are there any situations where a table would render less than 3 rows? In Kaleido I take it to imply that there won't be more than 3, but what are there possibilities of the table rendering 2 rows, and needing a 2row- skeleton loader? It may be how I'm understanding the Kaleido doc

lib/src/components/NxSkeletonLoader/NxSkeletonLoader.tsx
rpokorny: > I'm having a bit of trouble trying to understand the use of instance- why are we passing an empty object to the activeSkeletonLoaders Set instead of the target or animation?

The empty object is a stand-in for the `NxSkeletonLoader` component instance. Think of `activeSkeletonLoaders` as a set of currently-mounted `NxSkeletonLoader` react component instances.  Except, there is no "react component instance" object that we have access to that we could store in the set.  So instead, each one creates an object that we do have access to and stores it in the set as a stand-in.  The only data carried by the objects is their identity - none of the objects are considered to be equal to one another by the set.  

Thinking about it more,  I suppose the set could be replaced by a simple counter that gets incremented in each `useEffect` and decremented in each `useEffect` unregistration callback.  All we need to know is whether there are currently zero or non-zero `NxSkeletonLoaders` mounted.  I'll change it to that.

lib/src/components/NxTable/NxTableBody.tsx
rpokorny: > Are there any situations where a table would render less than 3 rows? In Kaleido I take it to imply that there won't be more than 3, but what are there possibilities of the table rendering 2 rows, and needing a 2row- skeleton loader? It may be how I'm interpreting the Kaleido doc

The assumption is that a table on which skeleton-loading is being used would ultimately have a dynamic number of rows that wouldn't be known until the load completes.  So three was the chosen placeholder number.

lib/src/components/SimpleComponents.tsx
```diff
@@ -19,10 +21,10 @@ export const NxThreatNumber = withClass('span', 'nx-threat-number');
 export const NxPageMain = withClass('main', 'nx-page-main');
 export const NxPageSidebar = withClass('aside', 'nx-page-sidebar');
 
-export const NxH1 = withClass('h1', 'nx-h1');
-export const NxH2 = withClass('h2', 'nx-h2');
-export const NxH3 = withClass('h3', 'nx-h3');
-export const NxH4 = withClass('h4', 'nx-h4');
+export const NxH1 = withClass('h1', 'nx-h1', undefined, <NxSkeletonLoader.Text />);
```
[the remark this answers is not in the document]
rpokorny: > are there cases in product that H1s need a skeleton loader?

it might as well have one since all the others do.  Going out of our way not to give it one seems like it adds needless risk in case one was ever needed for some reason.

lib/src/components/NxSkeletonLoader/NxSkeletonLoader.tsx
EmilyBermudez88: >The empty object is a stand-in for the NxSkeletonLoader component instance. Think of activeSkeletonLoaders as a set of currently-mounted NxSkeletonLoader react component instances. Except, there is no "react component instance" object that we have access to that we could store in the set. So instead, each one creates an object that we do have access to and stores it in the set as a stand-in. The only data carried by the objects is their identity - none of the objects are considered to be equal to one another by the set.

Thanks for the clarification!

lib/src/components/NxTable/NxTableBody.tsx
EmilyBermudez88: > > Are there any situations where a table would render less than 3 rows? In Kaleido I take it to imply that there won't be more than 3, but what are there possibilities of the table rendering 2 rows, and needing a 2row- skeleton loader? It may be how I'm interpreting the Kaleido doc
> 
> The assumption is that a table on which skeleton-loading is being used would ultimately have a dynamic number of rows that wouldn't be known until the load completes. So three was the chosen placeholder number.

Makes sense, thanks

EmilyBermudez88:
Spent time looking through this again and don't think I have any other questions or concerns atm. Looking good and really interesting to read through & see how it was implemented.

rpokorny:
> Is it the regular approval process for a spike ticket?

Yeah the approval is the same but what happens after is different.  Mainly, the PR will be closed instead of merged, and additional JIRA tickets will be created to move towards a full implementation.

EmilyBermudez88 (approved):

JasonDSwearingen (approved):

rpokorny:
Closing since it's a spike.  Will be carried through in other tickets

PR #876, opened 2022-12-16
Combobox closes dropdown on selection- RSC-1253

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1253

https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/m19LQYz

Current behaviour follows the recommendations on sketch, but not sure it's the cleanest implementation at this point. For `openState`, I decided not to make it a boolean, because if the openState is set to false on page load (which would make sense, as the dropdown would not be open on page load), all RTL tests querying for dropdown items fail (since the dropdown won't render on page load and would require the `onChange` prop to be called).


rpokorny:
I'm wondering if it would be clearer if the state value was called `hiddenBySelection` or something.  Then it could just be a boolean, initialized to `false`, that gets set to true when you select and set back to false on click or focus.  One major problem with the clarity of the current setup is that `openState` is just one factor in whether the dropdown is actually visibly open - there's are circumstances where `openState === 'open'` is true but the dropdown is not currently visible, and that could be confusing.

Additionally, the thing we're trying to track truly has just two values: either we're hiding the dropdown due to selection or we aren't.  So a three-value state field has some redundancy.

EmilyBermudez88:
> Additionally, the thing we're trying to track truly has just two values: either we're hiding the dropdown due to selection or we aren't. So a three-value state field has some redundancy.

Implemented here: [debcbf9](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/debcbf9f92788884dc90c49c99af31fe911d793e)

gallery/visualtests/nxCombobox.js
```diff
@@ -232,17 +232,44 @@ describe('NxCombobox', function() {
 
   it('passes a11y checks', a11yTest());
 
-  it('shows the dropdown when focused and hides the dropdown when not focused', async function() {
-    const inputSelector = `${basicExampleSelector} .nx-combobox__input input`,
-        dropdownMenuSelector = `${basicExampleSelector} .nx-dropdown-menu`,
-        [input, dropdownMenu] = await waitAndGetElements(inputSelector, dropdownMenuSelector);
+  describe('dropdown display behavior', function() {
+    it('shows the dropdown when initially focused', async function() {
```
rpokorny, line 236: This is only true in the case where the list of options is immediately available.  In the case where the options populate once you start typing, the dropdown _should not_ be visible until you start typing.  I wonder if this entire `describe` block should be duplicated, once for the "Basic Example" and once for the "Backend Query Example", with the first test in the latter changed to check that the dropdown opens once the user has typed.

gallery/visualtests/nxCombobox.js
```diff
@@ -232,17 +232,44 @@ describe('NxCombobox', function() {
...
+
+    it('closes dropdown when a selection is made', async function() {
+      const inputSelector = `${basicExampleSelector} .nx-combobox__input input`,
+          dropdownMenuSelector = `${basicExampleSelector} .nx-dropdown-menu`,
+          buttonSelector = `${dropdownMenuSelector} .nx-dropdown-button:first-child`,
+          [input, dropdownMenu, firstOptBtn] = await waitAndGetElements(
+              inputSelector, dropdownMenuSelector, buttonSelector);
+
+      await input.focus();
+      await firstOptBtn.click();
```
rpokorny, line 254: Also test that it closes on keyboard selection

gallery/visualtests/nxCombobox.js
```diff
@@ -232,17 +232,44 @@ describe('NxCombobox', function() {
...
+      expect(await isVisible(dropdownMenu)).toBe(false);
+    });
+
+    it('reopens dropdown after selection when input is refocused', async function() {
```
rpokorny, line 258: Also test that it reopens on click

lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -51,14 +51,14 @@ function NxComboboxRender<T extends string | number | DataItem<string | number,
       } = props,
       previousValue = usePrevious(value),
       newAttrs = omit(['trimmedValue'], attrs),
-
+      [hiddenBySelection, setHiddenBySelection] = useState(false),
```
rpokorny, line 54: I think a comment should be added explaining what this state variable is for

lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -99,6 +99,9 @@ function NxComboboxRender<T extends string | number | DataItem<string | number,
         doSearch(value);
       }
     }
+    if (matches.length || value) {
```
rpokorny, line 102: Is this condition necessary?

lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -112,6 +115,7 @@ function NxComboboxRender<T extends string | number | DataItem<string | number,
         onChange(elToFocusMatch.displayName, elToFocusMatch);
       }
       setFocusableBtnIndex(null);
+      setHiddenBySelection(true);
```
rpokorny, line 118: Should this line be here or in the `if (autoComplete...` condition above? My understanding is that we want to set `hiddenBySelection` if and only if we are firing `onChange` with its optional second argument.  In fact maybe a helper function should be written which combines those two actions, and all existing calls to `onChange(foo, bar)` replaced with that helper function.

lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -122,7 +126,7 @@ function NxComboboxRender<T extends string | number | DataItem<string | number,
   function handleOnChange(newVal: string) {
     setFocusableBtnIndex(null);
     onChange(newVal);
-
+    setHiddenBySelection(false);
```
rpokorny, line 129: Is there any case that this would execute and `hiddenBySelection` not already be false? This is essentially the `onChange` handler of the `<input>`, which I think could only fire if the `<input>` has focus, and it getting focus should already set `hiddenBySelection` to false, right?

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,12 +83,25 @@ describe('NxCombobox', function() {
         expect(inputElement).toHaveValue('Foo');
       });
 
-  it('sets aria-expanded on the input to true when focused',
+  it('sets aria-expanded on the input according to if the input is focused and when an option is selected',
```
rpokorny, line 86: This test name isn't very clear.  Since there are multiple ways for the relevant state change to happen, it might make sense to break it into several tests, but as an example one of those tests could be `it('sets aria-expanded to false when a dropdown item is selected by click'`.

lib/src/components/NxCombobox/NxCombobox.tsx
EmilyBermudez88: > Is there any case that this would execute and `hiddenBySelection` not already be false? This is essentially the `onChange` handler of the `<input>`, which I think could only fire if the `<input>` has focus, and it getting focus should already set `hiddenBySelection` to false, right?

This was kept for the scenario that a selection was made (closing the dropdown), and then the user editing the `<input>` value without shifting focus outside of `NxCombobox`

lib/src/components/NxCombobox/NxCombobox.tsx
EmilyBermudez88: > Is this condition necessary?

Thanks for pointing that out- no, it was for an old iteration and should have been removed. Removed here: [b9404ed](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/b9404edb82109eba303d62525fad164dc551dd4e)

lib/src/components/NxCombobox/NxCombobox.tsx
EmilyBermudez88: > I think a comment should be added explaining what this state variable is for

Good idea, added here: [e616eca](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/e616eca270fac3c4c24f0649b37d178c0f806f04)

lib/src/components/NxCombobox/NxCombobox.tsx
EmilyBermudez88: > Should this line be here or in the `if (autoComplete...` condition above? My understanding is that we want to set `hiddenBySelection` if and only if we are firing `onChange` with its optional second argument. In fact maybe a helper function should be written which combines those two actions, and all existing calls to `onChange(foo, bar)` replaced with that helper function.

Yep, it should have been, thanks for catching that. I did move it and created a helper function here: [f4ac895](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/f4ac89526e30ad86244aa8f94f1b1b257b670e13). I added the third parameter and a condition inside of `handleSelection` because `onChange(foo, bar)` is also called in `adjustBtnFocus` where we don't want to handle `hiddenBySelection`.

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,12 +83,25 @@ describe('NxCombobox', function() {
...
+        // initial focus of the input
+        inputElement.focus();
+        expect(inputElement).toHaveAttribute('aria-expanded', 'true');
+
+        // selection is made
+        await user.click(optionBtn);
         expect(inputElement).toHaveAttribute('aria-expanded', 'false');
+
+        // input is refocused after selection
+        inputElement.blur();
```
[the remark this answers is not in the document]
EmilyBermudez88: > Also test selection by keyboard and re-click by mouse.

Added in extra testing here: [2f37ff0](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/2f37ff0492da188d66841f51fcbe31a027c00630)

EmilyBermudez88:
I'm currently having this error come up when I click on an item in the dropdown, which I believe is triggered by the tooltip. This only occurs when I click on a dropdown item when `itemTooltip` is specified, and NOT when I select an item via `arrow keys` + `Enter`.

gallery/visualtests/nxCombobox.js
EmilyBermudez88: > Also test that it reopens on click

Added test here: [97cf579](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/97cf5797e53b63e273c24aeebec61743431fff6e)

gallery/visualtests/nxCombobox.js
EmilyBermudez88: > Also test that it closes on keyboard selection

Added test here: [97cf579](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/97cf5797e53b63e273c24aeebec61743431fff6e)

gallery/visualtests/nxCombobox.js
```diff
@@ -271,5 +296,24 @@ describe('NxCombobox', function() {
...
+          [input, dropdownMenu, firstOptBtn, basicExample] = await waitAndGetElements(
+              inputSelector, dropdownMenuSelector, buttonSelector, example);
+
+      await input.click();
+      await firstOptBtn.click();
+      expect(await isVisible(dropdownMenu)).toBe(false);
+
+      await basicExample.click();
+      expect(await isFocused(input)).toBe(false);
+      await input.click();
```
rpokorny, line 314: Crucially, we want to test that click reopens it _even if it never lost focus in the first place_.  ie. what you have here but without clicking the `basicExample` in between

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -391,6 +441,46 @@ describe('NxCombobox', function() {
...
+      await user.keyboard('[Enter]');
+      expect(inputElement).toHaveAttribute('aria-expanded', 'false');
+    });
+
+    it('sets aria-expanded to true when the input receives focus after dropdown selection and blur', async function() {
+      const user = userEvent.setup(),
+          { getByRole } = quickRender({ matches: [{ id: '1', displayName: 'Foo' }] }),
+          inputElement = getByRole('combobox');
+
+      await user.keyboard('[Tab]');
```
rpokorny, line 466: See https://testing-library.com/docs/user-event/convenience#tab

rpokorny:
> This only occurs when I click on a dropdown item when itemTooltip is specified, and NOT when I select an item via arrow keys + Enter. Now that showDropdown is also being affected by hiddenBySelection, the class nx-combobox--showable-dropdown is removed when something is selected, therefore setting the dropdown menu to display:none. Would you recommend I simply edit the CSS to have nx-combobox__menu set to visibility: hidden, and nx-combobox__dropdown-showable to visibility: visible, or would there be a better way to handle this? In previous iterations I had the matches conditionally render based on showDropdown, but this strays away from the original functionality, so I assume that will want to be left as is.

I'm not thinking of a reason why the dropdown would need to remain in the DOM when it is not showable, so I would try changing it to be rendered conditionally at that level rather than hidden in CSS.  However, just because I can't remember a reason that it is the way it is doesn't mean there isn't one, so test it carefully after making that change, and check with @cindyko226 to see if she remembers.

If removing it from the DOM doesn't work, yes you could try `visibility: hidden` but I'm not certain whether that will cause the element to block clicks on things behind it.  I don't think it will, but if it does, that can probably be addressed with `pointer-events: none`.

lib/src/components/NxCombobox/NxCombobox.tsx
rpokorny: > Yep, it should have been, thanks for catching that. I did move it and created a helper function here: [f4ac895](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/f4ac89526e30ad86244aa8f94f1b1b257b670e13). I added the third parameter and a condition inside of handleSelection because onChange(foo, bar) is also called in adjustBtnFocus where we don't want to handle hiddenBySelection.

Oh interesting, I hadn't thought about the arrow down case.  I think a cleaner way to do that would just be to use `onChange` directly in `adjustBtnFocus` and to avoid the conditional in `handleSelection`.  I also think `handleSelection` deserves a brief comment about what it does and when it should be used.

EmilyBermudez88: > Oh interesting, I hadn't thought about the arrow down case. I think a cleaner way to do that would just be to use `onChange` directly in `adjustBtnFocus` and to avoid the conditional in `handleSelection`. I also think `handleSelection` deserves a brief comment about what it does and when it should be used.


I noticed a discrepancy in examples of  when showDropdown is true vs. when the dropdown should actually be visible. If the example is not set to auto complete, `showDropdown` remains `true` when you click outside of the dropdown without selecting anything (since no selection was made to change `hiddenBySelection(true)` ), but  is `false` when the `autoComplete` prop is true,  because the function `setHiddenBySelection` is contained inside the condition `if (autoComplete && focusableBtnindex !-- null)` . 

If this is to be consistent across all examples, we would need to have `setHiddenBySelection(true)` outside of the autoComplete condition, while needing to leave `onChange` inside based on your previous comment in this thread. IF that's the case, is it necessary to have an additional helper function that can only be called inside `handleDropdownBtnClick`?   ... this may also lead to needing to rename the state, since it would also be set to true when input loses focus

EmilyBermudez88:
> > This only occurs when I click on a dropdown item when itemTooltip is specified, and NOT when I select an item via arrow keys + Enter. Now that showDropdown is also being affected by hiddenBySelection, the class nx-combobox--showable-dropdown is removed when something is selected, therefore setting the dropdown menu to display:none. Would you recommend I simply edit the CSS to have nx-combobox__menu set to visibility: hidden, and nx-combobox__dropdown-showable to visibility: visible, or would there be a better way to handle this? In previous iterations I had the matches conditionally render based on showDropdown, but this strays away from the original functionality, so I assume that will want to be left as is.
> 
> I'm not thinking of a reason why the dropdown would need to remain in the DOM when it is not showable, so I would try changing it to be rendered conditionally at that level rather than hidden in CSS. However, just because I can't remember a reason that it is the way it is doesn't mean there isn't one, so test it carefully after making that change, and check with @cindyko226 to see if she remembers.
> 
> If removing it from the DOM doesn't work, yes you could try `visibility: hidden` but I'm not certain whether that will cause the element to block clicks on things behind it. I don't think it will, but if it does, that can probably be addressed with `pointer-events: none`.

I checked with Cindy, and she can't think of a reason the dropdown should need to remain on the page when not showable, so I will work on removing it and test accordingly.

gallery/visualtests/nxCombobox.js
EmilyBermudez88: > Crucially, we want to test that click reopens it _even if it never lost focus in the first place_. ie. what you have here but without clicking the `basicExample` in between

I missed actually implementing the behaviour of re-opening the dropdown on click of the input, so that has been added here along with expanding the test: [1e7a194](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/1e7a1947a39353cf05e7ba45124d34d0d28e7cdf)

EmilyBermudez88:
> I'm not thinking of a reason why the dropdown would need to remain in the DOM when it is not showable, so I would try changing it to be rendered conditionally at that level rather than hidden in CSS.

Set conditional rendering of the dropdown & alert here, with Cindy's approval: [5f0517b](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/5f0517bc26bccd1922d2dfd4d4172be7833fabbe). Tests are failing, but will be resolved in future commits.

lib/src/components/NxCombobox/NxCombobox.tsx
EmilyBermudez88: Based on previous comment & discussion on Slack, conditional rendering implemented here: [5f0517b](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/5f0517bc26bccd1922d2dfd4d4172be7833fabbe)

> > Oh interesting, I hadn't thought about the arrow down case. I think a cleaner way to do that would just be to use `onChange` directly in `adjustBtnFocus` and to avoid the conditional in `handleSelection`. I also think `handleSelection` deserves a brief comment about what it does and when it should be used.
> 

Update of `handleSelection` function here: [63002d8](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/63002d8b32862f4b11bcc657fc3d629feedcf61e)

lib/src/components/NxCombobox/NxCombobox.scss
```diff
@@ -45,27 +43,9 @@
...
-        position: absolute;
-        right: initial;
-      }
-    }
-
-    &:focus-within {
-      @include show-dropdown;
+  &:focus-within {
+    .nx-dropdown-button.selected {
+      background-color: var(--nx-swatch-blue-90);
```
rpokorny, line 48: I feel like this probably doesn't need to be inside of `:focus-within`

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
EmilyBermudez88: > See https://testing-library.com/docs/user-event/convenience#tab

Thanks for forwarding the link. Fixed here: [f85b73a](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/f85b73a035d3ea670ef008a821346c94d3d7e53f)

lib/src/components/NxCombobox/NxCombobox.scss
EmilyBermudez88: > I feel like this probably doesn't need to be inside of `:focus-within`

removed here: [3246834](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/3246834ecfce508514dfa70948b9035699f52413)

gallery/visualtests/nxCombobox.js
EmilyBermudez88: > This is only true in the case where the list of options is immediately available. In the case where the options populate once you start typing, the dropdown _should not_ be visible until you start typing. I wonder if this entire `describe` block should be duplicated, once for the "Basic Example" and once for the "Backend Query Example", with the first test in the latter changed to check that the dropdown opens once the user has typed.

Makes sense! Added here: [f5745f9](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/f5745f9ad1c5136dc159b4eac0c839e66ac53348)

rpokorny:
> I'm currently having this error come up when I click on an item in the dropdown, which I believe is triggered by the tooltip.

This is fixed now, correct?

gallery/visualtests/nxCombobox.js
```diff
@@ -266,17 +266,216 @@ describe('NxCombobox', function() {
...
+      expect(await isFocused(input)).toBe(false);
+      expect(await isVisible(dropdownMenu)).toBe(false);
+      await input.click();
+      [dropdownMenu] = await waitAndGetElements(dropdownMenuSelector);
+      expect(await isFocused(input)).toBe(true);
+      expect(await isVisible(dropdownMenu)).toBe(true);
+    });
+  });
+
+  describe('dropdown display behavior with backend query', function() {
```
rpokorny, line 371:
```suggestion
  describe('dropdown display behavior with initially unpopulated list', function() {
```

gallery/visualtests/nxCombobox.js
```diff
@@ -266,17 +266,216 @@ describe('NxCombobox', function() {
...
+    it('shows the dropdown when focused and the user has typed a value', async function() {
+      const inputSelector = `${backendExampleSelector} .nx-combobox__input input`,
+          dropdownMenuSelector = `${backendExampleSelector} .nx-dropdown-menu`,
+          [input] = await waitAndGetElements(inputSelector);
+
+      await input.focus();
+      await getPage().keyboard.type('i');
+
+      const [dropdownMenu] = await waitAndGetElements(dropdownMenuSelector);
+      expect(await isVisible(dropdownMenu)).toBe(true);
```
rpokorny, line 381: The key difference from the predetermined list case is that the dropdown should not open before the user types a value.  This test should check that

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,18 +83,103 @@ describe('NxCombobox', function() {
         expect(inputElement).toHaveValue('Foo');
       });
 
-  it('sets aria-expanded on the input to true when focused',
+  it('sets aria-expanded to true on the input when initially focused', async function() {
```
rpokorny, line 86: Just like the predetermined list vs backend query cases in the functional test, this `aria-expanded` behavior is only true when the `matches` is populated, right? We should have another test for when it isn't

EmilyBermudez88:
> > I'm currently having this error come up when I click on an item in the dropdown, which I believe is triggered by the tooltip.
> 
> This is fixed now, correct?

Yep, fixed now.

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
EmilyBermudez88: > > Just like the predetermined list vs backend query cases in the functional test, this aria-expanded behavior is only true when the matches is populated, right? We should have another test for when it isn't
> 
> Should probably also have cases that check `aria-expanded` when the loading, empty, and error alerts are supposed to be present

Added in extra testing here: [1d7f7bf](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/1d7f7bfa6d42416ba7ba9ba6eebb44ad66b462d0)

gallery/visualtests/nxCombobox.js
EmilyBermudez88: > The key difference from the predetermined list case is that the dropdown should not open before the user types a value. This test should check that

added here: [46dd648](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/46dd648cb7c5a57ca0e0597587b9e59c7afccd09)

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,18 +83,122 @@ describe('NxCombobox', function() {
...
+  it('set aria-expanded to false on the input if there are no matches, even if focused', async function() {
+    const { rerender, getByRole } = quickRender(),
+        inputElement = getByRole('combobox');
+
+    expect(inputElement).toHaveAttribute('aria-expanded', 'false');
+
+    inputElement.focus();
+    expect(inputElement).toHaveAttribute('aria-expanded', 'false');
+
+    rerender(<NxCombobox {...minimalProps} value= {'foo'}/>);
```
rpokorny, line 108:
```suggestion
    rerender(<NxCombobox {...minimalProps} value="foo" />);
```

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,18 +83,122 @@ describe('NxCombobox', function() {
...
+    expect(inputElement).toHaveAttribute('aria-expanded', 'false');
```
rpokorny, line 109: Shouldn't this one be true? Shouldn't this show the loading alert?

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,18 +83,122 @@ describe('NxCombobox', function() {
...
+
+    rerender(<NxCombobox {...minimalProps} loading={true}/>);
+    expect(inputElement).toHaveAttribute('aria-expanded', 'false');
+
+    rerender(<NxCombobox { ...minimalProps } loadError={'boo'} />);
+    expect(inputElement).toHaveAttribute('aria-expanded', 'false');
```
rpokorny, line 115: Also test the cases where loading and error would result in `aria-expanded` being true: when `loading` or `loadError` are set and `value` is also set

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -83,18 +83,122 @@ describe('NxCombobox', function() {
         expect(inputElement).toHaveValue('Foo');
       });
 
-  it('sets aria-expanded on the input to true when focused',
+  it('sets aria-expanded to true on the input when initially focused and there are matches', async function() {
```
rpokorny, line 86: Consider grouping all of these in a `describe('aria-expanded'` block

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
EmilyBermudez88: > Shouldn't this one be true? Shouldn't this show the empty alert?

Aria-expanded was previously set according to `showDropdown && inputIsFocused`, and now `showDropdown`, so without matches, `aria-expanded` is set to false.
I just read through w3.org, and it says "when the combobox popup is not visible, the element with role combobox has aria-expanded set to false".  
Above that it mentions "the popup is an element that has a role listbox, tree, grid or dialog" (https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#:~:text=The-,popup,-is%20an%20element). If the empty message div has a parent element of `role="alert"`,  I think we should leave it as `aria-expanded= false`?

rpokorny: > Aria-expanded was previously set according to showDropdown && inputIsFocused, and now showDropdown, so without matches, aria-expanded is set to false.

Ah you're right.  In the previous impl it was false in that circumstance, and we aren't looking to change that

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
rpokorny: > Also test the cases where loading and error would result in aria-expanded being true: when loading or loadError are set and value is also set

I was wrong about this per https://github.com/sonatype/sonatype-react-shared-components/pull/876#discussion_r1063501298

rpokorny (approved):

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
EmilyBermudez88: > Consider grouping all of these in a `describe('aria-expanded'` block

Nested in a describe block here: [f396ba8](https://github.com/sonatype/sonatype-react-shared-components/pull/876/commits/f396ba8089204a5b15f7cdaa10c923778fa8ac6a)

PR #877, opened 2022-12-16
Upgrade Jest and Related Dependencies - RSC-1286

rpokorny:
https://issues.sonatype.org/browse/RSC-1286


EmilyBermudez88:
out of curiosity, how are you able to tell what are the related dependencies for jest (in the gallery's `package.json`, `@nivo/core` & `@babel/runtime` are updated as well)?

rpokorny:
> out of curiosity, how are you able to tell that @babel/runtime is a related dependencies for jest?

Because I happen to know that the only reason we use babel at all is for transpiling test files to a format that will run on node for testing.

lib/jest.config.js
```diff
@@ -12,11 +12,11 @@ module.exports = {
     "<rootDir>/src"
   ],
   "transform": {
-    "^.+\\.tsx?$": "ts-jest",
-    "^.+\\.jsx?$": "babel-jest"
+    "^.+\\.(t|j)sx?$": ["ts-jest", { tsconfig: '<rootDir>/tsconfig.test.json' }]
```
EmilyBermudez88, line 15: I figure I'd spend a bit of time reading into what the `jest.config.js` does exactly, and I'm still a bit unsure about `transform` here. From what I gather, it helps convert JSX or Typescript we may be using into JS for jest to be able to run, and that the syntax is `{string, pathToTransformer | [pathToTransformer, object] }` - I'm just a bit unsure here as to why the `tsconfig: ...` object was added here? Just for clarity, thanks!

EmilyBermudez88: >The config object with tsconfig is to tell typescript to use tsconfig.test.json instead of the default tsconfig.json.

Thanks for clarifying, that clears everything up!

EmilyBermudez88 (approved):

PR #879, opened 2022-12-19
NxStateful Accordion RTL Conversion - RSC-1285

rjp99:
https://issues.sonatype.org/browse/RSC-1285


rpokorny:
Because RTL doesn't do shallow rendering and therefore doesn't allow us to assert anything about one component being built on top of another component, we can't take for granted behaviors that a component inherits from whatever it wraps.  This is particularly inconvenient in the case of stateful components: we can no longer just say that they wrap the stateless version and leave it to the stateless version's unit tests to test most of the functionality.  Instead, we have to duplicated all of the stateless component's unit tests, as applicable, on the stateful version.

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,213 @@
...
+import { render } from '@testing-library/react';
+import NxButton from '../../../NxButton/NxButton';
 
 describe('NxStatefulAccordion', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxStatefulAccordion, {});
+
+  const quickRender = rtlRender(NxStatefulAccordion, {});
+  const renderEl = rtlRenderElement(NxStatefulAccordion, {});
 
   it('renders an NxAccordion with the specified props', function() {
```
rpokorny, line 21:
```suggestion
  it('renders a <details> with the specified props', function() {
```

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,213 @@
...
     const onToggle = jest.fn(),
-        component = getShallowComponent({ onToggle, className: 'foo', id: 'bar' });
+        component = renderEl({ onToggle, className: 'foo', id: 'bar' });
```
rpokorny, line 23: This classname isn't being tested

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,213 @@
...
   });
 
-  it('toggles the NxAccordion open prop when the NxAccordion onToggle callback is called', function() {
-    const component = getShallowComponent();
+  describe('onToggle', function() {
+    describe('when the accordion is currently closed', function() {
+      it('toggles the accordion open prop when the header is clicked', async function() {
+        const user = userEvent.setup(),
+            onToggle = jest.fn(),
+            el = quickRender({ onToggle,
```
rpokorny, line 111:
```suggestion
            el = quickRender({ 
              onToggle,
```

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,213 @@
...
+        )
+      });
+
+      expect(container.querySelector('SUMMARY')).toHaveAttribute('aria-controls', 'foo');
+    });
   });
 
-  it('toggles the NxAccordion open prop when the NxAccordion onToggle callback is called', function() {
-    const component = getShallowComponent();
+  describe('onToggle', function() {
```
rpokorny, line 106: The way these tests are organized is a little bit confusing.  I'd change this wrapping `describe` to be about "accordion header click" because the tests below test more than just `onToggle`, they test all of the behaviors that result from clicking the accordion.  Then, in the "when the accordion is currently closed" block, I'd separate the testing of `onToggle` and the state change into two different tests.  Then in the "when the accordion is currently open" block I'd follow the same pattern - basically make the tests identical except for the minimum changes to the setup and assertions.

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,213 @@
...
+        // click the accordion outside of the header
+        await user.click(container);
+
+        expect(onToggle).toHaveBeenCalledTimes(1);
+      });
     });
 
-    expect(component).toHaveProp('open', false);
+    describe('when a non-button element in the header with its own onClick handler is clicked', function() {
+      it('fires', async function() {
```
rpokorny, line 159:
```suggestion
      it('changes the state and fires onToggle', async function() {
```

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,213 @@
...
+
+        await user.click(title);
+
+        expect(onToggle).toHaveBeenCalledTimes(1);
+        expect(titleOnClick).toHaveBeenCalledTimes(1);
+      });
+    });
+
+    describe('when a button element in the header is clicked', function() {
+      it('does not fire', async function() {
```
rpokorny, line 184:
```suggestion
      it('does not change the state or fire onToggle', async function() {
```

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
rjp99: > This classname isn't being tested. That functionality is tested in the next test after this one, so the `className` here can be removed.

Removed here: 6078c7cc795259938902c153bc24d6657a67e910

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
rjp99: > The way these tests are organized is a little bit confusing. I'd change this wrapping `describe` to be about "accordion header click" because the tests below test more than just `onToggle`, they test all of the behaviors that result from clicking the accordion. Then, in the "when the accordion is currently closed" block, I'd separate the testing of `onToggle` and the state change into two different tests. Then in the "when the accordion is currently open" block I'd follow the same pattern - basically make the tests identical except for the minimum changes to the setup and assertions.

Like this?  7fe3a943195f7cb77776aec04c08b9a636b59083

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,250 @@
...
+        expect(el.getByRole('group')).toHaveAttribute('open');
+
+        await user.click(header);
+
+        expect(el.getByRole('group')).not.toHaveAttribute('open');
+      });
+    });
+
+    describe('when a non-button element in the header with its own onClick handler is clicked', function() {
+      it('changes the open state and fires onToggle', async function() {
```
rpokorny, line 196: I don't see assertions in this test about the open state

rpokorny: > I don't see assertions in this test about the open state

Same with the test below

EmilyBermudez88: > I don't see assertions in this test about the open state

Added in here: [7188b0a](https://github.com/sonatype/sonatype-react-shared-components/pull/879/commits/7188b0a816cd9498a090f26065b4f3ac8a75a0c1)

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,255 @@
...
+              <NxAccordion.Title>Foo</NxAccordion.Title>
+              <div className="nx-btn-bar">
+                <NxButton id="btn1" />
+                <NxButton id="btn2" onClick={btnOnClick} />
+              </div>
+            </NxAccordion.Header>
+          </NxStatefulAccordion>,
+        );
+        const btn1 = container.querySelector('button#btn1') as HTMLElement,
+            btn2 = container.querySelector('button#btn2') as HTMLElement;
```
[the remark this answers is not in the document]
rpokorny: > Can't we just use #btn1 and #btn2 as selectors?

I would think so.  Even "better" I think would be to use testids so then we can use RTL queries

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,255 @@
...
+    describe('when a non-button element in the header with its own onClick handler is clicked', function() {
+      it('changes the open state and fires onToggle', async function() {
+        const user = userEvent.setup();
+        const titleOnClick = jest.fn(),
+            onToggle = jest.fn();
+
+        const { container } = render(
+          <NxStatefulAccordion onToggle={onToggle}>
+            <NxAccordion.Header>
+              <NxAccordion.Title onClick={titleOnClick}>Foo</NxAccordion.Title>
```
rpokorny, line 204: This relies on the fact that `NxAccordion.Title` can actually take an `onClick` prop, which isn't really the point.  I think it'd be cleaner to wrap the "Foo" in a span with a testid and put the `onClick` handler on that

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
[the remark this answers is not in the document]
EmilyBermudez88: > > Can't we just use #btn1 and #btn2 as selectors?
> 
> I would think so. Even "better" I think would be to use testids so then we can use RTL queries

Updated tests here: [141fd74](https://github.com/sonatype/sonatype-react-shared-components/pull/879/commits/141fd742930b5eadea1da2665c234b418138d1cb)  (and subsequently removed accidental `console.log`)

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
EmilyBermudez88: > This relies on the fact that `NxAccordion.Title` can actually take an `onClick` prop, which isn't really the point. I think it'd be cleaner to wrap the "Foo" in a span with a testid and put the `onClick` handler on that

Makes sense, updated here: [141fd74](https://github.com/sonatype/sonatype-react-shared-components/pull/879/commits/141fd742930b5eadea1da2665c234b418138d1cb)

EmilyBermudez88:
I just noticed when I go onto the NxAccordion gallery page, this error comes up: 
Screenshot 2023-01-11 at 11 05 23 AM
When I remove the "Deprecated" `NxWarningAlerts`, the error disappears. However, I am not seeing this error `gallery.sonatype.dev`, and yet the gallery page has been untouched in this PR. I'm curious if others see this as well.

rpokorny:
> I just noticed when I go onto the NxAccordion gallery page, this error comes up:

Don't worry about that.  It's because the example description gets wrapped in a `<p>` and so really shouldn't have structured content within it.  We haven't gotten around to making a better solution.  I would assume you don't see the error on gallery.sonatype.dev because that site runs in prod mode which drops all of the sanity checks in the react code.

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,253 @@
...
-    expect(getShallowComponent({ defaultOpen: false })).toHaveProp('open', false);
-    expect(getShallowComponent({ defaultOpen: true })).toHaveProp('open', true);
+    expect(renderEl()).not.toHaveAttribute('open');
+    expect(renderEl({ defaultOpen: false })).not.toHaveAttribute('open');
+    expect(renderEl({ defaultOpen: true })).toHaveAttribute('open');
   });
 
-  it('toggles the NxAccordion open prop when the NxAccordion onToggle callback is called', function() {
-    const component = getShallowComponent();
+  it('renders non-header children in content wrapper', function() {
```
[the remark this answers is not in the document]
rpokorny: > was getting this console error (also seen in NxAccordion tests),

I think this is because `NxAccordion` has an unwritten rule that the `NxAccordion.Header` must be one of its immediate JSX descendants.  Your fix is good.  We should add a ticket to get that documented in the gallery page.

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,253 @@
...
+    const { container } = render(
+      <NxAccordion>
```
rpokorny, line 48: Shouldn't this be `NxStatefulAccordion` though?

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
```diff
@@ -4,46 +4,253 @@
...
+        <span className="bar">Bar</span>
+      </NxAccordion>
+    );
+
+    expect(container.querySelector('summary .bar')).not.toBeInTheDocument();
+    expect(container.querySelector('.bar')).toBeInTheDocument();
+  });
 
-    expect(component).toHaveProp('open', false);
+  describe('Header', function() {
```
rpokorny, line 60: This is just testing `NxAccordion.Header` which is the same component that is identically tested in `NxAccordion.test.tsx`.  So this block should be removed.

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
EmilyBermudez88: > Shouldn't this be `NxStatefulAccordion` though?

Yes it should, thanks for catching that: [e8d0404](https://github.com/sonatype/sonatype-react-shared-components/pull/879/commits/e8d040408cca55a32d06a3559633d56c41d14130)

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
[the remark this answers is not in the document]
EmilyBermudez88: Ah, that makes more sense now. Ticket created here: https://issues.sonatype.org/browse/RSC-1397

lib/src/components/NxAccordion/stateful/__tests__/NxStatefulAccordion.test.tsx
EmilyBermudez88: > This is just testing `NxAccordion.Header` which is the same component that is identically tested in `NxAccordion.test.tsx`. So this block should be removed.
> 
> Edit: actually the `aria-controls` tests are still relevant. Just the others should be removed.

Removed the first two unnecessary Header tests: [18e6fec](https://github.com/sonatype/sonatype-react-shared-components/pull/879/commits/18e6fec350c1c3710dea2ecbe8198a2453b99344)

rpokorny (approved):

PR #880, opened 2022-12-20
RSC-1234 - NxTransferListHalf RTL Conversion

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1234


lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -4,495 +4,537 @@
...
+      displayName: 'foo'
+    }, {
+      id: '2',
+      displayName: 'baz'
+    }];
 
-    expect(select.find(NxFontAwesomeIcon)).not.toExist();
-    expect(select.find('input.nx-transfer-list__checkbox')).not.toExist();
+    const { container } = quickRender({ items: dataItems }),
+        itemList = container.querySelector('.nx-transfer-list__item-list') as HTMLElement,
```
EmilyBermudez88, line 151: At this time, there are no roles assigned to `.nx-transfer-list__item-list` or `.nx-transfer-list__item`, so need to use `querySelector`. RSC-998 should improve roles, which will mean updating the tests

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -4,495 +4,537 @@
...
         items: [],
         footerContent: <div/>
       },
-      getShallow = getShallowComponent<Props<number>>(NxTransferListHalf, minimalProps),
-      getMounted = getMountedComponent<Props<number>>(NxTransferListHalf, minimalProps);
+      quickRender = rtlRender(NxTransferListHalf, minimalProps),
+      renderEl = rtlRenderElement(NxTransferListHalf, minimalProps);
+
+  beforeEach(function() {
+    // JSDOM is missing this function. https://github.com/jsdom/jsdom/issues/3002
```
EmilyBermudez88, line 30: Found the same solution online as `NxBreadCrumb` after getting `range.getBoundingClientRect is not a function`.

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -4,495 +4,537 @@
...
+    async function advanceTimers(time: number) {
+      await act(async () => { await jest.advanceTimersByTime(time); });
+    }
 
-    expect(onReorderItem).toHaveBeenCalledWith(1, 1);
-
-    moveUpButton.simulate('click');
+    it('renders move up and down buttons inside .nx-transfer-list__item when allowReordering is true',
+        async function() {
+          jest.useRealTimers();
```
EmilyBermudez88, line 314: I'm not sure this is the best resolution. On line 332, I want to check that the rendered buttons are indeed `Move Up` and `Move Down` buttons, and need to wait for the tooltip to initialize to confirm this. With `jest.useFakeTimers()` in the beforeEach, it causes this test to fail, and adding `await runTimers()` or `await advanceTimers(1000)` doesn't remedy the problem either. My understanding of jest's timers is still a bit shaky, so perhaps there is a better and cleaner way to resolve this issue than simply resetting the test to `jest.useRealTimers()`

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
rpokorny: > At this time, there are no roles assigned to .nx-transfer-list__item-list or .nx-transfer-list__item, so need to use querySelector. RSC-998 should improve roles, which will mean updating the tests

Please add either a comment on that ticket referencing this, and/or a code comment here referencing that ticket, to try to prevent us from forgetting

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
rpokorny: > Found the same solution online as NxBreadCrumb after getting range.getBoundingClientRect is not a function.

It's seeming like that's going to be necessary everywhere that `NxOverflowTooltip` is used.  Rather than having to mock that everywhere, that mock should be moved into `setupTests.ts`.  That file is configured using jest's [`setupFilesAfterEnv`](https://jestjs.io/docs/configuration#setupfilesafterenv-array)

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -4,495 +4,537 @@
...
 
-    const firstItem = list.find('.nx-transfer-list__item').at(0),
-        secondItem = list.find('.nx-transfer-list__item').at(1);
-    const moveDownButton = firstItem.find('.nx-btn').at(1);
-    const moveUpButton = secondItem.find('.nx-btn').at(0);
+    afterEach(runTimers);
 
-    expect(moveDownButton).toExist();
-    expect(moveUpButton).toExist();
+    async function runTimers() {
```
rpokorny, line 304: This is going to come up on every component that has tooltips.  Let's move these to a file of utility functions

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
rpokorny: > adding await runTimers() or await advanceTimers(1000) doesn't remedy the problem either. 

I've wrestled quite a bit with the idiosyncracies of jest's fake timers, and this test is the most bizarre one yet.  If I modify what you have here to use an appropriately-placed `await runTimers()` instead of `jest.useRealTimers()`, it fails in the following scenarios:
* When I run `yarn test-watch` with this as the only locally-modified file (meaning that this file is the only one who's tests run)
* When I run `yarn test-watch` and press 'a' to toggle its "run all tests" option
* When I run `yarn jest src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx`

However, it passes in every other scenario I try, including all of the following
* When I run `yarn test-watch` but with a few `console.log` statements added to `updateBatcher.ts`  These statements should have no meaningful effect on execution timing, but do cause a significant number of other tests to run besides the one in question
* When I run all tests with `yarn test`
* When I run just this test using `yarn jest --testNamePattern='renders move up and down buttons'`
* When I use `yarn test-watch`'s filter by test name feature with "renders move up and down buttons"

What I gather from all of this is that there must be some side effect that gets invoked simply from jest reading in one of the other files in our test suite (even if none of the tests in that file get executed), which causes this test to pass.  Only in execution modes where the other files aren't read at all does this test fail.  I will attempt to track down the side effect in question.

rpokorny: > When I run yarn test-watch but with a few console.log statements added to updateBatcher.ts These statements should have no meaningful effect on execution timing, but do cause a significant number of other tests to run besides the one in question

Actually, for at least that scenario, I am now seeing it fail sometimes and not others.  Still investigating.

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
EmilyBermudez88: > Please add either a comment on that ticket referencing this, and/or a code comment here referencing that ticket, to try to prevent us from forgetting

Added on the ticket here: https://issues.sonatype.org/browse/RSC-988, and in the code here: [76ab1ca](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/76ab1ca64f8cd9e8c7af3be8d0dce01bb7ee7706)

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
EmilyBermudez88: > It's seeming like that's going to be necessary everywhere that `NxOverflowTooltip` is used. Rather than having to mock that everywhere, that mock should be moved into `setupTests.ts`. That file is configured using jest's [`setupFilesAfterEnv`](https://jestjs.io/docs/configuration#setupfilesafterenv-array)

Moved into `setupTests.ts` here: [26ae3c2](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/26ae3c2df710029d6a89f10da24d594df39338e9). I kept getting an eslint error for `jest.fn()` saying `jest is not defined`. Andrew had a good suggestion of adding `setupTests.js` to the `.eslintrc` file in order to deal with this. If all looks good, then I'll also remove the same `beforeEach` in NxBreadcrumb.

rpokorny: > Moved into setupTests.ts here: [26ae3c2](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/26ae3c2df710029d6a89f10da24d594df39338e9). I kept getting an eslint error for jest.fn() saying jest is not defined. Andrew had a good suggestion of adding setupTests.js to the .eslintrc file in order to deal with this. If all looks good, then I'll also remove the same beforeEach in NxBreadcrumb.

Great.  I also made similar changes in RSC-1336.  You'll want to merge that branch into this one so you can more cleanly deal with the timing problems.

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
rpokorny: > Actually, for at least that scenario, I am now seeing it fail sometimes and not others. Still investigating.

I think it's fixed with the changes in RSC-1336.  Please merge that branch into this one.  After doing so, you'll want to retarget this PR at that branch.  Not because it's actually going to get merged there, but so that the diff stays accurate

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
rpokorny: > This is going to come up on every component that has tooltips. Let's move these to a file of utility functions

Done in RSC-1336

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -4,495 +4,537 @@
...
+        itemList = container.querySelector('.nx-transfer-list__item-list') as HTMLElement,
+        items = itemList.querySelectorAll('.nx-transfer-list__item');
+
+    expect(items.length).toBe(2);
+    expect(items[0]).toHaveTextContent('foo');
+    expect(items[1]).toHaveTextContent('baz');
   });
 
-  it('sets the item checked prop to isSelected', function() {
+  it('renders a move icon and checkbox input only when onItemChange is provided', function() {
```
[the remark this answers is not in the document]
EmilyBermudez88: > Do we have visual tests that adequately check the conditional around the icon? I think that would be the better way to test it than here

We didn't have any visual tests that explicitly laid out when the icon was visible or not, so added visual tests here: [ac29087](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/ac290870d4a087274abd1a214082c01838320147). Since we can't test which icon is being used when `isSelected` is true or false, I added clarity of which icon is rendered when.

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
EmilyBermudez88: > > This is going to come up on every component that has tooltips. Let's move these to a file of utility functions
> 
> Done in RSC-1336

Merged with main here: [5eed414](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/5eed41445b5e905b6c731b3ef177428f780c46e1) and removed redundant code here: [61e13d4](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/61e13d48b2a6541ff75e5394972af6c54badf465). Thanks for working on this in RSC-1336

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
EmilyBermudez88: > > Actually, for at least that scenario, I am now seeing it fail sometimes and not others. Still investigating.
> 
> I think it's fixed with the changes in RSC-1336. Please merge that branch into this one. After doing so, you'll want to retarget this PR at that branch. Not because it's actually going to get merged there, but so that the diff stays accurate

Merged with main here: [5eed414](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/5eed41445b5e905b6c731b3ef177428f780c46e1) and removed redundant code here: [61e13d4](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/61e13d48b2a6541ff75e5394972af6c54badf465). I checked to see if everything was passing after redundancies were removed, and all tests are passing.

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
EmilyBermudez88: > > Moved into setupTests.ts here: [26ae3c2](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/26ae3c2df710029d6a89f10da24d594df39338e9). I kept getting an eslint error for jest.fn() saying jest is not defined. Andrew had a good suggestion of adding setupTests.js to the .eslintrc file in order to deal with this. If all looks good, then I'll also remove the same beforeEach in NxBreadcrumb.
> 
> Great. I also made similar changes in RSC-1336. You'll want to merge that branch into this one so you can more cleanly deal with the timing problems.

I merged your changes here: [5eed414](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/5eed41445b5e905b6c731b3ef177428f780c46e1)

lib/src/components/NxTransferListHalf/__tests__/NxTransferListHalf.test.tsx
```diff
@@ -4,496 +4,507 @@
...
+            displayName: 'foobar'
+          }];
+
+          const { container, rerender } = quickRender({ allowReordering: true, items: dataItems}),
+              middleItem = container.querySelectorAll<HTMLElement>('.nx-transfer-list__item')[1],
+              buttons = within(middleItem).getAllByRole('button', { hidden: true });
+
+          // assert both a move up and move down button are present
+          expect(buttons.length).toBe(2);
+          await waitFor(() => expect(buttons[0]).toHaveAccessibleName('Move Up'));
```
rpokorny, line 302: Does this actually work now, without a call to `runTimers` before it? What about without the `waitFor`s?

EmilyBermudez88: You're right, I needed to replace that. Updated here and working: [ef7faff](https://github.com/sonatype/sonatype-react-shared-components/pull/880/commits/ef7faffad6cfd011c7ca3107d0dd3a235050ff34)

rpokorny (approved):

PR #881, opened 2022-12-20
Fix Build by Updating Puppeteer  - RSC-1327

rpokorny:
https://issues.sonatype.org/browse/RSC-1327

The is the best solution I could get for the latest functional/visual test breakage.  Unfortunately, it is still not a solution that is going to be stable going forward, but hopefully each time going forward that it breaks we can simply upgrade Puppeteer again to fix it again.

Here's why this is the best/only way forward:
* We need to support building on Linux, Mac (both Intel and M1) and ideally also Windows
* To get exactly matching snapshots across OSes, a docker container is used so that the test always runs in a Linux environment regardless of the host
* The docker environment matches the architecture of the host - it's Linux x64 on intel systems and Linux arm64 on M1 Macs.  M1 Macs _can_ run x64 docker containers but they're unstable and, from what I remember when we originally tried to get all of this working there, don't work well enough to run Chromium in the docker container
* Puppeteer's ideal way of running is to download Chromium itself.  Each version of puppeteer has exactly one version of chromium that it officially works with, and that is the version it downloads.  However, the downloads come from the Chromium projects official Google storage site, and are limited to what builds are available there.  *That site does not offer Linux arm64 downloads*.

The only known way to run puppeteer on Linux arm64 is to provide the Chromium version ourselves.  To do that, I know of two options: use the OS package manager within the docker container, or build Chromium ourselves, in Linux on an M1 Mac.

The latter option there stretches the limits of feasibility, so we're going with the former.  The downside there is that the version of Chromium installed in the docker container will change over time.  We'll just have to keep updating Puppeteer as that happens.

Another, completely different approach that I tried was getting puppeteer to run against the Chromium installation present in the selenium-chrome docker container that we already use for the SSR tests (and used to use for the applitools tests).  That approach seemed promising but ran into serious issues: the snapshots were all mysteriously misaligned (and therefore incorrect) and also the tests all seemed to run sequentially, drastically increasing build times.


cindyko226:
Confirmed the visual tests all passed with M1 Mac.

EmilyBermudez88 (approved):

PR #885, opened 2022-12-23
Add Missing overflowTooltip Prop to NxStatefulCheckbox - RSC-1260

rpokorny:
https://issues.sonatype.org/browse/RSC-1260

The way the stateful props for this compoment were written was very outdated and redundant, so I refactored it.


entroform (approved):
LGTM!
Just wondering, is this a new pattern we're moving forward with:
no longer having a separate file for stateful component prop types?

rpokorny:
> LGTM!
Just wondering, is this a new pattern we're moving forward with:
no longer having a separate file for stateful component prop types?

I think we mostly started to do it that way awhile ago.  We're definitely not entirely consistent about it though.  But when there's a lot of overlap between the stateful and stateless props (which is almost always the case), the types should probably be related

PR #889, opened 2022-12-29
Update Warning Colors - RSC-1124

rjp99:
https://issues.sonatype.org/browse/RSC-1124

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.9.0",
+  "version": "12.9.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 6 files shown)


rpokorny:
The documentation for the existing css variable states

> The color used on the NxAlert warning icon and for other equivalent icons and purposes.

Considering that, I think the existing variable should continue to be used for the icon rather than the border, and the border should either use a new variable or just use the swatch color directly.  I'm leaning towards the latter: our CSS variables are a public, documented part of our API surface and I don't know that it makes sense for this very specific border color to be part of the public API - just like the alert background color is not part of the public API.

rjp99:
> The documentation for the existing css variable states
> 
> > The color used on the NxAlert warning icon and for other equivalent icons and purposes.
> 
> Considering that, I think the existing variable should continue to be used for the icon rather than the border, and the border should either use a new variable or just use the swatch color directly. I'm leaning towards the latter: our CSS variables are a public, documented part of our API surface and I don't know that it makes sense for this very specific border color to be part of the public API - just like the alert background color is not part of the public API.

Updated here: 69ecbcc84ab41c1f9ca1f42b2c9f588d2b7f47fd

rpokorny (approved):

PR #890, opened 2023-01-04
Fix NxModal Horizontal Scrollbars Appearing in Safari 16.1+ - RSC-1332

entroform:
https://issues.sonatype.org/browse/RSC-1332


lib/src/components/NxModal/NxModal.scss
```diff
@@ -29,7 +29,6 @@
     flex-basis: auto;
     flex-direction: column;
     flex-shrink: 1;
-    overflow-y: hidden;
```
entroform, line 32: Removing this or setting x to hidden as well fixes it, but not sure if it that's a good solution.
`overflow: hidden`

rpokorny: I find it hard to believe that this change doesn't cause problems with scrolling form modals (I can't check of course as I have no access to Safari 16.1).  At the same time, the other solution you propose would seem like it shouldn't cause any problems especially in light of this disclaimer which `NxModal` has always carried:
image

So I'd go with the other solution.

entroform: > I find it hard to believe that this change doesn't cause problems with scrolling modals (I can't check of course as I have no access to Safari 16.1). At the same time, the other solution you propose would seem like it shouldn't cause any problems especially in light of this disclaimer which `NxModal` has always carried: image
> 
> So I'd go with the other solution.

Ah yeah I forgot to update the comment, but I went with the other solution because of that issue here:
14a2e69a4d0ae50738a80bfa0506ce0deb4ac164

rpokorny (approved):

rpokorny:
Actually, were you able to see investigate what exactly is overflowing and causing the scrollbar? Just ignoring that overflow with `overflow: hidden` makes me uneasy if we don't have a thorough understanding of what's going on

entroform:
> Actually, were you able to see investigate what exactly is overflowing and causing the scrollbar? Just ignoring that overflow with `overflow: hidden` makes me uneasy if we don't have a thorough understanding of what's going on

From my investigation yesterday it appears that the problem is with negative margin-right on nx-form and nx-modal-content, it doesn't remove the whitespace on the right causing it to overflow.

rpokorny:
> From my investigation yesterday it appears that the problem is with negative margin-right on nx-form and nx-modal-content, it doesn't remove the whitespace on the right causing it to overflow.

I wonder then, if a modal was constructed with some child content with `width: 100%`, if that would look right in Safari 16.1

entroform:
> > From my investigation yesterday it appears that the problem is with negative margin-right on nx-form and nx-modal-content, it doesn't remove the whitespace on the right causing it to overflow.
> 
> I wonder then, if a modal was constructed with some child content with `width: 100%`, if that would look right in Safari 16.1

It looks right when I tested it!

rpokorny (approved):

PR #893, opened 2023-01-09
Always use Fake Timers in Unit Test - RSC-1336

rpokorny:
https://issues.sonatype.org/browse/RSC-1336

### DIFF
--- a/lib/src/__testutils__/rtlUtils.tsx
+++ b/lib/src/__testutils__/rtlUtils.tsx
@@ -5,11 +5,13 @@
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
 import React, { ComponentType } from 'react';
-import { render, RenderResult } from '@testing-library/react';
+import { act, render, RenderResult } from '@testing-library/react';
 
 import '@testing-library/jest-dom';
+import realUserEvent from '@testing-library/user-event';
 import { path, pipe } from 'ramda';
 import { within } from '@testing-library/dom';
+import { Options } from '@testing-library/user-event/dist/types/options';
 
 export function rtlRender<P>(Component: ComponentType<P>, minimalProps: P) {
   return function renderWrapper(additionalProps?: Partial<P>): RenderResult {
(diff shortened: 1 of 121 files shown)


EmilyBermudez88:
When I'm running tests I get this at the end: 
Screenshot 2023-01-10 at 8 56 04 AM
Although I'm a tad confused as to why, since afterEach test all the timers are advanced.

rpokorny:
> When I'm running tests I get this at the end:

Weird. Which test command are you running when you see that?

EmilyBermudez88:
> > When I'm running tests I get this at the end:
> 
> Weird. Which test command are you running when you see that?

I got the error with `yarn run test`- I'm trying to go through to find which one is causing the problem with `yarn run test-watch` since it doesn't come up every time as you go from test to test

lib/src/setupTests.ts
```diff
@@ -6,9 +6,26 @@
...
 
 Enzyme.configure({ adapter: new Adapter() });
+
+beforeEach(function() {
+  // because so many of our components rely on NxTooltip which initializes asynchronously. If some tests
+  // use fake timers and some don't it can cause weird inter-test timing bugs
+  jest.useFakeTimers();
+
+  // JSDOM is missing this function. https://github.com/jsdom/jsdom/issues/3002
+  Range.prototype.getBoundingClientRect = jest.fn().mockReturnValue({
```
EmilyBermudez88, line 21: With this now in `setupTests.ts`, we should be able to remove it from `NxBreadcrumb` and `NxStatefulBreadcrumb`

rpokorny: > With this now in setupTests.ts, we should be able to remove it from NxBreadcrumb and NxStatefulBreadcrumb

[b2ed92d](https://github.com/sonatype/sonatype-react-shared-components/pull/893/commits/b2ed92dd6102df2564dc69cb103d08fd89ec40c9)

lib/src/__testutils__/rtlUtils.tsx
```diff
@@ -27,3 +29,18 @@ type RenderElementRetval<P> = (additionalProps?: Partial<P>) => HTMLElement | un
 export function rtlRenderElement<P>(Component: ComponentType<P>, minimalProps: P): RenderElementRetval<P> {
   return pipe(rtlRender(Component, minimalProps), path(['container', 'firstElementChild']));
 }
+
+export async function runTimers() {
```
EmilyBermudez88, line 33: I just remembered Combobox also has these functions defined, which can be removed & imported from `rtlUtils.tsx`

EmilyBermudez88 (approved):
Looks good! - just the one cleanup in Combobox here https://github.com/sonatype/sonatype-react-shared-components/pull/893#discussion_r1067453579

lib/src/__testutils__/rtlUtils.tsx
rpokorny: > I just remembered Combobox also has these functions defined, which can be removed & imported from rtlUtils.tsx

Thanks for catching these. f16e26b7a

PR #896, opened 2023-01-11
Avoid Scroll on Tree Label Mousedown - RSC-1342

rpokorny:
https://issues.sonatype.org/browse/RSC-1342


lib/src/components/NxTree/NxTreeItem.tsx
```diff
@@ -66,14 +66,19 @@ export default function NxTreeItem(props: ItemProps) {
...
   function focusSelf() {
     setFocusState('self');
 
     // If focus is within the tree, we want to move that focus to this item. Otherwise,
     // leave actual focus alone and just make this item the part of the tree that is _focusable_
     const treeRoot = parentKeyNavContext?.getTreeRoot();
     if (treeRoot?.contains(document.activeElement)) {
-      ref.current?.focus();
+      ref.current?.focus({ preventScroll: true });
+      getLabelElement()?.scrollIntoView({ block: 'nearest', inline: 'nearest' });
```
avnvbhatta, line 81: Is this line necessary if we already have `preventScroll: true`

rpokorny: > Is this line necessary if we already have preventScroll: true

Yes.  If we just prevent scroll entirely it gets bad for keyboard navigation - as you arrow-key through the tree scrolling containers fail to follow your progress.  Visually, the thing that _seems_ like it has focus is the tree item label, so this basically makes the automatic scrolling follow that instead of the entire tree item.

avnvbhatta (approved):

PR #897, opened 2023-01-12
Avoid Focus States Leaking into Active States on NxButton

rpokorny:
https://issues.sonatype.org/browse/RSC-1341


EmilyBermudez88 (approved):
Looks good!

PR #898, opened 2023-01-13
Dark Mode Binary Donut - RSC-1005

avnvbhatta:
JIRA: https://issues.sonatype.org/browse/RSC-1005


rpokorny:
I'm a little bit confused by the mockups for this ticket.  The main mockup background color is the tile background color, indigo-20.  The chart borders are also indigo-20, which means they wouldn't show up in their most common use case, on a tile.  But then the mockup has the other section with an indigo-35 background, on which the chart borders of course do show up.  What part of the UI is that supposed to represent?

rpokorny:
This example needs to be adjusted to use the dark mode colors: 
image

avnvbhatta:
> This example needs to be adjusted to use the dark mode colors: image

Added here: 8bc067454e86750f76633f5db9aa81a1786bde9a
The text required some updates based on the theme as well.
Also, the override for custom colors now disappears without the `!important` property.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartCustomExample.scss
```diff
@@ -9,6 +9,6 @@
     display: block;
     height: 45px;
     .nx-binary-donut-chart__arc {
-      stroke: var(--nx-swatch-purple-70);
+      stroke: var(--nx-swatch-purple-70) !important;
```
rpokorny, line 12:
```suggestion
      stroke: var(--nx-swatch-purple-70);

      @include nx-dark-mode-helpers.dark-mode {
        stroke: var(--nx-swatch-purple-70);
      }
```

avnvbhatta: Oh right, forgot about this :sweat_smile:

rpokorny:
We are adding the dark mode visual test snapshots in each ticket, so you should have those.  Also don't forget to get designer approval

rpokorny:
Since you were in favor of the variable-heavy approach to dark mode, would you want to adjust this PR to do it that way?

avnvbhatta:
> Since you were in favor of the variable-heavy approach to dark mode that I demoed in the button PR, would you want to adjust this PR to do it that way?

Oh yes, I was planning on using that approach. Used here: 4c1355b1e29bc744ddfbef584ee8effe8de32f7b

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -89,10 +91,11 @@ const NxBinaryDonutChartPage = () =>
             <NxTable.Cell>No</NxTable.Cell>
             <NxTable.Cell>
               The size of the hole in the donut, as a percentage of the donut's overall size.  The default value is 50.
-              The computation and use of this prop ignores the 1px white borders on the inside and outside of the donut.
-              That is, when the component is rendered at its default 30px width, a value of 50 on this prop corresponds
-              to 14px (<NxCode>((30px - (1px * 2)) * 50%)</NxCode>). That 14px is the radius of the inside of the actual
-              donut value arc and background, with the inner border being 1px further inside.
+              The computation and use of this prop ignores the 1px {themingEnabled ? 'dark blue' : 'white'} borders on
```
rpokorny, line 94: `themingEnabled` doesn't automatically mean dark mode, and I also don't really want our doc text to change between modes.  I'd rephrase this to be accurate in both modes.  Specifically, the border color isn't really important to the point of this sentence, so I'd just remove it.

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
```diff
@@ -172,8 +175,8 @@ const NxBinaryDonutChartPage = () =>
                         codeExamples={nxBinaryDonutChartBackgroundColorExample}
                         liveExample={NxBinaryDonutChartBackgroundColorExample}>
       Examples of <NxCode>NxBinaryDonutChart</NxCode>s on containers that have background colors which match parts
-      of the chart. This example shows that the chart is still discernable in this case due to the white borders
-      on the inside and outside of the donut.
+      of the chart. This example shows that the chart is still discernable in this case due to
+      the {themingEnabled ? 'dark blue' : 'white'} borders on the inside and outside of the donut.
```
rpokorny, line 179:
```suggestion
      the borders on the inside and outside of the donut.
```

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -343,6 +343,15 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-alert-error">
               The color used on the <NxCode>NxAlert</NxCode> error icon and for other equivalent icons and purposes.
             </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-binary-donut-chart-background-fill">
+              The color used on the <NxCode>NxBinaryDonutChart</NxCode> background.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-binary-donut-chart-background-stroke">
```
rpokorny, line 349: This should always match the tile background color, correct? We should probably just use that variable then and not make this additional one.

avnvbhatta: > This should always match the tile background color, correct? We should probably just use that variable then and not make this additional one.

Yup, changed here: 3c853a697ee170cbfbb5eb89bdff4b40691e67db

gallery/src/components/NxBinaryDonutChart/NxBinaryDonutChartPage.tsx
avnvbhatta: > `themingEnabled` doesn't automatically mean dark mode, and I also don't really want our doc text to change between modes. I'd rephrase this to be accurate in both modes. Specifically, the border color isn't really important to the point of this sentence, so I'd just remove it.

Makes sense: 1a8ce6bb84e87d74dd6ff86528ea40986aad356c

rpokorny (approved):

JasonDSwearingen:
+1

PR #899, opened 2023-01-13
Fix Screenreader Reading of Checkbox Checkmark - RSC-1346

rpokorny:
https://issues.sonatype.org/browse/RSC-1346


EmilyBermudez88 (approved):
The fix works for me!

PR #900, opened 2023-01-13
Dark Mode Buttons - RSC-1007

rpokorny:
https://issues.sonatype.org/browse/RSC-1007

In this branch I tried a different approach to dark mode: rather than using lots of `nx-dark-mode-helpers.dark-mode` mixin blocks, I created lots of very specific CSS variables that allow the component styles themselves to mostly ignore the presence of dark mode.  I am very interested in feedback on this approach, as I suspect we should be consistent in how we implement dark mode and should carefully consider the pros and cons.  Here are some pros and cons that I have thought of:

Pros:
1. Cleaner component style files
2. Better for a future possibility of more themes beyond just light and dark mode
3. Colors which get used in multiple places pretty much beg for a variable.  For instance, if a component uses the same color for its border and text, rather than having to override both of those properties separately in the mixin, it feels natural to want to create a variable that controls both of them, and then to just override that variable in the mixin.  Note that this only works with CSS variables and no SCSS variables, as it's reliant on the CSS cascade.  Doing this naively (e.g. locally within the component file) goes against our policy of making all CSS variables a part of our public, documented API.  So I figured why not go all the way and make them part of our public, documented API, and store them in the `_nx-variables.scss` and `_nx-colors.scss` files accordingly?

Cons:
1. Variable usages can get very specific and variable names get get accordingly unwieldy.  Basically everything about what the variable is for must be encoded into its name
2. All of these variables must be documented in the gallery


JasonDSwearingen:
Caught an error in the mockup. The Clicked state of the Tertiary Button should have a background of nx-indigo-25. The mockup has been updated accordingly.

rpokorny:
> Caught an error in the mockup. The Clicked state of the Tertiary Button should have a background of nx-indigo-25. The mockup has been updated accordingly.

Fixed in d595fb2ad.  The focus background was wrong too.  I had set it up to be indigo-25 but had failed to define indigo-25 in the swatch file.

JasonDSwearingen (approved):

EmilyBermudez88:
> Fixed in [d595fb2](https://github.com/sonatype/sonatype-react-shared-components/commit/d595fb2ad31870765933526ac205b3982d5ebeb2). The focus background was wrong too. I had set it up to be indigo-25 but had failed to define indigo-25 in the swatch file.

Looks like this is needed for the focus state of the secondary button background as well.

lib/src/base-styles/_nx-colors.scss
```diff
@@ -26,6 +26,24 @@
...
+  --nx-color-primary-button-text: var(--nx-swatch-white);
+  --nx-color-secondary-button: var(--nx-swatch-blue-40); //border and text
+  --nx-color-secondary-button-background: var(--nx-swatch-white);
+  --nx-color-secondary-button-background-hover: var(--nx-swatch-blue-90);
+  --nx-color-secondary-button-background-active: var(--nx-swatch-blue-95);
+  --nx-color-tertiary-button-border: var(--nx-swatch-grey-50);
+  --nx-color-tertiary-button-text: var(--nx-color-text-stark);
+  --nx-color-tertiary-button-background-focus: transparent;
+  --nx-color-tertiary-button-background-active: var(--nx-swatch-indigo-97);
+  --nx-color-error-button-background: var(--nx-color-alert-error);
```
EmilyBermudez88, line 41: In Sketch, the background of the error button is 40 in light mode, but 35 in dark mode

EmilyBermudez88:
> In this branch I tried a different approach to dark mode: rather than using lots of `nx-dark-mode-helpers.dark-mode` mixin blocks, I created lots of very specific CSS variables that allow the component styles themselves to mostly ignore the presence of dark mode. I am very interested in feedback on this approach, as I suspect we should be consistent in how we implement dark mode and should carefully consider the pros and cons.

I just spent some time looking over how this ticket was done and really like your idea. I like the idea of cleaner files and the re-usability if other themes are applied in the future (hadn't thought of that, great point). I personally don't mind the long and specific variable names. My only concern with it is the amount of variables that could be added to the docs & inside `_nx-variables.scss` and `_nx-colors.scss`. I'm not sure how many would need to be added as we deal with every ticket, but that list could become quite extensive and difficult to parse through.

rpokorny:
> Looks like this is needed for the focus state of the secondary button background as well.

5784970eb

lib/src/base-styles/_nx-colors.scss
rpokorny: > In Sketch, the background of the error button is 40 in light mode, but 35 in dark mode

I wasn't totally sure how/if to relate that to the `--nx-color-alert-error` variable in dark mode like it is in light mode. 
 Dark mode alerts are not yet implemented, but looking at their mockups it seems like they probably won't use the `--nx-color-alert-...` variables as all of the icons are the same dark color.  So for the `--nx-color-error-button-background` dark mode override, I just set it to the swatch color directly.  f93f61b83

EmilyBermudez88: > So for the `--nx-color-error-button-background` dark mode override, I just set it to the swatch color directly. 

Thanks for clarifying, makes sense!

avnvbhatta:
> https://issues.sonatype.org/browse/RSC-1007
> 
> In this branch I tried a different approach to dark mode: rather than using lots of `nx-dark-mode-helpers.dark-mode` mixin blocks, I created lots of very specific CSS variables that allow the component styles themselves to mostly ignore the presence of dark mode. I am very interested in feedback on this approach, as I suspect we should be consistent in how we implement dark mode and should carefully consider the pros and cons. Here are some pros and cons that I have thought of:
> 
> Pros:
> 
> 1. Cleaner component style files
> 2. Better for a future possibility of more themes beyond just light and dark mode
> 3. Colors which get used in multiple places pretty much beg for a variable.  For instance, if a component uses the same color for its border and text, rather than having to override both of those properties separately in the mixin, it feels natural to want to create a variable that controls both of them, and then to just override that variable in the mixin.  Note that this only works with CSS variables and no SCSS variables, as it's reliant on the CSS cascade.  Doing this naively (e.g. locally within the component file) goes against our policy of making all CSS variables a part of our public, documented API.  So I figured why not go all the way and make them part of our public, documented API, and store them in the `_nx-variables.scss` and `_nx-colors.scss` files accordingly?
> 
> Cons:
> 
> 1. Variable usages can get very specific and variable names get get accordingly unwieldy.  Basically everything about what the variable is for must be encoded into its name
> 2. All of these variables must be documented in the gallery

I like this approach better. As mentioned, while this does entail writing up more documentation for the variables used, it results in a lot cleaner files. However, if/when the documented variables get too large, we might need to figure out a way to organize them in the gallery, but that is a concern for later. Overall, I'd say we go through with this approach. 👍

EmilyBermudez88 (approved):

avnvbhatta (approved):

rpokorny:
Superceded by #1038

PR #903, opened 2023-01-17
NxProgressBar RTL conversion - RSC-1349

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1349

### DIFF
--- a/lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
+++ b/lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
@@ -124,10 +124,8 @@ describe('NxProgressBar', function() {
     });
 
     it('is overriden by labelError when it is specified', function() {
-      const el = renderEl()!,
-          elWithError = renderEl({ labelError: 'oops!' })!;
+      const elWithError = renderEl({ labelError: 'oops!' })!;
 
-      expect(el).toHaveTextContent('current progress');
       expect(elWithError).not.toHaveTextContent('current progress');
       expect(elWithError).toHaveTextContent('oops!');
     });
@@ -155,6 +153,7 @@ describe('NxProgressBar', function() {
 
     expect(elWithoutLabel).not.toHaveTextContent('current progress');
     expect(progressBarWithoutLabel).toHaveAccessibleName('current progress');
+
   });
 
   describe('showSteps', function() {
--- a/lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
+++ b/lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
@@ -69,29 +69,27 @@ describe('NxProgressBar', function() {
   });
 
   it('renders a counter when showCounter is not false', function() {
-    const counterSelector = '.nx-progress-bar__counter',
-        defaultEl = renderEl()!,
+    const defaultEl = renderEl()!,
         counterEl = renderEl({ showCounter: true })!,
         noCounterEl = renderEl({ showCounter: false })!;
 
-    expect(defaultEl.querySelector(counterSelector)).toBeInTheDocument();
-    expect(counterEl.querySelector(counterSelector)).toBeInTheDocument();
-    expect(noCounterEl.querySelector(counterSelector)).not.toBeInTheDocument();
+    expect(defaultEl).toHaveTextContent('0%');
+    expect(counterEl).toHaveTextContent('0%');
+    expect(noCounterEl).not.toHaveTextContent('0%');
   });
 
   it('renders a counter in all variants except inline', function() {
-    const counterSelector = '.nx-progress-bar__counter',
-        defaultEl = renderEl({ showCounter: true })!,
+    const defaultEl = renderEl({ showCounter: true })!,
         normalEl = renderEl({ variant: 'normal', showCounter: true })!,
         smallEl = renderEl({ variant: 'small', showCounter: true })!,
         fullEl = renderEl({ variant: 'full', showCounter: true })!,
         inlineEl = renderEl({ variant: 'inline', showCounter: true })!;
 
-    expect(defaultEl.querySelector(counterSelector)).toBeInTheDocument();
-    expect(normalEl.querySelector(counterSelector)).toBeInTheDocument();
-    expect(smallEl.querySelector(counterSelector)).toBeInTheDocument();
-    expect(fullEl.querySelector(counterSelector)).toBeInTheDocument();
-    expect(inlineEl.querySelector(counterSelector)).not.toBeInTheDocument();
+    expect(defaultEl).toHaveTextContent('0%');
+    expect(normalEl).toHaveTextContent('0%');
+    expect(smallEl).toHaveTextContent('0%');
+    expect(fullEl).toHaveTextContent('0%');
+    expect(inlineEl).not.toHaveTextContent('0%');
   });
 
   describe('label', function() {
@@ -164,7 +162,7 @@ describe('NxProgressBar', function() {
 
     it('doesn\'t render a counter when showSteps is true', function() {
       const el = renderEl({ showSteps: true, showCounter: true })!;
-      expect(el.querySelector('.nx-progress-bar__counter')).not.toBeInTheDocument();
+      expect(el).not.toHaveTextContent('0%');
     });
 
     it('adds the number of step elements equal to one less than max when true', function() {
--- a/lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
+++ b/lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
@@ -6,7 +6,7 @@
  */
 
 import React from 'react';
-import { within, render } from '@testing-library/react';
+import { within } from '@testing-library/react';
 import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
 import NxProgressBar, { Props } from '../NxProgressBar';
 
@@ -22,7 +22,7 @@ describe('NxProgressBar', function() {
     expect(renderEl()!.tagName).toBe('LABEL');
   });
 
-  it('renders a <progress> as a child', function() {
+  it('renders a <progress> element as a child', function() {
     const el = renderEl()!,
         progressBar = within(el).getByRole('progressbar');
     expect(progressBar).toBeInTheDocument();
@@ -30,7 +30,7 @@ describe('NxProgressBar', function() {
 
   it('forwards its ref to the <progress> element', function() {
     const ref = React.createRef<HTMLProgressElement>(),
-        el = render(<NxProgressBar {...minimalProps} ref={ref}/>),
+        el = quickRender({ ref: ref } as Partial<Props>),
         progressBar = el.getByRole('progressbar');
 
     expect(ref.current).toBe(progressBar);
@@ -46,7 +46,7 @@ describe('NxProgressBar', function() {
     expect(partialProgressBar).toHaveAttribute('value', '50');
   });
 
-  it('sets the max prop to 100 when unspecified and passes to progress element', function() {
+  it('sets the max prop to 100 when unspecified and passes it to the progress element', function() {
     const defaultEl = quickRender(),
         customMaxEl = quickRender({ max: 200 }),
         progressEl = defaultEl.getByRole('progressbar'),
@@ -69,8 +69,8 @@ describe('NxProgressBar', function() {
   });
 
   it('renders a counter when showCounter is not false', function() {
-    const counterSelector = '.nx-progress-bar__counter';
-    const defaultEl = renderEl()!,
+    const counterSelector = '.nx-progress-bar__counter',
+        defaultEl = renderEl()!,
         counterEl = renderEl({ showCounter: true })!,
         noCounterEl = renderEl({ showCounter: false })!;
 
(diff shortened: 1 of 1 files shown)


lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+
+  it('render a <label> as the top level element', function() {
+    expect(renderEl()!.tagName).toBe('LABEL');
+  });
 
-  describe('showSteps', function() {
-    it('adds an .nx-progress-bar__step-container when true', function() {
-      expect(getMountedComponent()).not.toContainMatchingElement('.nx-progress-bar__step-container');
-      expect(getMountedComponent({ showSteps: true })).toContainMatchingElement('.nx-progress-bar__step-container');
+  it('renders a <progress> element as a child', function() {
```
rpokorny, line 25: Doesn't match what' being tested
```suggestion
  it('renders an element with a role of "progressbar"', function() {
```

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+    const el = renderEl()!,
+        progressBar = within(el).getByRole('progressbar');
+    expect(progressBar).toBeInTheDocument();
+  });
+
+  it('forwards its ref to the <progress> element', function() {
```
rpokorny, line 31:
```suggestion
  it('forwards its ref to the progressbar', function() {
```

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+    const ref = React.createRef<HTMLProgressElement>(),
+        el = quickRender({ ref: ref } as Partial<Props>),
+        progressBar = el.getByRole('progressbar');
+
+    expect(ref.current).toBe(progressBar);
+  });
+
+  it('passes the value prop to the progress element', function() {
```
rpokorny, line 39:
```suggestion
  it('passes the value prop to the progressbar', function() {
```

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+    const noProgressEl = quickRender(),
+        partialProgressEl = quickRender({ value: 50 }),
+        noProgressBar = noProgressEl.getByRole('progressbar'),
+        partialProgressBar = partialProgressEl.getByRole('progressbar');
+
+    expect(noProgressBar).toHaveAttribute('value', '0');
+    expect(partialProgressBar).toHaveAttribute('value', '50');
+  });
+
+  it('sets the max prop to 100 when unspecified and passes it to the progress element', function() {
```
rpokorny, line 49:
```suggestion
  it('sets the max prop to 100 when unspecified and passes it to the progressbar', function() {
```

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+        inProgressWithMaxEl = renderEl({ value: 20, max: 40 });
+
+    expect(noProgressEl).toHaveTextContent('0%'),
+    expect(inProgressEl).toHaveTextContent('50%'),
+    expect(finishedEl).toHaveTextContent('100%');
+    expect(inProgressWithMaxEl).toHaveTextContent('50%');
+  });
+
+  it('renders a counter when showCounter is not false', function() {
+    const counterSelector = '.nx-progress-bar__counter',
```
rpokorny, line 72: Rather than looking up the class name it would be better to check whether the counter text content is present

rpokorny: > Rather than looking up the class name it would be better to check whether the counter text content is present

Same in the next test

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
-          partialBar = getMountedComponent({ showSteps: true, max: 12, value: 3 }),
-          fullBar = getMountedComponent({ showSteps: true, max: 12, value: 12 });
+    it('is overriden by labelError when it is specified', function() {
+      const el = renderEl()!,
+          elWithError = renderEl({ labelError: 'oops!' })!;
 
-      expect(zeroBar.find('.nx-progress-bar__step--below-value')).not.toExist();
-      expect(zeroBar.find('.nx-progress-bar__step--at-value')).not.toExist();
-      expect(zeroBar.find('.nx-progress-bar__step--above-value').length).toBe(11);
+      expect(el).toHaveTextContent('current progress');
```
rpokorny, line 132: Seems unnecessary given the other tests

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
-      expect(partialBar.find('.nx-progress-bar__step--at-value').length).toBe(1);
-      expect(partialBar.find('.nx-progress-bar__step--above-value').length).toBe(8);
+  it('assigns an aria-label when component\'s label text is not rendered', function() {
+    const progressBarWithLabel = quickRender().getByRole('progressbar');
+    const progressBarWithoutLabel = quickRender({ inlineCounter: true }).getByRole('progressbar');
 
-      // ensure correct ordering
-      expect(partialBar.find('.nx-progress-bar__step--at-value ~ .nx-progress-bar__step--below-value')).not.toExist();
-      expect(partialBar.find('.nx-progress-bar__step--above-value ~ .nx-progress-bar__step--at-value')).not.toExist();
+    expect(progressBarWithLabel).not.toHaveAttribute('aria-label', 'current progress');
```
rpokorny, line 148: Related to my first comment, here I think `aria-label` itself is an implementation detail, and what we really want to test is that the accessible name comes out correct one way or the other.  I'd have the `toHaveAccessibleName` matcher for that.

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+      expect(elWithoutSteps.childElementCount).toBe(2);
+      expect(within(elWithoutSteps).queryByRole('presenation')).not.toBeInTheDocument();
+
+      expect(elWithSteps.childElementCount).toBe(3);
+      expect(within(elWithSteps).getByRole('presentation')).toBeInTheDocument();
+    });
+
+    it('doesn\'t render a counter when showSteps is true', function() {
+      const el = renderEl({ showSteps: true, showCounter: true })!;
+      expect(el.querySelector('.nx-progress-bar__counter')).not.toBeInTheDocument();
```
rpokorny, line 167: Check by text content not class name

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
+          elWithSteps = renderEl({ showSteps: true })!;
+
+      expect(elWithoutSteps.childElementCount).toBe(2);
+      expect(within(elWithoutSteps).queryByRole('presenation')).not.toBeInTheDocument();
+
+      expect(elWithSteps.childElementCount).toBe(3);
+      expect(within(elWithSteps).getByRole('presentation')).toBeInTheDocument();
+    });
+
+    it('doesn\'t render a counter when showSteps is true', function() {
```
rpokorny, line 165: Of the three tests in this describe block, this is the only one that doesn't seem strictly stylistic to me.  I'd delete the other two and double check that they're covered by visual tests

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
EmilyBermudez88: > Related to my first comment, here I think `aria-label` itself is an implementation detail, and what we really want to test is that the accessible name comes out correct one way or the other. I'd use the `toHaveAccessibleName` matcher for that.

Added here [b7b7f69](https://github.com/sonatype/sonatype-react-shared-components/pull/903/commits/b7b7f69d02d8dddb436e4c31c6197b5385a4e45a). Upon further testing, I realize that the accessible name changes as well if the counter is visible or not, so it tests the accessible name of: 1. with label + counter, 2. with label no counter, and 3. with counter no label. Not sure if this should be split up into two tests -

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,176 @@
...
-  );
+  const minimalProps = {
+        value: 0,
+        label: 'current progress'
+      },
+      quickRender = rtlRender<Props>(NxProgressBar, minimalProps),
+      renderEl = rtlRenderElement<Props>(NxProgressBar, minimalProps);
+
+  it('render a <label> as the top level element', function() {
+    expect(renderEl()!.tagName).toBe('LABEL');
```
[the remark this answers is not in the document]
EmilyBermudez88: > Implementation detail IMO. I'd check that the accessible name of the `<progress>` element is correct and that that label text is also visible on the screen

Changed the first test to just test for visibility of label text here: [b7b7f69](https://github.com/sonatype/sonatype-react-shared-components/pull/903/commits/b7b7f69d02d8dddb436e4c31c6197b5385a4e45a). However, I'm now not so sure this test is needed. The rendering of the label text is tested in the `label` describe block.. The accessibility mentioned in this comment is dealt with further down in testing (as commented on in https://github.com/sonatype/sonatype-react-shared-components/pull/903#discussion_r1073793561)

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
EmilyBermudez88: > Seems unnecessary given the other tests

removed here: [4f5d6cf](https://github.com/sonatype/sonatype-react-shared-components/pull/903/commits/4f5d6cf8bfa9484b00e6e80a25b39276491dcfab)

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
EmilyBermudez88: > Of the three tests in this describe block, this is the only one that doesn't seem strictly stylistic to me. I'd delete the other two and double check that they're covered by visual tests

The third test is shown in visual tests (number of steps with assigned max value in label both visible). The role in the first test can't be tested visually, but if you don't think it's necessary in testing I can remove that test as well.

rpokorny: > The third test is shown in visual tests (number of steps with assigned max value in label both visible). The role in the first test can't be tested visually, but if you don't think it's necessary in testing I can remove that test as well.

`presentation` is the role for things that aren't relevant to accessibility technologies.  In fact `none` is an exact synonym for it in the role names.  So pretty much by definition anything with the `presentation` role is something we don't need to test in the unit tests.

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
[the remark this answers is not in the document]
rpokorny: > Changed the first test to just test for visibility of label text here: [b7b7f69](https://github.com/sonatype/sonatype-react-shared-components/pull/903/commits/b7b7f69d02d8dddb436e4c31c6197b5385a4e45a). However, I'm now not so sure this test is needed. The rendering of the label text is tested in the label describe block.. The accessibility mentioned in this comment is dealt with further down in testing (as commented on in https://github.com/sonatype/sonatype-react-shared-components/pull/903#discussion_r1073793561)

Yeah it's all covered elsewhere; remove this test

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
```diff
@@ -4,191 +4,182 @@
...
+        progressBarWithLabelAndCounter = within(elWithLabelAndCounter).getByRole('progressbar'),
+        elWithLabelAndNoCounter = renderEl({ showCounter: false })!,
+        progressBarWithLabelAndNoCounter = within(elWithLabelAndNoCounter).getByRole('progressbar'),
+        elWithoutLabel = renderEl({ inlineCounter: true})!,
+        progressBarWithoutLabel = within(elWithoutLabel).getByRole('progressbar');
 
-      expect(zeroBar.find('.nx-progress-bar__step--below-value')).not.toExist();
-      expect(zeroBar.find('.nx-progress-bar__step--at-value')).not.toExist();
-      expect(zeroBar.find('.nx-progress-bar__step--above-value').length).toBe(11);
+    expect(elWithLabelAndCounter).toHaveTextContent('current progress');
```
rpokorny, line 148: I'm a little bit confused about the new setup of this test, for two reasons.
1. didn't we already fully test the text content elsewhere? I wasn't expecting any `toHaveTextContent` assertions here
2. Why aren't we asserting `toHaveAccessibleName` on some of the renders? I would have expected that to be used on all of them.  In particular, since we are no longer testing that there's a `<label>` element involved, there is nothing automatically implying that the text content of a render will be the accessible name of the `progressbar` in that render.  So we should check the accessible names of all of the renders, regardless of whether they have visible text.

EmilyBermudez88: > I'm a little bit confused about the new setup of this test, for two reasons.
> 
> 1. didn't we already fully test the text content elsewhere? I wasn't expecting any `toHaveTextContent` assertions here
> 2. Why aren't we asserting `toHaveAccessibleName` on some of the renders? I would have expected that to be used on all of them.  In particular, since we are no longer testing that there's a `<label>` element involved, there is nothing automatically implying that the text content of a render will be the accessible name of the `progressbar` in that render.  So we should check the accessible names of all of the renders, regardless of whether they have visible text.

Makes sense re : 1, I will remove those assertions. 
I'm struggling with #2 though. I have tried asserting each of the 3 different renders have an accessible name (`expect(elWithLabelAndCounter).toHaveAccessibleName('current progress')`), which all come back as an empty string instead of 'current progress'.
 I tried to find the label element by `getByLabelText`, and can't seem to select the label that way either (even though I should be able to according to https://testing-library.com/docs/queries/bylabeltext/). Am I perhaps misunderstanding something about how the <label> element works for NxProgressBar?

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
EmilyBermudez88: > `presentation` is the role for things that aren't relevant to accessibility technologies. In fact `none` is an exact synonym for it in the role names. So pretty much by definition anything with the `presentation` role is something we don't need to test in the unit tests.

Thanks for clarifying there. So I've removed the two tests ([7258a9d](https://github.com/sonatype/sonatype-react-shared-components/pull/903/commits/7258a9d0e446193811bd5c98ce8978372037026e)) , and removed the `describe` block since it would just be holding one test. I did however create a new `describe` block for the counter to group the 4 tests together.

lib/src/components/NxProgressBar/__tests__/NxProgressBar.test.tsx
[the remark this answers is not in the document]
EmilyBermudez88: > Yeah it's all covered elsewhere; remove this test

removed here: [54af7f7](https://github.com/sonatype/sonatype-react-shared-components/pull/903/commits/54af7f742fd8bc2e4e0a7986e8537e9ba0013f6a)

rpokorny (approved):

PR #904, opened 2023-01-18
NxCollapsibleItems RTL Tests - RSC-1235

rpokorny:
https://issues.sonatype.org/browse/RSC-1235

### DIFF
--- a/lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
+++ b/lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
@@ -223,6 +223,13 @@ describe('NxCollapsibleItems', function() {
         expect(view.getByRole('listitem')).toBe(view.container.firstChild);
       });
 
+      it('renders a top-level element with the specified role', function() {
+        const view = quickRender({ role: 'menuitem' });
+
+        expect(view.queryByRole('listitem')).not.toBeInTheDocument();
+        expect(view.getByRole('menuitem')).toBe(view.container.firstChild);
+      });
+
       it('sets the specified classnames', function() {
         const el = renderEl({ className: 'foo' }),
             defaultEl = renderEl()!;
@@ -257,6 +264,7 @@ describe('NxCollapsibleItems', function() {
       const minimalProps = {
             children: <div id="test-id" className="bar" lang="en">foo</div>
           },
+          quickRender = rtlRender(NxCollapsibleItems.Child, minimalProps),
           renderEl = rtlRenderElement(NxCollapsibleItems.Child, minimalProps);
 
       it('renders an element like the children', function() {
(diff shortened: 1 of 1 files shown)


lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
```diff
@@ -32,262 +29,261 @@ describe('NxCollapsibleItems', function() {
...
-        <NxCollapsibleItems.Child id="test-id" className="bar" lang="en">foo</NxCollapsibleItems.Child>),
-          div = component.children().children();
+      const minimalProps = { children: 'foo' },
+          quickRender = rtlRender(NxCollapsibleItems.Child, minimalProps),
+          renderEl = rtlRenderElement(NxCollapsibleItems.Child, minimalProps);
 
-      it('renders a div with the listitem role and the nx-collapsible-items__child class', function() {
-        expect(component.hostNodes()).not.toExist();
-        expect(component.children().hostNodes()).not.toExist();
+      it('renders a top-level element with the listitem role', function() {
```
EmilyBermudez88, line 220: should you also test that the role can be overridden if another role is provided, like in line 182?

lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
```diff
@@ -32,262 +29,261 @@ describe('NxCollapsibleItems', function() {
...
 
-      expect(component.find('.nx-collapsible-items__trigger')).toHaveProp('aria-controls', childrenElId);
+      const childrenEl = document.getElementById(childrenElId)!,
+          childEl = screen.getByTestId('child');
+
+      expect(view.container).toContainElement(childrenEl);
+      expect(childrenEl).toContainElement(childEl);
     });
 
     it('sets aria-expanded iff both the isOpen prop is true and there are children', function() {
```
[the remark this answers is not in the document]
rpokorny: iff means "if and only if"

lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
```diff
@@ -32,262 +29,261 @@ describe('NxCollapsibleItems', function() {
...
+      await runTimers();
 
-      expect(tooltip).toExist();
-      expect(tooltip).toHaveProp('title', 'tip');
-      expect(tooltip).toHaveProp('placement', 'right');
+      const tooltip = screen.getByRole('tooltip');
 
-      expect(trigger).toExist();
+      expect(tooltip).toHaveTextContent('tip');
+      expect(tooltip.querySelector('.foo')).toBeInTheDocument();
```
cindyko226, line 150: Is this to test if the classname is set correctly?

rpokorny: > Is this to test if the classname is set correctly?

Yes but we don't have control over exactly which element the class gets set on, it's up to how MUI builds the tooltip.  So I can't assert that the class is on the tooltip, but I can assert that it's on some descendant of the tooltip.

cindyko226: Make sense, thanks.

lib/src/components/NxCollapsibleItems/__tests__/NxCollapsibleItems.test.tsx
rpokorny: > should you also test (both here and when the child is an element) that the role can be overridden if another role is provided, like in line 182?

There are actually a lot of nuances around the roles logic here that all deserve to be tested. Good catch. 0c825d96a

EmilyBermudez88 (approved):
LGTM!

PR #906, opened 2023-01-18
RTL Conversion - NxStatefulAlert - RSC-1343

avnvbhatta:
JIRA: https://issues.sonatype.org/browse/RSC-1343

### DIFF
--- a/lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
+++ b/lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
@@ -7,9 +7,7 @@
 import React, { ComponentType } from 'react';
 import { faBiohazard } from '@fortawesome/free-solid-svg-icons';
 
-import { userEvent } from '../../../../__testutils__/rtlUtils';
-
-import { rtlRender, rtlRenderElement } from '../../../../__testutils__/rtlUtils';
+import { rtlRender, rtlRenderElement, userEvent } from '../../../../__testutils__/rtlUtils';
 import NxStatefulAlert,
 { NxStatefulErrorAlert,
   NxStatefulWarningAlert,
--- a/lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
+++ b/lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
@@ -55,7 +55,7 @@ describe('NxStatefulAlert', function() {
       expect(el).toHaveAttribute('title', 'bar');
     });
 
-    it('renders a Close button even though the onClose prop is not specified', async function() {
+    it('renders a Close button that removes the alert from the DOM when clicked', async function() {
       const user = userEvent.setup(),
           view = quickRender(),
           closeBtn = view.getByRole('button', { name: 'Close' });
--- a/lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
+++ b/lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
@@ -55,14 +55,16 @@ describe('NxStatefulAlert', function() {
       expect(el).toHaveAttribute('title', 'bar');
     });
 
-    it('renders a Close button', async function() {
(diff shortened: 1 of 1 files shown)


rpokorny:
This seems like it duplicates the `NxAlert` tests which is necessary as we discussed, but I don't see anything testing the stateful behavior or the action of the close button.

avnvbhatta:
> This seems like it duplicates the `NxAlert` tests which is necessary as we discussed, but I don't see anything testing the stateful behavior or the action of the close button.

I assumed that by modifying [this part of the test](https://github.com/sonatype/sonatype-react-shared-components/pull/906/files#diff-7c044be1260d9d2488ddb8dcba0ed2ab360589f20c9203b4cc1a131c9dd06d07R58-R67) so that the alerts render the close button without specifying the `onClose` prop and clicking on the button would remove it from the DOM would suffice. 

What else do you think needs to be added?

rpokorny:
> I assumed that by modifying [this part of the test](https://github.com/sonatype/sonatype-react-shared-components/pull/906/files#diff-7c044be1260d9d2488ddb8dcba0ed2ab360589f20c9203b4cc1a131c9dd06d07R58-R67) so that the alerts render the close button without specifying the onClose prop and clicking on the button would remove it from the DOM would suffice.

What else do you think needs to be added?

Oh OK, that's closer than I had noticed.  I think that just needs to be adjusted so that it asserts the disappearance of the whole modal, not just the close button, and the name of the test could be improved.

avnvbhatta:
> > I assumed that by modifying [this part of the test](https://github.com/sonatype/sonatype-react-shared-components/pull/906/files#diff-7c044be1260d9d2488ddb8dcba0ed2ab360589f20c9203b4cc1a131c9dd06d07R58-R67) so that the alerts render the close button without specifying the onClose prop and clicking on the button would remove it from the DOM would suffice.
> > What else do you think needs to be added?
> 
> Oh OK, that's closer than I had noticed. I think that just needs to be adjusted so that it asserts the disappearance of the whole alert, not just the close button, and the name of the test could be improved.

Updated here: dcb51074b5c1d406a71033ab33b4c5fbb429cc2c

lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
```diff
@@ -4,56 +4,158 @@
...
+    it('renders a Close button even though the onClose prop is not specified', async function() {
+      const user = userEvent.setup(),
+          view = quickRender(),
+          closeBtn = view.getByRole('button', { name: 'Close' });
+
+      expect(view.container.querySelector('.nx-alert')).toBeInTheDocument();
+      expect(closeBtn).toBeInTheDocument();
+
+      await user.click(closeBtn);
+      expect(view.container.querySelector('.nx-alert')).not.toBeInTheDocument();
```
rpokorny, line 67: The `nx-alert` class is an implementation detail. IMO a better approach would be to check that the container element is empty.

lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
```diff
@@ -4,56 +4,158 @@
...
+      expect(renderEl()).toHaveAttribute('aria-atomic', 'true');
+    });
+
+    it('passes any other props to the top-level element', function() {
+      const el = renderEl({ id: 'foo', title: 'bar' });
+      expect(el).toHaveAttribute('id', 'foo');
+      expect(el).toHaveAttribute('title', 'bar');
+    });
+
+    it('renders a Close button even though the onClose prop is not specified', async function() {
```
rpokorny, line 58:
```suggestion
    it('renders a Close button that removes the alert from the DOM when clicked', async function() {
```

lib/src/components/NxAlert/stateful/__tests__/NxStatefulAlert.test.tsx
avnvbhatta: > The `nx-alert` class is an implementation detail. IMO a better approach would be to check that the container element is empty.

Changed here: c8e92ba8a07092dab3b8bb323be2bb8adb5b656b

rpokorny (approved):

PR #909, opened 2023-01-19
Convert NxStatefulDropdown, NxDropdown, AbstractDropdown, NxDropdownMenu tests to RTL - RSC-1230

entroform:
https://issues.sonatype.org/browse/RSC-1230


lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-    expect(button).toHaveClassName('.nx-dropdown__toggle');
-    expect(button).toHaveProp('variant', 'tertiary');
-    expect(button).toHaveProp('type', 'button');
-    expect(button.childAt(0)).toContainReact(<span className="nx-dropdown__toggle-label">dropdown</span>);
-    expect(icon).toHaveProp('icon', faCaretDown);
-  });
+  const quickRender = rtlRender(NxDropdown, minimalProps);
+  const renderEl = rtlRenderElement(NxDropdown, minimalProps);
 
   it('renders the button according to the supplied variant', function() {
```
rpokorny, line 26: Styling classes are implementation details.  The external behavior to test would be the actual styling, tested in visual tests.  If there aren't yet examples and visual tests of each of these variants, they should be added and this test removed.

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-    expect(component.find(NxButton)).toHaveProp('variant', 'error');
+    rerender(<NxDropdown {...minimalProps} variant="error" />);
+    expect(screen.getByRole('button')).toHaveClass('nx-btn--error');
   });
 
   it('correctly renders the menu based on isOpen prop', function() {
-    let component = getMountedComponent({ isOpen: true });
-    expect(component.find(NxDropdownMenu)).toExist();
+    const { container, rerender } = quickRender({ ...minimalProps, isOpen: true });
+    const menu = container.querySelector('.nx-dropdown-menu');
```
rpokorny, line 43: If `nx-dropdown-menu` had proper aria roles, I'd say you should use those.  I'm assuming the reason you're using the classname is because it doesn't have any roles.  A comment should be added pointing that out and referring to the JIRA ticket where that should be fixed (create the ticket if we don't already have one).

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-    component = getMountedComponent({ variant: 'error' });
-    expect(component.find(NxButton)).toHaveProp('variant', 'error');
+    rerender(<NxDropdown {...minimalProps} variant="error" />);
+    expect(screen.getByRole('button')).toHaveClass('nx-btn--error');
   });
 
   it('correctly renders the menu based on isOpen prop', function() {
-    let component = getMountedComponent({ isOpen: true });
-    expect(component.find(NxDropdownMenu)).toExist();
+    const { container, rerender } = quickRender({ ...minimalProps, isOpen: true });
```
rpokorny, line 42: One of the main features of `quickRender` is that it automatically includes the `minimalProps`; you only have to specify the extras
```suggestion
    const { container, rerender } = quickRender({ isOpen: true });
```

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-  it('disables the button (and the toggle fn) when the disabled prop is supplied', function() {
-    const toggleFn = jest.fn(),
-        component = getMountedComponent({ onToggleCollapse: toggleFn, disabled: true }),
-        button = component.find(NxButton);
+  it('disables the button (and the toggle fn) when the disabled prop is supplied', async function() {
+    const user = userEvent.setup();
+    const toggleFn = jest.fn();
+    renderEl({ onToggleCollapse: toggleFn, disabled: true });
+    const button = screen.getByRole('button');
+    expect(button).toHaveClass('disabled');
```
rpokorny, line 66: The classname is an impl detail and doesn't fully convey the a11y metadata we want to test
```suggestion
    expect(button).toBeDisabled();
 ```

File: lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx [hunk 5]
```diff
@@ -5,357 +5,301 @@
...
-  it('uses VDOM as label if supplied', function() {
-    const label = <NxFontAwesomeIcon icon={faTrash} />,
-        component = getMountedComponent({ label });
-    const button = component.find(NxButton);
+  it('renders react element as label if supplied', function() {
+    const label = <NxFontAwesomeIcon icon={faTrash} />;
+    const { container } = quickRender({ label });
 
-    expect(button.childAt(0)).toContainReact(label);
+    expect(container.querySelector('.nx-icon.fa-trash')).toBeInTheDocument();
```
rpokorny, line 77: These classes are impl details of `NxFontAwesomeIcon`.  A better approach would be something like
```
const view = quickRender({ label: <span data-testid="foo" /> }),
    toggle = view.getByRole('button');
    
expect(within(toggle).getByTestId('foo')).toBeInTheDocument();
```

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
 
-    expect(menu.childAt(1)).toMatchSelector(NxOverflowTooltip);
-    expect(menu.childAt(1).find('a')).toMatchElement(<a id="link2" key="2" className="nx-dropdown-link">Link2</a>);
+    const { queryAllByTestId } = quickRender({ children, isOpen: true });
 
-    expect(menu.childAt(2)).toMatchSelector(NxOverflowTooltip);
-    expect(menu.childAt(2).find('button')).toMatchElement(
-      <button id="link3" key="3" className="nx-dropdown-button">Link3</button>
-    );
+    const menuChildren = queryAllByTestId('menu-child');
```
rpokorny, line 96:
```suggestion
    const menuChildren = getAllByTestId('menu-child');
```

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
+  it('renders the children within the NxDropdownMenu in the specified order', async function() {
     const children = [
-      <a id="link1" key="1">Link1</a>,
-      <a id="link2" className="nx-dropdown-link" key="2">Link2</a>,
-      <button id="link3" className="nx-dropdown-button" key="3">Link3</button>,
-      <button id="link4" className="nx-dropdown-right-button" key="4">Link4</button>
+      <a data-testid="menu-child" id="link1" key="1">Link1</a>,
+      <a data-testid="menu-child" id="link2" className="nx-dropdown-link" key="2">Link2</a>,
+      <button data-testid="menu-child" id="link3" className="nx-dropdown-button" key="3">Link3</button>,
+      <button data-testid="menu-child" id="link4" className="nx-dropdown-right-button" key="4">Link4</button>
```
rpokorny, line 91: The ids and classes aren't really relevant to this test; they can be removed

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-    );
+    const menuChildren = queryAllByTestId('menu-child');
 
-    expect(menu.childAt(3)).toMatchElement(
-      <button id="link4" key="4" className="nx-dropdown-right-button">Link4</button>
-    );
+    expect(menuChildren[0] as HTMLElement).toHaveTextContent('Link1');
+    expect(menuChildren[1] as HTMLElement).toHaveTextContent('Link2');
+    expect(menuChildren[2] as HTMLElement).toHaveTextContent('Link3');
+    expect(menuChildren[3] as HTMLElement).toHaveTextContent('Link4');
```
rpokorny, line 101: These casts don't seem necessary

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
     expect(onToggleCollapse).not.toHaveBeenCalled();
 
-    act(() => {
-      document.dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
-    });
-    component!.update();
+    await user.click(document.body);
+    rerender(<NxDropdown {...minimalProps} onToggleCollapse={onToggleCollapse} isOpen={true} />);
```
rpokorny, line 112: What's the purpose of this `rerender`?

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
+
     expect(onToggleCollapse).toHaveBeenCalled();
   });
 
-  it('does not call onToggleCollapse if a click happens anywhere when the dropdown is closed', function() {
-    const onToggleCollapse = jest.fn(),
-        component = getMountedComponent({ onToggleCollapse }, { attachTo: container });
+  it('does not call onToggleCollapse if a click happens anywhere when the dropdown is closed', async function() {
```
rpokorny, line 117:
```suggestion
  it('does not call onToggleCollapse if a click happens anywhere aside from the toggle button when the dropdown is closed', async function() {
```

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-    expect(button).toHaveClassName('disabled');
-    button.simulate('click');
     expect(toggleFn).not.toHaveBeenCalled();
   });
 
-  it('uses VDOM as label if supplied', function() {
-    const label = <NxFontAwesomeIcon icon={faTrash} />,
-        component = getMountedComponent({ label });
-    const button = component.find(NxButton);
+  it('renders react element as label if supplied', function() {
```
rpokorny, line 73: We should probably also test when `label` is a plain string

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-    act(() => {
-      document.dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
-    });
-    component!.update();
+    await user.click(document.body);
+    rerender(<NxDropdown {...minimalProps} onToggleCollapse={onToggleCollapse} isOpen={true} />);
+
     expect(onToggleCollapse).toHaveBeenCalled();
```
rpokorny, line 114: Might be a good idea to check the number of calls here too, since the toggle code is known to be kind of touchy.
```suggestion
    expect(onToggleCollapse).toHaveBeenCalledTimes(1);
```

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
   });
 
   it('calls onToggleCollapse if ESC is pressed within the component while the dropdown is open', function() {
-    const onToggleCollapse = jest.fn(),
-        component = getMountedComponent({ onToggleCollapse, isOpen: true });
-
-    component.simulate('keyDown', { key: 'Escape', preventDefault: jest.fn() });
+    const onToggleCollapse = jest.fn();
+    const component = renderEl({ onToggleCollapse, isOpen: true })!;
+    fireEvent.keyDown(component, { key: 'Escape' });
```
rpokorny, line 161: Is there a reason `userEvent` isn't being used here?

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
+          isOpen: true,
+          onCloseKeyDown: e => e.preventDefault()
+        };
+
+        const { rerender } = quickRender(props);
+
+        const button = screen.getByRole('button');
+        fireEvent.keyDown(button, { key: 'Escape', bubbles: true });
+
+        rerender(<NxDropdown {...minimalProps} {...props} />);
```
rpokorny, line 213: Doesn't seem necessary

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
     expect(onToggleCollapse).not.toHaveBeenCalled();
 
-    act(() => {
-      document.dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
-    });
-    component!.update();
+    await user.click(document.body);
+    rerender(<NxDropdown {...minimalProps} {...props} />);
```
rpokorny, line 235: These rerenders also seem unnecessary. Are you under the impression that this is necessary to potentially call `onToggleCollapse`? I don't think it is

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
+    const component = renderEl({ isOpen: true, onCloseClick })!;
 
     expect(evt).toBeUndefined();
 
-    act(() => {
-      component.getDOMNode().dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
-    });
+    fireEvent.click(component);
```
rpokorny, line 267: Why not `userEvent`?

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
```diff
@@ -5,357 +5,301 @@
...
-        toggleBtn = component.find('button.nx-dropdown__toggle').getDOMNode();
+    const props: Partial<Props> = {
+      children: <button className="nx-dropdown-button" data-testid="dropdown-button">Foo</button>,
+      isOpen: true
+    };
+
+    const { rerender } = quickRender(props);
+
+    const menuBtn = screen.getByTestId('dropdown-button');
+    const toggleBtn = screen.getByRole('button', { name: 'dropdown-toggle'});
```
rpokorny, line 288:
```suggestion
    const toggleBtn = screen.getByRole('button', { name: 'dropdown-toggle' });
```

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,132 @@
...
+    await user.click(toggleButton);
 
-    act(() => {
-      element!.find(NxButton).getDOMNode().dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
+    let menu = container.querySelector('.nx-dropdown-menu');
+    expect(menu).toBeInTheDocument();
+
+    await act(async () => {
```
rpokorny, line 124: Another probably unnecessary `act`

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,132 @@
...
-    });
 
-    act(() => {
-      element.find(NxButton).getDOMNode().dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
-    });
-    element!.update();
-    expect(element!.find('.nx-dropdown-menu')).toExist();
+    const { container } = render(
```
rpokorny, line 111: Could use `quickRender`

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,132 @@
...
-import * as enzymeUtils from '../../../../__testutils__/enzymeUtils';
 
-import NxButton from '../../../NxButton/NxButton';
-import NxDropdown from '../../NxDropdown';
+import { act, render, screen, fireEvent } from '@testing-library/react';
+import { rtlRender, userEvent } from '../../../../__testutils__/rtlUtils';
+
 import NxStatefulDropdown, { Props } from '../NxStatefulDropdown';
 
 describe('NxStatefulDropdown', () => {
```
rpokorny, line 14: Since RTL can't test that `NxStatefulDropdown` depends on `NxDropdown`, these tests shouldn't assume that `NxStatefulDropdown` also implements all of the common behavior that it inherits from `NxDropdown`.  That is, ALL of `NxStatefulDropdown`'s behaviors must be tested in this file, including the behaviors that duplicate what's tested on `NxDropdown` directly.  This is how it is for our stateful components in general - their test files must duplicate a large amount of the tests from the corresponding stateless component.

lib/src/components/NxDropdownMenu/__tests__/NxDropdownMenu.test.tsx
```diff
@@ -5,25 +5,19 @@
...
-import { mount } from 'enzyme';
-import 'jest-enzyme';
 
-import { getShallowComponent } from '../../../__testutils__/enzymeUtils';
+import { render, within } from '@testing-library/react';
+import { rtlRenderElement } from '../../../__testutils__/rtlUtils';
 
 import NxDropdownMenu, { Props } from '../NxDropdownMenu';
 
 describe('NxDropdownMenu', function() {
```
rpokorny, line 14: Despite its name beginning with `Nx` (it probably shouldn't) this component is not publicly exposed and so shouldn't have tests.  The external behavior that is relevant to the two tests here is already covered in `NxDropdown`'s tests (e.g. the focus moving to the toggle when the dropdown is closed)

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > These classes are impl details of `NxFontAwesomeIcon`. A better approach would be something like
> 
> ```
> const view = quickRender({ label: <span data-testid="foo" /> }),
>     toggle = view.getByRole('button');
>     
> expect(within(toggle).getByTestId('foo')).toBeInTheDocument();
> ```
> 
> This also tests that the label ends up within the toggle button as opposed to somewhere else

Thanks, I updated it here: d227d8a9d808f9062c6dfd90d74a02982194c268

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > The ids and classes aren't really relevant to this test; they can be removed

Cleaned up here: b02bf2ba530f04db03a59c4b6c75c64294cbb19d

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > We should probably also test when `label` is a plain string

Added that test here: b02bf2ba530f04db03a59c4b6c75c64294cbb19d

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > Is there a reason `userEvent` isn't being used here?

Oh because the event need to be fired within the component.
Should I focus on the toggle and use userEvent on that?

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > These rerenders also seem unnecessary. Are you under the impression that this is necessary to potentially call `onToggleCollapse`? I don't think it is. It's not like in enzyme where we had to randomly call `component.update()` for reasons we never understood

Thanks for that clarification: I've removed it here: 875cc55fa5870d28a090056472f32c6d5c2f90bb

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > Doesn't seem necessary

Removed it here: 875cc55fa5870d28a090056472f32c6d5c2f90bb

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > What's the purpose of this `rerender`?

Removed it here: 875cc55fa5870d28a090056472f32c6d5c2f90bb

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
rpokorny: > Should I focus on the toggle and use userEvent on that?

I think that'd be preferable yeah

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > Could use `quickRender`

Fixed here: 239f969694ce3d65792056459da9a09afbd04d1b

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > Another probably unnecessary `act`

Removed: bd3d74ba6ed97d889feb074bff4ffc6ea724460b

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > > Should I focus on the toggle and use userEvent on that?
> 
> I think that'd be preferable yeah

Updated here: c79a3f815f7ed7b13d789a32448d3b5332fe1974

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > Why not `userEvent`?

Updated here: c79a3f815f7ed7b13d789a32448d3b5332fe1974

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > Styling classes are implementation details. The external behavior to test would be the actual styling, tested in visual tests. If there aren't yet examples and visual tests of each of these variants, they should be added and this test removed.

Removed variants test here: 5762a3468ac606409044e5c26218ec34dacf63a3

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > If `nx-dropdown-menu` had proper aria roles, I'd say you should use those. I'm assuming the reason you're using the classname is because it doesn't have any roles. A comment should be added pointing that out and referring to the JIRA ticket where that should be fixed (create the ticket if we don't already have one).

Added comment here: 9389efa4ca4eb23eabe01dab16a11cf70bc0f225

lib/src/components/NxDropdown/__tests__/AbstractDropdown.test.tsx
```diff
@@ -5,79 +5,68 @@
...
+
+import { screen, render } from '@testing-library/react';
+import { rtlRenderElement, userEvent } from '../../../__testutils__/rtlUtils';
 
 import AbstractDropdown, {
   AbstractDropdownProps,
   AbstractDropdownRenderToggleElement
 } from '../AbstractDropdown';
 
 describe('AbstractDropdown', () => {
```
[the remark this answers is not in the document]
entroform: > IMO `AbstractDropdown` shouldn't have unit tests at all anymore. It is an implementation detail of the components which use it. We wouldn't want to have to adjust our tests as a result of some internal rearrangement of which code is in `AbstractDropdown` vs `NxDropdown` is external behavior was unchanged.

Removed it here: 17c3f9e7d81dc688c3b765ba5c9bd7736913580a

lib/src/components/NxDropdownMenu/__tests__/NxDropdownMenu.test.tsx
entroform: > Despite its name beginning with `Nx` (it probably shouldn't) this component is not publicly exposed and so shouldn't have tests. The external behavior that is relevant to the two tests here is already covered in `NxDropdown`'s tests (e.g. the focus moving to the toggle when the dropdown is closed)

Removed it here: 430b2eb649c2cae5205dfa78e920f94ff10049dc

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > The classname is an impl detail and doesn't fully convey the a11y metadata we want to test

Hmm this failed for me because it doesn't have the disabled attribute.

rpokorny: > Hmm this failed for me because it doesn't have the disabled attribute.

Then that attribute should be added (or at least `aria-disabled` if the actual `disabled` attribute is a problem for some reason)

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > Since RTL can't test that `NxStatefulDropdown` depends on `NxDropdown`, these tests shouldn't assume that `NxStatefulDropdown` also implements all of the common behavior that it inherits from `NxDropdown`. That is, ALL of `NxStatefulDropdown`'s behaviors must be tested in this file, including the behaviors that duplicate what's tested on `NxDropdown` directly. This is how it is for our stateful components in general - their test files must duplicate a large amount of the tests from the corresponding stateless component.

Updated stateful test here: 6173c18d188592b3141d2c02f7f670e4d01b4715

lib/src/components/NxDropdown/__tests__/NxDropdown.test.tsx
entroform: > > Hmm this failed for me because it doesn't have the disabled attribute.
> 
> Then that attribute should be added (or at least `aria-disabled` if the actual `disabled` attribute is a problem for some reason)

Added that here: 260fe8b412cb9a7cd78cdc45abb36d00685ebfc0

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,311 @@
...
-    expect(component).toMatchSelector(NxDropdown);
-    expect(statelessDropdown).toHaveProp('label', 'dropdown');
-    expect(statelessDropdown).toHaveProp('variant', 'tertiary');
-    expect(statelessDropdown).toHaveProp('disabled', true);
-    expect(statelessDropdown).toHaveProp('toggleTooltip', 'a tooltip');
-    expect(statelessDropdown).toHaveProp('children', (<a>Hello</a>));
+    const dropdown = container.querySelector('.nx-dropdown');
+    const toggleButton = screen.getByRole('button', { name: 'dropdown-toggle'});
+
+    expect(dropdown).toHaveClass('extra-class');
```
rpokorny, line 33: Extra classes should be tested with this familiar pattern: https://github.com/sonatype/sonatype-react-shared-components/pull/909/files#diff-8db925aaf03e68533aaa4b6fa317b86cc22eb19cf52f14631de4c54e5851039bR36

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,311 @@
...
-    const component = getShallowComponent();
-    expect(component).toMatchSelector(NxDropdown);
-    expect(component).toHaveProp('label', 'dropdown');
-  });
+  const renderEl = rtlRenderElement<Props>(NxStatefulDropdown, minimalProps);
+  const quickRender = rtlRender<Props>(NxStatefulDropdown, minimalProps);
 
-  it('renders an NxDropdown component passing on the given props', function() {
-    const component = getShallowComponent({
+  it('renders an NxStatefulDropdown toggle element correctly with the given props', function() {
```
rpokorny, line 23: This seems like it's squashing together things that unrelated and tested separately on `NxDropdown` itself.  I'd want to see the tests here follow a similar pattern.  Don't follow the patterns present in the old `NxStatefulDropdown` tests, as those were designed around ensuring the that proper props are passed on to the child react component which is very different from what we're trying to test now.

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,311 @@
...
+
+    expect(dropdown).toHaveClass('extra-class');
+    expect(toggleButton).toHaveClass('nx-btn--tertiary');
+    expect(toggleButton).toHaveClass('disabled');
+    expect(toggleButton).toHaveAttribute('aria-disabled');
   });
 
-  it('opens the dropdown when the toggle is clicked', function() {
-    const mounted = getMountedComponent(undefined, { attachTo: container });
+  it('toggles the menu open and close when toggle is clicked', async function() {
```
rpokorny, line 39:
```suggestion
  it('toggles the menu open and closed when toggle is clicked', async function() {
```

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,311 @@
...
+
+    await user.click(toggleButton);
+
+    menu = container.querySelector('.nx-dropdown-menu');
+    expect(menu).not.toBeInTheDocument();
   });
 
-  it('closes the dropdown if the menu is open and the toggle is clicked', function() {
-    const mounted = getMountedComponent(undefined, { attachTo: container });
+  it('closes the dropdown when the Escape key is pressed on this component', async function() {
```
rpokorny, line 59: Should probably have a test that pressing ESC on an already-closed dropdown doesn't open it

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
```diff
@@ -5,192 +5,311 @@
...
-    const mounted = getMountedComponent(undefined, { attachTo: container });
+  it('closes the dropdown when an outside click happens', async function() {
+    const user = userEvent.setup();
 
-    act(() => {
-      mounted.find(NxButton).getDOMNode().dispatchEvent(new MouseEvent('click', { bubbles: true }));
-    });
-    mounted.update();
-    expect(mounted.find('.nx-dropdown-menu')).toExist();
+    const { container } = render(
```
rpokorny, line 80: This is fine but another option would be to still use `quickRender` or `renderEl` for the dropdown but then also have a separate `render()` statement for the button

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > Extra classes should be tested with this familiar pattern: https://github.com/sonatype/sonatype-react-shared-components/pull/909/files#diff-8db925aaf03e68533aaa4b6fa317b86cc22eb19cf52f14631de4c54e5851039bR36

Cleaned it up here: 041a235e97a4bd79499e548d3229ff2fbfaf70db

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > This is fine but another option would be to still use quickRender or renderEl for the dropdown but then also have a separate render() statement for the button

Thanks didn't know you could do that: a1fb4380312a886fdb4f7b946d7b797fbd73df25

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > Should probably have a test that pressing ESC on an already-closed dropdown doesn't open it

I added that test here: a88942639a230546e9e35bdb1d808c316a173f01

I've revised it a little here: ca629a0bcdd4105fd7e590a19f980ad231ebe4f8
Let me know what you think.

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
entroform: > This seems like it's squashing together things that are unrelated and tested separately on `NxDropdown` itself. I'd want to see the tests here follow a similar pattern. Don't follow the patterns present in the old `NxStatefulDropdown` tests, as those were designed around ensuring the that proper props are passed on to the child react component which is very different from what we're trying to test now.

I cleaned that up here: 041a235e97a4bd79499e548d3229ff2fbfaf70db

but realized there was already a disabled test below so I removed it and cleaned it up here:
5a150c1b2afd48ee1e108923be99f720bfef7a24

lib/src/components/NxDropdown/stateful/__tests__/NxStatefulDropdown.test.tsx
rpokorny: > I added that test here: https://github.com/sonatype/sonatype-react-shared-components/commit/a88942639a230546e9e35bdb1d808c316a173f01
>
> I've revised it a little here: https://github.com/sonatype/sonatype-react-shared-components/commit/ca629a0bcdd4105fd7e590a19f980ad231ebe4f8
Let me know what you think.

I think your first impl was missing an `await`.  But why'd you switch to `fireEvent` in the second attempt?

entroform: > > I added that test here: [a889426](https://github.com/sonatype/sonatype-react-shared-components/commit/a88942639a230546e9e35bdb1d808c316a173f01)
> > I've revised it a little here: [ca629a0](https://github.com/sonatype/sonatype-react-shared-components/commit/ca629a0bcdd4105fd7e590a19f980ad231ebe4f8)
> > Let me know what you think.
> 
> I think your first impl was missing an `await`. But why'd you switch to `fireEvent` in the second attempt?

Actually, I'm not sure why. I thought I had a reason to switch to fireEvent 😵 fixed it here:

872e45cc68a5ef5d3e6e4ea8dc7d96a2ddc4ad67

rpokorny (approved):

PR #911, opened 2023-01-19
NxPolicyViolationIndicator RTL Conversion - RSC-1348

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1348

### DIFF
--- a/lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
+++ b/lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
@@ -4,47 +4,89 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
-import 'jest-enzyme';
 
-import NxPolicyViolationIndicator, { Props } from '../NxPolicyViolationIndicator';
-import NxFontAwesomeIcon from '../../NxFontAwesomeIcon/NxFontAwesomeIcon';
+import React from 'react';
+import { render } from '@testing-library/react';
+import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
+import NxPolicyViolationIndicator from '../NxPolicyViolationIndicator';
 
 describe('NxPolicyViolationIndicator', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxPolicyViolationIndicator, {});
+  const quickRender = rtlRender(NxPolicyViolationIndicator, {}),
+      renderEl = rtlRenderElement(NxPolicyViolationIndicator, {});
 
-  it('sets the nx-policy-violation-indicator--unspecified class if no props are passed', function() {
-    expect(getShallowComponent()).toMatchSelector('.nx-policy-violation-indicator--unspecified');
+  it('adds specified classnames in addition to the defaults', function() {
+    const defaultEl = renderEl()!,
+        el = renderEl({ className: 'foo' })!;
+
+    expect(el).toHaveClass('foo');
+
+    for (const cls of Array.from(defaultEl.classList)) {
+      expect(el).toHaveClass(cls);
+    }
   });
 
-  it('sets the modifier class using the threatLevelCategory if it is provided', function() {
-    expect(getShallowComponent({ threatLevelCategory: 'low' })).toMatchSelector('.nx-policy-violation-indicator--low');
-    expect(getShallowComponent({ threatLevelCategory: 'low' }))
-        .not.toMatchSelector('.nx-policy-violation-indicator--unspecified');
+  it('adds additional attributes to the top-level element', function() {
+    const el = renderEl({ id: 'foo', lang: 'en' });
+
+    expect(el).toHaveAttribute('id', 'foo');
+    expect(el).toHaveAttribute('lang', 'en');
   });
 
-  it('sets the modifier class by converting the policyThreatLevel if it is provided', function() {
-    expect(getShallowComponent({ policyThreatLevel: 0 })).toMatchSelector('.nx-policy-violation-indicator--none');
-    expect(getShallowComponent({ policyThreatLevel: 1 })).toMatchSelector('.nx-policy-violation-indicator--low');
-    expect(getShallowComponent({ policyThreatLevel: 3 })).toMatchSelector('.nx-policy-violation-indicator--moderate');
-    expect(getShallowComponent({ policyThreatLevel: 5 })).toMatchSelector('.nx-policy-violation-indicator--severe');
-    expect(getShallowComponent({ policyThreatLevel: 9 })).toMatchSelector('.nx-policy-violation-indicator--critical');
-    expect(getShallowComponent({ policyThreatLevel: 9 }))
-        .not.toMatchSelector('.nx-policy-violation-indicator--unspecified');
+  it('applies a ref to the top level element', function() {
+    const ref = React.createRef<HTMLDivElement>(),
+        el = renderEl({ref: ref})!;
+
+    expect(ref.current).toBe(el);
   });
 
-  it('sets the modifier class using the threatLevelCategory if both props are provided', function() {
-    const component = getShallowComponent({ policyThreatLevel: 9, threatLevelCategory: 'low' });
+  it('sets the text according to threatLevelCategory if no text is provided', function() {
+    const { container, rerender } = render(
+      <NxPolicyViolationIndicator threatLevelCategory='none'></NxPolicyViolationIndicator>
+    );
+    expect(container).toHaveTextContent('none');
 
-    expect(component).toMatchSelector('.nx-policy-violation-indicator--low');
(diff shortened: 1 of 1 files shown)


lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
+  });
+
+  it('sets the text according to threatLevelCategory if no text is provided', function() {
+    const { container, rerender } = render(
+      <NxPolicyViolationIndicator threatLevelCategory='none'></NxPolicyViolationIndicator>
+    );
+    expect(container).toHaveTextContent('none');
+
+    rerender(<NxPolicyViolationIndicator threatLevelCategory='moderate'></NxPolicyViolationIndicator>);
+    expect(container).toHaveTextContent('moderate');
```
rpokorny, line 49:
```suggestion
    expect(renderEl({ threatLevelCategory: 'none' })).toHaveTextContent('none');
    expect(renderEl({ threatLevelCategory: 'moderate' })).toHaveTextContent('moderate');
```

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
   });
 
-  it('sets the modifier class using the threatLevelCategory if both props are provided', function() {
-    const component = getShallowComponent({ policyThreatLevel: 9, threatLevelCategory: 'low' });
+  it('overrides the default threatLevelCategory text if text is provided', function() {
+    const { container } = render(
+      <NxPolicyViolationIndicator threatLevelCategory='low'>Foo</NxPolicyViolationIndicator>
+    );
```
rpokorny, line 55:
```suggestion
    const el = renderEl({ threatLevelCategory: 'low', children: 'Foo' })!;
```
And then use `el` in the expectations not `container`

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
-    expect(component).toMatchSelector('.nx-policy-violation-indicator--low');
-    expect(component).not.toMatchSelector('.nx-policy-violation-indicator--critical');
-    expect(component).not.toMatchSelector('.nx-policy-violation-indicator--unspecified');
+    expect(container).not.toHaveTextContent('low');
+    expect(container).toHaveTextContent('Foo');
   });
 
-  it('adds aria attrs to help the icon show up for screen readers', function() {
-    const component = getShallowComponent({ policyThreatLevel: 9, threatLevelCategory: 'low' });
+  it('sets threatLevelCategory to unspecified if not defined', function() {
```
rpokorny, line 61:
```suggestion
  it('sets text content to unspecified if not defined', function() {
```

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
+    const unspecifiedEl = renderEl()!;
+    expect(unspecifiedEl).toHaveTextContent('unspecified');
+  });
+
+  it('sets the threatLevelCategory by converting policyThreatLevel if it is provided',
```
rpokorny, line 66:
```suggestion
  it('sets the text content to the the threat level category corresponding to the provided policyThreatLevel',
```

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
+        expect(renderEl({ policyThreatLevel: 0})!).toHaveTextContent('none');
+        expect(renderEl({ policyThreatLevel: 1})!).toHaveTextContent('low');
+        expect(renderEl({ policyThreatLevel: 3})!).toHaveTextContent('moderate');
+        expect(renderEl({ policyThreatLevel: 5})!).toHaveTextContent('severe');
+        expect(renderEl({ policyThreatLevel: 9})!).toHaveTextContent('critical');
+        expect(renderEl({ policyThreatLevel: 0})!).toHaveTextContent('none');
+        expect(renderEl()!).toHaveTextContent('unspecified');
+      });
+
+  it('renders an indicator according to threatLevelCategory over policyThreatNumber of both are provided', function() {
```
rpokorny, line 77:
```suggestion
  it('renders an indicator according to threatLevelCategory over policyThreatNumber if both are provided', function() {
```

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
+    expect(renderEl({ policyThreatLevel: 9, threatLevelCategory: 'low'})).toHaveTextContent('low');
```
rpokorny, line 78:
```suggestion
    expect(renderEl({ policyThreatLevel: 9, threatLevelCategory: 'low' })).toHaveTextContent('low');
```

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
+  });
+
+  it('adds aria-label to icon', function() {
```
rpokorny, line 81: `aria-label` is an impl detail, the accessible name is the external behavior we want to test.  Use `toHaveAccessibleName`.

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
```diff
@@ -4,47 +4,89 @@
...
+    expect(quickRender({ threatLevelCategory: 'low' }).getByRole('img'))
+        .toHaveAttribute('aria-label', 'threat level low');
+    expect(quickRender({ threatLevelCategory: 'severe'}).getByRole('img'))
+        .toHaveAttribute('aria-label', 'threat level severe');
+  });
 
-    expect(component.find(NxFontAwesomeIcon)).toHaveProp('aria-label', 'threat level low');
-    expect(component.find(NxFontAwesomeIcon)).toHaveProp('aria-hidden', false);
+  it('sets aria-hidden to false on icon', function() {
+    expect(quickRender().getByRole('img')).toHaveAttribute('aria-hidden', 'false');
```
rpokorny, line 90: The `aria-hidden` attribute itself is also an impl detail, we just want to make sure the image isn't hidden.  But there's not reason to think it would be, so this test seems kind of unnecessary to me (as does the presence of the `aria-hidden` attr in the implementation.  At any rate, the `getByRole('img')` calls in the test above ensure that the image is not hidden, as they exclude hidden elements by default.  So with that this test is definitely not necessary.

rpokorny: > as does the presence of the `aria-hidden` attr in the implementation

Apparently fontawesome icons have `aria-hidden="true"` by default so that explains that

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
EmilyBermudez88: >And then use el in the expectations not container

fixed here: [6adb242](https://github.com/sonatype/sonatype-react-shared-components/pull/911/commits/6adb242c1c0584669ac258fa1bc92d11b9dba698)

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
EmilyBermudez88: > `aria-label` is an impl detail, the accessible name is the external behavior we want to test. Use `toHaveAccessibleName`.
thanks, updated here: [c244209](https://github.com/sonatype/sonatype-react-shared-components/pull/911/commits/c24420967b560540051227a848bf623fe0b607ba)

lib/src/components/NxPolicyViolationIndicator/__tests__/NxPolicyViolationIndicator.test.tsx
EmilyBermudez88: >The aria-hidden attribute itself is also an impl detail, we just want to make sure the image isn't hidden. But there's not reason to think it would be, so this test seems kind of unnecessary to me (as does the presence of the aria-hidden attr in the implementation). At any rate, the getByRole('img') calls in the test above ensure that the image is not hidden, as they exclude hidden elements by default. So with that this test is definitely not necessary.

makes sense, I removed the test here: [c244209](https://github.com/sonatype/sonatype-react-shared-components/pull/911/commits/c24420967b560540051227a848bf623fe0b607ba)

rpokorny (approved):

PR #912, opened 2023-01-20
NxIconDropdown RTL Conversion - RSC-1231

cindyko226:
https://issues.sonatype.org/browse/RSC-1231


rpokorny:
Many of the tests here should duplicate those for `NxDropdown`.  Rather than have to check for the same issues in both PRs, lets wait until that one is done and then just copy that work here and make appropriate modifications to the tests around the toggle (the only part that's different between the two)

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
```diff
@@ -5,326 +5,298 @@
...
-    component!.update();
+    await user.click(document.body);
+
     expect(onToggleCollapse).toHaveBeenCalled();
   });
 
-  it('does not call onToggleCollapse if a click happens anywhere when the dropdown is closed', function() {
-    const onToggleCollapse = jest.fn(),
-        component = getMountedComponent({ onToggleCollapse }, { attachTo: container });
+  it('does not call onToggleCollapse if a click happens anywhere aside from the'
```
rpokorny, line 117:
```suggestion
  it('does not call onToggleCollapse if a click happens anywhere aside from the '
```

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
```diff
@@ -5,326 +5,298 @@
...
-    act(() => {
-      document.dispatchEvent(new MouseEvent('click', {
-        bubbles: true
-      }));
-    });
-    component!.update();
+    await user.click(document.body);
+
     expect(onToggleCollapse).toHaveBeenCalled();
   });
```
rpokorny, line 115: On thing I realize we're missing, here and probably also in `NxDropdown`, is a test that clicking the items within the dropdown 1. fires their click handlers, and 2. closes the dropdown.

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
```diff
@@ -5,326 +5,298 @@
...
-              onCloseKeyDown: e => e.preventDefault()
-            }, { attachTo: container });
-
-        act(() => {
-          component.find('button.nx-icon-dropdown__toggle').getDOMNode()
-              .dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
-        });
-        component.update();
+              onCloseKeyDown: e => e.preventDefault() })
+                .getByRole('button');
```
rpokorny, line 226:
```suggestion
              onCloseKeyDown: e => e.preventDefault() 
            }).getByRole('button');
```

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
```diff
@@ -5,326 +5,298 @@
...
+  it('does not call onToggleCollapse if a click happens when onCloseClick preventsDefault', async function() {
+    const user = userEvent.setup(),
+        onToggleCollapse = jest.fn(),
+        { getByRole, getByTestId } = quickRender({
+          children: <a data-testid="child">Foo</a>,
           onToggleCollapse,
           isOpen: true,
-          onCloseClick: e => e.preventDefault()
-        }, { attachTo: container });
+          onCloseClick: e => e.preventDefault() }),
```
rpokorny, line 242:
```suggestion
          onCloseClick: e => e.preventDefault() 
        }),
```

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
```diff
@@ -5,187 +5,233 @@
...
+import { rtlRender, rtlRenderElement, userEvent, runTimers } from '../../../../__testutils__/rtlUtils';
+import { act, within } from '@testing-library/react';
 
-import NxButton from '../../../NxButton/NxButton';
-import NxIconDropdown from '../../NxIconDropdown';
 import NxStatefulIconDropdown, { Props } from '../NxStatefulIconDropdown';
 
 describe('NxStatefulIconDropdown', () => {
-  let container: HTMLDivElement | null;
+  const childClickSpy = jest.fn();
```
rpokorny, line 14: Spies should be initialized in `beforeEach` or in the tests themselves.  `minimalProps` should just use a noop `() => {}` for required function props

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
```diff
@@ -5,187 +5,233 @@
...
-  it('closes the dropdown when the Escape key is pressed on this component', function() {
-    const page = (
-      <div>
-        <NxStatefulIconDropdown title="stateful dropdown" />
-      </div>
-    );
-    const element = mount(page, { attachTo: container });
-    element.find(NxButton).simulate('click');
-    expect(element.find('.nx-dropdown-menu')).toExist();
+  it('does not open the dropdown if a click happens anywhere aside from the'
```
rpokorny, line 129:
```suggestion
  it('does not open the dropdown if a click happens anywhere aside from the '
```

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
```diff
@@ -5,187 +5,233 @@
...
+        toggleBtn = getByRole('button');
+
+    await user.click(toggleBtn);
+
+    const menu = container.querySelector('.nx-dropdown-menu'),
+        children = getByTestId('child');
+
+    expect(menu).toBeInTheDocument();
+
+    await act(async () => { await user.click(children); });
```
rpokorny, line 165: `act` is unnecessary

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
```diff
@@ -5,187 +5,233 @@
...
+
+        await user.click(toggleBtn);
+
+        const menuBtn = getByTestId('dropdown-button');
+
+        menuBtn.focus();
+
+        expect(document.activeElement).toBe(menuBtn);
+
+        await act(async () => { await user.click(toggleBtn); });
```
rpokorny, line 233: Don't need `act`

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
cindyko226: Updated here c8b692165353a70352f64a1f27f767e82b6e3d22.

lib/src/components/NxIconDropdown/__tests__/NxIconDropdown.test.tsx
cindyko226: > probably also in NxDropdown

Added a note in the RSC-989 ticket to address this.



> a test that clicking the items within the dropdown 1. fires their click handlers, and 2. fires onToggleCollapse

Added here 654d52f7b11ba5fd3adbdb43c4f6607757d03843

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
cindyko226: When I removed the `act`, the tests also passed but there is console warning below, it seems the click fell outside the userEvent's built-in act handling, it happened on the second click which is to close the dropdown.
`Warning: An update to NxStatefulIconDropdown inside a test was not wrapped in act(...)`

rpokorny: > When I removed the act, the tests also passed but there is console warning below, it seems the click fell outside the userEvent's built-in act handling, it happened on the second click which is to close the dropdown.
Warning: An update to NxStatefulIconDropdown inside a test was not wrapped in act(...)

OK that makes sense after some similar problems that I had in `NxFilterDropdown`.  Basically any time the dropdown is being closed by the global, native click handler, we sometimes still need to use `act`.

lib/src/components/NxIconDropdown/stateful/__tests__/NxStatefulIconDropdown.test.tsx
rpokorny: > Don't need act

Actually, using click here at all really isn't right.  Click will inherently move the focus to the clicked element.  We want to test that focus moves to the toggle button even despite that.  The main way I can think of to accomplish that - and the way I did it in `NxFilterDropdown` - is to use the ESC key rather than the mouse to close the dropdown while the `menuBtn` is focused.

cindyko226: Make sense to me, updated here 2837602262ac9bf62d0739289962e6bc0b3be060, thanks.

rpokorny (approved):

PR #924, opened 2023-01-24
RTL Conversion - NxThreatIndicator - RSC-1355

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1355

`NxThreatIndicator` only renders an SVG. Based on threat levels, the appropriate modifier classes determine the color of the indicator. Instead of testing for classes (which are an implementation detail), the tests test if the SVG has the correct tooltips.


lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,104 @@
...
+    expect(tooltip).toHaveTextContent('Severe');
   });
 
-  it('sets the modifier class using the threatLevelCategory if it is provided', function() {
-    const icon = getShallowComponent({ threatLevelCategory: 'low' }).find(NxFontAwesomeIcon);
-
-    expect(icon).toMatchSelector('.nx-threat-indicator--low');
-    expect(icon).not.toMatchSelector('.nx-threat-indicator--unspecified');
-  });
+  it('renders a threat indicator if policyThreatLevel prop is provided', async function() {
```
rpokorny, line 31: Unless there's a condition that causes it not to render anything, these two tests don't really make sense to me

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,104 @@
...
-    expect(icon).toExist();
+    expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+    await user.hover(threatIndicator);
+    expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
   });
 
   it('should set the role to "presentation" when presentational prop is true', function() {
-    const icon = getShallowComponent({ presentational: true }).find(NxFontAwesomeIcon);
+    const el = quickRender({ presentational: true }),
+        threatIndicator = el.getByRole('presentation', { hidden: true });
```
rpokorny, line 96: I'm thinking the fact that they're hidden is a bug.  There's no point in us messing with the roles if they're always hidden, so presumably we didn't notice that they are when we implemented the `presentational` prop.  I suspect `aria-hidden={false}` should be added to the implementation.  That can be done in this ticket.

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,104 @@
...
-  const getShallowComponent = enzymeUtils.getShallowComponent(NxThreatIndicator, {});
+  const quickRender = rtlRender<Props>(NxThreatIndicator, {});
 
-  it('renders an faCircle NxFontAwesomeIcon with the `nx-threat-indicator` class', function() {
-    const icon = getShallowComponent().find(NxFontAwesomeIcon);
-
-    expect(icon).toHaveClassName('nx-threat-indicator');
-    expect(icon).toExist();
-    expect(icon).toHaveProp('icon', faCircle);
+  it('renders a threat indicator svg', function() {
```
rpokorny, line 15: The fact that it's an svg is also an implementation detail IMO.  The relevant accessibility check would be that it has `role="img"` (when `presentational` is not set)

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,104 @@
...
   });
 
   it('adds aria attrs to help the icon show up for screen readers', function() {
-    const component = getShallowComponent({ policyThreatLevel: 9, threatLevelCategory: 'low' });
-
-    expect(component.find(NxFontAwesomeIcon)).toHaveProp('aria-label', 'threat level low');
-    expect(component.find(NxFontAwesomeIcon)).not.toHaveProp('aria-hidden');
-  });
+    const { container } = quickRender({ policyThreatLevel: 9, threatLevelCategory: 'low' }),
+        threatIndicator = container.querySelector('svg')!;
```
rpokorny, line 53: Rather than querying for the svg everywhere, just use the top-level element of the render i.e. `container.firstElementChild`.  And of course, we have `rtlRenderElement`/`renderEl` to make that more convenient.

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
rpokorny: > I'm thinking the fact that they're hidden is a bug. There's no point in us messing with the roles if they're always hidden, so presumably we didn't notice that they are when we implemented the presentational prop. I suspect aria-hidden={false} should be added to the implementation. That can be done in this ticket.

Actually I now see that it's only hidden when we set `role="presentation"` and not in the other cases.  That seems redundant but fine (and it appears it's implemented in fontawesome, not by us).  That's fine I guess.  The cleanest way to test that, I think, would be to simply assert that it does _not_ render an element with role `img` when the `presentational` prop is true.

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Unless there's a condition that causes it not to render anything, these two tests don't really make sense to me

Agreed. Removed here: c5ebedea07e624cc6239aaac7bad0fe8fce4ba18

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > > I'm thinking the fact that they're hidden is a bug. There's no point in us messing with the roles if they're always hidden, so presumably we didn't notice that they are when we implemented the presentational prop. I suspect aria-hidden={false} should be added to the implementation. That can be done in this ticket.
> 
> Actually I now see that it's only hidden when we set `role="presentation"` and not in the other cases. That seems redundant but fine (and it appears it's implemented in fontawesome, not by us). The cleanest way to test that, I think, would be to simply assert that it does _not_ render an element with role `img` when the `presentational` prop is true.

Added new test here: 0185e49193add4d15213a9056fef1fc6c7f6cd45

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > The fact that it's an svg is also an implementation detail IMO. The relevant accessibility check would be that it has `role="img"` (when `presentational` is not set)

Changed here: 4c116c4eab2f01f2c1f3089e42a7d0c415cc10c2

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Rather than querying for the svg everywhere, just use the top-level element of the render i.e. `container.firstElementChild`. And of course, we have `rtlRenderElement`/`renderEl` to make that more convenient.

Changed to query by roles here: 4c116c4eab2f01f2c1f3089e42a7d0c415cc10c2

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,104 @@
...
-    expect(critical).toHaveProp('title', 'Critical');
+    expect(threatIndicator).toHaveAttribute('aria-label', 'threat level low');
+    expect(threatIndicator).not.toHaveAttribute('aria-hidden', true);
   });
 
-  it('should show custom tooltip title', function() {
-    const tooltip = getShallowComponent({ title: 'Extinction Level Threat' }).find(NxToolip);
-
-    expect(tooltip).toHaveProp('title', 'Extinction Level Threat');
+  describe('should have default tooltips', function() {
```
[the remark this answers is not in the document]
avnvbhatta: > Since there's a finite and small-ish number of possibilities here (11 threat levels numbers, six threat level categories) I think we should test all of them to make sure they have the correct tooltip text.

I agree. Added tests for all threat numbers here: f7f94af8862571762ca4971a06dada5d133d16a1

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
-  });
-
-  it('sets the nx-threat-indicator--unspecified class if no props are passed', function() {
-    const icon = getShallowComponent().find(NxFontAwesomeIcon);
+  const quickRender = rtlRender<Props>(NxThreatIndicator, {});
 
-    expect(icon).toMatchSelector('.nx-threat-indicator--unspecified');
+  it('renders a threat indicator', function() {
+    const el = quickRender(),
+        threatIndicator = el.getByRole('img', { hidden: true, queryFallbacks: true });
```
rpokorny, line 21: As far as I can tell this isn't hidden

rpokorny, line 21: Is there a reason you're not testing the `graphics-symbol` role?

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
 
   it('adds aria attrs to help the icon show up for screen readers', function() {
-    const component = getShallowComponent({ policyThreatLevel: 9, threatLevelCategory: 'low' });
+    const el = quickRender({ policyThreatLevel: 9, threatLevelCategory: 'low' }),
+        threatIndicator = el.getByRole('img', { hidden: true, queryFallbacks: true })!;
 
-    expect(component.find(NxFontAwesomeIcon)).toHaveProp('aria-label', 'threat level low');
-    expect(component.find(NxFontAwesomeIcon)).not.toHaveProp('aria-hidden');
+    expect(threatIndicator).toHaveAttribute('aria-label', 'threat level low');
+    expect(threatIndicator).not.toHaveAttribute('aria-hidden', true);
```
rpokorny, line 40: Assuming the `hidden: true` in the `getByRole` is not in fact needed and can be removed, this assertion can also be removed

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
   });
 
   it('adds aria attrs to help the icon show up for screen readers', function() {
-    const component = getShallowComponent({ policyThreatLevel: 9, threatLevelCategory: 'low' });
+    const el = quickRender({ policyThreatLevel: 9, threatLevelCategory: 'low' }),
+        threatIndicator = el.getByRole('img', { hidden: true, queryFallbacks: true })!;
 
-    expect(component.find(NxFontAwesomeIcon)).toHaveProp('aria-label', 'threat level low');
-    expect(component.find(NxFontAwesomeIcon)).not.toHaveProp('aria-hidden');
+    expect(threatIndicator).toHaveAttribute('aria-label', 'threat level low');
```
rpokorny, line 39: `toHaveAccessibleName` would be better

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
-    expect(severe).toHaveProp('title', 'Severe');
-    expect(critical).toHaveProp('title', 'Critical');
+  describe('should have default tooltips', function() {
+    describe('when policyThreatLevel prop is provided', function() {
+      allThreatLevelNumbers.forEach(threatNumber => {
+        it(`for policy threat level ${threatNumber}`, async function() {
+          const el = quickRender({ policyThreatLevel: threatNumber })!,
+              user = userEvent.setup(),
+              category = categoryByPolicyThreatLevel[threatNumber],
+              categoryTitleCase = category[0].toUpperCase() + category.slice(1),
```
rpokorny, line 50: This is too programmatic, and too reliant on the same data structures that drive the implementation.  As a test, this should just hardcode the relationships between the expected inputs and outputs.  e.g.
```
expect(getTooltipTextForProps({ policyThreatLevel: 1 })).toBe('Low');
```
and so on.  And you'd have to write the `getTooltipTextForProps` function.

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
+      });
+    });
+
+    describe('when threatLevelCategory prop is provided', function() {
+      allThreatLevelCategories.forEach(category => {
+        it(`for threat level category ${category}`, async function() {
+          const el = quickRender({ threatLevelCategory: category })!,
+              user = userEvent.setup(),
+              threatIndicator = el.getByRole('img', { hidden: true, queryFallbacks: true }),
+              categoryTitleCase = category[0].toUpperCase() + category.slice(1);
```
rpokorny, line 66: Also too programmatic

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
-    const icon = getShallowComponent({ className: 'foo', threatLevelCategory: 'low' }).find(NxFontAwesomeIcon);
+  describe('when presentational prop is set to true', function() {
+    let el: RenderResult;
 
-    expect(icon).toHaveClassName('foo');
-    expect(icon).toHaveClassName('nx-threat-indicator');
-    expect(icon).toHaveClassName('nx-threat-indicator--low');
-  });
+    beforeEach(() => {
+      el = quickRender({ presentational: true });
```
rpokorny, line 90: Using beforeEach in this way is an anti-pattern to some people.  As an alternative, consider shadowing `quickRender` within this `describe` block:
```
const quickRender = rtlRender(NxThreatIndicator, { ...minimalProps, presentational: true });
```

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
-    const icon = getShallowComponent({ presentational: true }).find(NxFontAwesomeIcon);
+    it('should set the role to "presentation"', function() {
+      const threatIndicator = el.getByRole('presentation', { hidden: true, queryFallbacks: true });
 
-    expect(icon).toHaveProp('role', 'presentation');
-  });
+      expect(threatIndicator).toBeInTheDocument();
+    });
+
+    it('should not set the role to "img"', function() {
```
rpokorny, line 108: Seems redundant with the previous test

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,115 @@
...
+      expect(el.queryByRole('img', { hidden: true, queryFallbacks: true })).not.toBeInTheDocument();
+    });
 
-  it('should set aria-label to undefined when presentational prop is true', function() {
-    const icon = getShallowComponent({ presentational: true }).find(NxFontAwesomeIcon);
+    it('should set aria-label to undefined', function() {
```
rpokorny, line 112: Cleaner: how about checking that it doesn't have an accessible name?

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > As far as I can tell this isn't hidden

The `svg` rendered on the browser doesn't have `aria-hidden`, but on RTL tests, the `svg` rendered always seem to have `aria-hidden`  even though the `presentational` prop isn't provided.
image

avnvbhatta: > Is there a reason you're not testing the `graphics-symbol` role?

Since we're using `queryFallbacks` and since `graphics-symbol` is a subclass of `img`, I thought it'd be safer to use the `img` role.

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > This is too programmatic, and too reliant on the same data structures that drive the implementation. As a test, this should just hardcode the relationships between the expected inputs and outputs. e.g.
> 
> ```
> expect(getTooltipTextForProps({ policyThreatLevel: 1 })).toBe('Low');
> ```
> 
> and so on. And you'd have to write the `getTooltipTextForProps` function.

Changed here: 8900977e452b45b41a5a7c46e7a9b8015a421cae

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Also too programmatic

Changed here: 8900977e452b45b41a5a7c46e7a9b8015a421cae

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Cleaner: how about checking that it doesn't have an accessible name?

c80d9a7fddba36db3191b374bed845810b2c64fd

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Seems redundant with the previous test

a2617d48cca0f385b3faa61fd2981a00885e8093

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: I was looking into it, and this might be because of Font Awesome's Auto Accessibility feature:
https://cdn.fontawesome.com/help#qa-autoa11y

rpokorny: > Since we're using queryFallbacks and since graphics-symbol is a subclass of img, I thought it'd be safer to use the img role.

It's safe in that it certainly enables the tests to pass but it means that they aren't as complete; they aren't testing that the `graphics-symbol` role is present

rpokorny: > I was looking into it, and this might be because of Font Awesome's Auto Accessibility feature:
https://cdn.fontawesome.com/help#qa-autoa11y

Well that's not good.  We don't want these icons to be hidden unless `presentational` is set.  And we also don't want behavior in tests to be differing from behavior in the browser, if we can at all help it.  This needs to be investigated more.

avnvbhatta: > > I was looking into it, and this might be because of Font Awesome's Auto Accessibility feature:
> > https://cdn.fontawesome.com/help#qa-autoa11y
> 
> Well that's not good. We don't want these icons to be hidden unless `presentational` is set. And we also don't want behavior in tests to be differing from behavior in the browser, if we can at all help it. This needs to be investigated more.

Added a config to disable auto a11y here: 504d0d9a10e4e39a12b178732a4200baf56b2350

avnvbhatta: > > Since we're using queryFallbacks and since graphics-symbol is a subclass of img, I thought it'd be safer to use the img role.
> 
> It's safe in that it certainly enables the tests to pass but it means that they aren't as complete; they aren't testing that the `graphics-symbol` role is present

Changed to use `graphics-symbol`. With this though, I don't think we need the `{ queryFallbacks: true }` anymore either.

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Assuming the `hidden: true` in the `getByRole` is not in fact needed and can be removed, this assertion can also be removed

Changed here: ce9d721aaf8ea9aed78dfad31a331f3577308de1

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > `toHaveAccessibleName` would be better

Changed here: ce9d721aaf8ea9aed78dfad31a331f3577308de1

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
rpokorny: > Added a config to disable auto a11y here: https://github.com/sonatype/sonatype-react-shared-components/commit/504d0d9a10e4e39a12b178732a4200baf56b2350

That disables it everywhere.  Is there a way to disable it per FA use?  Maybe just pass in `aria-hidden={false}` explicitly?

rpokorny: > That disables it everywhere. Is there a way to disable it per FA use? Maybe just pass in aria-hidden={false} explicitly?

Though I also wonder if it might just be a matter of throwing a `runTimers` in to wait for the tooltips to initialize.  [FA's docs](https://fontawesome.com/v5/docs/web/other-topics/accessibility#svg-with-javascript-semantic-icons) seem to suggest that any icon with a `title` will not be hidden

rpokorny: > With this though, I don't think we need the { queryFallbacks: true } anymore either.

Right I was expecting that to be removed.  I also think we should have one test that is about the fallback to the `img` role

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,137 @@
...
-    expect(icon).toExist();
-    expect(icon).toHaveProp('icon', faCircle);
-  });
-
-  it('sets the nx-threat-indicator--unspecified class if no props are passed', function() {
-    const icon = getShallowComponent().find(NxFontAwesomeIcon);
+  const quickRender = rtlRender<Props>(NxThreatIndicator, {});
 
-    expect(icon).toMatchSelector('.nx-threat-indicator--unspecified');
+  it('renders a threat indicator', function() {
```
rpokorny, line 14:
```suggestion
  it('renders an element with a role of graphics-symbol', function() {
```

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
```diff
@@ -4,116 +4,137 @@
...
-
-    expect(icon).toMatchSelector('.nx-threat-indicator--low');
-    expect(icon).not.toMatchSelector('.nx-threat-indicator--critical');
-    expect(icon).not.toMatchSelector('.nx-threat-indicator--unspecified');
+    await user.hover(threatIndicator);
+    const tooltip = await screen.findByRole('tooltip');
+    expect(tooltip).toHaveTextContent('Low');
   });
 
   it('adds aria attrs to help the icon show up for screen readers', function() {
```
rpokorny, line 30: Sets the accessible name based on the threat level category

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > > With this though, I don't think we need the { queryFallbacks: true } anymore either.
> 
> Right I was expecting that to be removed. I also think we should have one test that is about the fallback to the `img` role

Removed `queryFallback` here: b2399de83b33c306eba4b748c6d24a23aeedb192
Added test for `role="img"` fallback here: 11171e28206bfeda76b5182c66de5ab3e692fa9a

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > Sets the accessible name based on the threat level category

c20f4a137b2cb653adec3302cd5b3be12b67cb6e

lib/src/components/NxThreatIndicator/__tests__/NxThreatIndicator.test.tsx
avnvbhatta: > > That disables it everywhere. Is there a way to disable it per FA use? Maybe just pass in aria-hidden={false} explicitly?
> 
> Though I also wonder if it might just be a matter of throwing a `runTimers` in to wait for the tooltips to initialize. [FA's docs](https://fontawesome.com/v5/docs/web/other-topics/accessibility#svg-with-javascript-semantic-icons) seem to suggest that any icon with a `title` will not be hidden

The runTimers approach does work when dealing with `presentational={false}`. It doesn't work when `presentational={true}` since no tooltips are rendered and there's no `title` attribute. As a result the `aria-hidden` would be auto set to `true`.

I think the best approach is to add `aria-hidden={false}` because in either case when `presentational` is true/false, we don't wait the element to have `aria-hidden={true}`.

I've made the change here: 944940d46065a9f45c8502c910641074bd2b0c02

rpokorny: > It doesn't work when presentational={true} since no tooltips are rendered and there's no title attribute. As a result the aria-hidden would be auto set to true.

We _want_ the element to be hidden when `presentational` is true.  Now, it already sort of is due to the `role="presentation"`, but if FA also wants to add `aria-hidden="true"` for that case I have no problem with that.  So I'd revert https://github.com/sonatype/sonatype-react-shared-components/commit/944940d46065a9f45c8502c910641074bd2b0c02

avnvbhatta: > > It doesn't work when presentational={true} since no tooltips are rendered and there's no title attribute. As a result the aria-hidden would be auto set to true.
> 
> We _want_ the element to be hidden when `presentational` is true. Now, it already sort of is hidden due to the `role="presentation"`, but if FA also wants to add `aria-hidden="true"` for that case I have no problem with that. So I'd remove the explicit `aria-hidden={false}`

Removed explicit `aria-hidden` here: bf8ccdd52a2998b01c993a0b7dc6136eae77e18f
Removed auto a11y config here: b85bc20600e5686529105c450302af80c9607d50
Modified tests to use `runTimers` here: 7e322c3dd62ddbe4ec12aa83514365b9ec1309ac

rpokorny (approved):

PR #925, opened 2023-01-24
FontAwesomeIcon RTL - RSC-1417

achen5:
https://issues.sonatype.org/browse/RSC-1417


gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.10.10",
+  "version": "12.10.11",
```
rpokorny, line 3: No version bumps on these tickets.  The version was bumped once on the feature branch and the purpose of the feature branch is to avoid having to do more version management in each of these tickets (and avoid wasting a bunch of version numbers)

rpokorny:
I was curious to see what you'd come up with for this one.  As `NxFontAwesomeIndicator` is just a wrapper around a third-party component, and one that exposes the full API of that third-party component no less, there's not a super clean way to test it in RTL.  We can no longer just test that the props get passed along to `FontAwesomeIcon`, but to fully test all capabilities that are built into `FontAwesomeIcon` wouldn't be realistic either.  I think what you have here is a reasonable compromise.

gallery/package.json
achen5: > No version bumps on these tickets. The version was bumped once on the feature branch and the purpose of the feature branch is to avoid having to do more version management in each of these tickets (and avoid wasting a bunch of version numbers)

I reverted the version bumps here: [ec11651](https://github.com/sonatype/sonatype-react-shared-components/pull/925/commits/ec11651d843e04bd040a7a5dde04343891a6f59b)

rpokorny (approved):

PR #929, opened 2023-01-25
NxCollapsibleMultiSelect RTL Conversion - RSC-1372

cindyko226:
https://issues.sonatype.org/browse/RSC-1372

### DIFF
--- a/lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
+++ b/lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
@@ -4,355 +4,559 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React from 'react';
+import { rtlRender, rtlRenderElement, runTimers, userEvent } from '../../../../__testutils__/rtlUtils';
+import { createEvent, fireEvent, screen, within } from '@testing-library/react';
 
-import {getShallowComponent} from '../../../../__testutils__/enzymeUtils';
-import 'jest-enzyme';
-import NxCollapsibleMultiSelect, {Props, Option} from '../NxCollapsibleMultiSelect';
+import NxCollapsibleMultiSelect, { Props } from '../NxCollapsibleMultiSelect';
 import { NxTreeViewMultiSelect } from '../../../../index';
 
-import MultiSelectCounter from '../../../Counter/MultiSelectCounter';
-
 describe('NxCollapsibleMultiSelect', function() {
-  const requiredProps: Props = {
-    options: [
-      {id: 'foo', name: 'Foo'},
-      {id: 'bar', name: 'Bar'},
-      {id: null, name: 'Null'} // Ensure that `null` ids are valid
-    ],
-    children: 'Foobar',
-    name: 'foobar',
-    onChange: () => {}
-  };
-
-  const getShallow = getShallowComponent<Props>(NxCollapsibleMultiSelect, requiredProps);
+  const minimalProps: Props = {
+        options: [
+          {id: 'foo', name: 'Foo'},
+          {id: 'bar', name: 'Bar'},
+          {id: null, name: 'Null'} // Ensure that `null` ids are valid
+        ],
+        children: 'Foobar',
+        name: 'foobar',
+        onChange: () => {}
+      },
+      quickRender = rtlRender(NxCollapsibleMultiSelect, minimalProps),
+      renderEl = rtlRenderElement(NxCollapsibleMultiSelect, minimalProps);
 
   it('is aliased as NxTreeViewMultiSelect', function() {
     expect(NxCollapsibleMultiSelect).toBe(NxTreeViewMultiSelect);
   });
 
-  it('throws an error if a selectedId is supplied that is not part of the options', function() {
-    const expectedErr = 'You are attempting to select "ufo", but it is not part of the available options';
-    expect(() => {
-      getShallow({ selectedIds: new Set(['ufo']) });
-    }).toThrowError(expectedErr);
+  it('renders a top-level element with a group role', function() {
+    const view = quickRender();
+
+    expect(view.getByRole('group')).toBe(view.container.firstChild);
   });
 
-  it('properly renders component using only required props', function() {
-    const shallowRender = getShallow();
-
-    expect(shallowRender).toHaveDisplayName('AbstractCollapsibleItemsSelect');
-    expect(shallowRender).toHaveProp('name', 'foobar');
-    expect(shallowRender).toHaveProp('children', 'Foobar');
-    expect(shallowRender).toHaveProp('options', [
-      {id: 'foo', name: 'Foo'},
-      {id: 'bar', name: 'Bar'},
-      {id: null, name: 'Null'}
-    ]);
+  it('sets the specified id', function() {
+    expect(renderEl({ id: 'foo' })).toHaveAttribute('id', 'foo');
   });
 
-  it('passes props to AbstractCollapsibleItemsSelect', function() {
-    const optionalProps = {
-      isOpen: true,
-      disabled: true,
-      onToggleCollapse: jest.fn(),
-      disabledTooltip: 'test disabled tooltip',
-      optionTooltipGenerator: (option: Option) => option.name,
-      tooltipModifierClass: 'tooltip-test-class',
-      onFilterChange: jest.fn(),
-      filter: 'filter term',
-      filteredOptions: [{id: 'bar', name: 'Bar'}],
-      filterPlaceholder: 'test filter placeholder',
-      filterThreshold: 1,
-      id: 'someid'
-    };
-
-    const shallowRender = getShallow(optionalProps);
-    expect(shallowRender).toHaveDisplayName('AbstractCollapsibleItemsSelect');
-    expect(shallowRender).toHaveProp(optionalProps);
-  });
-
-  describe('renderOption prop', function () {
-    it('renders unchecked <NxCheckbox> if option is not selected', function () {
-      const shallowRender = getShallow({
-        selectedIds: new Set(['foo'])
-      });
-      const optionWrapper = shallowRender.renderProp('renderOption')({id: 'bar', name: 'Bar'});
-
-      expect(optionWrapper).toHaveDisplayName('ForwardRef(NxCheckbox)');
-      expect(optionWrapper).toHaveProp('isChecked', false);
+  describe('exceptions', function() {
+    beforeEach(function() {
+      // prevent RTL logging thrown exceptions
+      jest.spyOn(console, 'error').mockImplementation(() => {});
     });
 
-    it('renders checked <NxCheckbox> if option is selected', function () {
-      const shallowRender = getShallow({
-        selectedIds: new Set(['foo'])
-      });
-      const optionWrapper = shallowRender.renderProp('renderOption')({id: 'foo', name: 'Foo'});
+    it('throws an error if a selectedId is supplied that is not part of the options', function() {
+      const expectedErr = 'You are attempting to select "ufo", but it is not part of the available options';
 
-      expect(optionWrapper).toHaveDisplayName('ForwardRef(NxCheckbox)');
-      expect(optionWrapper).toHaveProp('isChecked', true);
-    });
-
-    it('renders unchecked <NxCheckbox> for all options if selectedIds prop is not provided', function () {
-      const shallowRender = getShallow();
-      const renderOption = shallowRender.renderProp('renderOption');
-
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('isChecked', false);
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('isChecked', false);
-      expect(renderOption({id: null, name: 'Null'})).toHaveProp('isChecked', false);
-    });
-
-    it('renders enabled <NxCheckbox> for all options if component is not disabled', function () {
-      const renderOption = getShallow().renderProp('renderOption');
-
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('disabled', false);
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('disabled', false);
-      expect(renderOption({id: null, name: 'Null'})).toHaveProp('disabled', false);
-    });
-
-    it('renders disabled <NxCheckbox> for all options if component is disabled', function () {
-      const shallowRender = getShallow({
-        disabled: true
-      });
-      const renderOption = shallowRender.renderProp('renderOption');
-
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('disabled', true);
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('disabled', true);
-      expect(renderOption({id: null, name: 'Null'})).toHaveProp('disabled', true);
-    });
-
-    it('uses option name as checkbox label', function () {
-      const renderOption = getShallow().renderProp('renderOption');
-
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('children', 'Foo');
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('children', 'Bar');
-      expect(renderOption({id: null, name: 'Null'})).toHaveProp('children', 'Null');
-    });
-
-    it('generates checkboxId using name prop and option name', function () {
-      const shallowRender = getShallow();
-      const renderOption = shallowRender.renderProp('renderOption');
-
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('checkboxId', 'nx-collapsible-items-select-foobar-foo');
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('checkboxId', 'nx-collapsible-items-select-foobar-bar');
-      expect(renderOption({id: null, name: 'Null'}))
-          .toHaveProp('checkboxId', 'nx-collapsible-items-select-foobar-null');
-    });
-
-    describe('onChange callback', function () {
-      it('properly calls component\'s onChange callback when an option is toggled', function() {
-        const onChangeSpy = jest.fn();
-        const shallowRender = getShallow({
-          onChange: onChangeSpy,
-          selectedIds: new Set(['bar'])
-        });
-        const renderOption = shallowRender.renderProp('renderOption');
-
-        expect(onChangeSpy).not.toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
+
+      expect(childrenElId).toBeDefined();
+
+      const childrenEl = document.getElementById(childrenElId)!,
+          childEl = view.getByRole('menu');
 
-      expect(optionWrapper).toHaveDisplayName('ForwardRef(NxCheckbox)');
-      expect(optionWrapper).toHaveProp('isChecked', true);
+      expect(view.container).toContainElement(childrenEl);
+      expect(childrenEl).toContainElement(childEl);
```
rpokorny, line 79: I'd think we want to assert that the element referenced by `aria-controls` has the `menu` role itself, not just contains an element that has it.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
+      expect(onToggleCollapseNoOptions).not.toHaveBeenCalled();
+      await user.click(noOptionsTrigger);
+      expect(onToggleCollapseNoOptions).not.toHaveBeenCalled();
 
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('disabled', false);
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('disabled', false);
-      expect(renderOption({id: null, name: 'Null'})).toHaveProp('disabled', false);
+      expect(onToggleCollapseOptions).not.toHaveBeenCalled();
+      await user.click(optionsTrigger);
+      expect(onToggleCollapseOptions).toHaveBeenCalled();
```
rpokorny, line 118: You don't really need two separate spies here.  With the clicks occurring in this order, even if the two components used the same `onToggleCollapse` callback you'd be sure that it wasn't until after the second click that it fired.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
-        disabled: true
+    describe('tooltip', function() {
+      describe('when component is disabled due to no options', function() {
+        it('has a default text constructed with the name prop', async function() {
+          const trigger = renderAndGetTrigger({ options: [] });
+
+          await runTimers();
+          expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+
+          fireEvent.mouseOver(trigger);
```
rpokorny, line 129: `userEvent` has a `hover` function that would be preferred

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
     });
+  });
 
-    it('uses option name as checkbox label', function () {
-      const renderOption = getShallow().renderProp('renderOption');
+  describe('filter', function() {
+    const filterView = (props?: Partial<Props>) => quickRender({
+      ...props,
+      filterThreshold: 2,
+      onFilterChange: jest.fn()
```
rpokorny, line 229: Set up mocks in individual tests or `beforeEach`
```suggestion
      onFilterChange: () => {}
```

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
-      expect(renderOption({id: null, name: 'Null'})).toHaveProp('children', 'Null');
+      expect(quickRender().queryByRole('textbox')).not.toBeInTheDocument();
+      expect(noFilterView.queryByRole('textbox')).not.toBeInTheDocument();
+      expect(filterView().getByRole('textbox')).toHaveAttribute('type', 'text');
     });
 
-    it('generates checkboxId using name prop and option name', function () {
-      const shallowRender = getShallow();
-      const renderOption = shallowRender.renderProp('renderOption');
+    it('sets the filter input placeholder as specified by filterPlaceholder, ' +
```
rpokorny, line 241: Should check the a11y name too

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
-        renderOption({id: 'bar', name: 'Bar'}).simulate('change');
-        expect(onChangeSpy).toHaveBeenCalledWith(new Set([]), 'bar');
+      expect(clearBtn).toBeInTheDocument();
+      expect(clearBtn).toHaveAccessibleName('Clear filter');
+    });
 
-        renderOption({id: null, name: 'Null'}).simulate('change');
-        expect(onChangeSpy).toHaveBeenCalledWith(new Set(['bar', null]), null);
-      });
+    it('clears the input when the clear button is clicked', async function() {
```
rpokorny, line 259:
```suggestion
    it('calls onFilterChange with the empty string when the clear button is clicked', async function() {
```

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
+      'default is "filter"', function() {
+      const viewWithPlaceholder = filterView({ filterPlaceholder: 'filter cats' });
 
-      expect(renderOption({id: 'foo', name: 'Foo'})).toHaveProp('checkboxId', 'nx-collapsible-items-select-foobar-foo');
-      expect(renderOption({id: 'bar', name: 'Bar'})).toHaveProp('checkboxId', 'nx-collapsible-items-select-foobar-bar');
-      expect(renderOption({id: null, name: 'Null'}))
-          .toHaveProp('checkboxId', 'nx-collapsible-items-select-foobar-null');
+      expect(filterView().getByRole('textbox')).toHaveAttribute('placeholder', 'filter');
+      expect(viewWithPlaceholder.getByRole('textbox')).toHaveAttribute('placeholder', 'filter cats');
     });
```
rpokorny, line 248: Needs a test that typing in the filter box triggers `onFilterChange`.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
+
+        expect(onChange).toHaveBeenCalledWith(new Set(['bar']), 'bar');
+
+        await user.click(options[1]);
 
-    it('renders disabled <NxCheckbox> if component is disabled', function () {
-      const shallowRender = getShallow({
-        disabled: true
+        expect(onChange).toHaveBeenCalledWith(new Set(['foo']), 'foo');
       });
```
rpokorny, line 378: Test having one or more already selected so we know the set containing the previous and new selections together gets constructed properly

rpokorny, line 378: Needs a test for unchecking.  See `NxFilterDropdown.test.tsx` for a way to combine this and my other comment here reasonably conveniently

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
```diff
@@ -4,355 +4,559 @@
...
+          const tooltip = screen.getByRole('tooltip');
 
-    describe('onChange callback', function () {
-      it('selects all options when no options are selected', function() {
-        const onChangeSpy = jest.fn();
-        const shallowRender = getShallow({
-          onChange: onChangeSpy,
-          selectedIds: new Set()
+          expect(tooltip).toHaveTextContent('customized-tooltip');
         });
```
rpokorny, line 396: Should test that the correct arguments are passed into `optionTooltipGenerator`

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/stateful/__tests__/NxStatefulCollapsibleMultiSelect.test.tsx
```diff
@@ -4,93 +4,543 @@
...
+
+      inputEl.focus();
+      fireEvent(inputEl, keyEvent);
+
+      expect(keyEvent.defaultPrevented).toBe(false);
     });
-    expect(shallowRender).toHaveProp('isOpen', true);
-    shallowRender.simulate('toggleCollapse');
-    expect(shallowRender).toHaveProp('isOpen', false);
   });
```
rpokorny, line 300: The stateful version of this component handles the filtering internally.  That needs to be tested.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Updated here. b8784b45f5ff1e74bfa592c0845358924c2cb56e

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Agree, updated here 3b5cf6e1f3e6a1c1529fc7fe5fd85aab4f8eaaaa

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Updated here 90a6389ea72a7c8eb19b4ef11c30b29abeb126be

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Added here 7b25dc023be059b3b34d7bd728f43772aeed5d6f

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Updated here e95cb8d950d397244d6057a9b7d28bdba430bbd1

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Updated here 83230a2cde6fac48d9d5c9c2a8533bef96565460

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Updated here 10a242ecd58c2723c1121e5cf860349fb5d9bca0

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/stateful/__tests__/NxStatefulCollapsibleMultiSelect.test.tsx
cindyko226: Updated here f0e6a20b5cc00424fde36d4b1d535f46452f155b

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: > userEvent has a hover function that would be preferred

I originally had the `userEvent` for hovering on the disabled trigger in this test but was not able to query the tooltip after waiting `runTimers`.
 
It seems the user-event `hover` does not trigger `mouseover` for disabled elements.
https://github.com/testing-library/user-event/issues/715

rpokorny: > I originally had the userEvent for hovering on the disabled trigger in this test but was not able to query the tooltip after waiting runTimers.

It seems the user-event hover does not trigger mouseover for disabled elements.
https://github.com/testing-library/user-event/issues/715

Ah. as that thread says, attribute-disabled buttons do not in fact dispatch mouse events.  Our implementation mitigates that by attaching the tooltip to a div which wraps the button.  Even though it would make the test aware of our implementation, I think it would be slightly preferable to adjust this test to user `user.hover()` on the wrapper div, rather than what you have here.  Neither approach is perfect from a test cleanliness perspective, but at least that way the event dispatching matches what happens in real life.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/stateful/__tests__/NxStatefulCollapsibleMultiSelect.test.tsx
rpokorny: > Updated here https://github.com/sonatype/sonatype-react-shared-components/commit/f0e6a20b5cc00424fde36d4b1d535f46452f155b

I think that needs to be fleshed out more.  For one thing, when the filter is active, we should check not just the number of options still present but which ones they are.  Additionally, more filtering cases should be checked.  For instance case-sensitivity, filtering by strings that match other parts besides the beginning of the word, and also something that tests the `fuzzyFilter` functionality of `NxStatefulCollapsibleMultiSelect` to at least some extent.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/stateful/__tests__/NxStatefulCollapsibleMultiSelect.test.tsx
```diff
@@ -4,93 +4,579 @@
...
+      expect(clearBtn).toHaveAccessibleName('Clear filter');
+    });
+
+    it('clears the input when the clear button is clicked', async function() {
+      const user = userEvent.setup(),
+          view = filterView(),
+          inputEl = view.getByRole('textbox');
+
+      await runTimers();
+
```
EmilyBermudez88, line 255: You should add in the user typing a value into the input, to confirm the clear button clears the input value

EmilyBermudez88:
It may be useful adding in a test that confirms the input has an attribute of `aria-controls` on the overall stateful & regular NxCollapsibleMultiSelect

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: Updated here 0940f9ea7961ef49aa45ea3665bad499f7853a60

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/stateful/__tests__/NxStatefulCollapsibleMultiSelect.test.tsx
cindyko226: Added more tests here 378b2fb2a90fe03deb2152edc9139c47c7c56a36

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/stateful/__tests__/NxStatefulCollapsibleMultiSelect.test.tsx
cindyko226: Thanks for catching it, updated here 4dae2b2d775d4ccb67236a44401f75d0d35f4d91

cindyko226:
> It may be useful adding in a test that confirms the input has an attribute of `aria-controls` on the overall stateful & regular NxCollapsibleMultiSelect

It is already included in both tests.

lib/src/components/NxCollapsibleItemsSelect/NxCollapsibleMultiSelect/__tests__/NxCollapsibleMultiSelect.test.tsx
cindyko226: As discussion, the RTL currently does not consider placeholder as an accessible name https://github.com/testing-library/dom-testing-library/issues/926 while it is defined in [HTML Accessibility API Mappings 1.0](https://www.w3.org/TR/html-aam-1.0/#input-type-text-input-type-password-input-type-number-input-type-search-input-type-tel-input-type-email-input-type-url-and-textarea-element-accessible-name-computation)(working draft), before RTL has any further updates, will just have the tests for checking it has placeholder attrs.

rpokorny (approved):

PR #930, opened 2023-01-25
NxBinaryDonutChart RTL Tests - RSC-1389

rpokorny:
https://issues.sonatype.org/browse/RSC-1389


lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
```diff
@@ -4,304 +4,224 @@
...
-        });
-
-    it('renders only a background circle path when value is negative', function() {
-      const circles = getShallowComponent({value: -50}).find('circle');
-      const paths = getShallowComponent({value: -50}).find('path');
-
-      expect(circles.length).toBe(0);
-      expect(paths.length).toBe(1);
-      expect(paths.at(0)).toHaveClassName('.nx-binary-donut-chart__background');
+    it('renders a correct arc path when 0 < value < 100', function() {
```
entroform, line 62: Just curious if we should have a test when the value is above maxVal?

entroform (approved):
Everything else looks good to me!

lib/src/components/NxBinaryDonutChart/__tests__/NxDonutChart.test.tsx
rpokorny: > Just curious if we should have a test when the value is above maxVal?

Yeah, and even more importantly one for when value is _at_ `maxVal`.  Granted, in either case there's a limited amount that it makes sense to test without visuals.  But nonetheless, [fe7675a](https://github.com/sonatype/sonatype-react-shared-components/pull/930/commits/fe7675a26bc30dcec2fc09117a6708ac3c60bcb0)

PR #931, opened 2023-01-25
RTL Tabs - RSC-1422

entroform:
https://issues.sonatype.org/browse/RSC-1422


gallery/visualtests/nxTabs.js
```diff
@@ -62,6 +63,16 @@ describe('NxTabs', function() {
             `${tabSimpleExampleSelector} .nx-tab:nth-child(2)`));
   });
 
+  it('shows tooltip when text is truncated', async function() {
```
rpokorny, line 66: See the third param to `hoverTest` in `testUtils`.  I believe it could simplify this test substantially.

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -5,103 +5,76 @@
...
-      <NxTabs onTabSelect={() => {}} activeTab={-1} className="test">
-        <NxTabList>
-          <NxTab>Tab 0</NxTab>
-        </NxTabList>
-        <NxTabPanel>Content 0</NxTabPanel>
-      </NxTabs>
-    );
+  it('merges any passed in className', function() {
+    const tabsWithAddedClassName = renderEl({ activeTab: -1, className: 'foo' });
+    const tabs = renderEl({ activeTab: -1 })!;
```
rpokorny, line 32: Why is `activeTab` needed in this test?

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -5,103 +5,76 @@
...
-        </NxTabList>
-        <NxTabPanel>Content 0</NxTabPanel>
-      </NxTabs>
-    );
+  it('renders no tab contents when none are active', function() {
+    const { getByRole, queryByRole } = quickRender({ activeTab: -1 });
 
-    expect(component.find(NxTab)).not.toHaveClassName('active');
-    expect(component.find(NxTabPanel)).toBeEmptyRender();
+    expect(getByRole('tab').classList.contains('active')).toBeFalsy();
```
rpokorny, line 64: Check `aria-selected` rather than this class

rpokorny: > Check aria-selected rather than this class

One way to do that is with the extra parameters to `queryByRole`. https://testing-library.com/docs/queries/byrole/

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -5,103 +5,76 @@
...
-  it('renders no tab contents when no active tab is specified', function () {
-    const component = mount(
-      <NxTabs onTabSelect={() => {}}>
-        <NxTabList>
-          <NxTab>Tab 0</NxTab>
-        </NxTabList>
-        <NxTabPanel>Content 0</NxTabPanel>
-      </NxTabs>
-    );
+  it('renders no tab contents when no active tab is specified', function() {
```
rpokorny, line 68: This doesn't really seem like a different test from the one above.  `activeTab=-1` and no `activeTab` at all seem like two different ways to specify the same thing.  Maybe these tests should be merged.

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -111,38 +84,47 @@ describe('NxTabs', function () {
...
 
-    component.find('[role="tab"]').simulate('click');
+    const { getByRole } = quickRender({
+      activeTab: 0,
+      onTabSelect
+    });
+
+    await user.click(getByRole('tab'));
 
     expect(onTabSelect).toHaveBeenCalledWith(0);
```
rpokorny, line 105: Probably more thorough to have the index of the clicked tab and the already-set `activeTab` be two different values

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -111,38 +84,47 @@ describe('NxTabs', function () {
...
-    component.find('[role="tab"]').simulate('keypress', { key: ' ' });
+    const firstTab = getAllByRole('tab')[0];
 
-    expect(onTabSelect).toHaveBeenCalledWith(0);
+    firstTab.focus();
+    await user.keyboard('{ArrowRight}{Enter}');
+
+    expect(onTabSelect).toHaveBeenCalledWith(1);
   });
 });
```
rpokorny, line 130: Needs tests of the `aria-controls` relationship between tabs and tabpanels

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -5,103 +5,76 @@
...
+  });
 
-    expect(component.prop('id')).toMatch('my-tabs');
+  it('renders using as specific id', function() {
+    const commponent = renderEl({
+      id: 'my-tabs',
+      activeTab: -1
+    })!;
+    expect(commponent.id).toMatch('my-tabs');
   });
```
rpokorny, line 59: Also test how an explicit id affects the generated ids of the children

rpokorny:
I'm thinking the individual test files for `NxTab`, `NxTabPanel`, and `NxTabList` can be removed.  Each of those components only really works in conjunction with `NxTabs`, and testing everything about all of them together (most of which we're already doing) in the `NxTabs` test file makes more sense to me

lib/src/components/NxTabs/__tests__/NxStatefulTabs.test.tsx
```diff
@@ -5,29 +5,81 @@
...
+    for (const cls of Array.from(tabs.classList)) {
+      expect(tabsWithAddedClassName).toHaveClass(cls);
+    }
   });
 
-  it('selects the default tab initially', function () {
-    const component = shallow(
+  it('selects no tab when none is provided', function() {
+    const { getByRole } = quickRender();
+    expect(getByRole('tab')).not.toHaveClass('active');
```
rpokorny, line 44: `aria-selected`

rpokorny:
Don't forget to apply most of my comments above to `NxStatefulTabs.test.tsx` as well as it has most of the same issues

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > Also test how an explicit id affects the generated ids of the children

Updated here: 109736a065caf6fe1852838fb3261f4c5eb097bf

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > Needs tests of the `aria-controls` and `aria-labelledby` relationships between tabs and tabpanels

Added here: 34ce2797159b5caf66689b8c31a31c1a92ef9a3f

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > This doesn't really seem like a different test from the one above. `activeTab=-1` and no `activeTab` at all seem like two different ways to specify the same thing. Maybe these tests should be merged.

Fixed: fc60025ae197c6b928bcd7b180351ad7dd1dc690

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > Probably more thorough to have the index of the clicked tab and the already-set `activeTab` be two different values

Updated here: 1b48c75f47c0586c2c307ad5c17a75c7ffa6e1cc

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -111,38 +84,47 @@ describe('NxTabs', function () {
       </NxTabs>
     );
 
-    expect(component.find('[role="tab"]').first()).not.toHaveClassName('active');
-    expect(component.find('[role="tab"]').last()).toHaveClassName('active');
-    expect(component.find('[role="tabpanel"]')).toHaveText('Content 1');
+    const tabs = getAllByRole('tab');
+    expect(tabs[1]).toHaveTextContent('Tab 1');
+    expect(tabs[0].classList.contains('active')).toBe(false);
+    expect(tabs[1].classList.contains('active')).toBe(true);
```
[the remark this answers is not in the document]
entroform: > `aria-selected`

Updated: 9ee3fecf501160880785f087d981f677b0a2c974

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > > Check aria-selected rather than this class
> 
> One way to do that is with the extra parameters to `queryByRole`. https://testing-library.com/docs/queries/byrole/

Ooo thanks! updated here: 9ee3fecf501160880785f087d981f677b0a2c974

lib/src/components/NxTabs/__tests__/NxStatefulTabs.test.tsx
entroform: > `aria-selected`

Updated as part of: 6bacf16984a6f52b3dcc2726b315329746467423

entroform:
> Don't forget to apply most of my comments above to `NxStatefulTabs.test.tsx` as well as it has most of the same issues

Thanks I updated NxStatefulTabs here: 6bacf16984a6f52b3dcc2726b315329746467423

entroform:
> I'm thinking the individual test files for `NxTab`, `NxTabPanel`, and `NxTabList` can be removed. Each of those components only really works in conjunction with `NxTabs`, and testing everything about all of them together (most of which we're already doing) in the `NxTabs` test file makes more sense to me

I removed them here: 11239546ffca3c1ed0b8ca5cfb460b455a2277e4

gallery/visualtests/nxTabs.js
entroform: > See the third param to `hoverTest` in `testUtils`. I believe it could simplify this test substantially.

Simplified here:  d9f6814fe4fdadf3b4d3d2b2adac68746afb588e

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -111,38 +84,98 @@ describe('NxTabs', function () {
...
+          <NxTabPanel>Content 1</NxTabPanel>
+        </NxTabs>
+      );
+    };
+
+    const { getByRole } = render(<Fixture />);
+    const getActiveTabControls = () => getByRole('tab', { selected: true }).getAttribute('aria-controls');
+
+    expect(getByRole('tabpanel').id).toEqual(getActiveTabControls());
+    expect(getByRole('tabpanel')).toHaveAccessibleName('Tab 0 Tab 0');
```
entroform, line 174: I noticed that the hidden part is part of the accessible name.
Do you think we should add `aria-hidden` to that?

rpokorny: > I noticed that the hidden part is part of the accessible name.
Do you think we should add aria-hidden to that?

Check to see if it's part of the accessible name in the actual browser - it might be different there due to the inclusion of the CSS.  If it's part of the accessible name there too though, then yeah we should fix that.

On the other hand if this is just an issue in the unit tests (due to the lack of CSS) I'd work around it by using a regex for the a11y name so that it doesn't have to be an exact, full match

entroform: > > I noticed that the hidden part is part of the accessible name.
> > Do you think we should add aria-hidden to that?
> 
> Check to see if it's part of the accessible name in the actual browser - it might be different there due to the inclusion of the CSS. If it's part of the accessible name there too though, then yeah we should fix that.
> 
> On the other hand if this is just an issue in the unit tests (due to the lack of CSS) I'd work around it by using a regex for the a11y name so that it doesn't have to be an exact, full match

Thanks, it looks okay on the browser:
I've updated it to use regex here:
10c589cc6b4ec332aa20394fd07bf3f83089954a

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -4,104 +4,77 @@
...
         </NxTabList>
         <NxTabPanel>Content 0</NxTabPanel>
       </NxTabs>
     );
 
-    expect(component.find(NxTab)).not.toHaveClassName('active');
-    expect(component.find(NxTabPanel)).toBeEmptyRender();
+    expect(container.firstElementChild!.id).toMatch('my-tabs');
+    expect(getByRole('tab')!.id).toMatch('my-tabs');
+    expect(getByRole('tabpanel')!.id).toMatch('my-tabs');
```
rpokorny, line 65: Maybe make sure that each of these doesn't exactly match `my-tabs` nor exactly match each other

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
```diff
@@ -111,38 +84,98 @@ describe('NxTabs', function () {
...
+    const firstTab = getAllByRole('tab')[0];
 
-    expect(onTabSelect).toHaveBeenCalledWith(0);
+    firstTab.focus();
+    await user.keyboard('{ArrowRight}{Enter}');
+
+    expect(onTabSelect).toHaveBeenCalledWith(1);
+  });
+
+  it('sets id and aria labelledby to panel to reference active tab aria-controls', async function() {
```
rpokorny, line 153: This test seems like it could be simplified.  Rather than using a stateful fixture and user interaction, why not just do two renders (or a render and rerender) with different props?

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > Maybe make sure that each of these doesn't exactly match `my-tabs` nor exactly match each other

yup: updated here: f42f83d3237e208639e1aabf8f86c0ab7c613e06

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
entroform: > This test seems like it could be simplified. Rather than using a stateful fixture and user interaction, why not just do two renders (or a render and rerender) with different props?

Yup, updated here: b51e5f8197b8378624b69466c5584a03f5fc9052

lib/src/components/NxTabs/__tests__/NxTabs.test.tsx
rpokorny: > yup: updated here: https://github.com/sonatype/sonatype-react-shared-components/commit/f42f83d3237e208639e1aabf8f86c0ab7c613e06

I misspoke slightly above.  The id of the `nx-tabs` itself _should_ exactly match "my-tabs", and this test should make sure of that.

entroform: > > yup: updated here: [f42f83d](https://github.com/sonatype/sonatype-react-shared-components/commit/f42f83d3237e208639e1aabf8f86c0ab7c613e06)
> 
> I misspoke slightly above. The id of the `nx-tabs` itself _should_ exactly match "my-tabs", and this test should make sure of that.

Ops, fixed here (Also updated stateful test): de8fa9f56f2a050b7784c21f760f02ec84bee096

rpokorny (approved):

PR #933, opened 2023-01-26
RTL Conversion - NxSearchDropdown - RSC-1369

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1369


rpokorny:
One thing I did in `NxTransferList` that I think makes sense here too is to add (an test on) a role of `group` to the top-level element

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
-        input = component.find(NxFilterInput);
+    const el = quickRender({ searchText: 'foo' }),
+        input = el.getByRole('searchbox');
 
-    expect(input).toHaveProp('value', 'foo');
+    expect(input).toHaveAttribute('value', 'foo');
   });
 
-  it('calls onSearchTextChange whenver the input\'s onChange event fires', function() {
+  it('calls onSearchTextChange whenver the input\'s onChange event fires', async function() {
```
rpokorny, line 63: This should be reworded in terms of user-facing events e.g. whenever the user types in the searchbox (and then other tests for other was the searchbox's value could change such as pressing ESC or clicking the clear button). Accordingly the test should use `userEvent` not `fireEvent`.

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
-    input.simulate('change', 'asdf');
-
-    expect(onSearchTextChange).toHaveBeenCalledWith('asdf');
+    fireEvent.change(input, { target: { value: 'aasd' }});
+    expect(onSearchTextChange).toHaveBeenCalledWith('aasd');
   });
 
-  it('calls onSearch whenver the input\'s onChange event fires with a value that differs after trimming, ' +
-      'passing the trimmed value', function() {
+  it('calls onSearch whenever the input\'s onChange event fires with a value that differs after trimming, ' +
```
rpokorny, line 74: Also should be in terms of user events and use `userEvent`.

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
-    expect(getShallow({ searchText: 'foo', disabled: undefined })).toContainMatchingElement(NxDropdownMenu);
-    expect(getShallow({ searchText: 'foo', disabled: null })).toContainMatchingElement(NxDropdownMenu);
-    expect(getShallow({ searchText: 'foo' })).toContainMatchingElement(NxDropdownMenu);
+  it('renders dropdown with role="menu" if there are matching elements', function() {
+    const el = quickRender({ searchText: 'foo', matches: [{ id: '1', displayName: '1' }]});
+
+    expect(el.getByRole('menu')).toBeInTheDocument();
   });
 
   it('sets aria-hidden to true on the dropdown menu if the searchText is empty or disabled is true', function() {
```
rpokorny, line 131: I'm not sure it's the most useful to have RTL search for the hidden menu.  Better maybe to just assert that there is no non-hidden menu in these cases

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
   it('sets aria-live on the dropdown menu to "polite"', function() {
-    expect(getShallow().find(NxDropdownMenu)).toHaveProp('aria-live', 'polite');
+    expect(quickRender().getByRole('alert', { hidden: true })).toHaveAttribute('aria-live', 'polite');
   });
 
   it('sets aria-busy on the dropdown menu if loading is true', function() {
-    expect(getShallow().find(NxDropdownMenu)).toHaveProp('aria-busy', false);
-    expect(getShallow({ loading: true }).find(NxDropdownMenu)).toHaveProp('aria-busy', true);
+    expect(quickRender().getByRole('alert', { hidden: true })).toHaveAttribute('aria-busy', 'false');
+    expect(quickRender({ loading: true }).getByRole('alert', { hidden: true })).toHaveAttribute('aria-busy', 'true');
```
rpokorny, line 151: Cleaner to construct a case where the menu isn't hidden.  Whether it's busy or not while it is hidden doesn't really matter to anything.

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
-        .toHaveProp('role', 'menu');
-  });
+    const dropdownMenu = quickRender({ searchText: 'asdf', error: 'foo' }).getAllByRole('alert')[0];
+    expect(dropdownMenu).toBeInTheDocument();
 
-  it('sets the alert role on the NxDropdownMenu when it is in loading, error, or empty states', function() {
-    expect(getShallow({ searchText: 'asdf', matches: [] }).find(NxDropdownMenu)).toHaveProp('role', 'alert');
-    expect(getShallow({ searchText: 'asdf', loading: true }).find(NxDropdownMenu)).toHaveProp('role', 'alert');
-    expect(getShallow({ searchText: 'asdf', error: 'foo' }).find(NxDropdownMenu)).toHaveProp('role', 'alert');
+    const errorAlert = within(dropdownMenu).getByRole('alert');
```
rpokorny, line 161: What are these two lines trying to test? Isn't `errorAlert` just going to end up being the same element as `dropdownMenu`?

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
+    expect(errorAlert).toBeInTheDocument();
   });
 
-  it('sets an id on the NxDropdownMenu and references it in the search boxes aria-controls', function() {
-    const component = getMounted({ searchText: 'foo' }),
-        filterInput = component.find(NxFilterInput).find('input'),
-        dropdown = component.find('.nx-dropdown-menu');
+  it('sets an id on the dropdown and references it in the search boxe\'s aria-controls', function() {
```
rpokorny, line 165:
```suggestion
  it('sets an id on the dropdown and references it in the search box\'s aria-controls', function() {
```

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
-    const loading = getShallow({ searchText: 'foo', loading: true }),
-        notLoading = getShallow({ searchText: 'foo' }),
-        explicitNotLoading = getShallow({ searchText: 'foo', loading: false });
+  it('shows loading text on the dropdown when loading prop is true', function() {
+    const el = quickRender({ searchText: 'foo', loading: true });
 
-    expect(loading.find(NxLoadWrapper)).toHaveProp('loading', true);
-    expect(notLoading.find(NxLoadWrapper)).toHaveProp('loading', undefined);
-    expect(explicitNotLoading.find(NxLoadWrapper)).toHaveProp('loading', false);
+    expect(el.getByRole('status')).toHaveTextContent('Loading…');
```
rpokorny, line 184: Should probably assert that this `status` is within the `alert`

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,225 +4,203 @@
...
+    const el = quickRender({ searchText: 'foo', error: 'bar' });
 
-    expect(error.find(NxLoadWrapper)).toHaveProp('error', 'bar');
-    expect(noError.find(NxLoadWrapper)).toHaveProp('error', undefined);
+    const dropdownMenu = el.getAllByRole('alert')[0];
+    expect(within(dropdownMenu).getByRole('alert')).toHaveTextContent('bar');
   });
 
-  it('fires onSearch with the searchText when the load wrappers retryHandler is triggered', function() {
+  it('fires onSearch with the searchText when the load wrappers retryHandler is triggered', async function() {
```
rpokorny, line 194:
```suggestion
  it('fires onSearch with the searchText when the retry button is clicked', async function() {
```

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -232,60 +210,48 @@ describe('NxSearchDropdown', function() {
           { id: '1', displayName: 'One' },
```
rpokorny, line 210: I can't comment that line, but the name of this test should be adjusted.  The load wrapper is an impl detail.  How about 'renders buttons with type "button" for each match'

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -232,60 +210,48 @@ describe('NxSearchDropdown', function() {
...
-    expect(notLoadingLoadWrapper.children().length).toBe(2);
-    expect(notLoadingLoadWrapper.childAt(0)).toMatchSelector('button.nx-dropdown-button');
-    expect(notLoadingLoadWrapper.childAt(0)).toHaveProp('type', 'button');
-    expect(notLoadingLoadWrapper.childAt(0)).toHaveText('One');
-    expect(notLoadingLoadWrapper.childAt(1)).toMatchSelector('button.nx-dropdown-button');
-    expect(notLoadingLoadWrapper.childAt(1)).toHaveText('Two');
+        el = quickRender({ matches, searchText: 'foo' }),
+        dropdown = el.getByRole('menu');
+
+    expect(dropdown.children.length).toBe(2);
```
rpokorny, line 216: Cleaner to count the `menuitem`s

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > What are these two lines trying to test? Isn't `errorAlert` just going to end up being the same element as `dropdownMenu`?

I am trying to assert that the `errorAlert (.nx-alert)` is going to be within the dropdown menu `(.nx-dropdown-menu)` that has a role of `alert` as well, based on this DOM structure:

image

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > This should be reworded in terms of user-facing events e.g. whenever the user types in the searchbox (and then other tests for other was the searchbox's value could change such as pressing ESC or clicking the clear button). Accordingly the test should use `userEvent` not `fireEvent`.

I've re-worded tests so that they are user-facing events as suggested: cb95291101d94947a0c5dc2843368a1dad4501bd

avnvbhatta:
> One thing I did in `NxTransferList` that I think makes sense here too is to add (an test on) a role of `group` to the top-level element

Added here: 4be373bf238b166d5797fe9080f15cd05b2cc3a4

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > Also should be in terms of user events and use `userEvent`. I _think_ the `initalSelectionStart` parameter on `userEvent.type` can be used to enter text (spaces for instance) in the input at positions other than the end.

I've had a lot of difficulty using `userEvent` in this component. Not sure if I'm not implementing it correctly, but I can't seem to change the input field with something like `user.type(input, '{backspace}'`. I also couldn't find documentation about  `initalSelectionStart`, so I tried using something like `user.type(input, '{arrowleft>4/} a'` to make 4 arrow left key presses to modify the start of the text, but it doesn't seem to work. I might need some help on this.

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > I'm not sure it's the most useful to have RTL search for the hidden menu. Better maybe to just assert that there is no non-hidden menu in these cases

Changed here: 00706381d6b2f3736a8182926ef36f1ec06d864e

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > Cleaner to construct a case where the menu isn't hidden. Whether it's busy or not while it is hidden doesn't really matter to anything.

Changed here: 845fc1b3735150e5436913ee828c14e482d1c901

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > Should probably assert that this `status` is within the `alert`

Added here: 063165dfc3227116141829180b041706b4eae005

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > Cleaner to count the `menuitem`s

Changed here: eaf2f40083f18c80c577e21f0f8eca4a141519b3

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > I can't comment that line, but the name of this test should be adjusted. The load wrapper is an impl detail. How about 'renders buttons with type "button" for each match'

Should be changed here: https://github.com/sonatype/sonatype-react-shared-components/pull/933/files#diff-8faf831d9f7a4427b8d4674137223bd7eb914e2b245eddc1308562fa0ce5a3f9R212-R218

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
rpokorny: > I've re-worded tests so that they are user-facing events as suggested: https://github.com/sonatype/sonatype-react-shared-components/commit/cb95291101d94947a0c5dc2843368a1dad4501bd

I notice you removed mention of the props function names (e.g. `onSearch` as well).  That is _not_ what I meant.  The component's props API is one external surface.  The user is another.  The goal of these tests is to make sure that when the user does something user-y that the props API gets triggered accordingly.  So the relevant props API parts should still be in the test names.

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
rpokorny: > I've had a lot of difficulty using userEvent in this component. Not sure if I'm not implementing it correctly, but I can't seem to change the input field with something like user.type(input, '{backspace}'. I also couldn't find documentation about initalSelectionStart, so I tried using something like user.type(input, '{arrowleft>4/} a' to make 4 arrow left key presses to modify the start of the text, but it doesn't seem to work. I might need some help on this.

Did you try `user.keyboard`? From the [user.type docs](https://testing-library.com/docs/user-event/utility#type):
 
> You should use [keyboard()](https://testing-library.com/docs/user-event/keyboard) if you want to just simulate pressing buttons on the keyboard.
> You can use type() if you just want to conveniently insert some text into an input field or textarea.

That said, initialSelectionStart` is only a parameter on `type`, and not on `keyboard`. So you'd have to figure out a different way for that part.  I'll look into it

rpokorny: > That said, initialSelectionStart is only a parameter on type, and not on keyboard. So you'd have to figure out a different way for that part. I'll look into it

See 0edb8b1bf

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
rpokorny: > I am trying to assert that the errorAlert (.nx-alert) is going to be within the dropdown menu (.nx-dropdown-menu) that has a role of alert as well, based on this DOM structure:

Having multiple nested `alert` elements was not deliberate I don't think, and should be fixed.  Adjusting the line that computes `dropdownMenuRole` in `NxSearchDropdown` so that it does not add the `alert` role in the error state should do it.

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > > I've re-worded tests so that they are user-facing events as suggested: [cb95291](https://github.com/sonatype/sonatype-react-shared-components/commit/cb95291101d94947a0c5dc2843368a1dad4501bd)
> 
> I notice you removed mention of the props function names (e.g. `onSearch`) as well. That is _not_ what I meant. The component's props API is one external surface. The user is another. The goal of these tests is to make sure that when the user does something user-y that the props API gets triggered accordingly. So the relevant props API parts should still be in the test names.

I see what you mean. That makes sense. Added back relevant keywords: e25c45117205cd3c01d45f0a3adb3263833a1a70

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > > I am trying to assert that the errorAlert (.nx-alert) is going to be within the dropdown menu (.nx-dropdown-menu) that has a role of alert as well, based on this DOM structure:
> 
> Having multiple nested `alert` elements was not deliberate I don't think, and should be fixed. Adjusting the line that computes `dropdownMenuRole` in `NxSearchDropdown` so that it does not add the `alert` role in the error state should do it.

Changed here: 551d7a62b46f821f20b029b3babed2169fa559af

rpokorny: > Changed here: https://github.com/sonatype/sonatype-react-shared-components/commit/551d7a62b46f821f20b029b3babed2169fa559af

I suspect `menu` isn't right for the error case.  I'd just not set a role on the `.nx-dropdown-menu` itself at all for that case

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,288 +4,255 @@
...
+    expect(onSearchTextChange).toHaveBeenCalledWith('a');
   });
 
-  it('calls onSearch whenver the input\'s onChange event fires with a value that differs after trimming, ' +
-      'passing the trimmed value', function() {
-    const onSearch = jest.fn(),
-        component = getShallow({ searchText: 'foo ', onSearch }),
-        input = component.find(NxFilterInput);
+  it('calls onSearch whenever the searchbox text changes with a value that differs after trimming, ' +
+    'passing the trimmed value '
```
rpokorny, line 84:
```suggestion
    'passing the trimmed value'
```

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,288 +4,255 @@
...
+  , async function() {
```
rpokorny, line 85: Odd formatting here

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > > Changed here: [551d7a6](https://github.com/sonatype/sonatype-react-shared-components/commit/551d7a62b46f821f20b029b3babed2169fa559af)
> 
> I suspect `menu` isn't right for the error case. I'd just not set a role on the `.nx-dropdown-menu` itself at all for that case

Changed and added test here: ae9b6c9dff97113a8bced5d2a1ba5ef1c34c72e2

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
```diff
@@ -4,85 +4,349 @@
...
+
+  it('calls onSearch whenever the input\'s onChange event fires with a value that differs after trimming, ' +
+  'passing the trimmed value', async function() {
+    const onSearch = jest.fn(),
+        el = quickRender({ defaultSearchText: 'foo ', onSearch }),
+        input = el.getByRole('searchbox');
+
+    expect(onSearch).not.toHaveBeenCalled();
+
+    fireEvent.change(input, { target: { value: 'foo' }});
```
rpokorny, line 106: Don't forget to use `userEvent` here too

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,288 +4,255 @@
...
+  it('sets the searchText as the value of the searchbox', function() {
+    const el = quickRender({ searchText: 'foo' }),
+        input = el.getByRole('searchbox');
 
-    expect(input).toHaveProp('value', 'foo');
+    expect(input).toHaveAttribute('value', 'foo');
   });
 
-  it('calls onSearchTextChange whenver the input\'s onChange event fires', function() {
+  it('calls onSearchTextChange whenver the user types in the searchbox', async function() {
```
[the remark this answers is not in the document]
avnvbhatta: > 

c4058a024987c9a730f1b0fa8e2cb2ac7227bb44

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > 

c4058a024987c9a730f1b0fa8e2cb2ac7227bb44

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > Odd formatting here

c4058a024987c9a730f1b0fa8e2cb2ac7227bb44

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
```diff
@@ -4,85 +4,349 @@
...
+    expect(el.getByRole('alert')).toHaveTextContent('asdfasdf');
+  });
+
+  it('renders dropdown with role="menu" if there are matching elements', function() {
+    const el = quickRender({ defaultSearchText: 'foo', matches: [{ id: '1', displayName: '1' }]});
+
+    expect(el.getByRole('menu')).toBeInTheDocument();
+  });
+
+  it('sets aria-hidden to true on the dropdown menu if the defaultSearchText is empty or disabled is true', function() {
```
rpokorny, line 155: Just test that no visible dropdown is present

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
```diff
@@ -4,85 +4,349 @@
...
+  });
+
+  it('shows error text when the error prop is provided', function() {
+    const el = quickRender({ defaultSearchText: 'foo', error: 'bar' });
+
+    const dropdownMenu = el.getAllByRole('alert')[0];
+    expect(within(dropdownMenu).getByRole('alert')).toHaveTextContent('bar');
+  });
+
+  it('fires onSearch with the defaultSearchText when the load wrappers retryHandler is triggered', async function() {
```
rpokorny, line 218: It should fire it with the current text input value, which isn't necessarily `defaultSearchText`

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
```diff
@@ -4,85 +4,349 @@
...
+    expect(onSelect).not.toHaveBeenCalled();
+
+    const menuItem = el.getByRole('menuitem', { name: 'Two' });
+
+    await user.click(menuItem);
+
+    expect(onSelect).toHaveBeenCalledWith({ id: '2', displayName: 'Two' });
+  });
+
+  it('calls onSearch with the current trimmed defaultSearchText if focus enters the  ' +
```
rpokorny, line 263:
```suggestion
  it('calls onSearch with the current trimmed text input value if focus enters the  ' +
```

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
```diff
@@ -4,288 +4,253 @@
...
-    const loading = getShallow({ searchText: 'foo', loading: true }),
-        notLoading = getShallow({ searchText: 'foo' }),
-        explicitNotLoading = getShallow({ searchText: 'foo', loading: false });
+  it('does not set a role on dropdown menu when error prop is provided', function() {
+    const el = quickRender({ searchText: 'foo', error: 'bar' });
 
-    expect(loading.find(NxLoadWrapper)).toHaveProp('loading', true);
-    expect(notLoading.find(NxLoadWrapper)).toHaveProp('loading', undefined);
-    expect(explicitNotLoading.find(NxLoadWrapper)).toHaveProp('loading', false);
+    expect(el.getByRole('alert').parentElement).not.toHaveAttribute('role');
```
rpokorny, line 192:
```suggestion
    expect(el.queryByRole('menu')).not.toBeInTheDocument();
```

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
avnvbhatta: > Don't forget to use `userEvent` here too

Yup, I've updated the stateful version to use `userEvent` here: aa5f41008258e8af27b6dc38fb812a28a69195f4

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
avnvbhatta: > It should fire it with the current (trimmed) text input value, which isn't necessarily `defaultSearchText`

Updated here: https://github.com/sonatype/sonatype-react-shared-components/pull/933/commits/aa5f41008258e8af27b6dc38fb812a28a69195f4#diff-90808a1737028207ca6500c1a317cec13f908eeab0b3fa61ae10a1bc93e9f19cR223-R244

lib/src/components/NxSearchDropdown/stateful/__tests__/NxStatefulSearchDropdown.test.tsx
avnvbhatta: > current trimmed defaultSearchText

Changed here: https://github.com/sonatype/sonatype-react-shared-components/pull/933/commits/aa5f41008258e8af27b6dc38fb812a28a69195f4#diff-90808a1737028207ca6500c1a317cec13f908eeab0b3fa61ae10a1bc93e9f19cR266-R267

lib/src/components/NxSearchDropdown/__tests__/NxSearchDropdown.test.tsx
avnvbhatta: > 

Changed here: 28a1b9550fa9cdea68b55393f6b4dc3b02023701

PR #934, opened 2023-01-27
RTL NxToggle - RSC-1385

entroform:
https://issues.sonatype.org/browse/RSC-1385


lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
```diff
@@ -21,46 +17,38 @@ describe('NxToggle', function() {
...
-    const checkbox = screen.getByRole<HTMLInputElement>('switch', { name: 'Enables whales' });
+  const quickRender = rtlRender<Props>(NxToggle, simpleProps);
+  const renderEl = rtlRenderElement(NxToggle, simpleProps);
 
-    expect(checkbox.tagName).toEqual('INPUT');
-
-    const label = checkbox.labels?.[0];
-
-    expect(label?.textContent).toEqual('Enables whales');
+  it('renders a swith with the correct label', function() {
```
rpokorny, line 23:
```suggestion
  it('renders a switch with the correct label', function() {
```

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
```diff
@@ -21,46 +17,38 @@ describe('NxToggle', function() {
...
 
-    const root = container.children[0];
+    expect(componentWithAddedClass).toHaveClass('foo');
 
-    expect(root).toHaveClass('foo');
-    expect(root).toHaveClass('nx-toggle');
+    for (const cls of Array.from(component.classList)) {
+      expect(componentWithAddedClass).toHaveClass(cls);
+    }
   });
```
rpokorny, line 43: Needs a test for general attributes on the top-level element

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
```diff
@@ -21,46 +17,38 @@ describe('NxToggle', function() {
...
 
   it('calls its onChange prop when the label is clicked', async function() {
     const user = userEvent.setup();
 
     const onChange = jest.fn();
 
-    const { container } = rtlRender({ onChange });
-    const label = container.querySelector('label') as HTMLLabelElement;
+    const { container } = quickRender({ onChange });
+    const label = container.querySelector('label')!;
```
rpokorny, line 51: I've wondered about this myself.  `<label>`s have a funny place in ARIA - they have no default role, may not be given a role, and may not be named.  They basically only exist to provide names to other elements, as far as aria is concerned.  At the same time, the fact that clicking a `<label>` will activate its associated form control is built into HTML.  These two facts together have caused me to conclude that we don't really need to test clicking on `<label>`s separately from clicking on their associated form fields.  So for this test I'd query the `switch` role and click that.  And not worry about the `<label>`

rpokorny:
Wait, why were any changes made here considering that these were already RTL tests? Improving existing RTL tests is not the goal of this push.  This ticket should have been closed without work.

entroform:
> Wait, why were any changes made here considering that these were already RTL tests? Improving existing RTL tests is not the goal of this push. This ticket should have been closed without work.

Oh part of this ticket is also converting the NxStatefulToggle to RTL (it's still in enzyme)

lib/src/components/NxToggle/__tests__/NxToggle.test.tsx
entroform: > Needs a test for general attributes on the top-level element

Added it here: 0368d86cfb8df2584b7119abd42356788ab4a7dd

lib/src/components/NxToggle/__tests__/NxStatefulToggle.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it('updates isChecked prop on NxToggle when the control is toggled', async function() {
+    const user = userEvent.setup();
+
+    const checkbox = quickRender(simpleProps).getByRole('switch')!;
+
+    expect(checkbox).not.toBeChecked();
+    await user.click(checkbox);
+    expect(checkbox).toBeChecked();
```
rpokorny, line 72: Toggle it the other way too

lib/src/components/NxToggle/__tests__/NxStatefulToggle.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it('calls its onChange prop when the input fires a change event', async function() {
```
rpokorny, line 75: This seems like basically the same thing as the test below.  I like that test's name better though.

rpokorny:
> I moved the stateful test next to the stateless one here

I disagree with that move.  The test file should be located relative to the component that it is testing.

entroform:
> > I moved the stateful test next to the stateless one here
> 
> I disagree with that move. The test file should be located relative to the component that it is testing.

Ops I think I got confused from NxTabs test, they have it on the same folder: b8bc6d0059d4a1e01b65e63634922e458685e1f6

lib/src/components/NxToggle/__tests__/NxStatefulToggle.test.tsx
entroform: > Toggle it the other way too

Updated that here: f04b9f3f6eba96ba9f0dfaaf1d0fdd6e15ba3cc4

lib/src/components/NxToggle/__tests__/NxStatefulToggle.test.tsx
entroform: > This seems like basically the same thing as the test below. I like that test's name better though.

Yup: f04b9f3f6eba96ba9f0dfaaf1d0fdd6e15ba3cc4

rpokorny (approved):

PR #935, opened 2023-01-27
RSC-1421-PageHeader - convert tests to RTL

achen5:
https://issues.sonatype.org/browse/RSC-1421


lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
-import * as enzymeUtils from '../../../__testutils__/enzymeUtils';
 import NxPageHeader from '../NxPageHeader';
-import AbstractNxPageHeader from '../../AbstractNxPageHeader/AbstractNxPageHeader';
 
-describe('NxPageHeader', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent(NxPageHeader, {});
+import { screen } from '@testing-library/react';
+import { rtlRenderElement } from '../../../__testutils__/rtlUtils';
+
+fdescribe('NxPageHeader', function() {
```
rpokorny, line 13:
```suggestion
describe('NxPageHeader', function() {
```

rpokorny: Jest has ways to filter which tests are running on the fly, so you don't generally need to do this.  For instance, in `yarn test-watch`, it has menu options allowing you to filter by filename or test name.  On the command line for `yarn jest`, you can also pass the filename as an arg or use the `--testNamePattern` option.

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
+  const renderEl = rtlRenderElement(NxPageHeader, {});
 
-  it('renders an AbstractNxPageHeader', function() {
-    expect(getShallowComponent()).toMatchSelector(AbstractNxPageHeader);
+ it('renders a simple NxPageHeader', function() {
+    renderEl();
+    const logo = screen.getByRole('img')
+    expect(logo).toBeInTheDocument();
+    
   });
```
rpokorny, line 22: One of the first tests we that we usually do is for whatever role is expected on the top-level element rendered by the component.  In this case that would be `<header>`.  The [ARIA in HTML](https://www.w3.org/TR/html-aria/#el-header) specs say that this element gets a role of `banner` unless it is included within certain other elements (which `NxPageHeader` shouldn't be).  So we should test for the `banner` role.

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
 
-  it('passes its links, homeLink, and children props to the AbstractNxPageHeader', function() {
-    const props = {
-          links: [{ name: 'foo', href: '#bar' }],
-          homeLink: '#home',
-          children: <button>Click Here For a Free iPhone</button>
-        },
-        component = getShallowComponent(props);
+  it('renders a custom logo when given logo prop', function() {
+    const logo = renderEl({ logo: { path: 'foo', alt: 'bar' } })?.querySelector("img");
```
rpokorny, line 25: This should be written in terms of RTL queries and roles, not CSS selectors and `querySelector`

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
+
+    expect(logo).toHaveAttribute('src', 'foo');
```
rpokorny, line 27: This test should test the `alt` attribute as well

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
+  });
+
+  it('renders logo link when given homeLink prop', function() {
+    renderEl({ homeLink: '#home'});    
 
-    expect(component).toHaveProp('links', props.links);
-    expect(component).toHaveProp('homeLink', props.homeLink);
-    expect(component.children()).toMatchElement(props.children);
+    const homeLink = screen.getByRole('link', {name: 'Home'}),
+      homeLinkLogo = homeLink.querySelector('img');
```
rpokorny, line 35: Use RTL role queries

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
-        component = getShallowComponent(props);
+  it('renders a custom logo when given logo prop', function() {
+    const logo = renderEl({ logo: { path: 'foo', alt: 'bar' } })?.querySelector("img");
+
+    expect(logo).toHaveAttribute('src', 'foo');
+    expect(renderEl({ logo: { path: 'foo', alt: 'bar' } })?.querySelectorAll("img").length).toEqual(1);
+  });
+
+  it('renders logo link when given homeLink prop', function() {
+    renderEl({ homeLink: '#home'});
```
rpokorny, line 32: The href should be tested to ensure it matches the `homeLink` prop

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
+    const link1 = screen.getByRole('link', {name: 'foo'}),
+      link2 = screen.getByRole('link', {name: 'baz'})
 
-    // this path comes from imgMock.ts
-    expect(logo).toHaveDisplayName('img');
-    expect(logo).toHaveProp('src', 'path/to/asset.png');
-    expect(logo).toHaveProp('alt', 'Sonatype');
-    expect(logo).toHaveClassName('nx-product__logo-image');
+    expect(link1).toBeInTheDocument();
+    expect(link2).toBeInTheDocument();
```
rpokorny, line 48: Test the `href`s

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
   });
```
rpokorny, line 49: We no longer want to test the `nx-page-header__link--current` class, but we do still want to test that the current link is identified.  The most appropriate way to do that would be to check for the `aria-current` attribute.  The implementation does not currently add that attribute, but improving it such that it does is something that I'd consider to be in-scope for this ticket.

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
     const props = {
           productInfo: {
             name: 'test app'
           }
         },
-        defaultProductInfoContent = getShallowComponent().prop('productInfoContent'),
-        richProductInfoContent = shallow(getShallowComponent(props).prop('productInfoContent'));
+        header = renderEl(),
+        headerWithProductName = renderEl(props),
+        headerWithProductNameText = headerWithProductName?.querySelector('div.nx-product__name');
```
rpokorny, line 59: The element tagname and classname are implementation details.  Just assert that the header overall contains the `name` text.

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
           productInfo: {
             name: 'test app',
             version: '1.2.3'
           }
         },
-        noVersionProductInfoContent = shallow(getShallowComponent(noVersionProps).prop('productInfoContent')),
-        richProductInfoContent = shallow(getShallowComponent(props).prop('productInfoContent'));
+        elNoProductVersion = renderEl(props),
+        elWithProductVersion = renderEl(propsWithVersion),
+        productVersion = renderEl(elWithProductVersion?.querySelector('.nx-product__version')!);
```
rpokorny, line 81: Also an impl detail.  Just check that the header overall contains the version text

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > One of the first tests we that we usually do is for whatever role is expected on the top-level element rendered by the component. In this case that would be `<header>`. The [ARIA in HTML](https://www.w3.org/TR/html-aria/#el-header) specs say that this element gets a role of `banner` unless it is included within certain other elements (which `NxPageHeader` shouldn't be). So we should test for the `banner` role.

Sounds good. I updated the test here: [a8d603f](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/a8d603f405e65072936359efebff634bc5b312fc)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
+import { screen } from '@testing-library/react';
+import { rtlRenderElement } from '../../../__testutils__/rtlUtils';
+
+fdescribe('NxPageHeader', function() {
+
+  const renderEl = rtlRenderElement(NxPageHeader, {});
 
-  it('renders an AbstractNxPageHeader', function() {
-    expect(getShallowComponent()).toMatchSelector(AbstractNxPageHeader);
+ it('renders a simple NxPageHeader', function() {
```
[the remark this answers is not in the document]
achen5: > RTL analyzes the DOM not the react component structure, so it doesn't (can't) check things about what react components are being used. So this test name needs to be updated.

The test was renamed due to changes from this commit: [a8d603f](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/a8d603f405e65072936359efebff634bc5b312fc)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > This should be written in terms of RTL queries and roles, not CSS selectors and `querySelector`

Fixed: [4ae11b1](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/4ae11b1ffaacfcf331ad8f3252b2685d95844a84)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > Use RTL role queries

Fixed here: [ee87f81](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/ee87f814b6d18dc223ec40e896a8ab7064b97fad)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > The href should be tested to ensure it matches the `homeLink` prop

Done: [8d4a342](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/8d4a342e95608fb3f69538c20b8edf28572cd92f)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > Test the `href`s

Done: [e5bb413](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/e5bb4133aed3551de758626a2fd3b37e2e852cfe)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > We no longer want to test the `nx-page-header__link--current` class, but we do still want to test that the current link is identified. The most appropriate way to do that would be to check for the [`aria-current`](https://www.w3.org/TR/wai-aria-1.2/#aria-current) attribute (which should be set to `page` on the link to the current page). The implementation does not currently add that attribute, but improving it such that it does is something that I'd consider to be in-scope for this ticket.

Fixed: [793a159](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/793a159d33a6f31bd28f62d24ca85e1cc040d8b1)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > The element tagname and classname are implementation details. Just assert that the header overall contains the `name` text.

Done: [15b9325](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/15b9325c914cc126a4aa0df665457670492f84c9)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > Also an impl detail. Just check that the header overall contains the version text

Done: [15b9325](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/15b9325c914cc126a4aa0df665457670492f84c9)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,119 @@
...
+    renderEl(props);
 
-    // this path comes from imgMock.ts
-    expect(logo).toHaveDisplayName('img');
-    expect(logo).toHaveProp('src', 'foo');
-    expect(logo).toHaveProp('alt', 'bar');
-    expect(logo).toHaveClassName('nx-product__logo-image');
+    const link = screen.getByRole('link',{name: 'foo'}),
+        homeLink = screen.getByRole('link', {name: 'Home'}),
+        homeLinkLogo = homeLink.querySelector('img'),
```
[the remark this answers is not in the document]
achen5: > Query by role

Done: [ee87f81](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/ee87f814b6d18dc223ec40e896a8ab7064b97fad)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
+import { rtlRenderElement } from '../../../__testutils__/rtlUtils';
 
 describe('NxPageHeader', function() {
-  const getShallowComponent = enzymeUtils.getShallowComponent(NxPageHeader, {});
 
-  it('renders an AbstractNxPageHeader', function() {
-    expect(getShallowComponent()).toMatchSelector(AbstractNxPageHeader);
+  const renderEl = rtlRenderElement(NxPageHeader, {});
+
+ it('renders a top-level header element', function() {
```
rpokorny, line 17:
```suggestion
 it('renders a top-level banner element', function() {
```

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
+    renderEl();
+    const header = screen.getByRole('banner');
+    expect(header).toBeInTheDocument();
```
rpokorny, line 20: Since the test specifically refers to the "top-level" element
```suggestion
    const el = renderEl();
    const header = screen.getByRole('banner');
    expect(header).toBe(el);
```

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
-  it('passes its links, homeLink, and children props to the AbstractNxPageHeader', function() {
-    const props = {
-          links: [{ name: 'foo', href: '#bar' }],
-          homeLink: '#home',
-          children: <button>Click Here For a Free iPhone</button>
-        },
-        component = getShallowComponent(props);
+  it('renders the default logo within the header', function() {
+    renderEl();
+    const logo = screen.getByRole('img');
```
rpokorny, line 25: To make sure that the image is within the header, as the test name says, what you really want to to is use RTL queries based off of the `banner` that the component renders, not off of the entire `screen`.  We already verified in the test above that the top-level element is the `banner`, so that means you can query off of that top-level element.  This is an extremely common use-case in RSC's RTL tests, and so we have a utility to help with it, which, to be consistent, you should really be using most everywhere that you're currently using `screen`.  That utility is the `rtlRender` function in `rtlUtils`.  It works similarly to `rtlRenderElement`: you pass it the component itself and the minimal props, and it returns a function (which we typically call `quickRender`).  That function can optionally take more props, and it renders the element to the page.  The difference is in what it returns. While `renderEl` returns the top-level element, `quickRender` returns an RTL `RenderResult` that's been modified to have all of its query functions bound to the rendered element rather than `screen`.  So using that infrastructure, the way I'd write this test would be something like this:
```
const view = quickRender(),
    logo = view.getByRole('img');

expect(logo).toBeInTheDocument();
```

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
+  });
+
+  it('renders a custom logo instead of the default one when given logo prop', function() {
+    renderEl({ logo: { path: 'foo', alt: 'bar' } });
+    const logo = screen.getByRole('img');
+
+    expect(logo).toHaveAttribute('src', 'foo');
+    expect(logo).toHaveAttribute('alt', 'bar');
+    // ensures there is only one logo
+    expect(screen.getAllByRole('img').length).toEqual(1);
```
rpokorny, line 36: `getByRole`, which you're using above, already verifies that there's only one `img`.  RTL `getBy…` queries throw an exception if they find more than one matching element.

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
+  });
+
+  it('renders logo link when given homeLink prop', function() {
+    renderEl({ homeLink: '#home'});
```
rpokorny, line 40:
```suggestion
    renderEl({ homeLink: '#home' });
```

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
+    renderEl({ homeLink: '#home'});    
+    const homeLink = screen.getByRole('link', {name: 'Home'}),
+      homeLinkLogo = screen.getByRole('img');
 
-    expect(component).toHaveProp('links', props.links);
-    expect(component).toHaveProp('homeLink', props.homeLink);
-    expect(component.children()).toMatchElement(props.children);
+    expect(homeLink).toBeInTheDocument();
+    expect(homeLink).toHaveAttribute('href','#home');
+    expect(homeLinkLogo).toBeInTheDocument();
```
rpokorny, line 46:
```suggestion
    expect(homeLink).toContainElement(homeLinkLogo);
```

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > To make sure that the image is within the header, as the test name says, what you really want to do is use RTL queries based off of the `banner` that the component renders, not off of the entire `screen`. We already verified in the test above that the top-level element is the `banner`, so that means you can query off of that top-level element. This is an extremely common use-case in RSC's RTL tests, and so we have a utility to help with it, which, to be consistent, you should really be using most everywhere that you're currently using `screen`. That utility is the `rtlRender` function in `rtlUtils`. It works similarly to `rtlRenderElement`: you pass it the component itself and the minimal props, and it returns a function (which we typically call `quickRender`). That function can optionally take more props, and it renders the element to the page. The difference is in what it returns. While `renderEl` returns the top-level element, `quickRender` returns an RTL `RenderResult` that's been modified to have all of its query functions bound to the rendered element rather than `screen`. So using that infrastructure, the way I'd write this test would be something like this:
> 
> ```
> const view = quickRender(),
>     logo = view.getByRole('img');
> 
> expect(logo).toBeInTheDocument();
> ```

Sounds good to me. I refactored the code here: [67d3ddb](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/67d3ddb6e80eea79e525764fca3eb48f940c9ff2)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > `getByRole`, which you're using above, already verifies that there's only one `img`. RTL `getBy…` queries throw an exception if they find more than one matching element.

I removed the test here: [e13c730](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/e13c730c802bc0e5b9c89013095c51200fedc0b2)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,117 @@
...
-    // this path comes from imgMock.ts
-    expect(logo).toHaveDisplayName('img');
-    expect(logo).toHaveProp('src', 'path/to/asset.png');
-    expect(logo).toHaveProp('alt', 'Sonatype');
-    expect(logo).toHaveClassName('nx-product__logo-image');
+    expect(link1).toBeInTheDocument();
+    expect(link1).toHaveAttribute('href', '#bar');
+    expect(link2).toBeInTheDocument();
+    expect(link2).toHaveAttribute('href', '#qux');
+    expect(link2).toHaveAttribute('aria-current', 'page');
```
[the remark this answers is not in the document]
achen5: > Should assert that the other link does not have `aria-current="page"`. Personally I'd split out the `aria-current` checks into a separate test

I created a separate test here: [72fb68c](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/72fb68cbc86215765364ed7a4725ec094ebb2342)

The non-current links are not supposed to have the `aria-current` attribute at all, so I went with `expect(link).not.toHaveAttribute('aria-current')` 
instead of 
`expect(link).toHaveAttribute('aria-current', 'page');`

If you prefer the latter, let me know

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
```diff
@@ -5,84 +5,133 @@
...
+      logo = view.getByRole('img');
+    expect(logo).toBeInTheDocument();
+  });
+
+  it('renders a custom logo instead of the default one when given logo prop', function() {
+    const view = quickRender({ logo: { path: 'foo', alt: 'bar' } }),
+      logo = view.getByRole('img');
+
+    expect(logo).toHaveAttribute('src', 'foo');
+    expect(logo).toHaveAttribute('alt', 'bar');
```
rpokorny, line 35: The global sidebar PR made me think of this but it applies here too.  Is it better to test the alt attribute specifically or just that this text becomes the logo's accessible name?

achen5:
> +1 aside from https://github.com/sonatype/sonatype-react-shared-components/pull/935/files#r1092361602 and you've got a bunch of eslint errors

Fixed here: [20faa4c](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/20faa4ca49698abed3f1b6bba0757ad54c897382)

lib/src/components/NxPageHeader/__tests__/NxPageHeader.test.tsx
achen5: > The global sidebar PR made me think of this but it applies here too. Is it better to test the alt attribute specifically or just that this text becomes the logo's accessible name?

Checking by logo's accessible name is slightly better than checking alt. I made the change here: [2fa9aa3](https://github.com/sonatype/sonatype-react-shared-components/pull/935/commits/2fa9aa37e99f43f7a3ed1ba836ec8199c45fb5bb)

PR #939, opened 2023-01-30
reactUtil RSC Tests - RSC-1409

rpokorny:
https://issues.sonatype.org/browse/RSC-1409

### DIFF
--- a/lib/src/util/__tests__/reactUtil.test.tsx
+++ b/lib/src/util/__tests__/reactUtil.test.tsx
@@ -4,13 +4,16 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
+import { render } from '@testing-library/react';
 import React, { ReactElement, ReactNode } from 'react';
-import { shallow } from 'enzyme';
-import 'jest-enzyme';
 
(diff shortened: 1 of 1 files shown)


lib/src/util/__tests__/reactUtil.test.tsx
```diff
@@ -118,22 +161,35 @@ describe('reactUtil', function() {
     });
 
     it('passes through elements as-is', function() {
```
entroform, line 163: Isn't this the same as line 26?

lib/src/util/__tests__/reactUtil.test.tsx
```diff
@@ -143,66 +199,98 @@ describe('reactUtil', function() {
     });
 
     it('wraps arrays that start with a string', function() {
-      const arr = ['bar', <span key="baz">baz</span>];
+      const arr = ['bar', <span key="baz">baz</span>],
+          result = ensureStartEndElements(arr) as ReactElement,
```
entroform, line 203: Should this be `ensureElement`?
Same as a few below

rpokorny: > Should this be ensureElement?
Same as a few below

Definitely.  Good catch. 423a4d257

entroform (approved):
LGTM!

PR #942, opened 2023-01-30
NxList RTL Tests - RSC-1391

cindyko226:
https://issues.sonatype.org/browse/RSC-1391


lib/src/components/NxList/NxListLinkItem.tsx
```diff
@@ -34,7 +34,8 @@ const NxListLinkItem = forwardRef<HTMLLIElement, NxListLinkItemProps>(
         <li ref={ref} className={liClassNames} {...attrs} aria-selected={selected} aria-current={selected}>
           <a aria-disabled={includesDisabledClass(aClassNames)}
              className={aClassNames}
-             href={href}
+             href={disabled ? undefined : href}
+             role={disabled ? 'link' : undefined}
```
cindyko226, line 38: For a11y specified in [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance), to disable <a> link
-  the link should not have href attr
-  have aria-disabled set to true
-  add a role="link" so that it is considered a link by screen readers

lib/src/components/NxList/__tests__/NxList.test.tsx
```diff
@@ -4,192 +4,216 @@
...
-import { act } from 'react-dom/test-utils';
-import { mount, ReactWrapper } from 'enzyme';
-import NxListButtonItem from '../NxListButtonItem';
 
 describe('NxList', function() {
+  const quickRender = rtlRender(NxList, {}),
+      renderEl = rtlRenderElement(NxList, {});
 
-  let mountContainers: HTMLElement[] = [];
+  // prevent RTL logging thrown warnings
```
rpokorny, line 17: Could you include a mention here of which warnings you're silencing?

lib/src/components/NxList/__tests__/NxList.test.tsx
```diff
@@ -4,192 +4,216 @@
...
 
-    expect(component.find('span')).toHaveText('Empty message');
+    await runTimers();
+
+    expect(ref.current).toBe(el);
   });
 
-  it('shows the emptyMessage when there are no children, no error, and not loading', async function() {
+  it('renders children correctly', function() {
+    const children = <NxList.Item>Cat</NxList.Item>,
```
rpokorny, line 64: Render more than one child, just to be safe

lib/src/components/NxList/__tests__/NxList.test.tsx
```diff
@@ -4,192 +4,216 @@
...
-  it('does not show the emptyMessage when isLoading', async function() {
-    const component = await mountAttached(<NxList emptyMessage="Empty message" isLoading />);
+      it('does not show the default empty message when there are children', function() {
+        const children = <NxList.Item>Cat</NxList.Item>,
+            { getByRole } = quickRender({ children });
 
-    expect(component).not.toIncludeText('Empty message');
-  });
+        expect(getByRole('list')).not.toHaveTextContent('This list is empty.');
+        expect(getByRole('listitem')).not.toHaveTextContent('This list is empty.');
```
rpokorny, line 97: I think checking the list item is redundant since you just checked the entire list of the line above
```suggestion
```

lib/src/components/NxList/__tests__/NxList.test.tsx
```diff
@@ -4,192 +4,216 @@
...
-      <NxList emptyMessage="Empty message" error="Errr" retryHandler={() => {}}>
-      </NxList>
-    );
+      it('does not show the default empty message when loading', function() {
+        const { getByRole } = quickRender({ isLoading: true });
 
-    expect(component).not.toIncludeText('Empty message');
-  });
+        expect(getByRole('list')).not.toHaveTextContent('This list is empty.');
+        expect(getByRole('listitem')).not.toHaveTextContent('This list is empty.');
```
rpokorny, line 104:
```suggestion
```

lib/src/components/NxList/__tests__/NxList.test.tsx
```diff
@@ -4,192 +4,216 @@
...
-              <NxList.Text>Bar</NxList.Text>
-            </NxList.Item>
-          </>
-        )
+      it('does not show the default empty message when error', function() {
+        const retryHandler = jest.fn(),
+            { getByRole } = quickRender({ error: 'Error message', retryHandler });
+
+        expect(getByRole('list')).not.toHaveTextContent('This list is empty.');
+        expect(getByRole('listitem')).not.toHaveTextContent('This list is empty.');
```
rpokorny, line 112:
```suggestion
```

lib/src/components/NxList/__tests__/NxList.test.tsx
cindyko226: I was going to asking a question about this, there is a console warning for not having `emptyMessage` when no children, no error and not loading, but is this warning necessary if we already have default empty message `This list is empty.` when `emptyMessage` prop is not provided? 
`console.warn('emptyMessage is required when no list items are to be rendered');`

lib/src/components/NxList/__tests__/NxListActions.test.tsx
```diff
@@ -4,24 +4,50 @@
...
+    const ref = React.createRef<HTMLDivElement>(),
+        el = renderEl({ ref });
+
+    expect(ref.current).toBe(el);
+  });
+
+  it('renders children correctly', function() {
+    const children = <NxButton>Foo</NxButton>,
+        view = quickRender({ children }),
+        el = view.container.querySelector('div.nx-list__actions'),
```
rpokorny, line 45: `el` is just the top-level element right? Use `view.container.firstElementChild` instead to avoid impl-detail classnames

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
```diff
@@ -5,76 +5,119 @@
...
 
-describe('NxListButtonItem', function() {
+describe('NxList.ButtonItem', function() {
+  const quickRender = rtlRender(NxList.ButtonItem, {}),
+      renderEl = rtlRenderElement(NxList.ButtonItem, {});
 
-  const minimalProps: NxListButtonItemProps = {};
-  const getShallow = getShallowComponent(NxList.ButtonItem, minimalProps);
-  const getMounted = getMountedComponent(NxList.ButtonItem, minimalProps);
+  it('renders a list containing a button', function() {
```
rpokorny, line 17:
```suggestion
  it('renders a listitem containing a button', function() {
```

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
```diff
@@ -5,76 +5,119 @@
...
-    expect(contentEl).toBeDisabled();
+    expect(el).toHaveAttribute('id', 'foo');
+    expect(el).toHaveAttribute('lang', 'en');
   });
 
-  it('renders a clickable list whose list item element has prop selected and is selected', function() {
-    const contentEl = getMounted({selected: true});
+  it('sets the specified classnames to the button in addition to the defaults ' +
+    'when buttonClassName prop is provided', function() {
+    const btn = quickRender({ buttonClassName: 'boo' }).getByRole('button'),
```
rpokorny, line 54: Since we have two classname related props we should test that each of them affects only the element that it's supposed to and not the other.  That is, in this test also set a `className` and assert that it doesn't end up on the button, and in 'sets the specified classnames to the list in addition to the defaults', add a `buttonClassName` and assert that it doesn't end up on the listitem.

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
```diff
@@ -5,76 +5,119 @@
...
   });
 
-  it('renders children correctly', function() {
-    const children = [
-      <NxList.Text key="1">Test Item 1 Text</NxList.Text>,
-      <NxList.Subtext key="2">Test Item 1 Subtext</NxList.Subtext>
-    ];
-    const contentEl = getMounted({children});
+  it('sets the specified attrs to the button ' +
+    'when buttonAttributes prop is provided', function() {
```
rpokorny, line 65: Similarly to what I said about classnames, test that the extra attrs and the top level and these attrs don't get mixed up

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
```diff
@@ -5,76 +5,119 @@
...
-    expect(contentEl).toExist();
-    expect(contentEl.find('span').at(0)).toMatchSelector('span.nx-list__text');
-    expect(contentEl.find('span').at(0)).toHaveText('Test Item 1 Text');
-    expect(contentEl.find('span').at(1)).toMatchSelector('span.nx-list__subtext');
-    expect(contentEl.find('span').at(1)).toHaveText('Test Item 1 Subtext');
+    expect(btn).toHaveAttribute('id', 'boo');
+    expect(btn).toHaveAttribute('value', 'test');
+  });
 
+  it('sets disabled and aria-disabled="true" to the button ' +
```
rpokorny, line 73: I don't think `aria-disabled` actually matters in this case; I'd only check `.toBeDisabled()` here.  We do however need a test that `aria-disabled` gets set if the caller passes in `disabled` as one of the classnames

lib/src/components/NxList/__tests__/NxListItem.test.tsx
```diff
@@ -5,35 +5,48 @@
...
-    expect(contentEl.find('li')).toHaveClassName('nx-list__item');
-    expect(contentEl.find('span').at(0)).toHaveText('Test Item 1 Text');
-    expect(contentEl.find('span').at(0)).toHaveClassName('nx-list__text');
-    expect(contentEl.find('span').at(1)).toHaveText('Test Item 1 Subtext');
-    expect(contentEl.find('span').at(1)).toHaveClassName('nx-list__subtext');
 
+describe('NxList.Item', function() {
+  const quickRender = rtlRender(NxList.Item, {}),
+      renderEl = rtlRenderElement(NxList.Item, {});
+
```
rpokorny, line 15: Should test that the top-level element is a `listitem`.

lib/src/components/NxList/__tests__/NxListLinkItem.test.tsx
```diff
@@ -4,76 +4,122 @@
...
 
-describe('NxListLinkItem', function() {
+describe('NxList.LinkItem', function() {
+  const minimalProps: NxListLinkItemProps & RefAttributes<HTMLLIElement> = {
+        href: 'www.sonatype.com'
+      },
+      quickRender = rtlRender(NxList.LinkItem, minimalProps),
+      renderEl = rtlRenderElement(NxList.LinkItem, minimalProps);
+
+  it('renders a list containing an anchor', function() {
```
rpokorny, line 20:
```suggestion
  it('renders a listitem containing a link', function() {
```

lib/src/components/NxList/__tests__/NxListLinkItem.test.tsx
```diff
@@ -4,76 +4,122 @@
...
+    expect(renderAndGetList({ selected: null })).not.toHaveAttribute('aria-current');
+
+    expect(renderAndGetList({ selected: true })).toHaveAttribute('aria-selected', 'true');
+    expect(renderAndGetList({ selected: true })).toHaveAttribute('aria-current', 'true');
+  });
+
+  it('renders children correctly', function() {
+    const children = [
+      <NxList.Text data-testid="text" key="1">Test Item 1 Text</NxList.Text>,
+      <NxList.Subtext data-testid="subtext" key="2">Test Item 1 Subtext</NxList.Subtext>
```
rpokorny, line 112: Pretty sure these should be indented two more levels

lib/src/components/NxList/__tests__/NxListSubtext.test.tsx
```diff
@@ -4,26 +4,46 @@
...
+    const ref = React.createRef<HTMLSpanElement>(),
+        el = renderEl({ ref });
+
+    expect(ref.current).toBe(el);
+  });
+
+  it('renders children correctly', function() {
+    const children = 'Book',
+        view = quickRender({ children }),
+        el = view.container.querySelector('span.nx-list__subtext');
```
rpokorny, line 44: Use `firstElementChild`

lib/src/components/NxList/__tests__/NxList.test.tsx
rpokorny: > I was going to asking a question about this, there is a console warning for not having emptyMessage when no children, no error and not loading, but is this warning necessary if we already have default empty message This list is empty. when emptyMessage prop is not provided?
console.warn('emptyMessage is required when no list items are to be rendered');

I think that's ok.  From what I can tell/recall, the default empty message isn't really supposed to be used but is there as a sort of safeguard.  So keep your code as-is but add the comment saying which message you're silencing

lib/src/components/NxList/__tests__/NxList.test.tsx
cindyko226: Updated here d48f9f648f10da27f8a30b98d9ab933ec292b005

lib/src/components/NxList/__tests__/NxList.test.tsx
cindyko226: Add comments here 536f555b49bb126863e05f2f2142febad218a267

lib/src/components/NxList/__tests__/NxListActions.test.tsx
cindyko226: Yes. Updated here f31d3f4a01144fd2931367a77813b963191affe1

lib/src/components/NxList/__tests__/NxListSubtext.test.tsx
cindyko226: Updated here cd95052b589983f78b3881028b57bbc9512c9c00

lib/src/components/NxList/__tests__/NxListLinkItem.test.tsx
cindyko226: To indent these two more levels, I would need to move the open bracket to the new line to avoid eslint complaints.
Updated here 9bf09bdb478fd4698d1c35aa2a57734639026e88.

lib/src/components/NxList/__tests__/NxListItem.test.tsx
cindyko226: Added here 999143c7e42cc36bc438640defd458e8ed027d0a

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
cindyko226: Ah right, make sense. Updated here eb1b5df24c158cfcbef7fc0843ac76e04ae9717e

lib/src/components/NxList/__tests__/NxListLinkItem.test.tsx
rpokorny: > To indent these two more levels, I would need to move the open bracket to the new line to avoid eslint complaints.
Updated here https://github.com/sonatype/sonatype-react-shared-components/commit/9bf09bdb478fd4698d1c35aa2a57734639026e88.

Huh, that's surprising.  It seems the react eslint plugin behaves differently from the base eslint indentation logic

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
cindyko226: Updated here 22441387eb16d63f3de01de20428262afc12a5ed

lib/src/components/NxList/__tests__/NxListButtonItem.test.tsx
cindyko226: Updated here 22441387eb16d63f3de01de20428262afc12a5ed

rpokorny (approved):

PR #943, opened 2023-01-30
RTL NxTable - RSC-1393 

entroform:
https://issues.sonatype.org/browse/RSC-1393


lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -55,25 +56,30 @@ describe('NxTable', function() {
...
+    const headerColumnHeaders = within(headerRows[0]).getAllByRole('columnheader');
+    const headerCells = within(headerRows[0]).queryAllByRole('cell');
+    const bodyRows = within(rowGroups[1]).getAllByRole('row');
+    const bodyColumnHeaders = within(bodyRows[0]).queryAllByRole('columnheader');
+    const bodyCells = within(bodyRows[0]).getAllByRole('cell');
 
-    expect(component.find('thead tr')).toHaveClassName('nx-table-row--header');
-    expect(component.find('thead tr')).toHaveClassName('nx-table-row');
+    expect(table).toBeInTheDocument();
+    expect(rowGroups.length).toBe(2);
```
rpokorny, line 69: There is a `toHaveLength` matcher

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
       </NxTable>
     );
 
-    expect(component.find('td')).toHaveProp('colSpan', 2);
-    expect(component.find('td')).toHaveClassName('nx-cell--meta-info');
+    expect(screen.getByRole('cell')).toHaveAttribute('colSpan', '2');
+    expect(screen.getByRole('cell')).toHaveClass('nx-cell--meta-info');
```
rpokorny, line 95: This is a presentational class; no need to test

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
+        </NxTableHead>
+        <NxTableBody emptyMessage="Empty" />
+      </NxTable>
+    );
 
-    expect(component!.find('td')).toHaveProp('colSpan', 2);
-    expect(component!.find('td')).toHaveClassName('nx-cell--meta-info');
-    expect(component!.find('td')).toHaveText('Empty');
+    expect(screen.getByRole('cell')).toHaveAttribute('colSpan', '2');
+    expect(screen.getByRole('cell')).toHaveClass('nx-cell--meta-info');
```
rpokorny, line 112:
```suggestion
```

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
-        <NxTableBody error="Error!"/>
+        <NxTableBody error="Error!" />
       </NxTable>
     );
 
-    expect(component.find('td')).toHaveProp('colSpan', 2);
-    expect(component.find('td')).toHaveClassName('nx-cell--meta-info');
-    expect(component.find('td')).toIncludeText('Error!');
+    expect(screen.getByRole('cell')).toHaveAttribute('colSpan', '2');
+    expect(screen.getByRole('cell')).toHaveClass('nx-cell--meta-info');
```
rpokorny, line 130:
```suggestion
```

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
         </NxTableHead>
-        <NxTableBody error="Error!"/>
+        <NxTableBody error="Error!" />
       </NxTable>
     );
 
-    expect(component.find('td')).toHaveProp('colSpan', 2);
-    expect(component.find('td')).toHaveClassName('nx-cell--meta-info');
-    expect(component.find('td')).toIncludeText('Error!');
+    expect(screen.getByRole('cell')).toHaveAttribute('colSpan', '2');
```
rpokorny, line 129: In JSX, `colSpan` uses camelCase, but isn't it usually entirely lowercase in the DOM?

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -55,25 +56,30 @@ describe('NxTable', function() {
...
-    expect(component.find('tbody tr')).not.toHaveClassName('nx-table-row--header');
-    expect(component.find('td.nx-cell--header')).not.toExist();
-
-    // colSpan should not be getting set on table cells
-    expect(component.find('td').findWhere(td => td.prop('colSpan') != null)).not.toExist();
+    expect(bodyCells[0]).not.toHaveAttribute('colSpan');
+    expect(bodyCells[1]).not.toHaveAttribute('colSpan');
   });
 
   it('Puts the correct colSpan on a loading meta-info cell', function() {
```
rpokorny, line 81: I feel like all of these `colspan` tests would make more sense within the `NxTable.Body` describe block

rpokorny:
`NxTableHead`, `NxTableBody`, etc are deprecated in favor of `NxTable.Head` etc.  I think the tests should all mainly use the modern names, and simple tests of the deprecated aliases should also be added

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
+          </NxTableBody>
+        </NxTable>
+      );
+
+      await waitFor(() => expect(screen.getByRole('rowgroup')).not.toHaveTextContent('Empty message'));
+    });
+  });
+
+  describe('NxTableRow', function() {
+    it('sets the rendered text content into RowContext provider label with cells separated by semi-colons', function() {
```
rpokorny, line 215: The context is an implementation detail.  Instead of interacting with it directly, test that the two features that it is intended to implement do work.  Those features being
1. That filter header cells end up as `<td>` elements rather than `<th>` elements. (or in terms of roles, `cell`s rather than `columnheader`s
2. That the accessible name on the row-click button is correct.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
+      expect(screen.getByTestId('context-dependent')).toHaveTextContent('false');
+    });
+  });
+
+  describe('NxTableCell', function() {
+    const renderHeaderComponent = (extraProps?: NxTableCellProps) => render(
+      <NxTable>
+        <NxTableHead>
+          <NxTableRow>
+            <HeaderContext.Provider value={true}>
```
rpokorny, line 338: The `HeaderContext` is also an implementation detail, and definitely should never be used in this way - the provider is set up in `NxTable.Head`

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
+              <NxTableCell { ...extraProps }/>
+            </HeaderContext.Provider>
+          </NxTableRow>
+        </NxTableHead>
+      </NxTable>
+    ).getByRole('columnheader') as HTMLElement;
+
+    describe('when the chevron prop is true', function() {
+      describe('when not isHeader', function() {
+        it('sets the button accessible name from the RowContext', function() {
```
rpokorny, line 348: Again, the context is an impl detail.  Test that the accessible name gets set from the row text content and that it gets set from the `clickAccessibleLabel` prop

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
+              </NxTableBody>
+            </NxTable>
+          );
+
+          expect(screen.getByRole('button')).toHaveAccessibleName('foobar');
+        });
+      });
+
+      describe('when isHeader', function() {
+        it('renders a visually hidden header cell with the text "Select Row"', function() {
```
rpokorny, line 366: It being visually hidden isn't really testable here so I'd remove that part of the test name

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -85,82 +91,369 @@ describe('NxTable', function() {
...
+          expect(component).toHaveTextContent('Select Row');
+        });
       });
     });
-    component.update();
 
-    expect(component.find('td')).toHaveProp('colSpan', 1);
-    expect(component.find('td')).toHaveClassName('nx-cell--meta-info');
-    expect(component.find('td')).toIncludeText('Error!');
+    describe('when the rowBtnIcon prop is set', function() {
```
rpokorny, line 378: Is this describe name wrong? It seems like all of these tests are about `isSortable`.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > There is a `toHaveLength` matcher

Thanks! I updated it here: 3a12e15fa0ec481d388ebca0c0d86293341b0ddf

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > This is a presentational class; no need to test

Removed here:
56b1813e19885eff729fc247da16d0e96c8571d2
d696c91b16f059c37753530144ee44115de00a36

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > In JSX, `colSpan` uses camelCase, but isn't it usually entirely lowercase in the DOM?

Oo for some reason it still passed so it looks like its not toHaveAttribute is not case sensitive?
I fixed that here: 90866a5d21352d7d363052acefeb0c54a20fb990

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > I feel like all of these `colspan` tests would make more sense within the `NxTable.Body` describe block

I moved it here: cabfc93c988fb7f09a313275f3cb1dc19d9c7cd9

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > The context is an implementation detail. Instead of interacting with it directly, test that the two features that it is intended to implement do work. Those features being
> 
> 1. That filter header cells end up as `<td>` elements rather than `<th>` elements. (or in terms of roles, `cell`s rather than `columnheader`s
> 2. That the accessible name on the row-click button is correct.

Thanks, Fixed that as part of: c54ac41c226c00680630c11145c503fe66946f46

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > Again, the context is an impl detail. Test that the accessible name gets set from the row text content and that it gets set from the `clickAccessibleLabel` prop

Fixed it as part of: c54ac41c226c00680630c11145c503fe66946f46

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > The `HeaderContext` is also an implementation detail, and definitely should never be used in this way - the provider is set up in `NxTable.Head`

Thanks! I adjusted it here: 09f61036e296efa83f8cce271d5ba1ed82bc5463

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > It being visually hidden isn't really testable here so I'd remove that part of the test name

Renamed here: 87e191edf2126ba712fdb6081718fb505989f131

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > > Is this describe name wrong? It seems like all of these tests are about isSortable.
> 
> But at the same time, I don't see tests for `rowBtnIcon`.

Cleaned that up and added rowBtnIcon tests here: 7ec076697da54552d9fc99478cb00360d11bf7b9

entroform:
> `NxTableHead`, `NxTableBody`, etc are deprecated in favor of `NxTable.Head` etc. I think the tests should all mainly use the modern names, and simple tests of the deprecated aliases should also be added

Thanks, I updated it here:
922fc5dcb5944a8e4fecec2d518e9a386e7d7786

Added simple test here, let me know if this is sufficient:
f481a47bf057ccf1f003670ef3ed8cecf1596c98

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,507 @@
...
+      );
+
+      expect(screen.getByRole('row')).toHaveAccessibleName('Foo Bar');
     });
-    component!.update();
 
-    expect(component!.find('td')).toHaveProp('colSpan', 2);
-    expect(component!.find('td')).toHaveClassName('nx-cell--meta-info');
-    expect(component!.find('td')).toHaveText('Empty');
+    it('sets the isFilterHeader flag into the RowContext, normalized to a boolean', function() {
```
rpokorny, line 250: Another overt use of an impl-detail context to avoid

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,507 @@
...
+
+        it('does not show button when rowBtnIcon is null', function() {
+          expect(within(renderTableCell({ rowBtnIcon: null })).queryByRole('button')).not.toBeInTheDocument();
+        });
+
+        it('show button when rowBtnIcon faEdit is set', function() {
+          expect(within(renderTableCell({ rowBtnIcon: faEdit })).getByRole('button')).toBeInTheDocument();
+        });
+
+        it('ignores the children and sort settings and adds the icon child wrapped in a nx-cell__row-btn',
```
rpokorny, line 406: `.nx-cell__row-btn` is no longer involved in this test

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,507 @@
...
+            function() {
+              const cell = renderTableCell({
+                rowBtnIcon: faEdit,
+                sortDir: 'asc',
+                children: <span>foo</span>
+              });
+
+              expect(within(cell).getByRole('button')).toBeInTheDocument();
+              expect(cell.innerHTML).not.toEqual('<span>foo</span>');
```
rpokorny, line 415: It's not clear to me what this line is trying to test.  I suspect it's not actually testing anything useful.  Btw though, you might be interested in the [`toContainHTML`](https://github.com/testing-library/jest-dom#tocontainhtml) matcher

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,507 @@
...
+            }
+        );
+
+        it('sets the nx-cell__row-btn aria-label from the RowContext', function() {
```
rpokorny, line 419: `nx-cell__row-btn` isn't involved in this test either

rpokorny, line 419: Externally, it's not from the `RowContext` it's from the `clickAccessibleName` or from the row text content (both should be tested)

entroform: > `nx-cell__row-btn` isn't involved in this test either

Cleaned up here: cec50dbad93958d240cdfabd25323834588f180f

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > `.nx-cell__row-btn` is no longer involved in this test

Cleaned up here: cec50dbad93958d240cdfabd25323834588f180f

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > It's not clear to me what this line is trying to test. I suspect it's not actually testing anything useful. Btw though, you might be interested in the [`toContainHTML`](https://github.com/testing-library/jest-dom#tocontainhtml) matcher

Fixed that here: cec50dbad93958d240cdfabd25323834588f180f

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > Another overt use of an impl-detail context to avoid

Removed it here: 2238321fbb01bf1a4b269b75bd00ccdc7d9156eb

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,446 @@
...
+            </NxTable.Row>
+          </NxTable.Head>
+          <NxTable.Body isLoading />
+        </NxTable>
+      );
 
-    // colSpan should not be getting set on table cells
-    expect(component.find('td').findWhere(td => td.prop('colSpan') != null)).not.toExist();
-  });
+      debug();
```
rpokorny, line 104:
```suggestion
```

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,446 @@
...
+        it('ignores the children and sort settings and renders a button instead',
+            function() {
+              const cell = renderTableCell({
+                rowBtnIcon: faEdit,
+                sortDir: 'asc'
+              });
+
+              const button = within(cell).getByRole('button');
+              expect(button).toBeInTheDocument();
+              expect(cell).not.toContainHTML('<span>foo</span>');
```
rpokorny, line 344: I still don't really know what this line is trying to test.  For one thing, doesn't the table cell under test contain "Boo" in the JSX, not "foo"? But additionally the span seems like an implementation detail.  If the goal is to assert that the cell does not contain its JSX children in this case, I'd just use `.not.toHaveTextContent` for that.  On the other hand if there's something I'm missing about what this line is trying to test, it needs a comment explaining it.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -4,163 +4,446 @@
...
+        const tooltip = await screen.findByRole('tooltip');
+        expect(tooltip).toBeInTheDocument();
+        expect(tooltip).toHaveTextContent('foo descending');
+      });
+
+      it('sets default button accessible name to match the tooltip', function() {
+        expect(
+            within(renderColumnHeader({ isSortable: true, children: 'foo' })
+            ).getByRole('button'))
+            .toHaveAccessibleName('foo unsorted');
```
rpokorny, line 431: Weird formatting in these three tests. I'd pull the element selection (`within(...).getByRole(...)`) out into a variable.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
entroform: > I still don't really know what this line is trying to test. For one thing, doesn't the table cell under test contain "Boo" in the JSX, not "foo"? But additionally the span seems like an implementation detail. If the goal is to assert that the cell does not contain its JSX children in this case, I'd just use `.not.toHaveTextContent` for that. On the other hand if there's something I'm missing about what this line is trying to test, it needs a comment explaining it.

Yeah I wanted to assert that the cell does not contain its JSX children:
Thanks for the suggestion, updated the test name and impl here:
2af1198e4537b02a8ad8237f0e06667307bf85c9

rpokorny: > Yeah I wanted to assert that the cell does not contain its JSX children:
Thanks for the suggestion, updated the test name and impl here:

But you're still checking for "foo" when the test setup has "Boo"

entroform: > > Yeah I wanted to assert that the cell does not contain its JSX children:
> > Thanks for the suggestion, updated the test name and impl here:
> 
> But you're still checking for "foo" when the test setup has "Boo"

I forgot I changed it to that: 5752a72adee7882dfcff10294deba663122e2d37

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -5,162 +5,448 @@
...
+
+        it('does not show button when rowBtnIcon is null', function() {
+          expect(within(renderTableCell({ rowBtnIcon: null })).queryByRole('button')).not.toBeInTheDocument();
+        });
+
+        it('show button when rowBtnIcon faEdit is set', function() {
+          expect(within(renderTableCell({ rowBtnIcon: faEdit })).getByRole('button')).toBeInTheDocument();
+        });
+
+        it('ignores the JSX children and sort settings and renders a button instead',
```
rpokorny, line 333: Considering that sort headers _also_ render buttons, is this test adequate? Maybe we should assert that the accessible name of the button doesn't include "Boo" and also doesn't include "ascending"

entroform: > Considering that sort headers _also_ render buttons, is this test adequate? Maybe we should assert that the accessible name of the button doesn't include "Boo" and also doesn't include "ascending"

Thanks, let me know if this works: 608d1323ce1366993ed3dae2d77c45402510ba4f

rpokorny (approved):

PR #945, opened 2023-01-31
NxSearchTransferList RTL Conversion- RSC-1384

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1384


lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
 
-  it('adds additional specified class names', function() {
-    const component = getShallow({ className: 'foo' });
+  it('adds additional attributes to the top-level element', function() {
+    const el = renderEl({ id: 'foo', lang: 'en' });
 
-    expect(component).toHaveClassName('foo');
-    expect(component).toHaveClassName('nx-search-transfer-list');
+    expect(el).toHaveAttribute('id', 'foo');
+    expect(el).toHaveAttribute('lang', 'en');
```
avnvbhatta, line 45: Since we're testing the "top-level" element, we should also test that `el` here is the top-level element?

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+          searchInput = quickRender({ onSearchTextChange }).getByRole('searchbox');
+
+      expect(onSearchTextChange).not.toHaveBeenCalled();
+
+      await user.type(searchInput, 'b');
+
+      expect(onSearchTextChange).toHaveBeenLastCalledWith('b');
+    });
+
+    it('calls onSearchTextChange when the ESC is pressed', async function() {
```
avnvbhatta, line 93:
```suggestion
    it('calls onSearchTextChange when the ESC key is pressed', async function() {
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+      expect(onSearchTextChange).toHaveBeenLastCalledWith('');
+    });
+
+    it('calls onSearchTextChange when the Clear Search button is clicked', async function() {
+      const user = userEvent.setup(),
+          onSearchTextChange = jest.fn(),
+          view = quickRender({ searchText: 'b', onSearchTextChange });
+
+      await runTimers();
+      const clearBtn = view.getByRole('button', { name: 'Clear search'});
```
avnvbhatta, line 111:
```suggestion
      const clearBtn = view.getByRole('button', { name: 'Clear search' });
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+      describe('when there are no matches', function() {
+        const viewNoMatches = (props?: Partial<Props>) => quickRender({ searchText: 'i', ...props });
+
+        it('renders a dropdown with role=alert when in empty, or loading states', function() {
+          expect(viewNoMatches().getByRole('alert')).toBeInTheDocument();
+          expect(viewNoMatches({ loading: true }).getByRole('alert')).toBeInTheDocument();
+        });
+
+        it('renders a dropdown with no role when in an error state', function() {
+          const dropdown = renderEl({ loadError: 'oops' })!.querySelector('.nx-search-dropdown__menu--error');
```
avnvbhatta, line 179: Maybe we can test the non-existence of the role instead?

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+          expect(viewNoMatches({ loading: true }).getByRole('alert')).toHaveAttribute('aria-busy', 'true');
+        });
+
+        it('renders an element with role=status within the dropdown when loading', function() {
+          const dropdown = viewNoMatches({ loading: true }).getByRole('alert');
+          expect(within(dropdown).getByRole('status')).toBeInTheDocument();
+        });
+
+        it('renders "Loading" text in the dropdown when loading is true', function() {
+          expect(viewNoMatches({ loading: true }).getByRole('status')).toHaveTextContent('Loading');
```
avnvbhatta, line 194:
```suggestion
          expect(viewNoMatches({ loading: true }).getByRole('status')).toHaveTextContent('Loading…');
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+        });
+
+        it('renders a retry button and error text when the error prop is provided', function() {
+          const alertEl = viewNoMatches({ loadError: 'oops' }).getByRole('alert');
+
+          expect(alertEl).toHaveTextContent('oops');
+          expect(within(alertEl).getByRole('button', { name: 'Retry' })).toBeInTheDocument();
+        });
+
+        it('renders a dropdown with "No Results Found" text there are no matching elements and not in error' +
```
avnvbhatta, line 208:
```suggestion
        it('renders a dropdown with "No Results Found" text if there are no matching elements and not in error' +
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+
+    it('sets the aria-controls on the input to the dropdown id', function() {
+      const view = quickRender({ searchText: 'b' }),
+          searchInput = view.getByRole('searchbox'),
+          dropdownId = view.getByRole('alert').getAttribute('id');
+
+      expect(searchInput).toHaveAttribute('aria-controls', dropdownId);
+    });
+
+    it('calls onSearchTextChange when the value in the search box is changed', async function() {
```
avnvbhatta, line 81:
```suggestion
    it('calls onSearchTextChange when the value in the searchbox is changed', async function() {
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+            retryBtn = quickRender({ loadError: 'oops', onSearch, searchText: 'item' })
+                .getByRole('button', { name: 'Retry' });
+
+        expect(onSearch).not.toHaveBeenCalled();
+
+        await user.click(retryBtn);
+        expect(onSearch).toHaveBeenCalledWith('item');
+      });
+
+      it('calls onSearch with the searchText when the dropdown or searchbox regain focus from elsewhere on the page' +
```
avnvbhatta, line 250:
```suggestion
      it('calls onSearch with the searchText when the dropdown or searchbox regain focus from elsewhere on the page ' +
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+    it('renders a fieldset element with role=group named "Items Added"', function() {
+      const transferListEl = quickRender().getByRole('group', { name: 'Items Added' });
+
+      expect(transferListEl).toBeInTheDocument();
+      expect(transferListEl.tagName).toBe('FIELDSET');
+    });
+
+    it('renders a default label of "Items Added" unless addedItemsLabel specified', function() {
+      const defaultTransferListEl = quickRender().getByRole('group', { name: 'Items Added' }),
+          customTransferListEl = quickRender({ addedItemsLabel: 'New Label' }).getByRole('group', { name: 'New Label'});
```
avnvbhatta, line 306:
```suggestion
          customTransferListEl = quickRender({ addedItemsLabel: 'New Label' }).getByRole('group', { name: 'New Label' });
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+    ];
+
+    it('renders a fieldset element with role=group named "Items Added"', function() {
+      const transferListEl = quickRender().getByRole('group', { name: 'Items Added' });
+
+      expect(transferListEl).toBeInTheDocument();
+      expect(transferListEl.tagName).toBe('FIELDSET');
+    });
+
+    it('renders a default label of "Items Added" unless addedItemsLabel specified', function() {
```
avnvbhatta, line 304:
```suggestion
    it('renders a default label of "Items Added" unless addedItemsLabel is specified', function() {
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+      });
+    });
+
+    describe('filtering', function() {
+      it('sets value of the input according to the addedItemsFilter', function() {
+        expect(quickRender().getByRole('textbox')).toHaveValue('');
+        expect(quickRender({ addedItemsFilter: 'b' }).getByRole('textbox')).toHaveValue('b');
+      });
+
+      it('filter items which contain the addedItemsFiler value by case-insensitive substrintg match', function() {
```
avnvbhatta, line 409:
```suggestion
      it('filter items which contain the addedItemsFiler value by case-insensitive substring match', function() {
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
```diff
@@ -5,160 +5,489 @@
...
+            filterInput = quickRender({ onAddedItemsFilterChange }).getByRole('textbox');
+
+        expect(onAddedItemsFilterChange).not.toHaveBeenCalled();
+
+        await user.type(filterInput, 'b');
+
+        expect(onAddedItemsFilterChange).toHaveBeenLastCalledWith('b', expect.objectContaining({ target: filterInput}));
+      });
+
+      it('calls onAddedItemsFilterChange with an empty string when the ESC is pressed', async function() {
```
avnvbhatta, line 442:
```suggestion
      it('calls onAddedItemsFilterChange with an empty string when the ESC key is pressed', async function() {
```

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
EmilyBermudez88: > Since we're testing the "top-level" element, we should also test that `el` here is the top-level element?

Sounds good! Tested here: [fc4cbc1](https://github.com/sonatype/sonatype-react-shared-components/pull/945/commits/fc4cbc14aa6082c2f3a01d511fc56153d7da99f4)

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
EmilyBermudez88: > Maybe we can test the non-existence of the role instead?

I had left this as a query selector since I figured it was best to test that the dropdown doesn't have either the menu or alert role. But if it's in an alert state, the dropdown contains an alert, so `getByRole('alert')` would return true. However, I'm happy to change it as well and just assert that role=menu isn't present!

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
EmilyBermudez88: Resolved here: [16238e8](https://github.com/sonatype/sonatype-react-shared-components/pull/945/commits/16238e83e1c393aff746e7306e3581ee639a470b)

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
EmilyBermudez88: resolved here: [16238e8](https://github.com/sonatype/sonatype-react-shared-components/pull/945/commits/16238e83e1c393aff746e7306e3581ee639a470b)

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
avnvbhatta: > > Maybe we can test the non-existence of the role instead?
> 
> I had left this as a query selector since I figured it was best to test that the dropdown doesn't have either the menu or alert role. But if it's in an alert state, the dropdown contains an alert, so `getByRole('alert')` would return true. However, I'm happy to change it as well and just assert that role=menu isn't present!

Yeah I think it's better to change it so that we don't query internal class details. We'd also be consistent with other tests where we're not as concerned with implementation details.

avnvbhatta (approved):
LGTM besides small comment https://github.com/sonatype/sonatype-react-shared-components/pull/945#discussion_r1093716708

lib/src/components/NxSearchTransferList/__tests__/NxSearchTransferList.test.tsx
EmilyBermudez88: > > > Maybe we can test the non-existence of the role instead?
> > 
> > 
> > I had left this as a query selector since I figured it was best to test that the dropdown doesn't have either the menu or alert role. But if it's in an alert state, the dropdown contains an alert, so `getByRole('alert')` would return true. However, I'm happy to change it as well and just assert that role=menu isn't present!
> 
> Yeah I think it's better to change it so that we don't query internal class details. We'd also be consistent with other tests where we're not as concerned with implementation details.

Makes sense, added here: [384b76e](https://github.com/sonatype/sonatype-react-shared-components/pull/945/commits/384b76eb3390b70799304e36d7a6935fb679b883)

PR #946, opened 2023-01-31
RTL Conversion - NxGlobalSidebar - RSC-1419

avnvbhatta:
JIRA: https://issues.sonatype.org/browse/RSC-1419


rpokorny:
I don't think there are separate tickets for `NxGlobalSidebarNavigation` and `NxGlobalSidebarNavigationLink`, so those need to be part of this ticket

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
```diff
@@ -5,98 +5,101 @@
...
-      logoAltText: 'alt text',
-      logoLink: '/ref'
-    };
-    getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxGlobalSidebar, minimalProps);
-  });
+  const minimalProps: Props = {
+    isOpen: false,
+    toggleOpenIcon: faCrow,
+    toggleCloseIcon: faBiohazard,
+    onToggleClick: jest.fn,
```
rpokorny, line 18:
```suggestion
    onToggleClick: () => {},
```

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
```diff
@@ -5,98 +5,101 @@
...
-  it('renders an aside with nx-global-sidebar class', function() {
-    expect(getShallowComponent()).toMatchSelector('aside.nx-global-sidebar');
-  });
+  const quickRender = rtlRender<Props>(NxGlobalSidebar, minimalProps);
 
-  it('adds classes based on isOpen state', function() {
-    expect(getShallowComponent({ isOpen: true })).toMatchSelector('.nx-global-sidebar.open');
-    expect(getShallowComponent({ isOpen: false })).toMatchSelector('.nx-global-sidebar.closed');
+  it('renders a top-level element with role="complementary"', function() {
+    expect(quickRender().getByRole('complementary')).toBeInTheDocument();
```
rpokorny, line 27: Tests about the "top-level" element should check that `view.container.firstElementChild` is the element you're looking at

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
```diff
@@ -5,98 +5,101 @@
...
+    const view = quickRender(),
+        link = view.getByRole('link'),
+        logo = view.getByRole('img');
 
-    expect(link).toHaveProp('href', '/ref');
-    expect(logo).toHaveProp('src', 'logoImg');
-    expect(logo).toHaveProp('alt', 'alt text');
+    expect(link).toHaveAttribute('href', '/ref');
+    expect(logo).toHaveAttribute('src', 'logoImg');
+    expect(logo).toHaveAttribute('alt', 'alt text');
```
rpokorny, line 37: I wonder if it would be better to check this as the accessible name of the image rather than specifically the alt text

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
avnvbhatta: > Tests about the "top-level" element should check that `view.container.firstElementChild` is the element you're looking at

Added here: bb0dd225751fb3de34eea9bf2d344495195a83f4

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
avnvbhatta: > I wonder if it would be better to check this as the accessible name of the image rather than specifically the alt text

That's a good idea. Changed here: 4944f73634008b984e99e18cb855427dd94c1e2d

avnvbhatta:
> I don't think there are separate tickets for `NxGlobalSidebarNavigation` and `NxGlobalSidebarNavigationLink`, so those need to be part of this ticket

The tests have now been added.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
```diff
@@ -5,98 +5,105 @@
...
+    expect(view.getByTestId('div1')).toBeInTheDocument();
+    expect(view.getByTestId('p1')).toBeInTheDocument();
+    expect(view.getByTestId('div2')).toBeInTheDocument();
   });
 
-  it('calls onToggleClick when toggle button is pressed', function() {
-    const component = getShallowComponent(),
-        btn = component.find(NxButton);
+  it('calls onToggleClick when toggle button is pressed', async function() {
+    const onToggleClick = jest.fn(() => {}),
```
EmilyBermudez88, line 100:
```suggestion
    const onToggleClick = jest.fn(),
```

EmilyBermudez88, line 100: Comparing tests to the docs, there is no documentation for `onToggleClick` in the gallery docs (even though it's listed as a prop in `types` and in all the examples). Probably a separate ticket, but I'm assuming we need to add it in

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,61 @@
...
   });
 
   it('passes additional attributes to <footer>', function() {
-    expect(getShallowComponent({ id: 'test-id' })).toHaveProp('id', 'test-id');
-    expect(getShallowComponent({ lang: 'en-ca' })).toHaveProp('lang', 'en-ca');
+    expect(renderEl({ id: 'test-id' })).toHaveAttribute('id', 'test-id');
+    expect(renderEl({ lang: 'en-ca' })).toHaveAttribute('lang', 'en-ca');
   });
 
   it('renders the Support div with a link with the passed href and text and does not render when empty', function() {
```
EmilyBermudez88, line 31: Instead of asserting that the div is in the DOM, perhaps the test could look at if the component renders link in a variety of situations: (if neither are provided, if the supportLink is not provided but the supportText is,  if the supportLink is provided and support text isn't, and yes if both are provided). And then still assert the href attribute and that the text is correct.

EmilyBermudez88:
For all the footer content that accepts both string or JSX, there should probably be a second test confirming everything renders correctly when you use JSX as a prop as well.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,61 @@
...
   });
 
   it('renders the Release div with a text and version number and does not render when empty', function() {
-    const releaseEmpty = getShallowComponent().find('div.nx-global-sidebar__release'),
-        release = getShallowComponent({ releaseText: 'React Shared Components: 3.1.4' })
-            .find('div.nx-global-sidebar__release');
+    const { container: viewWithoutReleaseText} = quickRender(),
+        viewWithReleaseText = quickRender({ releaseText: 'React Shared Components: 3.1.4' });
+
+    expect(viewWithReleaseText.getByText('React Shared Components: 3.1.4')).toBeVisible();
```
EmilyBermudez88, line 46:
```suggestion
    expect(viewWithReleaseText.getByText('React Shared Components: 3.1.4')).toBeInTheDocument();
```

And for the other `toBeVisible` assertions

EmilyBermudez88:
For `NxGlobalSidebarNavigation`, I think we should also test that the element has role=navigation (even if it already is a <nav> element)

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarNavigationLink.test.tsx
```diff
@@ -20,33 +18,80 @@ describe('NxGlobalSidebarNavigationLink', function() {
...
-    expect(getShallowComponent()).toMatchSelector(NxOverflowTooltip);
   });
 
   it('renders an <a> with the passed href and classes', function() {
-    const link = getShallowComponent({ className: 'extra-class' }).find('a');
-    expect(link).toMatchSelector('a.nx-global-sidebar__navigation-link.nx-text-link');
-    expect(link).toMatchSelector('.extra-class');
-    expect(link).toHaveProp('href', '#someurl');
+    const link = quickRender({ className: 'extra-class' }).getByRole('link');
+    expect(link).toHaveClass('extra-class');
```
EmilyBermudez88, line 78: maybe test classes as we have in other tests with the comparison of an element with default classes vs a custom class name

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
avnvbhatta: > Comparing tests to the docs, there is no documentation for `onToggleClick` in the gallery docs (even though it's listed as a prop in `types` and in all the examples). Probably a separate ticket (?), but I'm assuming we need to add it in

Yep, I see that it's missing. Created a ticket for it https://issues.sonatype.org/browse/RSC-1449

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,61 @@
...
 
-    expect(releaseEmpty).not.toExist();
-    expect(release).toHaveText('React Shared Components: 3.1.4');
+    expect(viewWithoutReleaseText.querySelector('.nx-global-sidebar__release')).not.toBeInTheDocument();
```
rpokorny, line 48: This is, of course an impl detail, but I can see how it can be challenging to know how to test the negative case when it just doesn't add anything.  How about tests for each optional prop _by themselves_ which assert not only that the render includes the relevant text but that it includes _only_ that text (and anything else that's supposed to be there by default, if there is any).  And then aside from those tests of the props individually have one other test where they're all used together and we assert that the text from all of them is present.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,61 @@
...
+import NxGlobalSidebarFooter, { NxGlobalSidebarFooterProps as Props } from '../NxGlobalSidebarFooter';
+import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
 
 describe('NxGlobalSidebarFooter', function() {
   const minimalProps: Props = {},
-      getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxGlobalSidebarFooter, minimalProps);
+      quickRender = rtlRender<Props>(NxGlobalSidebarFooter, minimalProps),
+      renderEl = rtlRenderElement<Props>(NxGlobalSidebarFooter, minimalProps);
+
+  it('passes additional classes specified as props to <footer>', function() {
```
rpokorny, line 15: Nothing about it being a `<footer>` is being asserted here.

Looking at what roles we might want to look for, I find some interesting things in the specs.  `<footer>` has a default role of `contentinfo` unless it is a descendant of a of certain other elements, include those with `role="complementary"` (which the sidebar overall has).  In those cases, it just has a role of `generic` (like a div).  Further, a `<footer>` that is a descendant of a `complementary` element _may not_ be given the `contentinfo` role manually.  Looking at the definition of the `contentinfo` role, it does seem like a fit for our footer, and yet we are blocked from applying it by the rules in the specs.

It makes me think maybe we have the role for the sidebar itself wrong.  Maybe the sidebar should be a `group`, an extra wrapping `complementary` element should be added around the sidebar header, navigation, and extra content area (but not the footer), and the footer should be allowed to gain the `contentinfo` role that it would then naturally have.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarNavigation.test.tsx
```diff
@@ -4,20 +4,26 @@
...
+import { rtlRenderElement } from '../../../__testutils__/rtlUtils';
 
 describe('NxGlobalSidebarNavigation', function() {
   const minimalProps: Props = {},
-      getShallowComponent = enzymeUtils.getShallowComponent<Props>(NxGlobalSidebarNavigation, minimalProps);
+      renderEl = rtlRenderElement<Props>(NxGlobalSidebarNavigation, minimalProps);
 
-  it('renders a <nav> with nx-global-sidebar__navigation class and those passed as props', function() {
-    expect(getShallowComponent({ className: 'class-A' })).toMatchSelector('nav.nx-global-sidebar__navigation.class-A');
+  it('passes additional classes specified as props to <nav>', function() {
```
rpokorny, line 14: `<nav>` isn't being tested here.  Check for the `navigation` role.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarNavigationLink.test.tsx
```diff
@@ -20,33 +18,80 @@ describe('NxGlobalSidebarNavigationLink', function() {
...
+      await user.hover(link);
+
+      expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+    });
 
-  it('renders an NxOverflowTooltip as container', function () {
-    expect(getShallowComponent()).toMatchSelector(NxOverflowTooltip);
   });
 
   it('renders an <a> with the passed href and classes', function() {
```
rpokorny, line 76:
```suggestion
  it('renders a link with the passed href and classes', function() {
```

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > Instead of asserting that the div is in the DOM, perhaps the test could look at if the component renders link in a variety of situations: (if neither are provided, if the supportLink is not provided but the supportText is, if the supportLink is provided and support text isn't, and yes if both are provided). And then still assert the href attribute and that the text is correct.

I've modified the tests (also incudes your suggestions) here: 4d234960d5a980fe8bfd11ddb392c5085d58bb8c

avnvbhatta:
> For all the footer content that accepts both string or JSX, there should probably be a second test confirming everything renders correctly when you use JSX as a prop as well.

Please check: 4d234960d5a980fe8bfd11ddb392c5085d58bb8c

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarNavigationLink.test.tsx
avnvbhatta: > maybe test classes as we have in other tests with the comparison of an element with default classes vs a custom class name

Added here: 56471ee4b334cff22b1734ff484818241c41cd87

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > And for the other `toBeVisible` assertions

I went with `toBeVisible` since it already asserts that the element is in the document. Furthermore, it also checks to see if the text is actually visible to the users. https://github.com/testing-library/jest-dom#tobevisible

Let me know what you think.

avnvbhatta:
> NxGlobalSidebarNavigation

Added here: 29a5e37f3f19c8e851e36ca3c1c6c4b0d2e42bc3

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,61 @@
...
+        supportLink = viewWithSupportText.getByRole('link');
 
-    expect(supportEmpty).not.toExist();
-    expect(support).toContainMatchingElement('span.nx-global-sidebar__support-text');
-    expect(support.find('span.nx-global-sidebar__support-text')).toHaveText('Support for RSC');
-    expect(supportLink).toHaveProp('href', '#supporturl');
+    expect(viewWithSupportText.getByText('Support for RSC')).toBeVisible();
+    expect(supportLink).toHaveAttribute('href', '#supporturl');
+
+    expect(viewWithoutSupportText.querySelector('.nx-global-sidebar__support')).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
avnvbhatta: > Seems like it'd be better to query by role for `link` here

Changed here: 74df155d607da545c62b64d775c9e736e1a70bf4

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > This is, of course an impl detail, but I can see how it can be challenging to know how to test the negative case when it just doesn't add anything. How about tests for each optional prop _by themselves_ which assert not only that the render includes the relevant text but that it includes _only_ that text (and anything else that's supposed to be there by default, if there is any). And then aside from those tests of the props individually have one other test where they're all used together and we assert that the text from all of them is present.

Assuming that you suggested to take out the `querySelector` tests for negative cases, and only make organizational changes, I *think* I made changes based on your suggestions. 

Can you please check 4d234960d5a980fe8bfd11ddb392c5085d58bb8c to make sure we're on the same page?

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarNavigation.test.tsx
avnvbhatta: > `<nav>` isn't being tested here. Check for the `navigation` role.

Added here: 29a5e37f3f19c8e851e36ca3c1c6c4b0d2e42bc3

rpokorny: > Added here: https://github.com/sonatype/sonatype-react-shared-components/commit/29a5e37f3f19c8e851e36ca3c1c6c4b0d2e42bc3

That's good but you didn't remove the mentions of `<nav>` from the other test names (which are not in fact testing for the `nav` tagName)

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,131 @@
...
+      });
 
-  it('renders the Release div with a text and version number and does not render when empty', function() {
-    const releaseEmpty = getShallowComponent().find('div.nx-global-sidebar__release'),
-        release = getShallowComponent({ releaseText: 'React Shared Components: 3.1.4' })
-            .find('div.nx-global-sidebar__release');
+      it('renders the Support div with a link with the passed href and jsx ' +
+      'if supportText and supportLink props are provided', function() {
+        const jsx =
+          <>
```
rpokorny, line 47: Why the fragment?

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,131 @@
...
+        productTagLine: 'Powered by PLAID VILLAIN'
+      });
 
-  it('renders the Created By text by default and does not render when set to false', function() {
-    const createdByTrue = getShallowComponent().find('div.nx-global-sidebar__created-by'),
-        createdByFalse = getShallowComponent({ showCreatedBy: false }).find('div.nx-global-sidebar__created-by');
+      expect(view.getByText('Support for RSC')).toBeVisible();
+      expect(view.getByRole('link')).toHaveTextContent(/^Support for RSC$/);
+      expect(view.getByText('React Shared Components: 3.1.4')).toBeVisible();
+      expect(view.getByText('Powered by PLAID VILLAIN')).toBeVisible();
```
rpokorny, line 126: Include the "Created by" line in this test as well

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > Nothing about it being a `<footer>` is being asserted here.
> 
> Looking at what roles we might want to look for, I find some interesting things in the specs. `<footer>` has a default role of `contentinfo` unless it is a descendant of of certain other elements, including those with `role="complementary"` (which the sidebar overall has). In those cases, it just has a role of `generic` (like a div). Further, a `<footer>` that is a descendant of a `complementary` element _may not_ be given the `contentinfo` role manually. Looking at the definition of the `contentinfo` role, it does seem like a fit for our footer, and yet we are blocked from applying it by the rules in the specs.
> 
> It makes me think maybe we have the role for the sidebar itself wrong. Maybe the sidebar should be a `group`, an extra wrapping `complementary` element should be added around the sidebar header, navigation, and extra content area (but not the footer), and the footer should be allowed to gain the `contentinfo` role that it would then naturally have.

I think what you're saying makes a lot of sense. Wrapping the sidebar header, navigation and extra content with a `role="complementary"` div and not the footer would naturally give the footer a role of `contentInfo`, which would alleviate most of our concerns. However, this might be a bit tricky to achieve with the current implementation. 

Here's how the sidebar is structured currently:
```
    <aside className={classes} id={id} aria-label="global sidebar">
      <div className="nx-global-sidebar__header">
        <a className="nx-global-sidebar__product-info nx-global-sidebar__expanded-content"
           href={logoLink}>
          { logo }
        </a>
        { toggleButton }
      </div>
      { children }
    </aside>
```

Wrapping the header is quite easy, but the navigation section (rendered through `NxGlobalSidebarNavigation`) and the footer (rendered through `NxGlobalSidebarFooter`), both are rendered here as `children` of `NxGlobalSidebar`. The extra content area is also rendered as part of the footer.

So wrapping the header, navigation and extra content with one role and leaving out the footer, might not be possible. Thoughts?

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarNavigation.test.tsx
avnvbhatta: > > Added here: [29a5e37](https://github.com/sonatype/sonatype-react-shared-components/commit/29a5e37f3f19c8e851e36ca3c1c6c4b0d2e42bc3)
> 
> That's good but you didn't remove the mentions of `<nav>` from the other test names (which are not in fact testing for the `nav` tagName)

Removed here: dffc31526a06b0b950b8b23f9c3f3316cf973987

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > Why the fragment?

No reason. Something about fragments just make it look more jsx-y

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > Include the "Created by" line in this test as well

Added here: 84425bca6c7f0554988dca551aed9fa25edeca2c

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
EmilyBermudez88: Looks good!

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
rpokorny: > Yep, that makes sense since it is asserting it's in the document. It may be splitting hairs a bit, but do we need to assert the text is visible? (aka is there a situation where the content is in the document but not visible?). As I said, it's probably unnecessary. I'm good with toBeVisible based on your suggestion!

I don't have a strong opinion on this, but consider that assertions of visibility are of limited effectiveness when our stylesheets aren't loaded (and the browser engine doesn't actually do layout anyway)

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
rpokorny: > So wrapping the header, navigation and extra content with one role and leaving out the footer, might not be possible. Thoughts?

Ah I see, you'll have to dig through the react children which is kind of ugly and inflexible (it requires the `NxGlobalSidebarFooter` to be a direct JSX child of the `NxGlobalSidebar` without even a fragment in between).  It's not unprecedented though, and in fact we have a utility that's perfect for this case.  See `childUtil.ts#splitOutFirst`.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
rpokorny: > No reason. Something about fragments just make it look more jsx-y

I'm just thinking that without it this assigment would be one line rather than four.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > > So wrapping the header, navigation and extra content with one role and leaving out the footer, might not be possible. Thoughts?
> 
> Ah I see, you'll have to dig through the react children which is kind of ugly and inflexible (it requires the `NxGlobalSidebarFooter` to be a direct JSX child of the `NxGlobalSidebar` without even a fragment in between). It's not unprecedented though, and in fact we have a utility that's perfect for this case. See `childUtil.ts#splitOutFirst`.

Oh yes, the `childUtil.ts#splitOutFirst` would definitely work, but one concern is that the extra content area (support, release, product and created by divs) are rendered as a part of  `NxGlobalSidebarFooter`, and as a result would be within the `<footer>`. 

I'm wondering if the extra wrapping `complementary` element should be added to only wrap the header and navigation, and leave the extra content to be rendered within the `<footer>`.

rpokorny: > Oh yes, the childUtil.ts#splitOutFirst would definitely work, but one concern is that the extra content area (support, release, product and created by divs) are rendered as a part of NxGlobalSidebarFooter, and as a result would be within the <footer>.

I'm not sure what you mean.  Of these four children, I'm saying the first three should be wrapped in a complementary
image

avnvbhatta: > > Oh yes, the childUtil.ts#splitOutFirst would definitely work, but one concern is that the extra content area (support, release, product and created by divs) are rendered as a part of NxGlobalSidebarFooter, and as a result would be within the .
> 
> I'm not sure what you mean. Of these four children, I'm saying the first three should be wrapped in a complementary image

I had misunderstood what you said by extra content. I thought you were talking about the support, release text etc.
I have made the following changes:
Change sidebar role and update tests: 18cd56b8d6623037686896e014c4708233648f6a
Add test for extra content: 9b5e56aaab6af29e5a73b8eb9479e1d27dc6602d
Update documentation about NxGlobalSidebarFooter to be direct child: e5a9b960be168164c0cc447b2b850736bae5fd82

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.tsx
```diff
@@ -47,18 +49,22 @@ const NxGlobalSidebar: FunctionComponent<Props> = function NxGlobalSidebar(props
...
   return (
-    <aside className={classes} id={id} aria-label="global sidebar">
-      <div className="nx-global-sidebar__header">
-        <a className="nx-global-sidebar__product-info nx-global-sidebar__expanded-content"
-           href={logoLink}>
-          { logo }
-        </a>
-        { toggleButton }
+    <aside className={classes} id={id} aria-label="global sidebar" role="group">
+      <div role="complementary">
```
rpokorny, line 57: If you make this the `aside` and the other element a `div` then you don't need to be explicit with the role here.

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
```diff
@@ -5,98 +5,200 @@
...
-    expect(getShallowComponent({ isOpen: false })).toMatchSelector('.nx-global-sidebar.closed');
+  describe('children', function() {
+    describe('when only the NxGlobalSidebarNavigation is a provided as a child', function () {
+      it('renders an element with role="navigation" within an element with role="complementary"', function() {
+        const children =
+          <NxGlobalSidebarNavigation key="1">
+            <NxGlobalSidebarNavigationLink icon={faCrow} text="testLink" href="testLink"/>
+          </NxGlobalSidebarNavigation>,
+            view = quickRender({ ...minimalProps, children }),
+            group = view.getByRole('group'),
```
rpokorny, line 45: Since the test above confirms that the top-level element is the `group`, I think you can skip doing `within(group).getByRole('complementary')` and instead just do `view.getByRole('complementary')`

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
```diff
@@ -5,98 +5,200 @@
...
+                                 supportLink="https://github.com/sonatype/sonatype-react-shared-components"
+                                 releaseText="Release 3.1.4"
+                                 productTagLine="Powered by PLAID VILLAIN" />,
+            view = quickRender({ ...minimalProps, children }),
+            group = view.getByRole('group'),
+            footer = within(group).getByRole('contentinfo');
+
+        expect(group).toBeInTheDocument();
+        expect(footer).toBeInTheDocument();
+      });
```
rpokorny, line 70: Maybe confirm that the footer is not inside the `complementary`

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.tsx
avnvbhatta: > If you make this the `aside` and the other element a `div` then you don't need to be explicit with the role here.

Changed here: bc41d7e723b1445e48ec2ec12229e666e5bc4aec

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
avnvbhatta: > Since the test above confirms that the top-level element is the `group`, I think you can skip doing `within(group).getByRole('complementary')` and instead just do `view.getByRole('complementary')`

Changed here: 91ffc6b835712f1865e7a7c49f391747870d40e1

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebar.test.tsx
avnvbhatta: > Maybe confirm that the footer is not inside the `complementary`

Good point. Added here: 32fd0c4c07b15486ce3d783467113284475e9dc6

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
```diff
@@ -4,63 +4,140 @@
...
+      renderEl = rtlRenderElement<Props>(NxGlobalSidebarFooter, minimalProps);
+
+  it('renders a top-level <footer> element with role="contentinfo"', function() {
+    const view = quickRender(),
+        group = view.getByRole('contentinfo');
 
-  it('renders an <footer> with nx-global-sidebar__footer class and those passed as props', function() {
-    expect(getShallowComponent({ className: 'test-class' }))
-        .toMatchSelector('footer.nx-global-sidebar__footer.test-class');
+    expect(group.tagName).toBe('FOOTER');
```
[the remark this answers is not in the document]
avnvbhatta: > I don't think we need to worry about the tagName anymore now that we've got the role correct

Remove tagName: 072a192e35ad3b5054199de31dc8249991f4fd48

lib/src/components/NxGlobalSidebar/__tests__/NxGlobalSidebarFooter.test.tsx
avnvbhatta: > > No reason. Something about fragments just make it look more jsx-y
> 
> I'm just thinking that without it this assigment would be one line rather than four.

Changed here: c134dd4dde59c96fd44c92cd2229bbc9c09e3f56

rpokorny (approved):

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.10.12",
```
rpokorny, line 3: This doesn't seem right

rpokorny:
I think there's unrelated changes showing up in the diff currently

avnvbhatta:
> I think there's unrelated changes showing up in the diff currently

Everything should be fixed now.

PR #949, opened 2023-02-03
Hide text filter input - Collapsible Multi/Radio Selects - RSC-1439

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1439

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.10.13",
+  "version": "12.10.14",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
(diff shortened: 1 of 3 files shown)


entroform:
Just wondering if we should add a test for this.

avnvbhatta:
> Just wondering if we should add a test for this.

Yep, but we decided not to add enzyme tests. I've created another ticket for it:
https://issues.sonatype.org/browse/RSC-1457

entroform (approved):
LGTM :)

rpokorny:
> Yep, but we decided not to add enzyme tests. I've created another ticket for it:
https://issues.sonatype.org/browse/RSC-1457

Seeing as the fix here was CSS I have to think the only test that can properly verify this would be a visual test, not RTL

PR #953, opened 2023-02-07
useFuzzyFilter RTL - RSC-1453

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1453

### DIFF
--- a/lib/src/util/__tests__/useFuzzyFilter.test.tsx
+++ b/lib/src/util/__tests__/useFuzzyFilter.test.tsx
@@ -83,12 +83,13 @@ describe('useFuzzyFilter', function () {
     expect(outputItems[3]).toHaveTextContent('Moped');
   });
 
-  it('returns unchanged input if filter was set to empty string', async function () {
+  it('returns unchanged input if filter was set to empty string', function () {
     const view = renderWithFilter(),
         outputItems = view.getAllByTestId('result'),
         inputEl = view.getByRole('textbox');
 
     expect(inputEl).toHaveValue('');
+
     expect(outputItems.length).toBe(4);
     expect(outputItems[0]).toHaveTextContent('Bicycle');
     expect(outputItems[1]).toHaveTextContent('Motorcycle');
@@ -100,12 +101,13 @@ describe('useFuzzyFilter', function () {
     const user = userEvent.setup(),
         view = renderWithFilter(),
         inputEl = view.getByRole('textbox');
+    let outputItems = view.getAllByTestId('result');
 
+    expect(outputItems.length).toBe(4);
     await user.type(inputEl, '[Space]');
 
-    const outputItems = view.getAllByTestId('result');
+    outputItems = view.getAllByTestId('result');
     expect(outputItems.length).toBe(4);
-    expect(inputEl).toHaveValue(' ');
   });
 
   it('filters input and returns new filter term', async function() {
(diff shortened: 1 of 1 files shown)


lib/src/util/__tests__/useFuzzyFilter.test.tsx
```diff
@@ -51,50 +45,129 @@ describe('useFuzzyFilter', function () {
...
+        outputItems = view.getAllByTestId('result');
+
+    expect(outputItems.length).toBe(4);
+    expect(outputItems[0]).toHaveTextContent('Bicycle');
+    expect(outputItems[1]).toHaveTextContent('Motorcycle');
+    expect(outputItems[2]).toHaveTextContent('Skateboard');
+    expect(outputItems[3]).toHaveTextContent('Moped');
   });
 
   it('returns unchanged input if filter was set to empty string', function () {
```
EmilyBermudez88, line 86: I'm unsure if this test is working as intended. useFuzzyFilter initially sets filterTerm as an empty string, so I don't THINK setFilterTerm needs to be called for this test? But if that's the case, is this simply the same test as the before it?

lib/src/util/__tests__/useFuzzyFilter.test.tsx
```diff
@@ -51,50 +45,129 @@ describe('useFuzzyFilter', function () {
...
+
+    return (
+      <>
+        <input onChange={onChange} value={filterTerm.toString()}/>
+        { output.map((result) => <span data-testid="result" key={result.id}>{result.name}</span>) }
+      </>
+    );
+  }
+
+  const renderWithFilter = rtlRender<Props>(FixtureWithFilter, {input, options});
```
avnvbhatta, line 64:
```suggestion
  const renderWithFilter = rtlRender<Props>(FixtureWithFilter, { input, options });
```

lib/src/util/__tests__/useFuzzyFilter.test.tsx
avnvbhatta: > I'm unsure if this test is working as intended. useFuzzyFilter initially sets filterTerm as an empty string, so I don't THINK setFilterTerm needs to be called for this test? But if that's the case, is this simply the same test as the before it?

I'm in favor of removing this test. Looks like either way we're testing the same "filter is not set" vs "filter is empty", which basically looks to me, like the same thing.

lib/src/util/__tests__/useFuzzyFilter.test.tsx
```diff
@@ -51,50 +45,129 @@ describe('useFuzzyFilter', function () {
...
+    let outputItems = view.getAllByTestId('result');
+
+    expect(outputItems.length).toBe(4);
+    await user.type(inputEl, 'scooter');
+
+    outputItems = view.getAllByTestId('result');
+    expect(outputItems.length).toBe(1);
+    expect(outputItems[0]).toHaveTextContent('Moped or Scooter');
+    expect(inputEl).toHaveValue('scooter');
+  });
```
avnvbhatta, line 158: I don't think this test is necessary. For an approximate string matcher, it's bound to find 'scooter' regardless of where it appears.

EmilyBermudez88: > I don't think this test is necessary. For an approximate string matcher, it's bound to find 'scooter' regardless of where it appears.

I was unsure about this one as well. I used `fuzzyFilter.test.ts` as a guideline for tests, and it had this test included. Remove?

avnvbhatta: > > I don't think this test is necessary. For an approximate string matcher, it's bound to find 'scooter' regardless of where it appears.
> 
> I was unsure about this one as well. I used `fuzzyFilter.test.ts` as a guideline for tests, and it had this test included. Remove from `useFuzzyFilter.test.ts`?

I looked a bit into this and since our fuse.js options set the `ignoreLocation` to true, I think the test should stay. https://fusejs.io/api/options.html#ignorelocation

EmilyBermudez88: > I looked a bit into this and since our fuse.js options set the `ignoreLocation` to true, I think the test should stay. https://fusejs.io/api/options.html#ignorelocation

Thanks for clarifying!

PR #955, opened 2023-02-07
RTL Conversion - NxModal - RSC-1450

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1450

### DIFF
--- a/lib/src/components/NxModal/__tests__/NxModal.test.tsx
+++ b/lib/src/components/NxModal/__tests__/NxModal.test.tsx
@@ -12,7 +12,7 @@ import { render, within } from '@testing-library/react';
 describe('NxModal', function() {
   const minimalProps: Props = {
     children: 'A message to show in a modal',
-    onClose: jest.fn()
+    onClose: () => {}
   };
 
   type PropsWithRef = Props & RefAttributes<HTMLDialogElement>;
--- a/lib/src/components/NxModal/__tests__/NxModal.test.tsx
+++ b/lib/src/components/NxModal/__tests__/NxModal.test.tsx
@@ -188,7 +188,7 @@ describe('NxModal', function() {
   });
 
   describe('NxModal.Header', function() {
-    it('makes a <header> tag with the nx-modal-header class', function() {
+    it('renders an element with role="banner"', function() {
       const view = render(
         <NxModal { ...minimalProps }><NxModal.Header>Test</NxModal.Header></NxModal>
       );
--- a/lib/src/components/NxModal/__tests__/NxModal.test.tsx
+++ b/lib/src/components/NxModal/__tests__/NxModal.test.tsx
@@ -4,79 +4,63 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React from 'react';
-import {mount, shallow} from 'enzyme';
-import 'jest-enzyme';
-
-import { getMountedComponent } from '../../../__testutils__/enzymeUtils';
+import React, { RefAttributes } from 'react';
 import NxModal, { Props } from '../NxModal';
-import NxTooltip from '../../NxTooltip/NxTooltip';
-import NxButton from '../../NxButton/NxButton';
-import { Tooltip } from '@material-ui/core';
-import { runTimers } from '../../../__testutils__/rtlUtils';
+import { rtlRender, rtlRenderElement, runTimers, userEvent } from '../../../__testutils__/rtlUtils';
+import { render, within } from '@testing-library/react';
 
 describe('NxModal', function() {
-  const dummyCloseHandler = jest.fn();
   const minimalProps: Props = {
     children: 'A message to show in a modal',
-    onClose: dummyCloseHandler
+    onClose: jest.fn()
   };
 
-  const getMounted = getMountedComponent<Props>(NxModal, minimalProps),
-      getModal = (props?: Partial<Props>) => getMounted(props).find('dialog');
+  type PropsWithRef = Props & RefAttributes<HTMLDialogElement>;
+  const quickRender = rtlRender<PropsWithRef>(NxModal, minimalProps),
+      renderEl = rtlRenderElement<PropsWithRef>(NxModal, minimalProps);
 
-  it('renders an nx-modal-backdrop <dialog> containing an nx-modal <div>', function () {
-    const nxModal = getModal();
+  it('renders a top-level element with role="dialog"', function () {
+    const view = quickRender();
 
-    expect(nxModal).toMatchSelector('dialog.nx-modal-backdrop');
-    expect(nxModal.children()).toMatchSelector('div.nx-modal');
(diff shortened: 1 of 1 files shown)


lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -4,79 +4,64 @@
...
-import { runTimers } from '../../../__testutils__/rtlUtils';
+import { rtlRender, rtlRenderElement, runTimers, userEvent } from '../../../__testutils__/rtlUtils';
+import { render, within } from '@testing-library/react';
 
 describe('NxModal', function() {
-  const dummyCloseHandler = jest.fn();
   const minimalProps: Props = {
     children: 'A message to show in a modal',
-    onClose: dummyCloseHandler
+    onClose: jest.fn()
```
EmilyBermudez88, line 15:
```suggestion
    onClose: () => {}
```

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -4,79 +4,64 @@
...
   });
 
-  it('includes any passed in attributes to the nx-modal div', function() {
-    const nxModal = getModal({ id: 'modal-id', lang: 'en_US' });
-
-    expect(nxModal.find('.nx-modal').prop('id')).toEqual('modal-id');
-    expect(nxModal.find('.nx-modal').prop('lang')).toEqual('en_US');
-  });
+  it('includes any passed in attributes to the modal div', function() {
+    const el = renderEl({ id: 'modal-id', lang: 'en_US' })?.firstElementChild;
```
EmilyBermudez88, line 57:
```suggestion
    const el = renderEl({ id: 'modal-id', lang: 'en_US' })!.firstElementChild;
```

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -96,93 +81,83 @@ describe('NxModal', function() {
...
+          user = userEvent.setup();
 
+      quickRender({ onCancel: mockCallBack });
       expect(mockCallBack).not.toHaveBeenCalled();
-      component.simulate('keyDown', createEvent());
+
+      await user.keyboard('[Escape]');
+
       expect(mockCallBack).toHaveBeenCalledTimes(1);
       expect(mockCallBack.mock.calls[0][0].type).toBe('cancel');
```
EmilyBermudez88, line 107: I saw this in Jest's docs at https://jestjs.io/docs/mock-functions, but curious if you know how this is constructed? I think the first [0] refers to the first call made and the second [0] refers to the argument passed, but I'm just a bit unsure about how it comes up first with `mock.calls`- is there a way in JS/React to parse down into events like that to get the type?

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -96,93 +81,83 @@ describe('NxModal', function() {
...
-      component.simulate('keyDown', createEvent('q'));
-      component.simulate('keyDown', createEvent('Q'));
       expect(mockCallBack).not.toHaveBeenCalled();
     });
 
-    it('calls stopPropagation and stopImmediatePropagation on Escape keydowns', function() {
-      const component = getModal({ onClose: jest.fn() }),
-          escEvent = createEvent(),
-          otherEvent = createEvent('q');
+    it('executes onCancel method ONLY when pressing ESC key', async function () {
```
EmilyBermudez88, line 124: does onCancel get called when you close the alert by clicking the close button in the modal footer?

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -4,79 +4,64 @@
...
 
-    expect(nxModal).toMatchSelector('dialog.nx-modal-backdrop');
-    expect(nxModal.children()).toMatchSelector('div.nx-modal');
+    expect(dialog).toBeInTheDocument();
+    expect(dialog).toBe(view.container.firstElementChild);
   });
 
   it('renders children nodes within the modal', function() {
-    const nxModal = getModal({ children: <div className="bar"/> });
+    const view = quickRender({ children: <div className="bar" data-testid="test-div" /> }),
```
EmilyBermudez88, line 31:
```suggestion
    const view = quickRender({ children: <div data-testid="test-div" /> }),
```
Probably fine without the className included

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -96,93 +81,83 @@ describe('NxModal', function() {
...
+    });
   });
 
   it('moves focus back to the previously focused element when closed', async function() {
     function Fixture({ modalOpen }: { modalOpen: boolean }) {
       return (
         <>
-          <button id="test-btn">Test</button>
+          <button data-testid="test-btn">Test</button>
           { modalOpen && <NxModal onCancel={jest.fn()}><button id="cancel-btn">Close</button></NxModal> }
```
entroform, line 161: I wonder if the `id="cancel-btn"` is needed and if not could be removed.

lib/src/components/NxModal/__tests__/NxModal.test.tsx
avnvbhatta: > does onCancel get called when you close the alert by clicking the close button in the modal footer? And if that is true, should there be a test proving the modal doesn't close when clicking on the dialog element/modal backdrop?

`onCancel` doesn't get called when a button in the modal footer is clicked.

lib/src/components/NxModal/__tests__/NxModal.test.tsx
avnvbhatta: > I wonder if the `id="cancel-btn"` is needed and if not could be removed.

Yup, can certainly be removed. df890b63f1dccfd8a2311c44c231a9bdbacd2ba2

lib/src/components/NxModal/__tests__/NxModal.test.tsx
EmilyBermudez88: > `onCancel` doesn't get called when a button in the modal footer is clicked.

Thanks for clarifying!

lib/src/components/NxModal/__tests__/NxModal.test.tsx
avnvbhatta: > I saw this in Jest's docs at https://jestjs.io/docs/mock-functions, but curious if you know how this is constructed? I think the first [0] refers to the first call made and the second [0] refers to the argument passed, but I'm just a bit unsure about how it comes up first with `mock.calls`- is there a way in JS/React to parse down into events like that to get the type?

I think you're right. `mock.calls` looks to be an array of the mock calls made. The call itself is another array that stores the arguments passed into it. I think this is just the structure that jest follows. I think we can just use the Event interface to get the type.

entroform (approved):
This looks good to me,
just need to change the base branch to `RSC-1438_password-manager-nx-modal`
😸

PR #957, opened 2023-02-08
Update Visual Tests - NxCollapsible Multi/Radio Select - RSC-1457

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1457

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.10.15",
+  "version": "12.10.16",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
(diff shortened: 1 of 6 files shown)


EmilyBermudez88:
I think `moveMouseAway()` needs to be added to the tests? The background colour of the trigger seems to be off and when I run the test again with `moveMouseAway()` it fails, so there may be a bit of colour on the background that's hard to tell on the snapshot?

avnvbhatta:
> I think `moveMouseAway()` needs to be added to the tests? The background colour of the trigger seems to be off and when I run the test again with `moveMouseAway()` it fails, so there may be a bit of colour on the background that's hard to tell on the snapshot?

Added `moveMouseAway` here: 58eb84265417e0486609d96a27c2e07cbd4b6a71

EmilyBermudez88 (approved):
LGTM!

PR #958, opened 2023-02-08
Deprecate NxDropdown variant prop - RSC-1431

entroform:
https://issues.sonatype.org/browse/RSC-1431

### DIFF
--- a/gallery/src/components/NxDropdown/NxDropdownPage.tsx
+++ b/gallery/src/components/NxDropdown/NxDropdownPage.tsx
@@ -87,7 +87,7 @@ const NxDropdownPage = () =>
                 Defaults to <NxCode>"tertiary"</NxCode>
                 <NxWarningAlert>
                   Deprecated: This prop should not be used.
-                  It produces button styling that does not look good.
+                  It produces button styling that does not comply with design standards.
                 </NxWarningAlert>
               </NxTable.Cell>
             </NxTable.Row>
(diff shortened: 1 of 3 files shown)


gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -85,6 +85,11 @@ const NxDropdownPage = () =>
               <NxTable.Cell>
                 What type of button to render for the dropdown.
                 Defaults to <NxCode>"tertiary"</NxCode>
+
+                <NxWarningAlert>
+                  Deprecated: This prop should not be used.
+                  It produces button styling that does not look good.
```
avnvbhatta, line 91:
```suggestion
                  It produces button styling that does not comply with design standards.
```

gallery/src/components/NxDropdown/NxDropdownPage.tsx
```diff
@@ -85,6 +85,11 @@ const NxDropdownPage = () =>
               <NxTable.Cell>
                 What type of button to render for the dropdown.
                 Defaults to <NxCode>"tertiary"</NxCode>
+
```
avnvbhatta, line 88:
```suggestion
```

gallery/src/components/NxDropdown/NxDropdownPage.tsx
avnvbhatta: Let me know your thoughts on this. Might be a bit better than "does not look good".

entroform: >Let me know your thoughts on this. Might be a bit better than "does not look good".
sounds good to me!

PR #959, opened 2023-02-08
NxTile RTL Conversion- RSC-1451

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1451

### DIFF
--- a/lib/src/components/NxTile/__tests__/NxTile.test.tsx
+++ b/lib/src/components/NxTile/__tests__/NxTile.test.tsx
@@ -10,7 +10,7 @@ import { render, within } from '@testing-library/react';
 
 import NxTile from '../NxTile';
 
-function testMergedClassNames(Component: ComponentType<{}>) {
+function testMergedClassNames(Component: ComponentType) {
   return () => {
     const renderEl = rtlRenderElement(Component, {});
 
--- a/lib/src/components/NxTile/__tests__/NxTile.test.tsx
+++ b/lib/src/components/NxTile/__tests__/NxTile.test.tsx
@@ -4,68 +4,100 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import React from 'react';
-import { shallow } from 'enzyme';
-import 'jest-enzyme';
+import React, { ComponentType } from 'react';
+import { render, screen } from '@testing-library/react';
+import { rtlRenderElement, runTimers, userEvent } from '../../../__testutils__/rtlUtils';
 
 import NxTile from '../NxTile';
 
+function test(Component: ComponentType<{}>, tagName: string, ...classNames: string[]) {
+  return () => {
+    const el = render(<Component/>).container.firstElementChild!;
+    expect(el.tagName).toBe(tagName.toUpperCase());
+
+    for (const className of classNames) {
+      expect(el).toHaveClass(className);
+    }
+  };
+}
+
 describe('NxTile', function() {
-  it('makes a <section> tag with an nx-tile class', function() {
-    expect(shallow(<NxTile/>)).toMatchSelector('section.nx-tile');
-  });
+  it('makes a <section> tag with an nx-tile class', test(NxTile, 'section', 'nx-tile'));
 });
 
 describe('NxTile.Header', function() {
-  it('makes a <header> tag with an nx-tile-header class', function() {
-    expect(shallow(<NxTile.Header/>)).toMatchSelector('header.nx-tile-header');
-  });
+  it('makes a <header> tag with an nx-tile-header class', test(NxTile.Header, 'header', 'nx-tile-header'));
 });
 
 describe('NxTile.HeaderTitle', function() {
-  it('makes a <div> tag with an nx-tile-header__title class', function() {
-    expect(shallow(<NxTile.HeaderTitle/>).children()).toMatchSelector('div.nx-tile-header__title');
+  const renderEl = rtlRenderElement(NxTile.HeaderTitle, {});
+
+  it('makes a <div> tag with an nx-tile-header__title class', test(NxTile.HeaderTitle, 'div', 'nx-tile-header__title'));
+
+  it('renders a tooltip when the title text is too long to be fully seen', async function() {
+  // Mock layout methods that NxOverflowTooltip relies on in order to make it think overflow is occurring
+    jest.spyOn(Element.prototype, 'getBoundingClientRect').mockReturnValue({
+      x: 0,
+      y: 0,
+      width: 1,
+      height: 1,
+      top: 0,
+      right: 1,
+      bottom: 1,
+      left: 0
+    } as DOMRect);
+
+    // mock that text extends 1px farther than container
+    jest.spyOn(Range.prototype, 'getBoundingClientRect').mockReturnValue({
+      x: 0,
+      y: 0,
+      width: 2,
+      height: 1,
+      top: 0,
+      right: 2,
+      bottom: 1,
+      left: 0
+    } as DOMRect);
+
+    const el = renderEl({ children: <h2>Title</h2>})!,
+        user = userEvent.setup();
+
+    await user.hover(el);
+    await runTimers();
+
+    expect(screen.getByRole('tooltip')).toBeInTheDocument();
+    expect(screen.getByRole('tooltip')).toHaveTextContent('Title');
   });
 });
 
 describe('NxTile.Headings', function() {
-  it('makes a <hgroup> tag with an nx-tile-header__headings class', function() {
-    expect(shallow(<NxTile.Headings/>)).toMatchSelector('hgroup.nx-tile-header__headings');
-  });
+  it('makes a <hgroup> tag with an nx-tile-header__headings class',
+      test(NxTile.Headings, 'hgroup', 'nx-tile-header__headings'));
 });
 
 describe('NxTile.HeaderSubtitle', function() {
-  it('makes a <h3> tag with an nx-tile-header__subtitle class', function() {
-    expect(shallow(<NxTile.HeaderSubtitle/>)).toMatchSelector('h3.nx-tile-header__subtitle');
-  });
+  it('makes a <h3> tag with an nx-tile-header__subtitle class',
+      test(NxTile.HeaderSubtitle, 'h3', 'nx-tile-header__subtitle'));
 });
 
 describe('NxTile.HeaderActions', function() {
-  it('makes a <div> tag with an nx-tile__actions class', function() {
-    expect(shallow(<NxTile.HeaderActions/>)).toMatchSelector('div.nx-tile__actions');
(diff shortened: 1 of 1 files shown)


lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,100 @@
...
 });
 
 describe('NxTile.HeaderTitle', function() {
-  it('makes a <div> tag with an nx-tile-header__title class', function() {
-    expect(shallow(<NxTile.HeaderTitle/>).children()).toMatchSelector('div.nx-tile-header__title');
+  const renderEl = rtlRenderElement(NxTile.HeaderTitle, {});
+
+  it('makes a <div> tag with an nx-tile-header__title class', test(NxTile.HeaderTitle, 'div', 'nx-tile-header__title'));
+
+  it('renders a tooltip when the title text is too long to be fully seen', async function() {
```
EmilyBermudez88, line 37: I added in this test because the docs specifically mention that a tooltip will be applied when the title text is too long, which is unlike `NxTile.HeaderSubtitle` which just wraps the text. I can remove if it seems unncessary

entroform: Nice catch! I think it's neat that you spy on the getBoundingClientRect to trigger the overflow. I wonder though if that's testing too much into the implementation details and it might be safer to add a functional test for this. Let me know what you think!

EmilyBermudez88: > Nice catch! I think it's neat that you spy on the getBoundingClientRect to trigger the overflow. I wonder though if that's testing too much into the implementation details and it might be safer to add a functional test for this. Additionally that will be more future proof because we would be testing the behaviour and if the implementation detail changes we don't have to rewrite this test 😸. Let me know what you think!

Good point. And having looked at the functional tests, there IS a test for the tooltip to appear already. So test has been removed: [de395c4](https://github.com/sonatype/sonatype-react-shared-components/pull/959/commits/de395c41b0125999f64b00b5e78f0008671ec14e)

lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,100 @@
...
-import React from 'react';
-import { shallow } from 'enzyme';
-import 'jest-enzyme';
+import React, { ComponentType } from 'react';
+import { render, screen } from '@testing-library/react';
+import { rtlRenderElement, runTimers, userEvent } from '../../../__testutils__/rtlUtils';
 
 import NxTile from '../NxTile';
 
+function test(Component: ComponentType<{}>, tagName: string, ...classNames: string[]) {
```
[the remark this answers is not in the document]
EmilyBermudez88: > Hmm I wonder if testing the specific classname and tagName is testing implementation details. You could test the role for some of these, header etc
> 
> ^ I don't think we should test the specific class name and tag (impl details), we should test some of them to render elements with the correct role. It might be a good idea to refactor this function for testing class merging with a descriptive name like `testMergePassedInClassName` 🤔

I definitely see your point, but I'm struggling to find another solution here. None of these elements minus `NxTile.Header` have an accessible role to query for (even `NxTile` you can't query for role region or section). Since all these elements are essentially containers to hold something else, do you have any recommendations to test these without classes/tagNames? 

If I were to use jsx to build a full modal to test its components (and then query for textContent etc), I think I'd be testing more that it renders children more than anything else. 

Any thoughts are appreciated!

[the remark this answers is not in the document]
entroform: > > Hmm I wonder if testing the specific classname and tagName is testing implementation details. You could test the role for some of these, header etc
> > ^ I don't think we should test the specific class name and tag (impl details), we should test some of them to render elements with the correct role. It might be a good idea to refactor this function for testing class merging with a descriptive name like `testMergePassedInClassName` 🤔
> 
> I definitely see your point, but I'm struggling to find another solution here. None of these elements minus `NxTile.Header` have an accessible role to query for (even `NxTile` you can't query for role region or section). Since all these elements are essentially containers to hold something else, do you have any recommendations to test these without classes/tagNames?
> 
> If I were to use jsx to build a full modal to test its components (and then query for textContent etc), I think I'd be testing more that it renders children more than anything else.
> 
> Any thoughts are appreciated!

Hi Emily, I think it may be safe to delete/ignore tests for some of the wrappers that don't have accessible roles and just have RTL tests for ones that does, I think that might be sufficient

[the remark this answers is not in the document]
avnvbhatta: > > > Hmm I wonder if testing the specific classname and tagName is testing implementation details. You could test the role for some of these, header etc
> > > ^ I don't think we should test the specific class name and tag (impl details), we should test some of them to render elements with the correct role. It might be a good idea to refactor this function for testing class merging with a descriptive name like `testMergePassedInClassName` 🤔
> > 
> > 
> > I definitely see your point, but I'm struggling to find another solution here. None of these elements minus `NxTile.Header` have an accessible role to query for (even `NxTile` you can't query for role region or section). Since all these elements are essentially containers to hold something else, do you have any recommendations to test these without classes/tagNames?
> > If I were to use jsx to build a full modal to test its components (and then query for textContent etc), I think I'd be testing more that it renders children more than anything else.
> > Any thoughts are appreciated!
> 
> Hi Emily, I think it may be safe to delete/ignore tests for some of the wrappers that don't have accessible roles and just have RTL tests for ones that does, I think that might be sufficient

I agree with Andrew. I think we have pretty good coverage for NxTile through visual tests as well. So for most of the wrapper tests, we can get rid of.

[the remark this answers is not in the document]
EmilyBermudez88: Thanks for both your thoughts! I removed the implementation details, but kept the generic testing that classNames can be added to the components (as mentioned by Andrew in the first comment of the thread).
[9bf028a](https://github.com/sonatype/sonatype-react-shared-components/pull/959/commits/9bf028a69b0c9fd27f2b90c0f9333d069c95bd16)

lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,92 @@
...
 
 describe('NxTile.HeaderSubtitle', function() {
-  it('makes a <h3> tag with an nx-tile-header__subtitle class', function() {
-    expect(shallow(<NxTile.HeaderSubtitle/>)).toMatchSelector('h3.nx-tile-header__subtitle');
+  it('renders a top level element with role=heading', function() {
+    const view = quickRender(NxTile.HeaderSubtitle),
+        subtitle = view.getByRole('heading');
+
+    expect(subtitle).toBeInTheDocument();
+    expect(subtitle).toBe(view.container.firstElementChild);
```
entroform, line 62: Same thing here (See comment below).

lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,92 @@
...
+
+    expect(customEl).toHaveClass('foo');
+
+    for (const cls of Array.from(defaultEl.classList)) {
+      expect(customEl).toHaveClass(cls);
+    }
+  };
+}
+
+function quickRender(Component: ComponentType<{}>) {
```
entroform, line 27: I think we can just use render from RTL directly and remove this function

lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,92 @@
...
  */
-import React from 'react';
-import { shallow } from 'enzyme';
-import 'jest-enzyme';
+import { ComponentType } from 'react';
+import { rtlRender, rtlRenderElement } from '../../../__testutils__/rtlUtils';
 
 import NxTile from '../NxTile';
 
+function testMergedClassNames(Component: ComponentType<{}>) {
```
entroform, line 12:
```suggestion
function testMergedClassNames(Component: ComponentType) {
```

lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,92 @@
...
 
 describe('NxTile.SubsectionHeader', function() {
-  it('makes a <header> tag with an nx-tile-subsection__header class', function() {
-    expect(shallow(<NxTile.SubsectionHeader/>)).toMatchSelector('header.nx-tile-subsection__header');
+  it('renders a top level element with role=banner', function() {
+    const view = quickRender(NxTile.SubsectionHeader),
+        subsectionHeader = view.getByRole('banner');
+
+    expect(subsectionHeader).toBeInTheDocument();
+    expect(subsectionHeader).toBe(view.container.firstElementChild);
```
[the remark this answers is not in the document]
EmilyBermudez88: > I'm not sure if this is needed, I think just checking if the banner role is sufficient, we can rename the test to simply: "renders a banner element"

Makes sense, simplified the test here: [76d7a29](https://github.com/sonatype/sonatype-react-shared-components/pull/959/commits/76d7a2945d216d072ec76ec467e06c731a24a0d0)

lib/src/components/NxTile/__tests__/NxTile.test.tsx
EmilyBermudez88: > I think we can just use render from RTL directly and remove this function

Yep, that's cleaner, updated with render here: [76d7a29](https://github.com/sonatype/sonatype-react-shared-components/pull/959/commits/76d7a2945d216d072ec76ec467e06c731a24a0d0)

lib/src/components/NxTile/__tests__/NxTile.test.tsx
EmilyBermudez88: > Same thing here (See comment below).

[76d7a29](https://github.com/sonatype/sonatype-react-shared-components/pull/959/commits/76d7a2945d216d072ec76ec467e06c731a24a0d0)

entroform (approved):
Yay looks good to me, thanks Emily!

lib/src/components/NxTile/__tests__/NxTile.test.tsx
```diff
@@ -4,68 +4,85 @@
...
-  it('makes a <header> tag with an nx-tile-subsection__header class', function() {
-    expect(shallow(<NxTile.SubsectionHeader/>)).toMatchSelector('header.nx-tile-subsection__header');
+  it('renders a banner element', function() {
+    const { container } = render(<NxTile.SubsectionHeader />);
+    const subsectionHeader = within(container).getByRole('banner');
+
+    expect(subsectionHeader).toBeInTheDocument();
   });
+
+  it('madds specified classnames in addition to the defaults',
```
avnvbhatta, line 82:
```suggestion
  it('adds specified classnames in addition to the defaults',
```

avnvbhatta (approved):

PR #961, opened 2023-02-09
Fire onSearch for NxCombobox when focused with mouse- RSC-1434

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1434

It seems as though `evt.relatedTarget` doesn't always return an element when clicking into the combobox (it only returns an element if that element is focusable). I replaced the `comingFromOutsidePage` with a state value that tracks if the document is focused (document.hasFocus() : https://developer.mozilla.org/en-US/docs/Web/API/Document/hasFocus


lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -87,29 +87,36 @@ function NxComboboxRender<T extends string | number | DataItem<string | number,
...
-      const comingFromOutsidePage = evt.relatedTarget === null,
+      const comingFromOutsidePage = pageIsFocused === false,
           comingFromChildNode = evt.relatedTarget instanceof Node && evt.currentTarget.contains(evt.relatedTarget);
 
       if (!(comingFromOutsidePage || comingFromChildNode)) {
         doSearch(value);
       }
     }
-
+    handlePageFocus();
```
avnvbhatta, line 113: Instead of the extra `handlePageFocus` method, we can replace this with `setPageIsFocused(document.hasFocus());`

EmilyBermudez88: > Instead of the extra `handlePageFocus` method, we can replace this with `setPageIsFocused(document.hasFocus());`

Thanks, that makes much more sense! [00e3956](https://github.com/sonatype/sonatype-react-shared-components/pull/961/commits/00e3956f5313c10c247493aa3b5ed53aea7f4fbc)

lib/src/components/NxCombobox/NxCombobox.tsx
```diff
@@ -87,29 +87,31 @@ function NxComboboxRender<T extends string | number | DataItem<string | number,
...
   // the component and then comes back to it the search should be retried automatically
   function handleComponentFocus(evt: FocusEvent<HTMLDivElement>) {
-    setInputIsFocused(true);
 
+    setInputIsFocused(true);
     if (loadError) {
       // check if this is focus coming into the component from somewhere else on the page, not just moving between
       // children of this component and not from focus coming into the browser from some other window
-      const comingFromOutsidePage = evt.relatedTarget === null,
+      const comingFromOutsidePage = pageIsFocused === false,
```
entroform, line 101: Just wondering if it would be simpler if we remove pageIsFocused state
and just have `!document.hasFocus()`

Something like:
```suggestion
      const comingFromOutsidePage = !document.hasFocus(),
```

EmilyBermudez88: > Just wondering if it would be simpler if we remove `pageIsFocused` state entirely and just have `!document.hasFocus()` here
> 

So I just tested this, which doesn't work unfortunately. I assume it's because the second you click back into the page, the document now has focus and updates immediately (instead of the state variable which is updated after the fact), and doSearch is called again.

avnvbhatta (approved):
PR title needs to be updated. But overall, +1!

lib/src/components/NxCombobox/NxCombobox.tsx
entroform: > > Just wondering if it would be simpler if we remove `pageIsFocused` state entirely and just have `!document.hasFocus()` here
> 
> So I just tested this, which doesn't work unfortunately. I assume it's because the second you click back into the page, the document now has focus and updates immediately (instead of the state variable which is updated after the fact), and doSearch is called again.

Ah I see

EmilyBermudez88:
I completely forgot about unit tests, so fixing some failures now. Testing pointed out that before, `onSearch` was only called in an errorState when the combobox receives focus when the focus is shifted from a different focusable component. We didn't notice this when tabbing through, but if you click on the example tile then tab into NxCombobox, it won't call `onSearch` either. So the tests that asserted `onSearch` wasn't called when the combobox was focused now fail. Is it ok to instead assert that they were called initially on focus, but then confirm there are no ADDITIONAL calls after that?

lib/src/components/NxCombobox/__tests__/NxCombobox.test.tsx
```diff
@@ -405,17 +405,15 @@ describe('NxCombobox', function() {
...
     retryBtn.focus();
-    expect(onSearch).not.toHaveBeenCalled();
+    expect(onSearch).toHaveBeenCalledTimes(1);
 
     inputElement.focus();
-    expect(onSearch).not.toHaveBeenCalled();
+    expect(onSearch).toHaveBeenCalledTimes(1);
   });
 
   it('does not call onSearch if focus moves into the component from an outside window while there is an error',
```
entroform, line 419: Should the name of this test be updated?

entroform (approved):
LGTM!

PR #967, opened 2023-02-14
RTL Conversion - NxStatefulGlobalSidebar - RSC-1491

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1491


EmilyBermudez88:
I think we should also include a test that confirms the sidebar opens and closes when the toggle button is clicked

avnvbhatta:
> I think we should also include a test that confirms the sidebar opens and closes when the toggle button is clicked

Added it here: b132d4b7d6da8183bdc6179cd730a03fe68b2439

EmilyBermudez88 (approved):
Looks good!

PR #969, opened 2023-02-14
Remove Enzyme As A Dependency - RSC-1424

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1424


lib/src/__testutils__/rtlUtils.tsx
```diff
@@ -7,7 +7,6 @@
 import React, { ComponentType } from 'react';
 import { act, render, RenderResult } from '@testing-library/react';
 
-import '@testing-library/jest-dom';
```
rpokorny, line 10: This isn't still necessary?

rpokorny: > This isn't still necessary?

Nvm I see you added it in `setupTests`

gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.10.21",
```
rpokorny, line 3: Version bumps not expected; this is being merged to the feature branch

rpokorny (approved):
+1 aside from the version bumps

gallery/package.json
avnvbhatta: > Version bumps not expected; this is being merged to the feature branch

It was only to trigger CI. Reverted here: c078212807fd8ba661742406d1282efbed3d16d5

PR #973, opened 2023-02-15
NxDescriptionList does not use additive className - RSC-1456

entroform:
https://issues.sonatype.org/browse/RSC-1456

### DIFF
--- a/lib/src/components/NxDescriptionList/NxDescriptionList.tsx
+++ b/lib/src/components/NxDescriptionList/NxDescriptionList.tsx
@@ -6,6 +6,7 @@
  */
 import React, { forwardRef, useRef } from 'react';
 import useMergedRef from '@react-hook/merged-ref';
+import classnames from 'classnames';
 
 import withClass from '../../util/withClass';
 import useEmptyComponent from '../../util/useEmptyComponent';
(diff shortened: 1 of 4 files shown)


lib/src/components/NxDescriptionList/NxDescriptionList.tsx
```diff
@@ -17,24 +18,26 @@ import { Props, propTypes } from './types';
...
+        <ul className={classNames} ref={ref} { ...otherProps }>
+          <li ref={emptyChildRef} className="nx-list__item nx-list__item--empty">
+            <span className="nx-list__text">
+              {emptyMessage || 'This list is empty.'}
+            </span>
+          </li>
+        </ul>
+      );
+
+      return isEmpty ? emptyList : <dl ref={ref} className="nx-list nx-list--description-list" { ...otherProps } />;
```
EmilyBermudez88, line 39: the `className` prop needs to be applied here too (right now it's just targeting the empty lists!)

entroform: > the `className` prop needs to be applied here too (right now it's just targeting the empty lists!)

Nice catch! fixed here: 4bd43eeea473b69d4efce966d25eab8756a8876a

EmilyBermudez88 (approved):
Looks good!

PR #976, opened 2023-02-21
 Dark Mode: NxGlobalHeader - RSC-1023

entroform:
https://issues.sonatype.org/browse/RSC-1023


JenniKobylski (approved):
Lovely! Awesome work.

lib/src/base-styles/_nx-global-header.scss
```diff
@@ -7,16 +7,24 @@
...
+  background-color: var(--nx-color-global-header-background);
+  border-bottom: 1px solid;
+  border-color: var(--nx-color-global-header-border);
   display: flex;
   grid-area: header; // see nx-page's grid-template
   height: 42px;
   padding: 0 var(--nx-spacing-6x);
 
+  @include nx-dark-mode-helpers.dark-mode {
+    background-color: var(--nx-color-global-header-background);
```
EmilyBermudez88, line 23: you should be able to get rid of the background-color & border-color here. _nx-colors.scss will take care of it!

lib/src/base-styles/_nx-colors.scss
```diff
@@ -59,6 +59,10 @@
   --nx-color-subsection-border: var(--nx-color-border);
   --nx-color-text-dark: var(--nx-color-text-stark);
 
+  // NxGlobalHeader
+  --nx-color-global-header-background: var(--nx-swatch-white);
+  --nx-color-global-header-border: var(--nx-color-border);
```
EmilyBermudez88, line 64: Quick question- in these scenarios, is it better to use the basic swatch color, or a variable name like `--nx-color-border`? I had this thought in the tooltip ticket but kept it as a swatch color. Just to know for the future!

lib/src/base-styles/_nx-colors.scss
```diff
@@ -76,5 +80,9 @@
     //NxBinaryDonutChart variables
     --nx-color-binary-donut-chart-background-fill: var(--nx-swatch-indigo-30);
     --nx-color-binary-donut-chart-arc-stroke: var(--nx-swatch-pink-80);
+
+    // NxGlobalHeader
+    --nx-color-global-header-background: var(--nx-swatch-indigo-20);
```
rpokorny, line 85: If I'm not mistaken, the header background color is always the same as the tile background color (`--nx-color-component-background`) so I don't think this variable is necessary.

lib/src/base-styles/_nx-colors.scss
```diff
@@ -76,5 +80,9 @@
...
+    --nx-color-global-header-border: var(--nx-swatch-indigo-15);
```
rpokorny, line 86: From the way the notes on the mockup are worded it seems to me like it is not a coincidence that the header border matches the page background in dark mode.  As such, I think `--nx-color-site-background` should be used here instead of using the swatch directly.

lib/src/base-styles/_nx-global-header.scss
entroform: > you should be able to get rid of the `background-color` & `border-color` here.` _nx-colors.scss` will take care of it!

Ops forgot to remove that! Thanks!
removed it here: 3d943e5eb19da2a899947b4772f6da631aad7d80

lib/src/base-styles/_nx-colors.scss
entroform: > From the way the notes on the mockup are worded it seems to me like it is not a coincidence that the header border matches the page background in dark mode. As such, I think `--nx-color-site-background` should be used here instead of using the swatch directly.

Ah I see, I fixed that here: ef41aa01e135d07b51bbcd85a6470c8590920ab8

lib/src/base-styles/_nx-colors.scss
entroform: > If I'm not mistaken, the header background color is always the same as the tile background color (`--nx-color-component-background`) so I don't think this variable is necessary.

Removed it here: ef41aa01e135d07b51bbcd85a6470c8590920ab8

lib/src/base-styles/_nx-colors.scss
entroform: > Quick question- in these scenarios, is it better to use the basic swatch color, or a variable name like `--nx-color-border`? I had this thought in the tooltip ticket but kept it as a swatch color. Just to know for the future!

Yeah, it's tricky. My thought was that maybe we wanted to stick with its semantic associations(variables) as much as possible.
And in this case, it was already associated with `--nx-color-border` before, so I kept it.
I think we use swatches if there is no new or previous semantic or general associations

EmilyBermudez88: thanks!

lib/src/base-styles/_nx-global-header.scss
EmilyBermudez88: > > you should be able to get rid of the `background-color` & `border-color` here.` _nx-colors.scss` will take care of it!
> 
> Ops forgot to remove that! Thanks! removed it here: [3d943e5](https://github.com/sonatype/sonatype-react-shared-components/commit/3d943e5eb19da2a899947b4772f6da631aad7d80)

You should be able to remove the `border-color` as well from the dark-mode helper

EmilyBermudez88 (approved):
LGTM minus the one comment I left

lib/src/base-styles/_nx-global-header.scss
entroform: > > > you should be able to get rid of the `background-color` & `border-color` here.` _nx-colors.scss` will take care of it!
> > 
> > 
> > Ops forgot to remove that! Thanks! removed it here: [3d943e5](https://github.com/sonatype/sonatype-react-shared-components/commit/3d943e5eb19da2a899947b4772f6da631aad7d80)
> 
> You should be able to remove the `border-color` as well from the dark-mode helper

Ops, here it is: 64be5391553bf7eb61523602922953160dba908b

PR #980, opened 2023-02-23
Update NxPageHeader Color Variables to Match Kaleido- RSC-1496

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1496


entroform (approved):
LGTM!

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -17,7 +17,7 @@
   box-sizing: content-box;
 
   background-color: var(--nx-swatch-white);
-  border-bottom: 1px solid var(--nx-swatch-grey-90);
+  border-bottom: 1px solid var(--nx-swatch-indigo-90);
```
entroform, line 20: Just wondering if this should use `--nx-color-border`

lib/src/base-styles/_nx-page-header.scss
```diff
@@ -63,7 +63,7 @@
   }
 
   &:active, &.nx-page-header__link--current {
-    border-bottom-color: var(--nx-swatch-teal-60);
+    border-bottom-color: var(--nx-swatch-blue-40);
```
entroform, line 66: And this `--nx-color-link`, double check with the designers.
Semantically, it sounds like it should match with `--nx-color-link` 🤔

lib/src/base-styles/_nx-page-header.scss
EmilyBermudez88: > Just wondering if this should use `--nx-color-border`

Thanks for catching that! [ad970fa](https://github.com/sonatype/sonatype-react-shared-components/pull/980/commits/ad970faf3b34c45366b877f1c9378b4370cfabcb)

lib/src/base-styles/_nx-page-header.scss
EmilyBermudez88: > And this `--nx-color-link`, double check with the designers. Semantically, it sounds like it should match with `--nx-color-link` 🤔

I did confirm with Jason, and changed here: [ad970fa](https://github.com/sonatype/sonatype-react-shared-components/pull/980/commits/ad970faf3b34c45366b877f1c9378b4370cfabcb)

PR #983, opened 2023-02-24
NxButton & NxSegmentedButton Interaction States- RSC-1461

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1461

### DIFF
--- a/lib/src/base-styles/_nx-btn.scss
+++ b/lib/src/base-styles/_nx-btn.scss
@@ -29,9 +29,8 @@
   }
 
   &:focus:not(:active) {
-    border-color: var(--nx-color-interactive-border-focus);
-    box-shadow: var(--nx-box-shadow-focus);
-    outline: 0;
+    outline-offset: -5px;
+    outline: 2px solid var(--nx-swatch-blue-30);
   }
 
   &:active {
@@ -54,8 +53,13 @@
   border-color: transparent;
   color: var(--nx-swatch-white);
 
-  &:hover, &:focus {
+  &:hover {
     background-color: var(--nx-swatch-blue-30);
+  }
+
+  &:focus:not(:active) {
+    outline-offset: -5px;
+    outline: 2px solid var(--nx-swatch-white);
   }
 
   &:active {
@@ -69,19 +73,17 @@
   color: var(--nx-color-text-dark);
 
   &:hover {
-    background-color: transparent;
-    border-color: var(--nx-color-interactive-border-hover);
+    background-color: var(--nx-swatch-indigo-90);
   }
 
-  &:focus {
-    background-color: transparent;
-    border-color: var(--nx-color-interactive-border-focus);
+  &:focus:not(:active) {
+    outline-offset: -5px;
+    outline: 2px solid var(--nx-swatch-grey-40);
   }
 
   &:active {
     background-color: var(--nx-swatch-indigo-97);
-    border-color: var(--nx-color-interactive-border-hover);
-    box-shadow: none;
+    border-color: var(--nx-swatch-grey-60);
   }
 }
 
@@ -90,18 +92,17 @@
   border-color: transparent;
   color: var(--nx-swatch-white);
 
-  &:focus {
-    background-color: var(--nx-swatch-red-35);
+  &:hover {
+    background-color: var(--nx-swatch-red-30);
   }
 
-  &:hover {
-    background-color: var(--nx-swatch-red-35);
+  &:focus:not(:active) {
+    outline-offset: -5px;
+    outline: 2px solid var(--nx-swatch-white);
   }
 
   &:active {
     background-color: var(--nx-swatch-pink-40);
-    border-color: transparent;
-    box-shadow: none;
   }
 
   .nx-icon {
@@ -141,7 +142,6 @@
     &, &:hover, &:focus {
       background-color: transparent;
       border-color: transparent;
-      box-shadow: none;
     }
   }
 }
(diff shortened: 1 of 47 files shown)


lib/src/base-styles/_nx-colors.scss
```diff
@@ -17,8 +17,8 @@
 
   --nx-color-interactive-background-selected: var(--nx-swatch-blue-90);
   --nx-color-interactive-background-active: var(--nx-swatch-blue-95);
-  --nx-color-interactive-background-hover: var(--nx-swatch-indigo-95);
-  --nx-color-interactive-background-disabled: var(--nx-swatch-grey-95);
+  --nx-color-interactive-background-hover: var(--nx-swatch-grey-90);
```
EmilyBermudez88, line 20: tables, lists and dropdowns are the only components that use this variable, and all have `--nx-swatch-grey-90` as their background hover colour according to updated Kaleido mockups, so updated here

lib/src/base-styles/_nx-btn.scss
```diff
@@ -90,18 +92,17 @@
...
-  &:focus {
-    background-color: var(--nx-swatch-red-35);
+  &:hover {
+    background-color: var(--nx-swatch-red-30);
   }
 
-  &:hover {
-    background-color: var(--nx-swatch-red-35);
+  &:focus:not(:active) {
+    outline-offset: -5px;
```
avnvbhatta, line 100: Looks like we're using `-5px` in multiple places. We could probably use a local variable.

EmilyBermudez88: > Looks like we're using `outline-offset: -5px` in multiple places. We could probably use a local variable.

Makes sense! What are your thoughts about maybe creating a @mixin as well to hold both the outline-offset & outline property to clean things up?

avnvbhatta: > > Looks like we're using `outline-offset: -5px` in multiple places. We could probably use a local variable.
> 
> Makes sense! What are your thoughts about maybe creating a @mixin as well to hold both the outline-offset & outline property to clean things up?

My only hesitation with a mixin is the outline color. If the outline colors were to be consistent across all components, that would work well, but we have different outline colors for different components.

rpokorny: > My only hesitation with a mixin is the outline color. If the outline colors were to be consistent across all components, that would work well, but we have different outline colors for different components.

Mixins can have parameters and those parameters can have default values

rpokorny: > Mixins can have parameters and those parameters can have default values

Also consider `@extend` with a [placeholder selector](https://sass-lang.com/documentation/style-rules/placeholder-selectors)

lib/src/base-styles/_nx-btn.scss
```diff
@@ -24,14 +24,14 @@
...
     background-color: var(--nx-swatch-blue-90);
+    color: var(--nx-swatch-blue-30);
   }
 
   &:focus:not(:active) {
-    border-color: var(--nx-color-interactive-border-focus);
-    box-shadow: var(--nx-box-shadow-focus);
-    outline: 0;
+    outline-offset: -5px;
+    outline: 2px solid var(--nx-swatch-blue-30);
```
EmilyBermudez88, line 34: @avnvbhatta I followed up with Jason today and they said to stick with `nx-swatch-blue-30` here - `nx-swatch-blue-40` didn't pass accessibility contrast when also in a hover state, which is why the outline was made slightly darker. They also caught that the text is supposed to darken as well on hover, so updated here: [a583477](https://github.com/sonatype/sonatype-react-shared-components/pull/983/commits/a58347772ba59f1dbbc711146a5f1702328ac96f)

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > Also consider `@extend` with a [placeholder selector](https://sass-lang.com/documentation/style-rules/placeholder-selectors)

Thanks for pointing me towards that, it seems like a better idea to use that to keep our CSS output a bit cleaner. Implemented here: [6da3385](https://github.com/sonatype/sonatype-react-shared-components/pull/983/commits/6da33851020820429f2aee544dccf6256896ebf3)

rpokorny: > Thanks for pointing me towards that, it seems like a better idea to use that to keep our CSS output a bit cleaner. Implemented here: [6da3385](https://github.com/sonatype/sonatype-react-shared-components/pull/983/commits/6da33851020820429f2aee544dccf6256896ebf3)

I'd actually keep the placeholder class inside of the `_nx-btn.scss` file.  I've seen the scss compiler get weird with duplicated output when a placeholder class is imported and extended in multiple files.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -35,6 +35,10 @@
     outline-color: var(--nx-swatch-blue-30);
   }
 
+  &:focus-visible {
```
EmilyBermudez88, line 38: When the button is focused and then clicked, the browser sets a temporary default blue outline around the button (I assume because we are using `&:focus:not(:active)`). I discussed this with Jason, and removed the outline here since we already have `:focus` and `:active` states defined for all the buttons, and should therefore pass accessibility.

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > I'd actually keep the placeholder class inside of the `_nx-btn.scss` file. I've seen the scss compiler get weird with duplicated output when a placeholder class is imported and extended in multiple files.

Sounds good! Moved here: [f100fd7](https://github.com/sonatype/sonatype-react-shared-components/pull/983/commits/f100fd732874aacc5454c65637dc62f77116be66)

avnvbhatta (approved):

PR #984, opened 2023-02-27
Fix NxToggle Border Color - RSC-1499

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1499


EmilyBermudez88:
I looked at all the states to double check, and saw that the disabled state background is off slightly as well. Sketch has it as `--nx-swatch-grey-90` and the gallery has it as `--nx-swatch-grey-95`

avnvbhatta:
> I looked at all the states to double check, and saw that the disabled state background is off slightly as well. Sketch has it as `--nx-swatch-grey-90` and the gallery has it as `--nx-swatch-grey-95`

Updated here: 58370f7aeece8d0c67322e730f91ac217addf661
BTW, I imagine the disabled background is now updated for every component. So I've changed it in the `_nx-colors.scss` file.

EmilyBermudez88:
> BTW, I imagine the disabled background is now updated for every component. So I've changed it in the `_nx-colors.scss` file.

I took a quick peak at buttons and text input to confirm, and they're both `nx-swatch-grey-90` so that should work!

EmilyBermudez88 (approved):

PR #988, opened 2023-02-28
Interaction States Tabs- RSC-1480

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1480

Thoughts so far: 

FOCUS STATE:
NxTab is already using border-bottom for the blue underline when active. If we just use an offset outline for the focus state though, we can’t have rounded corners since the bottom border doesn’t use a border-radius. 
Therefore, I tried adding a ::after pseudo element for the outline. Because the outline needs to maintain the same width as the tab, but 48px tall instead of 56px tall, I extended the width of the ::after and then applied outline-offset to make it work. (Not very clean right now clearly).

HOVER STATE:
For the same reason of using the border-bottom for the underline, I think we need to ALSO use a pseudo element for the hover state. The grey background with rounded corners is shorter than the overall height of NxTab. 
Therefore used another pseudo element with a smaller height. The problem I am running into here is how to get it behind `NxTab`. It seems as though the background only shows up when opacity is less than 100%, even if I try to use z-index . Odd, since the background of NxTab is transparent, and reading online it seems as though setting the z-index to -1 should do the trick. 

I feel like there should be a cleaner way to fix this. Also, I noticed that the border-radius of the :before and :after elements are different (which I assume has to do with the fact that the outline is offset )

### DIFF
--- a/lib/src/components/NxTabs/NxTabs.scss
+++ b/lib/src/components/NxTabs/NxTabs.scss
@@ -42,26 +42,58 @@
   margin: 0 var(--nx-spacing-4x) -1px 0;
   // subtract border thickness for consistent spacing
   padding: var(--nx-spacing-4x) var(--nx-spacing-6x) calc(var(--nx-spacing-4x) - 2px) var(--nx-spacing-6x);
+  position: relative;
   text-align: center;
   min-width: 80px;
   max-width: 240px;
+  background-color: transparent;
+  opacity: 99%;
 
   &.active {
     @include nx-text-helpers.semi-bold;
 
-    border-bottom-color: var(--nx-swatch-blue-40);
+    border-bottom: 2px solid var(--nx-swatch-blue-40);
     color: var(--nx-swatch-blue-40);
   }
 
-  &:hover {
+  &:hover:not(.active) {
     @include nx-text-helpers.semi-bold;
+    color: var(--nx-color-text-dark);
+
+    &:before {
+      content: '';
+      position: absolute;
+      top: 4px;
+      left: 0;
+      height: calc(100% - 8px);
+      width: 100%;
+      border-radius: 6px;
+      background-color: var(--nx-swatch-grey-97);
+      z-index: -1;
+    }
   }
 
   &:focus {
-    border-bottom-width: 1px;
-    box-shadow: var(--nx-box-shadow-focus);
-    margin-bottom: 0;
-    outline: solid 1px var(--nx-color-interactive-border-focus);
+    // border-bottom-width: 1px;
+    // box-shadow: var(--nx-box-shadow-focus);
+    // margin-bottom: 0;
+    &:after {
+      content: '';
+      position: absolute;
+      top: 0;
+      left: -4px;
+      height: 100%;
+      // can't control the outline's offset top/bottom with left/right, so need to make the width
+      // larger so outline-offset meets the element's edge
+      width: calc(100% + 8px);
+      outline: 1px solid var(--nx-swatch-blue-40);
+      outline-offset: -5px;
+      border-radius: 6px;
+    }
+  }
+
+  &:focus-visible {
+    outline: none;
   }
 
   // a hidden copy of the text, always at semi-bold weight, to enfore consistent element width
--- a/lib/src/base-styles/_nx-color-swatches.scss
+++ b/lib/src/base-styles/_nx-color-swatches.scss
@@ -38,6 +38,7 @@
   --nx-swatch-grey-85: hsl(var(--nx-swatch-grey-hs), 85%);
   --nx-swatch-grey-90: hsl(var(--nx-swatch-grey-hs), 90%);
   --nx-swatch-grey-95: hsl(var(--nx-swatch-grey-hs), 95%);
+  --nx-swatch-grey-97: hsl(var(--nx-swatch-grey-hs), 97%);
   --nx-swatch-white: hsl(var(--nx-swatch-grey-hs), 100%);
 
   // Teals


rpokorny:
> It seems as though the background only shows up when opacity is less than 100%, even if I try to use z-index 

The problem there is that you need to make the tab a stacking context.  The cleanest way to do that is to set `position: relative` on the tab (which you've done) and _also_ set a `z-index` on the tab that is any value other than `auto` - zero in this case is fine as the tab itself shouldn't stack any special way relative to its surroundings.

What happens if you don't do that is that the stacking context of some parent element (possibly the whole page) is used, and so setting `z-index: -1` on the `::before` causes it to go behind the tile.

Incidentally, setting `opacity` to a value other than 1 is another way to create a stacking context (though with other undesired visual effects) which is why you were able to discover that workaround.  But using the cleaner strategy that I described above will avoid messing up the colors and so is preferred.

See https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

rpokorny:
> Because the outline needs to maintain the same width as the tab, but 48px tall instead of 56px tall, I extended the width of the ::after and then applied outline-offset to make it work. (Not very clean right now clearly).

That sounds janky and as you mentioned has an undesired effect on the border radius.  Why not just set the `width` and `height` of the focus box to exactly the size you want it to visually be, and avoid `outline-offset`?

rpokorny:
The outline and focus boxes have the same size, so it might be simpler to implement them on the same element.  That is, you've already got the sizing right on the `::before`, so why not use that same element for the focus as well rather than the goofy stuff with the outline-offset?

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -42,26 +42,58 @@
...
+    border-bottom: 2px solid var(--nx-swatch-blue-40);
     color: var(--nx-swatch-blue-40);
   }
 
-  &:hover {
+  &:hover:not(.active) {
     @include nx-text-helpers.semi-bold;
+    color: var(--nx-color-text-dark);
+
+    &:before {
```
rpokorny, line 63: Using single-colon syntax to reference pseudo-elements is deprecated, from a time before the distinction between pseudo-elements and pseudo-classes was invented.  The double-colon syntax is the more correct way.

EmilyBermudez88:
> The problem there is that you need to make the tab a stacking context. The cleanest way to do that is to set `position: relative` on the tab (which you've done) and _also_ set a `z-index` on the tab that is any value other than `auto` - zero in this case is fine as the tab itself shouldn't stack any special way relative to its surroundings.

Thanks for pointing this out! I had tried setting the tab to z-index:0 at some point, but I bet I had forgotten to put position:relative on it at the time, which is why I thought it didn't work. Thanks for clearing that up!

EmilyBermudez88:
>Why not just set the width and height of the focus box to exactly the size you want it to visually be, and avoid outline-offset? There's also not really any need to use outline in this case (except maybe consistency with other components); this one could be a border. 

>The outline and focus boxes have the same size, so it might be simpler to implement them on the same element. That is, you've already got the sizing right on the ::before, so why not use that same element for the focus as well rather than the goofy stuff with the outline-offset?

Thanks for the suggestions, it has definitely cleaned things up. Implemented both the background change and the border on one `::before` element[0de0bcb](https://github.com/sonatype/sonatype-react-shared-components/pull/988/commits/0de0bcb689fb7f38114eedef82ba3ae2fa8a5605)

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -42,26 +41,49 @@
...
   min-width: 80px;
   max-width: 240px;
+  background-color: transparent;
+  z-index: 0;
+
+  &::before {
+    box-sizing: border-box;
+    border-radius: 6px;
+    content: '';
+    height: calc(100% - 8px);
```
avnvbhatta, line 55: The height of the focus "box" is `48px` on Kaleido. It's `46px` here.

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -42,26 +41,49 @@
...
+    top: 4px;
+    width: 100%;
+    z-index: -1;
+  }
 
   &.active {
     @include nx-text-helpers.semi-bold;
 
-    border-bottom-color: var(--nx-swatch-blue-40);
+    border-bottom: 2px solid var(--nx-swatch-blue-40);
```
avnvbhatta, line 66: We can use `--nx-color-interactive-border-focus` here, as well as for the `color` property.

lib/src/components/NxTabs/NxTabs.scss
```diff
@@ -42,26 +41,49 @@
...
+    }
   }
 
   &:focus {
-    border-bottom-width: 1px;
-    box-shadow: var(--nx-box-shadow-focus);
-    margin-bottom: 0;
-    outline: solid 1px var(--nx-color-interactive-border-focus);
+    &::before {
+      border: 1px solid var(--nx-swatch-blue-40);
```
avnvbhatta, line 81: `--nx-color-interactive-border-focus` here as well.

lib/src/components/NxTabs/NxTabs.scss
EmilyBermudez88: > The height of the focus "box" is `48px` on Kaleido. It's `46px` here.

Thank you for catching that! Changed here: [6135f18](https://github.com/sonatype/sonatype-react-shared-components/pull/988/commits/6135f18a24ea74dcb7b377ae3f7876fbc9a35bdd)

lib/src/components/NxTabs/NxTabs.scss
EmilyBermudez88: > We can use `--nx-color-interactive-border-focus` here, as well as for the `color` property.

Updated for both: [6135f18](https://github.com/sonatype/sonatype-react-shared-components/pull/988/commits/6135f18a24ea74dcb7b377ae3f7876fbc9a35bdd)

lib/src/components/NxTabs/NxTabs.scss
EmilyBermudez88: > `--nx-color-interactive-border-focus` here as well.

[6135f18](https://github.com/sonatype/sonatype-react-shared-components/pull/988/commits/6135f18a24ea74dcb7b377ae3f7876fbc9a35bdd)

lib/src/base-styles/_nx-colors.scss
```diff
@@ -16,9 +16,9 @@
   --nx-color-form-element-border-pristine: var(--nx-swatch-indigo-90);
 
   --nx-color-interactive-background-selected: var(--nx-swatch-blue-90);
-  --nx-color-interactive-background-active: var(--nx-swatch-blue-95);
-  --nx-color-interactive-background-hover: var(--nx-swatch-indigo-95);
-  --nx-color-interactive-background-disabled: var(--nx-swatch-grey-95);
+  --nx-color-interactive-background-active: var(--nx-swatch-grey-95);
+  --nx-color-interactive-background-hover: var(--nx-swatch-grey-90);
+  --nx-color-interactive-background-disabled: var(--nx-swatch-grey-90);
```
avnvbhatta, line 21: I don't think you'll require these changes once you get latest changes from the feature branch.

EmilyBermudez88: > I don't think you'll require these changes once you get latest changes from the feature branch.

Yep, I merged the feature branch before I saw this comment, and looking at the diffs its cleared these changes.

PR #989, opened 2023-03-01
Interaction States: NxTextLink - RSC-1483

entroform:
https://issues.sonatype.org/browse/RSC-1483


lib/src/base-styles/_nx-text-link.scss
```diff
@@ -37,8 +37,8 @@
     // was changed to 4px to get it visually close to
     // var(--nx-border-radius) or (6px).
     border-radius: 4px;
-    outline: 1px solid var(--nx-color-interactive-border-focus);
-    outline-offset: 1px;
+    outline: 2px solid var(--nx-color-interactive-border-focus);
+    outline-offset: 2px;
```
EmilyBermudez88, line 41: Looking at Sketch, it seems as though the focus border is tighter around the text (and if you focus on the example of NxCode inside of a NxTextLink, it leaves a small white space between the code's background colour and the focus border)

EmilyBermudez88:
The hover state text colour is getting overridden by the focus state colour. Looks like it was doing this before, but in Sketch now the hover state is darker when hovered & focused

EmilyBermudez88:
I may be wrong, but should the focus border disappear on click? The 'click state' shows no border in Sketch, but is 'click' referring to active here?

entroform:
> I may be wrong, but should the focus border disappear on click? The 'click state' shows no border in Sketch, but is 'click' referring to active here?

Yeah this one is tricky, because clicking it adds focus state, so I guess the mockup could be interpreted different ways here. I'll double check with @JasonDSwearingen when we do the design approval

entroform:
> The hover state text colour is getting overridden by the focus state colour. Looks like it was doing this before, but in Sketch now the hover state is darker when hovered & focused

Thanks for catching this!: 4f6cb5ff74e359fd96a014b54665106990f09d84

lib/src/base-styles/_nx-text-link.scss
entroform: > Looking at Sketch, it seems as though the focus border is tighter around the text (and if you focus on the example of NxCode inside of a NxTextLink, it leaves a small white space between the code's background colour and the focus border). So we may not need the 2px of `outline-offset`

Yup! I removed it here: 0ee5deb101caa7d5c4dac776f57c1b10c1a8998e

rpokorny:
> Yeah this one is tricky, because clicking it adds focus state, so I guess the mockup could be interpreted different ways here. I'll double check with @JasonDSwearingen when we do the design approval

We've established in recent-ish design review meetings that the Click state in the mockups refers to both `:active` and `:active:focus`

entroform:
> > Yeah this one is tricky, because clicking it adds focus state, so I guess the mockup could be interpreted different ways here. I'll double check with @JasonDSwearingen when we do the design approval
> 
> We've established in recent-ish design review meetings that the Click state in the mockups refers to both `:active` and `:active:focus`

Thanks for clarifying that: Updated 14f13a707175c92ff73cbf1eb0f159aec0e23312

JasonDSwearingen:
+1

avnvbhatta (approved):
LTGM!

PR #990, opened 2023-03-01
Interaction States Icon Only Btn - RSC-1472

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1472

I updated the snapshots for NxDrawer here as well, since it was waiting on  `nx-btn--secondary` & `nx-btn--icon-only`. NxCloseBtn does not have its own set of interaction states, but borrows them from `nx-btn--icon-only`, so those snapshots updated here as well.


lib/src/base-styles/_nx-btn.scss
```diff
@@ -135,9 +135,22 @@
...
 
 .nx-btn--icon-only {
-  @extend .nx-btn--tertiary;
-
+  background-color: transparent;
   border-color: transparent;
+  color: var(--nx-color-text-dark);
+
+  &:hover {
+    background-color: var(--nx-swatch-grey-90);
```
avnvbhatta, line 143: There's a variable for this: `--nx-color-interactive-background-hover`

lib/src/base-styles/_nx-btn.scss
```diff
@@ -135,9 +135,22 @@
...
+  }
+
+  &:focus:not(:active) {
+    outline: 2px solid var(--nx-color-interactive-border-focus);
+    outline-offset: -2px;
+  }
+
+  &:active {
+    background-color: var(--nx-swatch-grey-95);
```
avnvbhatta, line 152: Need to confirm with @JasonDSwearingen , but it looks like Lists also have the same active background color of `grey-95`. In that case we'll need to update the `--nx-color-interactive-background-active` variable to be `grey-95` instead of `blue-95`.

lib/src/components/NxIconDropdown/NxIconDropdown.scss
```diff
@@ -21,7 +21,6 @@
   margin-left: 0;
 
   &.open {
-    background-color: var(--nx-swatch-grey-95);
-    border-color: var(--nx-swatch-grey-05);
+    background-color: var(--nx-swatch-blue-90);
```
avnvbhatta, line 24: Also a variable already defined for it: `--nx-color-interactive-background-selected`

EmilyBermudez88: > Also a variable already defined for it: `--nx-color-interactive-background-selected`

thanks for catching that! [cfef41f](https://github.com/sonatype/sonatype-react-shared-components/pull/990/commits/cfef41f493519717ac4aad50260c5ef22b4942fb)

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > There's a variable for this: `--nx-color-interactive-background-hover`

Excellent, much better: [cfef41f](https://github.com/sonatype/sonatype-react-shared-components/pull/990/commits/cfef41f493519717ac4aad50260c5ef22b4942fb)

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > Need to confirm with @JasonDSwearingen , but it looks like Lists also have the same active background color of `grey-95`. In that case we'll need to update the `--nx-color-interactive-background-active` variable to be `grey-95` instead of `blue-95`.

Updated the variable here after you confirmed with Jason: [10665ef](https://github.com/sonatype/sonatype-react-shared-components/pull/990/commits/10665ef8ef0855f0098a5a91672d2c362ad86bac)

avnvbhatta (approved):

PR #992, opened 2023-03-03
Light Mode Interaction States - NxTable - RSC-1479

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1479


entroform (approved):
Looks good to me!

PR #996, opened 2023-03-06
Light Mode Interaction States - NxFormSelect - RSC-1470

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1470

### DIFF
--- a/gallery/visualtests/__image_snapshots__/nx-form-select-js-nx-form-select-disabled-looks-disabled-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-form-select-js-nx-form-select-disabled-looks-disabled-1.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:9af33dd92a0ffbccdb3970f914593f906accdca587623333e5e841a7298c2844
-size 5845
+oid sha256:4129187cf61760fa8721badfe1fb398a455440cacd1f7a0321b384df991ca9cc
+size 5819
--- a/gallery/visualtests/__image_snapshots__/nx-form-select-js-nx-form-select-simple-has-a-blue-border-when-focused-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-form-select-js-nx-form-select-simple-has-a-blue-border-when-focused-1.png
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:abeeff9b326ad9197454722d687d85592aed24c2d503ed15894c5e9de87d6192
-size 7083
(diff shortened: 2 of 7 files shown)


entroform:
In the mockup it shows that the active state does not have focus styling,
I just learned recently that it was intended that way even for `:active:focus`

Screenshot 2023-03-06 at 3 18 06 PM

It is covered by the select menu so it's not really a big deal but just curious if it's worth the effort to adjust to match the spec.

rpokorny:
> It is covered by the select menu so it's not really a big deal

That's only true on mac

avnvbhatta:
> > It is covered by the select menu so it's not really a big deal
> 
> That's only true on mac

The `active` state on the `select` behaves quite strangely. With this selector,
```
      &:active:focus {
        outline: none;
      }
```
on Chrome, forcing the `active` and `focus` doesn't render the inner outline, but clicking on the select still renders it. Now sure why.

entroform:
> > > It is covered by the select menu so it's not really a big deal
> > 
> > 
> > That's only true on mac
> 
> The `active` state on the `select` behaves quite strangely. With this selector,
> 
> ```
>       &:active:focus {
>         outline: none;
>       }
> ```
> 
> on Chrome, forcing the `active` and `focus` doesn't render the inner outline, but clicking on the select still renders it. Not sure why.

Hmm yeah, that's so weird. Playing around with it, it looks like active state is not even registered.
not sure if there's anything we can do about it?

rpokorny:
> on Chrome, forcing the active and focus doesn't render the inner outline, but clicking on the select still renders it. Not sure why.

Are you saying it behaves as expected IRL and only behaves strangely in the dev tools? I wouldn't worry too much about that; bugs in the dev tools aren't unheard of and aren't really a concern for real users.

avnvbhatta:
> > on Chrome, forcing the active and focus doesn't render the inner outline, but clicking on the select still renders it. Not sure why.
> 
> Are you saying it behaves as expected IRL and only behaves strangely in the dev tools? I wouldn't worry too much about that; bugs in the dev tools aren't unheard of and aren't really a concern for real users.

I think it behaves strangely IRL. Clicking on the select doesn't register an active state. So the `&:active:focus` doesn't select anything IRL. Only when the element is forced to active and focused through dev tools does the `&:active:focus` selector works.

avnvbhatta:
@JasonDSwearingen suggested that as long as it's a visual thing and it doesn't affect screenreaders, the current implementation is fine.

entroform:
> @JasonDSwearingen suggested that as long as it's a visual thing and it doesn't affect screenreaders, the current implementation is fine.

I was wondering if we should still remove the focus border in css, for other OS/browsers?
Or if it's fine like this should the mockup then be updated to match?

entroform (approved):
🚀

PR #997, opened 2023-03-07
Light Mode Interaction States - NxFilterInput - RSC-1469

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1469


EmilyBermudez88:
All looks good but I think Sketch may have a mistake. It says the close button should be 2px less vertically (as you've done), but the icon in Sketch is still 32px x 32px

EmilyBermudez88 (approved):

PR #998, opened 2023-03-07
Interaction states color picker - RSC-1464

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1464


lib/src/components/NxColorPicker/NxColorPicker.scss
```diff
@@ -43,14 +39,15 @@
   }
 
   &__color {
-    stroke: none;
     fill: var(--nx-selectable-color-dark);
+    stroke: var(--nx-selectable-color-dark);
+    stroke-width: 1px;
   }
-
+
```
entroform, line 46:
```suggestion
 
```

entroform:
It looks like the color circle diameter grew by one pixel: `25px` instead of `24px` in the mockup

EmilyBermudez88:
> It looks like the color circle diameter grew by one pixel: `25px` instead of `24px` in the mockup

Ok, that is what I suspected but was having trouble telling. My thought would be to shrink the radius of both `<circle>` elements to compensate for the stroke being added now. What do you think?

entroform:
> > It looks like the color circle diameter grew by one pixel: `25px` instead of `24px` in the mockup
> 
> Ok, that is what I suspected but was having trouble telling. My thought would be to shrink the radius of both `<circle>` elements to compensate for the stroke being added now. What do you think?

Hmm yeah, it looks like that's the way (changing the radius)!

EmilyBermudez88:
> > > It looks like the color circle diameter grew by one pixel: `25px` instead of `24px` in the mockup
> > 
> > 
> > Ok, that is what I suspected but was having trouble telling. My thought would be to shrink the radius of both `<circle>` elements to compensate for the stroke being added now. What do you think?
> 
> Hmm yeah, it looks like that's the way (changing the radius)!

Fixed here, thanks for catching it: [6eab0ba](https://github.com/sonatype/sonatype-react-shared-components/pull/998/commits/6eab0ba8c29b00cb534f8f37a46fe76ffc3422b1)

lib/src/components/NxColorPicker/NxColorPicker.scss
EmilyBermudez88: fixed in this commit: [6eab0ba](https://github.com/sonatype/sonatype-react-shared-components/pull/998/commits/6eab0ba8c29b00cb534f8f37a46fe76ffc3422b1)

entroform:
> > > > It looks like the color circle diameter grew by one pixel: `25px` instead of `24px` in the mockup
> > > 
> > > 
> > > Ok, that is what I suspected but was having trouble telling. My thought would be to shrink the radius of both `<circle>` elements to compensate for the stroke being added now. What do you think?
> > 
> > 
> > Hmm yeah, it looks like that's the way (changing the radius)!
> 
> Fixed here, thanks for catching it: [6eab0ba](https://github.com/sonatype/sonatype-react-shared-components/pull/998/commits/6eab0ba8c29b00cb534f8f37a46fe76ffc3422b1)

Thanks Emily, that looks good! just need to update the snapshots again i think

entroform (approved):
LGTM! Just need to update the snapshots 🚀

EmilyBermudez88:
> LGTM! Just need to update the snapshots 🚀

Thanks- I could have sworn I pushed them up but clearly didn't! [5d50588](https://github.com/sonatype/sonatype-react-shared-components/pull/998/commits/5d50588e6c79c83e996e39b544c4e1f35720e851)

PR #1001, opened 2023-03-08
Light Mode Interaction States - NxCombobox - RSC-1465

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1465


EmilyBermudez88:
I just noticed (related to this ticket, but I missed it in the dropdown ticket), that on hover, the items in NxDropdownMenu are supposed to have the colour `--nx-color-text-dark`. Putting a ticket in now to fix that, which will fix this.

EmilyBermudez88 (commented):
on line 245 for the visual tests, 'it has a light blue border and light blue background when clicked' should now be 'it has a light grey background when clicked'

avnvbhatta:
> on line 245 for the visual tests, 'it has a light blue border and light blue background when clicked' should now be 'it has a light grey background when clicked'

Updated here: 3540c120fca773189285f0a825a38e443500d25e

EmilyBermudez88:
One final thing that I'm unsure about- in the dropdown for combobox, if an item is 'selected' with the arrow keys, does the text need to darken as well? This wasn't something needed for NxDropdownMenu, but it seems as though that behaviour exists in NxList, so not sure if it's relevant here (since NxDropdown is supposed to copy relevant styles from NxList)

avnvbhatta:
> 



> One final thing that I'm unsure about- in the dropdown for combobox, if an item is 'selected' with the arrow keys, does the text need to darken as well? This wasn't something needed for NxDropdownMenu, but it seems as though that behaviour exists in NxList, so not sure if it's relevant here (since NxDropdown is supposed to copy relevant styles from NxList)

I would consult with Jason about it. There isn't exactly a "selected" state for dropdowns, and the combobox behaves a bit differently in terms of how we interact with the menu items. But regardless, it seems like a good idea to add it. I would still confirm with Jason, and if he agrees, then Kaleido needs to be updated as well. Currently, there aren't mocks for combobox menu items.

avnvbhatta:
> One final thing that I'm unsure about- in the dropdown for combobox, if an item is 'selected' with the arrow keys, does the text need to darken as well? This wasn't something needed for NxDropdownMenu, but it seems as though that behaviour exists in NxList, so not sure if it's relevant here (since NxDropdown is supposed to copy relevant styles from NxList)

I talked to @JasonDSwearingen and made the change here: 25b2ba448ea291561aa49c227fc3c98577c00903

EmilyBermudez88 (approved):
LGTM!

PR #1005, opened 2023-03-09
fix NxDropdownMenu btn on hover - RSC 1515

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1515

The text within the dropdown toggle was failing consistently in one snapshot, so updated that here as well as part of fixing the mistake with NxDropdown.

### DIFF
--- a/gallery/visualtests/__image_snapshots__/nx-dropdown-js-nx-dropdown-default-nx-dropdown-when-open-has-a-dark-grey-border-when-not-focused-with-expanded-menu-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-dropdown-js-nx-dropdown-default-nx-dropdown-when-open-has-a-dark-grey-border-when-not-focused-with-expanded-menu-1.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ba127798056e9b86d6d75963c75d5245126bc646a25993462db61a93d63fa9ed
-size 13591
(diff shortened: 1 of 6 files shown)


avnvbhatta:
Since we've modified the hover state, I was expecting some of the hover based snapshot changes, but I don't see them.
Also the current changed snapshot only has the "Filter" placeholder text changed?

EmilyBermudez88:
> Since we've modified the hover state, I was expecting some of the hover based snapshot changes, but I don't see them. Actually, looking at the test, it looks like we don't have a hover test when the dropdown is open. Perhaps we should add a hover test.

There weren't any tests that deal with the interaction state of the dropdown menu, but that's a good point. I'll add them in.


> Also the current changed snapshot only has the "Filter" placeholder text changed?

Yep that was it. And it was the only snapshot that had this happen (all other text colours of the placeholder passed). I tested it 5 or 6 times and consistently that one failed. It must have had a glitch when I initially ran the tests the first time.

EmilyBermudez88:
> > Since we've modified the hover state, I was expecting some of the hover based snapshot changes, but I don't see them. Actually, looking at the test, it looks like we don't have a hover test when the dropdown is open. Perhaps we should add a hover test.
> 
> There weren't any tests that deal with the interaction state of the dropdown menu, but that's a good point. I'll add them in.
> 

Added here! [5e7f9cf](https://github.com/sonatype/sonatype-react-shared-components/pull/1005/commits/5e7f9cfa211ad602da2329e595cf0b35283c2130)

avnvbhatta (approved):
LGTM

PR #1008, opened 2023-03-09
Interaction States - NxBreadcrumb - RSC-1460

entroform:
https://issues.sonatype.org/browse/RSC-1460

The mockups have non integer values on the button but that's being fixed.


avnvbhatta:
For the snapshots, besides the ones with the icon dropdown, are those expected differences? I double checked the height and colors, but look like they are the same on prod as well.

entroform:
> For the snapshots, besides the ones with the icon dropdown, are those expected differences? I double checked the height and colors, but look like they are the same on prod as well.

Yeah some of them have no changes, I deleted the tests in bulk and re run it. But I'll revert the ones with no differences!

avnvbhatta (approved):

PR #1009, opened 2023-03-10
Interaction States - NxTag - RSC-1481

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1481

### DIFF
--- a/gallery/visualtests/nxTag.js
+++ b/gallery/visualtests/nxTag.js
@@ -6,6 +6,8 @@
  */
 const { setupBrowser } = require('./testUtils');
 
+const OUTSET = 6;
+
 describe('NxTag', function() {
   const {
     focusTest,
@@ -13,8 +15,9 @@ describe('NxTag', function() {
     hoverTest,
     simpleTest,
     waitAndGetElements,
-    checkScreenshotCoordinates,
     checkScreenshot,
+    checkScreenshotCoordinates,
+    moveMouseAway,
     scrollIntoView,
     wait,
     a11yTest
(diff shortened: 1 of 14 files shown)


gallery/visualtests/nxTag.js
```diff
@@ -46,17 +49,22 @@ describe('NxTag', function() {
...
+    it('has a blue outline and dark grey border when focused and hovered',
+        focusAndHoverTest(selector, undefined, OUTSET));
 
-    it('has a dark blue/grey background when clicked', async function() {
-      const [targetElement] = await waitAndGetElements(selector);
+    it('has a darker background and no border when selected', async function() {
+      const [targetElement, lastElement] = await waitAndGetElements(selector, lastTagSelector);
 
       await targetElement.click();
+      //remove focus from tag for snapshot
```
EmilyBermudez88, line 65: I couldn't figure out how to blur the element here - I tried blurring NxTag and the icon inside, but both left the focus outline in the snapshot. Any other ideas to remove the border are appreciated!

rpokorny: > I couldn't figure out how to blur the element here - I tried blurring NxTag and the icon inside, but both left the focus outline in the snapshot. Any other ideas to remove the border are appreciated!

NxSelectableTag is built around a checkbox input.  That's what you'd need to blur

EmilyBermudez88: > > I couldn't figure out how to blur the element here - I tried blurring NxTag and the icon inside, but both left the focus outline in the snapshot. Any other ideas to remove the border are appreciated!
> 
> NxSelectableTag is built around a checkbox input. That's what you'd need to blur

Ah of course, that makes sense. Thanks for the suggestion/feedback!

entroform:
It looks good, but I did notice that we have the same issue in Safari:
I wonder if we should go back to using the position absolute solution 🤔 

Screenshot 2023-03-13 at 4 30 51 PM

EmilyBermudez88:
>It looks good, but I did notice that we have the same issue in Safari:
I wonder if we should go back to using the position absolute solution 🤔

Makes sense to me! Back to `::before` here: [3be9e74](https://github.com/sonatype/sonatype-react-shared-components/pull/1009/commits/3be9e74fee37f20ef4b2d8eb5e39d2455122e95b)

entroform (approved):
🚀

PR #1016, opened 2023-03-15
Fix NxFileUpload focus border on safari - RSC-1536

entroform:
https://issues.sonatype.org/browse/RSC-1536
https://issues.sonatype.org/browse/RSC-1468

It looks like this approach with box-shadow works and looks consistently well across different browsers.
I wonder if we should change our other implementation to use this instead of outline?

### DIFF
--- a/lib/src/components/NxFileUpload/NxFileUpload.scss
+++ b/lib/src/components/NxFileUpload/NxFileUpload.scss
@@ -98,20 +98,18 @@
     border-style: inherit;
     border-width: inherit;
     border-radius: inherit;
+    border: none;
     color: var(--nx-swatch-indigo-30);
     cursor: pointer;
     display: inline-flex;
     font-size: var(--nx-font-size);
     height: inherit;
     margin-left: auto;
-    margin-right: -1px; // outline to overlap with parent border
+    margin-right: -1px; // "outline" to overlap with parent border
     padding: 0 var(--nx-spacing-3x);
 
-    outline: 1px solid transparent;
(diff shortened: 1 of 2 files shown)


EmilyBermudez88:
I've been playing around, and wondering what your thoughts would be using `border` instead of box-shadow? In order to keep the same size of the dismiss button when focused, we could just decrease the padding on the L & R.

EmilyBermudez88:
If we keep the box-shadow solution, what are your thoughts on including NxPolicyThreatSlider in this fix as well? (as mentioned in your slack thread [here](https://sonatype.slack.com/archives/CJUF9G7T2/p1678893187164679))

entroform:
> I've been playing around, and wondering what your thoughts would be using `border` instead of box-shadow? In order to keep the same size of the dismiss button when focused, and decrease the padding on the L & R (although I'm not sure if there was a specific reason to avoid borders in the first place)

Thanks Emily!
Yeah, I was also thinking about moving back to using border,
Although, if it turns out that this is a lot simpler than decreasing the paddings,
I don't think it hurts sticking with box shadow solution for this one

avnvbhatta (approved):

lib/src/components/NxFileUpload/NxFileUpload.scss
```diff
@@ -98,20 +98,18 @@
     border-style: inherit;
```
EmilyBermudez88, line 98: If there's no border, I THINK we can get rid of `border-style` & 'border-width`

PR #1019, opened 2023-03-16
Add <caption> to NxTable - RSC-1266

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1266


gallery/src/components/NxTable/NxTablePage.tsx
```diff
@@ -74,6 +74,14 @@ export default function NxTablePage() {
               </NxTable.Row>
             </NxTable.Head>
             <NxTable.Body>
+              <NxTable.Row>
+                <NxTable.Cell>caption</NxTable.Cell>
```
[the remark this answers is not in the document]
avnvbhatta: > Why a prop and not something like `<NxTable.Caption>`?

I was thinking from a _label_ perspective. Similar to how the `NxFormGroup` and `NxFieldset` uses the `label` prop. I personally also just prefer the prop. Just set it on the table and no need to worry about placement and what not. 

Regardless, I'm open to suggestions.

lib/src/components/NxTable/__tests__/NxTable.test.tsx
```diff
@@ -46,6 +46,23 @@ describe('NxTable', function() {
...
+    });
+
+    it('assigns the accessible-name of the table when caption prop is provided', function() {
+      render(<NxTable caption="table-title"></NxTable>);
+      expect(screen.getByRole('table', { name: 'table-title'})).toBeInTheDocument();
+    });
+
+    it('does not render a <caption> when caption prop is not provided', function() {
+      render(<NxTable></NxTable>);
+      expect(document.querySelectorAll('caption').length).toBe(0);
```
EmilyBermudez88, line 62: To match how we use querySelector in other RTL tests, we could do:
```
const {container} = quickRender();
expect(container.querySelector(‘caption’)).not.toBeInTheDocument();
```

avnvbhatta: > To match how we use querySelector in other RTL tests, we could do:
> 
> ```
> const {container} = quickRender();
> expect(container.querySelector(‘caption’)).not.toBeInTheDocument();
> ```

Changed here: 175e2b42d7514aa7d377717ea61d9ab493a8ed02

PR #1022, opened 2023-03-17
Dark Mode - NxCounter - RSC-1015

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1015


lib/src/base-styles/_nx-colors.scss
```diff
@@ -80,6 +80,10 @@
   --nx-color-code-border: var(--nx-swatch-pink-95);
   --nx-color-code-text: var(--nx-swatch-grey-20);
 
+  // NxCounter
+  --nx-color-counter: var(--nx-swatch-blue-10);
```
EmilyBermudez88, line 84: So this was part of RSC-1511, but in Kaleido the light mode background colour of NxCounter is grey-05 instead of blue-10. Maybe confirm with Jason on this one about which colour we should use here?

EmilyBermudez88 (approved):
LGTM!

PR #1023, opened 2023-03-17
Update final interaction states snapshots - RSC-1534

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1534

Confirming all snapshots are updated & old ones are removed before the final merge.


entroform:
Oo I believe there are some colour variables that are also no longer in use like `--nx-color-interactive-shadow-focus`
I think we should deprecate them here before we merge

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -205,15 +205,21 @@ const CssVariablesPage = () => {
...
+            </NxWarningAlert>
           </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-box-shadow-dropdown">
             The standard box-shadow to apply to dropdown elements.
           </PropertyDocItem>
           <PropertyDocItem propertyVar="--nx-drop-shadow-focus">
-            Again, the standard shadow to apply to focused elements in addition to a border or outline. On elements
-            where <NxCode>box-shadow</NxCode> isn't supported, such as SVG elements, use this instead
-            as the value of a <NxCode>filter</NxCode> <NxCode>drop-shadow</NxCode>.
+            <NxWarningAlert>
```
entroform, line 217: This is just a small thing but I wonder  what you think about just having the deprecate message inside the alert and the rest outside.

EmilyBermudez88: > This is just a small thing but I wonder what you think about just having the deprecate message inside the alert and the rest outside.

I agree with you. However, the other deprecated examples had both the 'deprecated' reasoning & definition inside the alert. I'm happy to change all of them for consistency sake - what do you think?

entroform: Ah okay, I guess we can keep it the same with other ones for consistency sake 👍

avnvbhatta (approved):
🚀

PR #1029, opened 2023-03-22
Fix browser hotkeys being prevented when navigating NxTabs and NxTree - RSC-1507

entroform:
https://issues.sonatype.org/browse/RSC-1507


EmilyBermudez88:
I'm not sure why this happened from the changes made, but it seems that when you are focused on a tab and press `tab`, if there is no focusable element inside of it then it skips focusing the tabpanel (but if you then `shift` + `tab` to go back, it DOES focus on the tabpanel). Before it always shifted focus to the tabpanel, no matter whether the content had a focusable element inside of it or not.

rpokorny:
> I'm not sure why this happened from the changes made, but it seems that when you are focused on a tab and press tab, if there is no focusable element inside of it then it skips focusing the tabpanel (but if you then shift + tab to go back, it DOES focus on the tabpanel). Before it always shifted focus to the tabpanel, no matter whether the content had a focusable element inside of it or not.

We focus the next element that we want focused but then when the tab key event finishes bubbling the browser processes it - by moving the focus _again_.  This is presumably why we were calling `preventDefault` there in the first place

lib/src/components/NxTree/NxTreeItem.tsx
```diff
@@ -136,12 +131,12 @@ export default function NxTreeItem(props: ItemProps) {
   function onKeyDown(evt: KeyboardEvent<HTMLLIElement>) {
     switch (evt.key) {
       case 'ArrowUp':
-        stopPropAndPreventDefault(evt);
```
rpokorny, line 139: Won't this result in the page scrolling when the user presses the Up arrow here? (and similarly for the other directions)

rpokorny:
I think the approach in this PR is risky; we were preventing default in each of these places for a reason.  A safer approach IMO would be to check for modifier keys and to ignore the event if any are pressed (except in any case where we are deliberately handling any modifier keys, but I'm not sure we do any of that).

entroform:
> I think the approach in this PR is risky; we were preventing default in each of these places for a reason. A safer approach IMO would be to check for modifier keys and to ignore the event if any are pressed (except in any case where we are deliberately handling any modifier keys, but I'm not sure we do any of that).

Thanks Ross, Yeah I wasn't sure about this approach.
I added a check for modifier keys here (ignoring caps lock), let me know what you  think!
7a9ce6dfa1adbdc9a84f397ef043174a621f25a5

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -78,7 +86,7 @@ const NxTab = function NxTabElement(props: NxTabProps) {
     const nextElement = getNextElementFromEventKey(event.key);
 
     if (nextElement) {
-      event.preventDefault();
+      preventDefaultWhenModifierKeyIsNotPressed(event.nativeEvent);
       (nextElement as HTMLElement).focus();
```
rpokorny, line 90: This will still move the focus on e.g. Shift+ArrowRight.  I'd think we wouldn't want that.  I know for instance it drives me crazy that Sketch reacts to Ctrl+Alt+Right (by changing to an entirely different mockup) because I use that key combination to switch workspaces.  IMO our navigations should only be based on _modifierless_ key events.

lib/src/components/NxTabs/NxTab.tsx
```diff
@@ -46,6 +47,13 @@ const NxTab = function NxTabElement(props: NxTabProps) {
     }
   }
 
+  function preventDefaultWhenModifierKeyIsNotPressed(evt: KeyboardEvent) {
```
rpokorny, line 50: This functionality should probably be moved to a utility.  In this PR we have two places that need basically the same logic, and there are or will be other places in RSC that will as well as we implement more advanced keynav.

entroform: > This functionality should probably be moved to a utility. In this PR we have two places that need basically the same logic, and there are or will be other places in RSC that will as well as we implement more advanced keynav.

Thanks, I moved it to it's own file here: 56973dc6a829f285e56a1ce4295a8c7d3221040c

lib/src/components/NxTabs/NxTab.tsx
entroform: > This will still move the focus on e.g. Shift+ArrowRight. I'd think we wouldn't want that. I know for instance it drives me crazy that Sketch reacts to Ctrl+Alt+Right (by changing to an entirely different mockup) because I use that key combination to switch workspaces. IMO our navigations should only be based on _modifierless_ key events.

Changed navigations based on modifierless key events here: 56973dc6a829f285e56a1ce4295a8c7d3221040c

lib/src/util/keyboardUtil.ts
```diff
@@ -0,0 +1,13 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+const MODIFIER_KEYS = ['Alt', 'AltGraph', 'Control', 'Meta', 'OS', 'Shift'];
+
+export function modifierKeyIsPressed(event: React.KeyboardEvent | KeyboardEvent) {
+  return MODIFIER_KEYS.some((key) => event.getModifierState(key));
+}
```
rpokorny, line 13: Let's unit test this

entroform: > Let's unit test this

Thanks, I added unit test here: 9442f123c76db9a1ed20f5afe4cf53905c205866

lib/src/util/__tests__/keyboardUtil.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+import React, { useState } from 'react';
+import { rtlRender } from '../../__testutils__/rtlUtils';
+import { userEvent } from '../../__testutils__/rtlUtils';
+import { modifierKeyIsPressed } from '../keyboardUtil';
+
+describe('modifierKeyIsPressed', function () {
+  const Fixture = () => {
+    const [isPressed, setIsPressed] = useState(false);
+    const handleKeyDown = (event: React.KeyboardEvent) => {
+      event.preventDefault();
```
rpokorny, line 16: Why do we need to `preventDefault` here?

lib/src/util/__tests__/keyboardUtil.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    expect(targetEl).toHaveTextContent('false');
+    targetEl.focus();
+
+    await user.keyboard('{ArrowRight}');
+    expect(targetEl).toHaveTextContent('false');
+
+    await user.keyboard('{Control}');
+    expect(targetEl).toHaveTextContent('true');
+
+    await user.keyboard('{ArrowUp}');
```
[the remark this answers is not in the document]
entroform: > There's a way to press like Shift+ArrowUp, right? We should test that. I'd also add tests for all of the other modifier keys.

Oh yes, added that to the test here: 9dd9c177b067db05b807b1004197afc197fb8091

lib/src/util/__tests__/keyboardUtil.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  it('should only return true when a modifier key is pressed', async () => {
+    const user = userEvent.setup();
+    const { getByTestId } = quickRender();
+
+    const targetEl = getByTestId('target');
+    expect(targetEl).toHaveTextContent('false');
+    targetEl.focus();
+
+    // "OS" is not playing nicely with js-dom
```
rpokorny, line 34: What were the issues?  That's one of the more realistic keys so it'd be a shame to not be able to test it

rpokorny: > That's one of the more realistic keys

Well, maybe.  MDN claims that means the Windows key, but at least on my Linux system that's not the case.  Maybe it is on Windows though?

lib/src/util/__tests__/keyboardUtil.test.tsx
rpokorny: > Oh that's so userEvent.keyboard doesn't lose focus on the fixture.

What's causing it to lose focus without this?  If your test was pressing Tab I'd assume it's that, but that's not what you're doing.

entroform: > > Oh that's so userEvent.keyboard doesn't lose focus on the fixture.
> 
> What's causing it to lose focus without this? If your test was pressing Tab I'd assume it's that, but that's not what you're doing.

I'm not entirely sure but `userEvent.keyboard` gives me this error:
`Cannot read properties of null (reading 'nodeType')` (Sounded like it lost focus)
The error occur when attempting {Modifier+ArrowLeft} after {Modifier+ArrowUp} is pressed.

rpokorny: > I'm not entirely sure but userEvent.keyboard gives me this error:
Cannot read properties of null (reading 'nodeType') (Sounded like it lost focus)
The error occur when attempting {Modifier+ArrowLeft} after {Modifier+ArrowUp} is pressed.

Upgrading to the latest `@testing-library/user-event` seems to fix that

lib/src/util/__tests__/keyboardUtil.test.tsx
entroform: Hmm yeah it is returning false on the test 😿 
It looks like Mac uses the Meta key, but it's strange that it is still returning false when
artificially setting the type to "OS".

I'm curious, what's the difference between "Super" or "Hyper" key vs "Meta" key in Linux?

rpokorny: > I'm curious, what's the difference between "Super" or "Hyper" key vs "Meta" key in Linux?

Long story short, it completely depends on your configuration.  From the sound of it the Windows key is typically mapped to Meta in the KDE desktop environment (what I use) but not in GNOME or others which are also common (where it is mapped to Super).   See https://askubuntu.com/a/19565 .

tl;dr
> Furthermore, the definition of modifier keys depends on:
> * The keyboard itself
> * The keyboard layout
> * The driver
> * The operating system's implementation

entroform: > > I'm curious, what's the difference between "Super" or "Hyper" key vs "Meta" key in Linux?
> 
> Long story short, it completely depends on your configuration. From the sound of it the Windows key is typically mapped to Meta in the KDE desktop environment (what I use) but not in GNOME or others which are also common (where it is mapped to Super). See https://askubuntu.com/a/19565 .
> 
> tl;dr
> 
> > Furthermore, the definition of modifier keys depends on:
> > 
> > * The keyboard itself
> > * The keyboard layout
> > * The driver
> > * The operating system's implementation

Thanks!

entroform: > As explained in that post, the key names come from a certain ridiculous keyboard from years past: image

I would love to have that keyboard just because 😆

rpokorny: > I would love to have that keyboard just because

Same

lib/src/util/__tests__/keyboardUtil.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    for (const modifier of without(['OS'], MODIFIER_KEYS)) {
```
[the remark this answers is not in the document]
rpokorny: > Hardcoded it here! https://github.com/sonatype/sonatype-react-shared-components/commit/4c56a8da6340b2d0febabbf12908ebe1ab9d06ac

The comment about why "OS" was missing was still valuable.  Though it'd be worth checking that again now that we've upgraded `user-event`.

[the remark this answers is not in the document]
entroform: > > Hardcoded it here! [4c56a8d](https://github.com/sonatype/sonatype-react-shared-components/commit/4c56a8da6340b2d0febabbf12908ebe1ab9d06ac)
> 
> The comment about why "OS" was missing was still valuable. Though it'd be worth checking that again now that we've upgraded `user-event`.

💯 
I tested it after upgrading and sadly, still got the same error 😭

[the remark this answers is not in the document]
rpokorny: > I tested it after upgrading and sadly, still got the same error sob

I think you should put the comment back though.

rpokorny (approved):
+1 aside from https://github.com/sonatype/sonatype-react-shared-components/pull/1029#discussion_r1176705627

lib/src/util/__tests__/keyboardUtil.test.tsx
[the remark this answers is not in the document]
entroform: > > I tested it after upgrading and sadly, still got the same error sob
> 
> I think you should put the comment back though.

Thanks, added comment back here: 96afd8098c56068ab8326b4d440eb96937d0c917

PR #1031, opened 2023-03-22
Fix NxCounter Height & Colour - RSC-1511

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1511


entroform:
Hmm the uneven padding makes this weird, but yeah It looks like that's how it was set up in Sketch

entroform (approved):

PR #1035, opened 2023-03-24
policy violation indicator dark mode - RSC-1035

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1035


lib/src/components/NxPolicyViolationIndicator/NxPolicyViolationIndicator.scss
```diff
@@ -27,14 +28,15 @@
   display: inline-flex;
 
   .nx-icon {
-    color: var(--nx-swatch-white);
+    color: var(--nx-color-policy-indicator-icon-background);
```
entroform, line 31: I think it might be more consistent/accurate if we rename this `--nx-color-policy-indicator-icon`, removing the `-background`. Background to is more for like "container" or "layer" background , and this feels like it's more related to "foreground" element so we can just go with icon.

entroform (approved):
LGTM!

PR #1037, opened 2023-03-24
Fixed stacked modals focus cycling - RSC-1542

entroform:
https://issues.sonatype.org/browse/RSC-1542


lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -183,10 +189,29 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
...
+      if (dialogEl && document.activeElement) {
+        const focusIsInsideDialogModal = document.activeElement.closest('dialog[aria-modal="true"][open]');
+        if (!focusIsInsideDialogModal) {
+          const firstFocusableElement = getFirstVisibleFocusableElement(dialogEl);
+          if (firstFocusableElement) {
+            firstFocusableElement.focus();
+          }
+        }
+      }
+    };
```
avnvbhatta, line 204: Is there a scenario when the focus goes out of the dialog modal? I'm not entirely sure if this is required?

entroform: > Is there a scenario when the focus goes out of the dialog modal? I'm not entirely sure if this is required?

There is, for example, when focused on the submit button on a form and submits, the submit button is removed and the focus moves out to body(or null) which moves it out of the modal

lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -148,6 +148,12 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       const dialogEl = dialogRef.current!;
 
+      // Don't manage tab cycling when the focus is not within the dialog
+      // so that there is no key nav conflict when another modal is open.
+      if (!dialogEl.contains(document.activeElement)) {
```
rpokorny, line 153: Does this negate the need for the `preventDefault` on line 168.  Or alternatively, would it have been better to adjust around that line rather than add this condition here?

lib/src/components/AbstractDialog/AbstractDialog.tsx
rpokorny: > There is, for example, when focused on the submit button on a form and submits, but there is a validation error, the submit button is removed and the focus moves out to body(or null) which moves it out of the modal

I suspect it would be cleaner to implement this with `focusout`, using `relatedTarget` to detect whether the focus is leaving the modal or just moving around within it

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -290,4 +291,59 @@ describe('NxModal', function() {
...
+
+    // should cycle tabbable elements in modal 2 properly
+    const modal2Tabbables = getAllByTestId('tabbable-modal-2');
+    expect(modal2Tabbables[0]).toHaveFocus();
+    await user.tab();
+    expect(modal2Tabbables[1]).toHaveFocus();
+    await user.tab();
+    expect(modal2Tabbables[0]).toHaveFocus();
+  });
 });
```
rpokorny, line 349: Need a test for the scenario described in https://github.com/sonatype/sonatype-react-shared-components/pull/1037/files#r1154650216

lib/src/components/AbstractDialog/AbstractDialog.tsx
entroform: > > There is, for example, when focused on the submit button on a form and submits, but there is a validation error, the submit button is removed and the focus moves out to body(or null) which moves it out of the modal
> 
> I suspect it would be cleaner to implement this with `focusout`, using `relatedTarget` to detect whether the focus is leaving the modal or just moving around within it

Oo yeah that would be cleaner: a777480372daee2c407a4f1464e25901c40d741d

lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -148,23 +148,30 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
...
-        const activeFocusIsNullOrDialog
-          = [document.body, dialogEl, null].includes(document.activeElement as HTMLElement);
-        const activeFocusIsInDialog = !activeFocusIsNullOrDialog && dialogEl.contains(document.activeElement);
+        const focusIsDialogElement = document.activeElement === dialogEl;
 
-        if (!activeFocusIsInDialog) {
+        // Prevent native tab cycling
+        // when focus is the dialog element, so it does not leak
+        // (In the case where there is no focusable elements inside the dialog)
+        if (focusIsDialogElement) {
```
rpokorny, line 168: To me it would be slightly cleaner to think about if instead of this special case, we adjusted `getFirstVisibleFocusableElement` and `getLastVisibleFocusableElement` to return the `dialogEl` in the case that there are no other focusable elements.  That matches the behavioral model more closely: if nothing else in the modal is focusable, then the modal itself is focusable.  And then I think the correct behavior would occur automatically from the conditionals after this one (which themselves could be further cleaned up a little) and this conditional could be removed.

entroform: > To me it would be slightly cleaner to think about if instead of this special case, we adjusted `getFirstVisibleFocusableElement` and `getLastVisibleFocusableElement` to return the `dialogEl` in the case that there are no other focusable elements. That matches the behavioral model more closely: if nothing else in the modal is focusable, then the modal itself is focusable. And then I think the correct behavior would occur automatically from the conditionals after this one (which themselves could be further cleaned up a little) and this conditional could be removed.

Thanks! That simplified things a lot and definitely cleaner: c9979e1e4fe14938a2b279e719635ef644f15bdf

entroform: Also removed early exit here: 73be804f6399907868d8c567432252bad4f88b71

lib/src/components/NxModal/__tests__/NxModal.test.tsx
entroform: > Need a test for the scenario described in https://github.com/sonatype/sonatype-react-shared-components/pull/1037/files#r1154650216

Added that test and simplified AbstractDialog further here: 5846ccae98a610cc71ff31f4265d6e958d143b65

entroform: Looks like I'll have to write a function test to test the form one (display: none) doesn't behave nicely in jest-dom

entroform:
> It occurs to me that the gallery example here has the two modals as siblings in the DOM. It's also possible to render an `NxModal` from within another `NxModal`. We should construct an example that ensures that that works correctly too, both in regards to the issue in this PR and just in general.

Thanks!
I added a nested example (and fixed the behaviour) here: 4f531583eb8fa7ea21095fd37986d496b3999c40
^Also cleaned up the stacked example.

Going to add the functional test in another commit

gallery/visualtests/nxModal.js
```diff
@@ -142,6 +143,35 @@ describe('NxModal', function() {
     });
   });
 
+  describe('Nested NxModal', function() {
```
rpokorny, line 146: Should probably have a visual test for this example too

lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -45,10 +47,10 @@ const getFocusableElements = (element: HTMLElement) =>
   Array.from(element.querySelectorAll(FOCUSABLE_ELEMENTS_SELECTOR)) as HTMLElement[];
 
 const getFirstVisibleFocusableElement = (element: HTMLElement) =>
-  find(isVisible)(getFocusableElements(element));
+  find(isVisible)(getFocusableElements(element)) || element;
```
rpokorny, line 50: Minor
```suggestion
  find(isVisible, getFocusableElements(element)) || element;
```

lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -148,45 +147,62 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
...
-            && lastFocusableElement
-          ) {
-            lastFocusableElement.focus();
-            event.preventDefault();
+      // Only manage tab cycling when the focus is within the dialog
+      // and not nested inside another dialog
+      // so that there is no keynav conflict when another modal is open.
+      if (
+        document.activeElement
+        && document.activeElement.closest('dialog[aria-modal="true"]') === dialogEl
```
rpokorny, line 155:
```suggestion
        && document.activeElement.closest(DIALOG_MODAL_SELECTOR) === dialogEl
```

lib/src/components/NxModal/__tests__/NxModal.test.tsx
```diff
@@ -290,4 +291,157 @@ describe('NxModal', function() {
...
+    expect(buttons[1]).toHaveFocus();
+    await user.click(buttons[1]);
+
+    // Focus is removed
+    expect(document.activeElement).toBe(document.body);
+
+    await user.tab();
+    expect(buttons[0]).toHaveFocus();
+  });
 });
```
rpokorny, line 447: One more thing that I think needs testing: closing the second modal and having focus return to the first, and having tab and/or shift+tab work properly after

gallery/visualtests/nxModal.js
entroform: > Should probably have a visual test for this example too

Thanks, added that here: 5e819f7c58f6538c1e0290bed72c203480b9bcd0

lib/src/components/NxModal/__tests__/NxModal.test.tsx
entroform: > One more thing that I think needs testing: closing the second modal and having focus return to the first, and having tab and/or shift+tab work properly after

Updated tests here: 4a237738dfcecc6e04ee5b53b9a3c135b80755a5

gallery/visualtests/nxModal.js
```diff
@@ -22,12 +22,13 @@ describe('NxModal', function() {
...
       wideExampleSelector = '#nx-modal-wide-example',
       stackedExampleSelector = '#nx-modal-stacked-example',
+      nestedExampleSelector = '#nx-modal-nested-example',
       escClosingExampleSelector = '#nx-modal-esc-example';
 
   async function openModal(exampleSelector) {
     const openModalBtnSelector = `${exampleSelector} button`,
         closeModalBtnSelector =
-            `${exampleSelector} .nx-footer .nx-btn-bar .nx-btn:not(.nx-btn--primary):not(.nx-btn-tertiary)`;
+        `${exampleSelector} .nx-footer .nx-btn-bar .nx-btn:not(.nx-btn--primary):not(.nx-btn-tertiary)`;
```
rpokorny, line 31: Remaining autoformat changes

lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -45,10 +47,10 @@ const getFocusableElements = (element: HTMLElement) =>
   Array.from(element.querySelectorAll(FOCUSABLE_ELEMENTS_SELECTOR)) as HTMLElement[];
 
 const getFirstVisibleFocusableElement = (element: HTMLElement) =>
-  find(isVisible)(getFocusableElements(element));
+  find(isVisible, getFocusableElements(element)) || element;
 
 const getLastVisibleFocusableElement = (element: HTMLElement) =>
-  findLast(isVisible)(getFocusableElements(element));
+  findLast(isVisible)(getFocusableElements(element)) || element;
```
rpokorny, line 53: This one too ```suggestion
  findLast(isVisible, getFocusableElements(element)) || element;
```

File: lib/src/components/AbstractDialog/AbstractDialog.tsx [hunk 11]
```diff
@@ -148,45 +147,62 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
...
+      const dialogEl = dialogRef.current;
+      if (dialogEl) {
+        const departingFocusIsInsideDialog = !!(
+          event.target
+          && (event.target as HTMLElement).closest(DIALOG_MODAL_SELECTOR) === dialogEl
+        );
+        const receivingFocusIsInsideDialog
+          = !!(
+            event.relatedTarget
+            && (event.relatedTarget as HTMLElement).closest(DIALOG_MODAL_SELECTOR)
```
rpokorny, line 188: Isn't this supposed to be 
```suggestion
            && (event.relatedTarget as HTMLElement).closest(DIALOG_MODAL_SELECTOR) === dialogEl
```

lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -148,45 +147,62 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
...
+          );
```
rpokorny, line 189: These two statements follow a very similar pattern but are formatted fairly differently.  I prefer the formatting on the first one.

gallery/visualtests/nxModal.js
entroform: > Remaining autoformat changes

Thanks fixed: bc426e2939b3d22c6d9e91dcf0b72563b966cccc

lib/src/components/AbstractDialog/AbstractDialog.tsx
entroform: > These two statements follow a very similar pattern but are formatted fairly differently. I prefer the formatting on the first one.

Fixed: 8eb61a1a01d8388378b93f387f2ffb84dee6057b

gallery/visualtests/nxModal.js
rpokorny: > Thanks fixed: https://github.com/sonatype/sonatype-react-shared-components/commit/bc426e2939b3d22c6d9e91dcf0b72563b966cccc

There was more than just that line.  Check the overall PR diff

lib/src/components/AbstractDialog/AbstractDialog.tsx
entroform: > Isn't this supposed to be

Oh this was intentional change to fix the focusing issue in modals that are siblings.

gallery/visualtests/nxModal.js
entroform: > > Thanks fixed: [bc426e2](https://github.com/sonatype/sonatype-react-shared-components/commit/bc426e2939b3d22c6d9e91dcf0b72563b966cccc)
> 
> There was more than just that line. Check the overall PR diff

Ops, fixed: c23e1604fde46358a504a4d6a25a9bdafefb0c41

lib/src/components/AbstractDialog/AbstractDialog.tsx
rpokorny: > Oh this was intentional change to fix the focusing issue in modals that are siblings.

Could you help me understand what the issue was and why this helped?  I'd think that regardless of any sibling modals, we'd still want to trap focus within this modal, and would want `focusIsLeavingDialog` to be true even if the `relatedTarget` is in a sibling modal.

entroform: > > Oh this was intentional change to fix the focusing issue in modals that are siblings.
> 
> Could you help me understand what the issue was and why this helped? I'd think that regardless of any sibling modals, we'd still want to trap focus within this modal, and would want `focusIsLeavingDialog` to be true even if the `relatedTarget` is in a sibling modal.

Sorry I forgot to add a comment, I'll do that now:

The problem was that when you open a nested modal, this is triggered because the relatedTarget parent dialog is not the wrapping modal's dialog and `getFirstVisibleFocusableElement(dialogEl).focus();` is triggered. 

For some reason, in Chrome, this forces the focus back on the parent dialog instead of the nested dialog even though the getFirstVisibleFocusableElement on the nested dialog was triggered after. 🤔 

So this workaround only triggers when the receiving focus is not inside any modal dialog, which happens for when the focused element is hidden via CSS.

rpokorny: > Sorry I forgot to add a comment, I'll do that now:
>
> The problem was that when you open a nested modal, this is triggered because the relatedTarget parent dialog is not the wrapping modal's dialog and getFirstVisibleFocusableElement(dialogEl).focus(); is triggered.
>
> For some reason, in Chrome, (This works fine in Firefox strangely), this forces the focus back on the parent dialog instead of the nested dialog even though the getFirstVisibleFocusableElement on the nested dialog was triggered after. 
>
> So this workaround only triggers when the receiving focus is not inside any modal dialog, which happens for when the focused element is hidden via CSS.

Let me check my understanding: so when one modal is nested inside another, and a `focusout` event occurs within the inner one, I imagine it triggers this handler on both modals.  And the handler for the outer modal thinks focus is leaving, since the nearest modal is the child and not itself, and so it steals the focus.  Is that right?

I wonder if a better solution would be to use `stopPropagation` in the inner modal's handler of for this event.

entroform: > > Sorry I forgot to add a comment, I'll do that now:
> > The problem was that when you open a nested modal, this is triggered because the relatedTarget parent dialog is not the wrapping modal's dialog and getFirstVisibleFocusableElement(dialogEl).focus(); is triggered.
> > For some reason, in Chrome, (This works fine in Firefox strangely), this forces the focus back on the parent dialog instead of the nested dialog even though the getFirstVisibleFocusableElement on the nested dialog was triggered after.
> > So this workaround only triggers when the receiving focus is not inside any modal dialog, which happens for when the focused element is hidden via CSS.
> 
> Let me check my understanding: so when one modal is nested inside another, and a `focusout` event occurs within the inner one, I imagine it triggers this handler on both modals. And the handler for the outer modal thinks focus is leaving, since the nearest modal is the child and not itself, and so it steals the focus. Is that right?
> 
> I wonder if a better solution would be to use `stopPropagation` in the inner modal's handler of for this event.

Oh it's not a problem when the modals are nested inside, the problem occur when the modals are siblings.
So when the second modal opens, focusout event occurs on the first modal and calls the focus back,
It looks like looking at the dev tool that it calls it twice, both from the first modal handler, then the on mount focus gets called from the second modal but that didn't override the focusout handlers for some reason (and only on Chrome)

Going to dig into this more 🤔

entroform: After investigating this more, it looks like that is the case:
When you open a second modal that is sibling from the other modal,
focusout is triggered and due to a race condition, steals the focus from the first focus on mount (in Chrome).

The fix here works because now the focus out handler is only triggered when the focus goes out of _any_ dialog modal. (This is a scenario that happens when the focus leaves an element when it is hidden via CSS) which was the purpose for that handler anyways.

I cleaned up the code and added a comment, let me know what your thoughts on this 👍

rpokorny: > After investigating this more, it looks like that is the case:
When you open a second modal that is sibling from the other modal,
focusout is triggered and due to a race condition, steals the focus from the first focus on mount (in Chrome).
>
> The fix here works because now the focus out handler is only triggered when the focus goes out of any dialog modal. (This is a scenario that happens when the focus leaves an element when it is hidden via CSS) which was the purpose for that handler anyways.
>
> I cleaned up the code and added a comment, let me know what your thoughts on this +1

OK I think I understand now.  How about just changing the variable name to `receivingFocusIsInsideAnyDialogModal`.  "Any" is more accurate than "Another" here if I understand correctly.

rpokorny (approved):

PR #1038, opened 2023-03-24
NxButton Dark Mode - RSC-1007

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1007


avnvbhatta:
These two snapshots are incorrect and need to be checked:

- gallery/visualtests/__image_snapshots__/nx-segmented-button-js-nx-segmented-button-primary-nx-button-primary-nx-button-dropdown-section-has-a-lighter-blue-background-when-opened-and-not-focused-1.png 

- gallery/visualtests/__image_snapshots__/nx-segmented-button-js-nx-segmented-button-secondary-nx-button-secondary-nx-button-dropdown-section-has-a-light-blue-background-when-opened-and-not-focused-1.png

EmilyBermudez88:
> Looks like the focus states for the primary segmented buttons is a bit off: https://github.com/sonatype/sonatype-react-shared-components/blob/6dd08a415671584a9e787ed06f48836599409abe/gallery/visualtests/__image_snapshots_dark__/nx-segmented-button-js-nx-segmented-button-primary-nx-button-primary-nx-button-dropdown-section-has-a-white-inner-outline-when-focused-1-dark.png
> 
> https://github.com/sonatype/sonatype-react-shared-components/blob/6dd08a415671584a9e787ed06f48836599409abe/gallery/visualtests/__image_snapshots_dark__/nx-segmented-button-js-nx-segmented-button-primary-nx-button-primary-nx-button-main-section-has-a-white-inner-

Yeah this was something I noticed as well and not sure how to fix at this point. Because the border width is now 2px instead of 1px, the outline is closer to the border line than it used to be. But I don't think we can adjust the offset of an outline on just one side. I'll fiddle with this and see if I can come up with something

gallery/visualtests/__image_snapshots_dark__/nx-segmented-button-js-nx-segmented-button-tertiary-nx-button-tertiary-nx-button-dropdown-section-has-a-darker-grey-inner-outline-and-light-indigo-background-when-opened-1-dark.png
rpokorny, line 1: The mockups aren't clear about how the dropdown toggle should look when it is focused and the dropdown is expanded

EmilyBermudez88: > The mockups aren't clear about how the dropdown toggle should look when it is focused and the dropdown is expanded

I had talked to Jason about that, and they said it should still have the interior focus border when focused & expanded. I will follow up with them to make sure they update the mocks.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -6,48 +6,75 @@
...
   text-decoration: none;
 
   &:hover {
-    background-color: var(--nx-swatch-blue-90);
+    background-color: var(--nx-color-secondary-button-background-hover);
   }
 
   &.nx-btn--secondary:hover {
-    color: var(--nx-swatch-blue-30);
+    color: var(--nx-color-secondary-button-text-hover);
```
rpokorny, line 63: Why is this in a separate block and not the `&:hover` block above? Note that we do still support raw HTML buttons which use the `nx-btn` classes directly, and in that support we also do still say that a default `nx-btn` is a secondary `nx-btn`.  That is, `<button class="nx-btn">` and `<button class="nx-btn nx-btn--secondary">` should behave identically.  I suspect the only reason this isn't causing a bug is because `--nx-color-secondary-button-text-hover` is the same as `--nx-color-secondary-button-text` (what the button has without this line) anyway.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -6,48 +6,75 @@
...
 
-%button-focus-outline {
+%transparent-button-focus-outline {
+  outline-offset: calc(0px - var(--nx-border-width-button) - 4px);
+  outline: 2px solid;
+}
+
+%solid-button-focus-outline {
   outline-offset: -5px;
   outline: 2px solid;
```
rpokorny, line 18: Consider that the CSS generated by these placeholder classes ends up like this: 
image

There's actually no point in including the `outline` property in a placeholder or mixin because it gets applied to `nx-btn` itself, a class which all more specialized buttons should also have.  That style can just be right in `.nx-btn:focus:not(active)`.  Then if that is removed from these placeholders, also consider that `solid-button-focus-outline` is only used in one place, so the `outline-offset` can just be moved to that place and that placeholder class can be removed.  `transparent-button-focus-outline` still has some value in reuse, but I think a placeholder class is probably more complex than necessary for it at that point.  I'd just use a scss variable to store the calc expression.

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -52,49 +53,111 @@
...
 
 .nx-btn--primary:not([disabled]) {
-  &.nx-segmented-btn__main-btn {
-    border-right-color: var(--nx-swatch-blue-60);
+  &.nx-segmented-btn__main-btn, &.nx-segmented-btn__dropdown-btn {
+    border-color: var(--nx-color-primary-segmented-button-divider);
+    position: relative;
+
+    &::after {
+      border: 2px solid transparent;
```
rpokorny, line 61: Are all of the two's and four's in here interrelated? If so they should be based on a common variable.  Either a SCSS local variable, or if it makes sense (is this 2px the standard focus border width") as a documented CSS variable

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > Why is this in a separate block and not the `&:hover` block above? Note that we do still support raw HTML buttons which use the `nx-btn` classes directly, and in that support we also do still say that a default `nx-btn` is a secondary `nx-btn`. That is, `<button class="nx-btn">` and `<button class="nx-btn nx-btn--secondary">` should behave identically. I suspect the only reason this isn't causing a bug is because `--nx-color-secondary-button-text-hover` is the same as `--nx-color-secondary-button-text` (what the button has without this line) anyway.

This had stemmed from a problem that arose in the interaction state button ticket. For the secondary button, when hovered, the text color shifts slightly from `blue-40` to `blue-30` to pass contrast. By adding the color change in the `&:hover` block, it also affected the text color on hover of all other buttons. Would you suggest listing it inside of the `&:hover` block and then override it in the other hover states?

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
EmilyBermudez88: > Are all of the two's and four's in here interrelated? If so they should be based on a common variable. Either a SCSS local variable, or if it makes sense (is this 2px the standard focus border width") as a documented CSS variable

The 4px in the calc expression has no significance. It was a way to set the `outline-offset` to either -5px or -6px depending on the border width to keep the appearance that there was 2px of space between the border & outline. However, the 2px IS the standard focus border width, so I'll make a documented variable for that

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > There's actually no point in including the `outline` property in a placeholder or mixin because it gets applied to `nx-btn` itself, a class which all more specialized buttons should also have. That style can just be right in `.nx-btn:focus:not(active)`. Then if that is removed from these placeholders, also consider that `solid-button-focus-outline` is only used in one place, so the `outline-offset` can just be moved to that place and that placeholder class can be removed. `transparent-button-focus-outline` still has some value in reuse, but I think a placeholder class is probably more complex than necessary for it at that point. I'd just use a scss variable to store the calc expression instead of a placeholder there.

Great, that makes a lot of sense! Cleaned up here: [02ff31b](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/02ff31b5d7d3b5f482891c51ed331880c29238d5)

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
EmilyBermudez88: >>Are all of the two's and four's in here interrelated? If so they should be based on a common variable. Either a SCSS local variable, or if it makes sense (is this 2px the standard focus border width") as a documented CSS variable

>The 4px in the calc expression has no significance. It was a way to set the outline-offset to either -5px or -6px depending on the border width to keep the appearance that there was 2px of space between the border & outline. However, the 2px IS the standard focus border width, so I'll make a documented variable for that 👍

[aa61db3](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/aa61db3fbd2838c46c2c41b636b92ab76293240f)

lib/src/base-styles/_nx-btn.scss
rpokorny: > Would you suggest listing it inside of the &:hover block and then override it in the other hover states?

Yes

lib/src/base-styles/_nx-btn.scss
```diff
@@ -6,48 +6,67 @@
  */
 @use '../scss-shared/nx-container-helpers';
 @use '../scss-shared/nx-text-helpers';
+@use '../scss-shared/nx-dark-mode-helpers';
 
-%button-focus-outline {
-  outline-offset: -5px;
-  outline: 2px solid;
+$transparent-button-outline-offset: calc(0px - var(--nx-border-width-button) - 4px);
```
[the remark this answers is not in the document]
rpokorny: > To make it a local variable, inaccessible outside this file:

Actually, this cleaned up even better than I had realized it would, and this variable is only used in one place, which means it really doesn't need to be a variable at all

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
rpokorny: > [aa61db3](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/aa61db3fbd2838c46c2c41b636b92ab76293240f)

I'm guessing there are a bunch of other places beyond the scope of this ticket where that variable could/should be used.  If so we should put in a tech debt ticket to update them

lib/src/base-styles/_nx-btn.scss
rpokorny: > > Would you suggest listing it inside of the &:hover block and then override it in the other hover states?
>
> Yes

To be clear, after your explanation I can see that there is already a bug here in light mode: on raw HTML buttons with the `nx-btn` class, the hover text color is the same as the non-hover text color when it should be 10% lighter.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -63,75 +82,44 @@
...
   &:focus:not(:active) {
-    @extend %button-focus-outline;
-    outline-color: var(--nx-swatch-grey-40);
+    outline-color: var(--nx-color-tertiary-button-focus-outline);
   }
 
   &:active {
-    background-color: var(--nx-swatch-indigo-97);
-    border-color: var(--nx-swatch-grey-60);
+    background-color: var( --nx-color-tertiary-button-background-active);
```
rpokorny, line 102:
```suggestion
    background-color: var(--nx-color-tertiary-button-background-active);
```

lib/src/base-styles/_nx-colors.scss
```diff
@@ -59,6 +59,39 @@
...
+  --nx-color-error-button-background: var(--nx-color-alert-error);
+  --nx-color-error-button-background-hover: var(--nx-swatch-red-30);
+  --nx-color-error-button-background-active: var(--nx-swatch-pink-40);
+  --nx-color-error-button-focus-outline: var(--nx-swatch-white);
+  --nx-color-error-button-text: var(--nx-color-error-button-focus-outline);
+  --nx-color-disabled-button-background: var(--nx-color-interactive-background-disabled);
+  --nx-color-disabled-button-border: var(--nx-color-interactive-border-disabled);
+  --nx-color-disabled-button-text: var(--nx-color-text-disabled);
+  --nx-color-primary-segmented-button-divider: var(--nx-swatch-blue-60);
+  --nx-color-button-icon-only-background-active: var(--nx-color-interactive-background-active);
```
rpokorny, line 91: I'm not seeing this one used anywhere

gallery/visualtests/__image_snapshots_dark__/nx-segmented-button-js-nx-segmented-button-tertiary-nx-button-tertiary-nx-button-dropdown-section-has-a-darker-grey-inner-outline-and-light-indigo-background-when-opened-1-dark.png
EmilyBermudez88: > > The mockups aren't clear about how the dropdown toggle should look when it is focused and the dropdown is expanded
> 
> I had talked to Jason about that, and they said it should still have the interior focus border when focused & expanded. I will follow up with them to make sure they update the mocks.

Mocks should now be updated 👍

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -52,49 +53,111 @@
 }
 
 .nx-btn--primary:not([disabled]) {
-  &.nx-segmented-btn__main-btn {
-    border-right-color: var(--nx-swatch-blue-60);
+  &.nx-segmented-btn__main-btn, &.nx-segmented-btn__dropdown-btn {
+    border-color: var(--nx-color-primary-segmented-button-divider);
+    position: relative;
+
+    &::after {
```
rpokorny, line 62: Is there a reason this isn't entirely defined in the focus block? It has no visible effect until the `border-color` is set, correct?

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -52,49 +53,111 @@
...
+      border: var(--nx-focus-outline-width) solid transparent;
+      border-radius: 4px;
+      box-sizing: border-box;
+      content: '';
+      height: calc(100% - 4px);
+      position: absolute;
+      left: 2px;
+      top: 2px;
+      width: calc(100% - 4px);
```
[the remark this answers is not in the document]
rpokorny: > Although they're separate from the focus outline width, these 2's and 4's are still related to each other right? Maybe a scss local variable would be appropriate for them

Oh I think I see, is it to make it easier to implement the border-radius overrides below?

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
EmilyBermudez88: > I'm guessing there are a bunch of other places beyond the scope of this ticket where that variable could/should be used. If so we should put in a tech debt ticket to update them

I was thinking the same thing. Ticket to update: https://issues.sonatype.org/browse/RSC-1554

EmilyBermudez88:
> >Looks like the focus states for the primary segmented buttons is a bit off:

>So I tried using a pseudo element just for the primary segmented button which seems to work well. ([9056066](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/9056066a767c99b2bb994e33f51e5cb2025ff984)) I did notice a discrepancy in the light/dark mode primary segmented button outline though- that in light mode, the outline is offset 1px more than in dark mode, so the light mode primary segmented button snapshots will now need to be updated as well. I will have to confirm with Jason tomorrow for how for they'd like the outline offset. I'll update the snapshots once you both give me the thumbs up on this approach

So Jason confirmed that we are going to change the focus outline of the primary / error button to 2px from the button's edge instead of 3px. Code updated here: [6bd4ad1](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/6bd4ad1f8c2f45359bc965d0206b7d9d1fba5ed3).

lib/src/base-styles/_nx-btn.scss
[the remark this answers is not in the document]
EmilyBermudez88: > > To make it a local variable, inaccessible outside this file:
> 
> Actually, this cleaned up even better than I had realized it would, and this variable is only used in one place, which means it really doesn't need to be a variable at all

Removed here: [deac080](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/deac0808835df98a9e5aada5b9e11842e073737b)

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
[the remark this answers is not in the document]
EmilyBermudez88: > Although they're separate from the focus outline width, these 2's and 4's are still related to each other right? Maybe a scss local variable would be appropriate for them

Yeah this one I had been waiting on Jason's confirmation about the focus outline's spacing. So initially updated here[6bd4ad1](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/6bd4ad1f8c2f45359bc965d0206b7d9d1fba5ed3) until I saw your comment re: local variables, and then fixed here: [deac080](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/deac0808835df98a9e5aada5b9e11842e073737b)

rpokorny:
Minor mockup bug: *some* of the icons in the icon-only button mocks are using `HSL(231, 41%, 90%)` instead of the `nx-indigo-90` color variable (`nx-indigo-90` is `hsl(231, 43%, 90%)`)

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > > > Would you suggest listing it inside of the &:hover block and then override it in the other hover states?
> > 
> > 
> > Yes
> 
> To be clear, after your explanation I can see that there is already a bug here in light mode: on raw HTML buttons with the `nx-btn` class, the hover text color is the same as the non-hover text color when it should be 10% lighter.

Updated here: [db199c1](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/db199c1c73ec523021856baca2e132fef8235309)

lib/src/base-styles/_nx-colors.scss
EmilyBermudez88: > I'm not seeing this one used anywhere

Thanks for catching that! Removed it here: [28a6d48](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/28a6d48eb4720af4dfbcd9b0464828021c1505ef)

lib/src/base-styles/_nx-btn.scss
```diff
@@ -6,48 +6,67 @@
...
+  border-color: transparent;
+  color: var(--nx-color-#{$variant}-button-text);
+
+  &:hover {
+    background-color: var(--nx-color-#{$variant}-button-background-hover);
+  }
+
+  &:focus:not(:active) {
+    outline-color: var(--nx-color-#{$variant}-button-focus-outline);
+    outline-offset: -5px;
```
rpokorny: Shouldn't this be `-4px`? Also could it be a computation based on some variables? The amount by which the focus borders are supposed to be inset is always `2px` right? So maybe that should be a CSS variable as well, and this `outline-offset` could be calculated from that variable and `--nx-focus-outline-width`

lib/src/base-styles/_nx-btn.scss
```diff
@@ -6,48 +6,64 @@
...
-    color: var(--nx-swatch-blue-30);
+    background-color: var(--nx-color-secondary-button-background-hover);
+    color: var(--nx-color-secondary-button-text-hover);
   }
 
   &:focus:not(:active) {
-    @extend %button-focus-outline;
-    outline-color: var(--nx-swatch-blue-30);
+    outline: var(--nx-focus-outline-width) solid var(--nx-color-secondary-button-focus-outline);
+    outline-offset: calc(0px - var(--nx-border-width-button) - 4px);
```
rpokorny, line 57: The `4px` here would also be affected by what I said in https://github.com/sonatype/sonatype-react-shared-components/pull/1038/files#r1158637782

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
EmilyBermudez88: > Is there a reason this isn't entirely defined in the focus block? It has no visible effect until the `border-color` is set, correct?

Yeah, I'm not sure why I put that there initially, it definitely makes more sense the way you've suggested: [83cd0fa](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/83cd0fa6313451b676280186f02bc3e0b3a2aa20). (Then realized I needed to be more specific for the border-radius definitions in the individual buttons, so moved those into a individual focus blocks here [2fa7ece](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/2fa7ecec9411e9260f7166e4ec0b0ccb2148f2f7))

gallery/visualtests/__image_snapshots_dark__/nx-segmented-button-js-nx-segmented-button-tertiary-nx-button-tertiary-nx-button-dropdown-section-has-a-darker-grey-inner-outline-and-light-indigo-background-when-opened-1-dark.png
rpokorny: > Mocks should now be updated +1

Now they aren't clear about how it should look when expanded and hovered :smile:

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -6,7 +6,7 @@
  */
  @use '../../scss-shared/nx-dark-mode-helpers';
 
- $primary-btn-focus-outline-spacing: 2px;
+ $-primary-btn-focus-outline-spacing: 2px;
```
rpokorny, line 9: You can declare variables inside of scss blocks to scope them just to that block, which removes the need to make their names so detailed.

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -77,34 +98,24 @@
...
-    padding-left: var(--nx-spacing-6x);
-
-    &::after {
-      border-top-right-radius: 0;
-      border-bottom-right-radius: 0;
+    @include main-btn-styling;
+
+    &:focus:not(:active) {
+      &::after {
+        border-top-right-radius: 0;
```
rpokorny, line 105: Can't these parts be included in the mixins as well?

rpokorny: > Can't these parts be included in the mixins as well?

Though I guess there's no reason to move them there, since they don't get repeated

EmilyBermudez88: > Can't these parts be included in the mixins as well?

I debated on that, and ultimately left them out because of the fact that the mixin is used for disabled buttons in dark mode as well and seemed odd to include it there (even if it doesn't actually affect disabled buttons)

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > Shouldn't this be `-4px`? Also could it be a computation based on some variables? The amount by which the focus borders are supposed to be inset is always `2px` right? So maybe that should be a CSS variable as well, and this `outline-offset` could be calculated from that variable and `--nx-focus-outline-width`

This had been partially resolved previously ([deac080](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/deac0808835df98a9e5aada5b9e11842e073737b)) but created the css variable for the 2px inset here as well: [913d7f1](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/913d7f1ffccd46447a83a70c425b8bbd1cabc07e)

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
EmilyBermudez88: > You can declare variables inside of scss blocks to scope them just to that block, which removes the need to make their names so detailed.

With the new css variable made (--nx-focus-outline-inset), this no longer seemed necessary so removed entirely here: [913d7f1](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/913d7f1ffccd46447a83a70c425b8bbd1cabc07e)

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
```diff
@@ -88,11 +86,11 @@
         border-radius: 4px;
```
rpokorny, line 86: I bet this could/should be `calc(var(--nx-border-radius) - var(--nx-focus-outline-inset))`

gallery/visualtests/__image_snapshots_dark__/nx-segmented-button-js-nx-segmented-button-tertiary-nx-button-tertiary-nx-button-dropdown-section-has-a-darker-grey-inner-outline-and-light-indigo-background-when-opened-1-dark.png
rpokorny: > Now they aren't clear about how it should look when expanded and hovered smile

See https://sonatype.slack.com/archives/CJUF9G7T2/p1680710192650639?thread_ts=1680708320.712499&cid=CJUF9G7T2

EmilyBermudez88:
> Might be a preexisting bug but I think the segmented button dropdown toggles are an incorrect width. They're 41px wide including the divider but sketch has them at 36px. The main issue seems to be that the right padding is 4x when it should be 3x. There might be another pixel or so of adjustment needed to get it exactly the way it's supposed to be though

Yep, the 4px was an accident when I was looking into something and forgot to change it back. I updated to adjust to the border width of the button, so all segmented buttons should now be 36px. [db5f995](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/db5f995c544283c4bd5ecb5824fc8127b9154e5b)

EmilyBermudez88:
> Minor mockup bug: **some** of the icons in the icon-only button mocks are using `HSL(231, 41%, 90%)` instead of the `nx-indigo-90` color variable (`nx-indigo-90` is `hsl(231, 43%, 90%)`)

I reached out to Jason on this, and they said that it should be fixed now.

lib/src/components/NxSegmentedButton/NxSegmentedButton.scss
EmilyBermudez88: > I bet this could/should be `calc(var(--nx-border-radius) - var(--nx-focus-outline-inset))`

Fixed here [7062fd7](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/7062fd7b606ff6ee1864fdb40d44e8132ff9c94a)

rpokorny (approved):

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -426,6 +435,99 @@ const CssVariablesPage = () => {
...
+            <ColorDocRow colorVar="--nx-color-tertiary-button-border-active">
+              The border color of tertiary buttons on activation/click.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-tertiary-button-focus-outline">
+              The color of the focus outline for tertiary buttons.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-tertiary-button-text">
+              The text color of tertiary buttons.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-secondary-button-background">
```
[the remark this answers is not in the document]
EmilyBermudez88: > Duplicate

Thanks! [0a1bf61](https://github.com/sonatype/sonatype-react-shared-components/pull/1038/commits/0a1bf61ad566697c6c01286b12f0a8f5734eae82)

PR #1041, opened 2023-03-28
update NxDivider color in dark mode - RSC-1543

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1543


avnvbhatta:
Looks good, but we might also want to confirm with @JasonDSwearingen about the horizontal divider based on `--nx-border-secondary` that's used on lists and tables.

avnvbhatta (approved):

PR #1045, opened 2023-03-29
adjust placeholder text color- RSC-1514

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1514

I confirmed with Jason that in the ticket's description, the placeholder text is being changed to `nx-swatch-grey-40`, not `nx-swatch-indigo-40`


lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -116,6 +116,11 @@
       }
     }
 
+    .nx-text-input__input {
+      &::placeholder {
```
EmilyBermudez88, line 120: before the placeholder text wasn't specifically set for disabled, so now it follows the mocks (grey-60 when disabled, grey-40 when not)

PR #1046, opened 2023-03-29
Dark Mode - NxCollapsibleItems - RSC-1014

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1014


EmilyBermudez88 (commented):
`--nx-color-text-stark` for `.nx-collpasible-items__trigger` on line 107 & 170 😄

lib/src/base-styles/_nx-colors.scss
```diff
@@ -157,6 +164,11 @@
     // NxCounter
     --nx-color-counter: var(--nx-swatch-indigo-90);
     --nx-color-counter-active-text: var(--nx-swatch-indigo-05);
+
+    // NxCollapsibleItems
+    --nx-color-collapsible-items-background-active: var(--nx-swatch-indigo-30);
```
EmilyBermudez88, line 169: As per our discussion yesterday, we should be able to reuse this variable for other components like the Icon Only Btn & Pagination, so we could make this one `--nx-color-interactive-background-active'

avnvbhatta:
> `--nx-color-text-stark` for `.nx-collpasible-items__trigger` on line 107 & 170 😄

Check: 0a72640b0d9042b20277f824bf4c5c22f7db72f1

avnvbhatta:
> I think we need to merge main into the RSC-928 & here, because the close icon on the filter input is positioned incorrectly again

main has been merged. Should be fixed now.

lib/src/base-styles/_nx-colors.scss
avnvbhatta: > As per our discussion yesterday, we should be able to reuse this variable for other components like the Icon Only Btn & Pagination, so we could make this one `--nx-color-interactive-background-active'

The issue with that is that `--nx-color-interactive-background-active` has already been defined for light mode (`grey-95`), but for collapsible items, the correct color is `grey-97` =/

avnvbhatta:
> If you look at the mocks between light & dark mode, the disabled states look different. In light mode, only the trigger has the disabled colour, but in dark mode the children also have the disabled colour applied. I think we need to confirm what design is looking for there.

I reached out to Jason and they said the dark mode disabled state was incorrect. The mockup has been updated.

gallery/visualtests/__image_snapshots_dark__/nx-collapsible-items-js-nx-collapsible-items-nx-collapsible-items-with-action-content-does-not-close-the-collapsible-items-when-the-action-content-is-expanded-1-dark.png
rpokorny, line 1: It seems like this is dependent on `NxIconDropdown`'s dark mode conversion which doesn't appear to be done yet

gallery/visualtests/__image_snapshots_dark__/nx-collapsible-items-js-nx-collapsible-items-nx-collapsible-items-with-checkbox-radio-children-looks-right-expanded-1-dark.png
rpokorny, line 1: Also dependent on checkboxes

lib/src/base-styles/_nx-colors.scss
```diff
@@ -27,7 +27,7 @@
   --nx-color-text-stark: var(--nx-swatch-grey-05);
   --nx-color-text: var(--nx-swatch-grey-30);
   --nx-color-text-disabled: var(--nx-color-interactive-border-disabled); // match disabled text and border
-  --nx-color-text-placeholder: var(--nx-swatch-grey-60);
+  --nx-color-text-placeholder: var(--nx-swatch-grey-40);
```
rpokorny, line 30: This seems like it would have a much wider affect than just this PR.  Did it not cause additional visual test diffs?

avnvbhatta: > This seems like it would have a much wider affect than just this PR. Did it not cause additional visual test diffs?

We made this change based on [RSC-1514](https://issues.sonatype.org/browse/RSC-1514). Since this change is for light mode and `main` already has the newer snapshots, this shouldn't have any effect on the dark mode feature branch.

rpokorny: > We made this change based on [RSC-1514](https://issues.sonatype.org/browse/RSC-1514). Since this change is for light mode and main already has the newer snapshots, this shouldn't have any effect on the dark mode feature branch.

Ok so it's just that only the dark mode branch has this variable for it? Has that commit on main been merged into the dark mode feature branch? Theoretically this should have been done in that merge commit not an unrelated dev branch

avnvbhatta: > > We made this change based on [RSC-1514](https://issues.sonatype.org/browse/RSC-1514). Since this change is for light mode and main already has the newer snapshots, this shouldn't have any effect on the dark mode feature branch.
> 
> Ok so it's just that only the dark mode branch has this variable for it? Has that commit on main been merged into the dark mode feature branch? Theoretically this should have been done in that merge commit not an unrelated dev branch

Yes, only the dark mode feature branch has this variable. 

You're right. I think it'd be cleaner if I changed this in the merge commit. I'll try to push this change to the feature branch first then merge it back to this branch.

avnvbhatta: > > > We made this change based on [RSC-1514](https://issues.sonatype.org/browse/RSC-1514). Since this change is for light mode and main already has the newer snapshots, this shouldn't have any effect on the dark mode feature branch.
> > 
> > 
> > Ok so it's just that only the dark mode branch has this variable for it? Has that commit on main been merged into the dark mode feature branch? Theoretically this should have been done in that merge commit not an unrelated dev branch
> 
> Yes, only the dark mode feature branch has this variable.
> 
> You're right. I think it'd be cleaner if I changed this in the merge commit. I'll try to push this change to the feature branch first then merge it back to this branch.

Reverted commit here: 0966c69426e0c74e49f6d4f87533acdbd635f678

EmilyBermudez88:
> > If you look at the mocks between light & dark mode, the disabled states look different. In light mode, only the trigger has the disabled colour, but in dark mode the children also have the disabled colour applied. I think we need to confirm what design is looking for there.
> 
> I reached out to Jason and they said the dark mode disabled state was incorrect. The mockup has been updated.

Thanks for clarifying that.  Although I guess I'm not wondering about what the Collapsible Radio & Multi Select would look like disabled? 🤔  This may be intentional, but looks a bit odd to me
Screenshot 2023-04-05 at 2 38 05 PM
Screenshot 2023-04-05 at 2 38 17 PM

avnvbhatta:
> > > If you look at the mocks between light & dark mode, the disabled states look different. In light mode, only the trigger has the disabled colour, but in dark mode the children also have the disabled colour applied. I think we need to confirm what design is looking for there.
> > 
> > 
> > I reached out to Jason and they said the dark mode disabled state was incorrect. The mockup has been updated.
> 
> Thanks for clarifying that. Although I guess I'm not wondering about what the Collapsible Radio & Multi Select would look like disabled? 🤔 This may be intentional, but looks a bit odd to me

We'll need to wait on the radio/checkbox PR to be merged for this. The disabled styles are inherited from radio/checkbox. Ideally, these should look like this when the PR is merged:
image

gallery/visualtests/__image_snapshots_dark__/nx-collapsible-items-js-nx-collapsible-items-nx-collapsible-items-with-action-content-does-not-close-the-collapsible-items-when-the-action-content-is-expanded-1-dark.png
avnvbhatta: > It seems like this is dependent on `NxIconDropdown`'s dark mode conversion which doesn't appear to be done yet

Updated snapshots here: c5ca437396b9a1e7379f3ba56d8ca0ebfcaa47a2

gallery/visualtests/__image_snapshots_dark__/nx-collapsible-items-js-nx-collapsible-items-nx-collapsible-items-with-checkbox-radio-children-looks-right-expanded-1-dark.png
avnvbhatta: > Also dependent on radios

Updated snapshots here: c5ca437396b9a1e7379f3ba56d8ca0ebfcaa47a2

lib/src/base-styles/_nx-colors.scss
```diff
@@ -203,6 +207,8 @@
 
     --nx-color-site-background: var(--nx-swatch-indigo-15);
 
+    --nx-color-interactive-background-hover: var(--nx-swatch-indigo-10);
+    --nx-color-interactive-background-selected: var(--nx-swatch-purple-30);
```
rpokorny, line 211: This group of properties is already present below, and the hover one is duplicated there

avnvbhatta: > This group of properties is already present below, and the hover one is duplicated there

Good catch: 2e28079a218dc08e6a4518bab2982d42c047d29b

EmilyBermudez88:
It doesn't show in the mocks, but is there an active state for clickable items? If not, we need to override this behaviour from NxButton

Screenshot 2023-04-17 at 11 50 20 AM

avnvbhatta:
> Looking at the disabled example vs the mock, the children are different. I believe in past conversations it was confirmed that it was a mock mistake (that the children should be the regular indigo-90 instead of the disabled text color?)

You might be looking at a "scratch" file on Sketch as mentioned [here](https://sonatype.slack.com/archives/CJUF9G7T2/p1681758014062559). The correct URL for the mockup is [here](https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/DEoQwJW), and I believe the current implementation is correct.

avnvbhatta:
> It doesn't show in the mocks, but is there an active state for clickable items? If not, we need to override this behaviour from NxButton
> 

Updated here: https://github.com/sonatype/sonatype-react-shared-components/pull/1046/commits/84137b6b74414b13fcd145663959f7a308418346

EmilyBermudez88 (approved):
LGTM!

PR #1047, opened 2023-03-31
Dark Mode: NxSearchDropdown - RSC-1039

entroform:
https://issues.sonatype.org/browse/RSC-1039


EmilyBermudez88:
It may be worth updating the hover state of the close button, since the icon only button has a different hover state colour than the filter input

EmilyBermudez88:
There are a few discrepancies in the mocks that may need to be fixed and/or confirmed with design:

gallery/visualtests/__image_snapshots_dark__/nx-search-dropdown-js-nx-search-dropdown-looks-right-with-the-long-variant-in-error-1-dark.png
rpokorny, line 1: This text color is clearly incorrect.  Have alerts been converted yet? If not I'd think we need to pause this ticket until they are

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -62,6 +62,10 @@
     &::placeholder {
       color: var(--nx-color-text-placeholder);
     }
+
+    &[disabled] {
+      color: var(--nx-color-text-disabled);
```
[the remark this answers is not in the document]
EmilyBermudez88: > Was this a bug with text inputs more generally, and not specifically `NxSearchDropdown`?

You're probably right about that. I did the dark mode text input ticket, and set the placeholder color to the disabled state, but didn't think to change the actual text color if there was a value inside.

gallery/visualtests/__image_snapshots_dark__/nx-search-dropdown-js-nx-search-dropdown-looks-right-with-the-long-variant-in-error-1-dark.png
EmilyBermudez88: > This text color is clearly incorrect. Have alerts been converted yet? If not I'd think we need to pause this ticket until they are

NxAlert is now merged so you're good to keep going 👍

entroform: > > This text color is clearly incorrect. Have alerts been converted yet? If not I'd think we need to pause this ticket until they are
> 
> NxAlert is now merged so you're good to keep going 👍

Thanks! Snapshots are updated here: e92216c2deb96fc3563e0a6cb231c85fe6d67890

entroform:
> There are a few discrepancies in the mocks that may need to be fixed and/or confirmed with design:
> 
> 1. In NxTextInput, the text colour in dark mode is `indigo-80`, but here is `indigo-90`
> 2. ‘Text’ in the ‘Non-Empty’ examples still have the placeholder colour
> 3. Loading Spinner: in light mode, we use `--nx-color-text`, and in dark mode we use `--nx-color-text-stark` (even though the all other options use `--nx-color-text`, as does the submit mask)
>    All the code here follows what's already been implemented elsewhere, but maybe good to check with design that the mocks are wrong, or update accordingly.

I believe these might have all been addressed, since I merge the PR with the feature branch.
And the mockup have been updated as well: non-empty.
It looks like the submit mask this is a some sketch error and Jason is fixing it but yes it should be nx-color-text.
Thanks for finding these 😸

EmilyBermudez88:
>I believe these might have all been addressed, since I merge the PR with the feature branch.
And the mockup have been updated as well: non-empty.
It looks like the submit mask this is a some sketch error and Jason is fixing it but yes it should be nx-color-text.
Thanks for finding these 😸

Submit mask & the non-empty examples have been fixed. But all the examples that include a dropdown have the text set to indigo-90 or its HEX value #DBDEF0

lib/src/components/NxTextInput/NxTextInput.scss
```diff
@@ -112,7 +116,7 @@
   &.disabled {
     .nx-text-input__input {
       &::placeholder {
-        color: var(--nx-color-interactive-border-disabled);
+        color: var(--nx-color-text-disabled);
```
EmilyBermudez88, line 119: actually I think this is a mistaken duplicate from when I did NxTextInput. You should be able to remove it

entroform: > actually I think this is a mistaken duplicate from when I did NxTextInput. You should be able to remove it

Fixed here! 21c3b0680a80d9ee52e1610f2b55bbeffe0db407

entroform:
> > I believe these might have all been addressed, since I merge the PR with the feature branch.
> > And the mockup have been updated as well: non-empty.
> > It looks like the submit mask this is a some sketch error and Jason is fixing it but yes it should be nx-color-text.
> > Thanks for finding these 😸
> 
> Submit mask & the non-empty examples have been fixed. But all the examples that include a dropdown have the text set to indigo-90 or its HEX value #DBDEF0. Also the two examples under 'Component' have a border colour of indigo-60 instead of the default indigo-70

These should be fixed now 👍

EmilyBermudez88 (approved):

PR #1048, opened 2023-04-03
Dark Mode:  Checkboxes Radios - RSC-1010

entroform:
Checkboxes:
https://issues.sonatype.org/browse/RSC-1010

Radios:
https://issues.sonatype.org/browse/RSC-1037


entroform:
I'm tempted to change all the `--nx-color ... -disabled` to `--nx-color-disabled` but maybe it's better like this for now?
Maybe something we can revisit later.
Let me know if you have any thoughts on this 👍

avnvbhatta:
> I'm tempted to change all the `--nx-color ... -disabled` to `--nx-color-disabled` but maybe it's better like this for now? Maybe something we can revisit later. Let me know if you have any thoughts on this 👍

I think it's a good idea. Less maintenance as well. I think we'll only have 2 disabled variables instead of the 4 we have right now. I'd say we do it in another PR though.

gallery/visualtests/__image_snapshots_dark__/nx-radio-js-nx-radio-attribute-disabled-nx-radio-and-attribute-disabled-checked-looks-disabled-by-default-1-dark.png
avnvbhatta, line 1: This looks like it's incorrect. 
Should look like:
image

Instead it's:
image

avnvbhatta:
The check icon in a disabled + checked checkbox has color `indigo-25` in the mockup, but is `indigo-20` here. I would double check with Jason on this. I feel like the mockup is incorrect.

avnvbhatta:
The disabled items have the wrong color. Should be `indigo-40`. Currently it's the same as text color `(indigo-80)`.
image

lib/src/base-styles/_nx-colors.scss
```diff
@@ -92,13 +94,17 @@
   --nx-color-counter: var(--nx-swatch-grey-05);
   --nx-color-counter-active-text: var(--nx-swatch-white);
 
+  // Radio Checkbox
+  --nx-color-radio-checkbox-selected: var(--nx-swatch-blue-40);
```
rpokorny, line 98:
```suggestion
  --nx-color-radio-checkbox-selected: var(--nx-swatch-blue-40);

```

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -6,6 +6,19 @@
  */
 @use '../scss-shared/nx-container-helpers';
 @use '../scss-shared/nx-text-helpers';
+@use '../scss-shared/nx-dark-mode-helpers';
+
+:root {
+  --nx-radio-checkbox-border-width: 1px;
```
rpokorny, line 12: Why is this here and not in `_nx-variables.scss`?

entroform: > Why is this here and not in `_nx-variables.scss`?

Totally forgot to move it there: 2353f6e1bcb99869300b48b54245e083c1f0d210

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -30,17 +43,19 @@
     cursor: pointer;
 
     .nx-radio-checkbox__control {
-      border-color: var(--nx-color-interactive-border-hover);
+      @include inset-box-shadow-border(var(--nx-color-interactive-border-hover));
```
rpokorny, line 38: Why is this being done with box-shadow?

entroform: > Why is this being done with box-shadow?

To fix it offsetting the checkmark icon, since it is now 2px in dark mode instead of 1

rpokorny: > Why is this being done with box-shadow?

If it's about positioning the checkmark, consider that the border isn't visible when the checkmark is there, and really could be removed entirely in that state

lib/src/scss-shared/_nx-focus-border-helpers.scss
```diff
@@ -6,12 +6,12 @@
  */
 
 // For dynamically calculating focus border styles based on height and width
-// Border size is 2px border and 4px here represents total space in each axis
+// Border size is 2px border + 2px gap = 2 * 4px here to represents total space in each axis
```
rpokorny, line 9: This seems like it would affect light mode too wouldn't it?

lib/src/base-styles/_nx-radio-checkbox.scss
entroform: > > Why is this being done with box-shadow?
> 
> If it's about positioning the checkmark, consider that the border isn't visible when the checkmark is there, and really could be removed entirely in that state

Hmm, that would also complicate the focus styling because it also offsets the calculation for that .. i think

lib/src/scss-shared/_nx-focus-border-helpers.scss
entroform: > This seems like it would affect light mode too wouldn't it?

I think in light mode it is also 2px for focus borders.

entroform: > This seems like it would affect light mode too wouldn't it?

Oh yeah, this was also one of the reasons why I was trying out box-shadow, it simplified the left, top calculation on the ::after

rpokorny: > I think in light mode it is also 2px for focus borders.

Yes but I was trying to understand how this change doesn't result in a difference to light mode.  I now see that this is essentially cancelled out by the `box-sizing: border-box` that's applied to the `::after` in this PR.  I suspect you could get rid of both that change and this change together with no visual difference.

rpokorny: > Oh yeah, this was also one of the reasons why I was trying out box-shadow, it simplified the left, top calculation on the ::after

OK that's a decent reason.  Please document that in a code comment near the point where the box-shadow stuff is implemented.

entroform:
> The check icon in a disabled + checked checkbox has color `indigo-25` in the mockup, but is `indigo-20` here. I would double check with Jason on this. I feel like the mockup is incorrect.

Thanks, just talked with Jason.
This should be updated now on Sketch

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -6,17 +6,22 @@
...
-  height: $-height;
-  width: $-width;
+  height: $-size;
+  width: $-size;
+
+  border-radius: $-size;
+}
+
+.nx-radio__outer-circle {
+  stroke-width: var(--nx-radio-checkbox-border-width);
```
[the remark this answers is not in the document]
entroform: > This isn't working correctly. If you remove the `border-radius` you can see that the `nx-radio__outer-circle` is actually thicker than what's visible with that border radius, and gets cut off. Conversely, if you add `display: none` to `.nx-radio__outer-circle` you can see the visible border increase towards the inside. Overall I think there's a real problem in terms of how the outer circle is drawn: it's effectively drawn by _both_ the box-shadow and the svg `<circle>`, with each of those two drawing slightly different things neither of which are quite right.
> 
> Part of the problem is that the `r` on the `.nx-radio__outer-circle` needs to be decreased by 0.5 in dark mode to accommodate the increased `stroke-width`. That could be challenging as that property only became CSS-controllable in SVG 2 which has unclear browser support. So the best path forward may depend on what you find in terms of browser support for that.
> 
> I think a bigger re-think is needed here. Some options I can suggest:
> 
> * introduce a wrapper element around the `nx-radio-checkbox__control` that just establishes the positioning context for the focus border, but which doesn't draw anything itself.  Then the borders on the `nx-radio-checkbox__control` can go back to being drawn the way they used to be.
> * Switch to `svg` for `NxCheckbox` so that it matches `NxRadio` in that regard, using a `<path>` to draw its border.  As I write this I realize it's probably the harder option.
> * **If the radius of the `nx-radio__outer-circle` cannot be controlled via CSS** then you'll probably need to stick with something more like your current `box-shadow` solution, but getting rid of `nx-radio__outer-circle` as it's essentially just in the way right now.

Thanks, did a refactor here: c48a8acbcfd31f99314d410d180ece3c046a6577
I still need to update visual tests and add comment.

entroform:
> The disabled items have the wrong color. Should be `indigo-40`. Currently it's the same as text color `(indigo-80)`. image

Thanks, I fixed the disabled state as part of this refactor.
c48a8acbcfd31f99314d410d180ece3c046a6577

There's some weird border color on the mockup, currently sorting that out with Jason.

lib/src/scss-shared/_nx-focus-border-helpers.scss
entroform: > > Oh yeah, this was also one of the reasons why I was trying out box-shadow, it simplified the left, top calculation on the ::after
> 
> OK that's a decent reason. Please document that in a code comment near the point where the box-shadow stuff is implemented.

Added Comment here: 64a881d4a98b35a73759f8e0e3ebb680923d9562

gallery/visualtests/__image_snapshots_dark__/nx-radio-js-nx-radio-attribute-disabled-nx-radio-and-attribute-disabled-checked-looks-disabled-by-default-1-dark.png
entroform: > This looks like it's incorrect. Should look like: image
> 
> Instead it's: image

This was fixed as part of this refactor:
64a881d4a98b35a73759f8e0e3ebb680923d9562

The border thing is still unconfirmed waiting for Jason

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -230,6 +230,9 @@ const CssVariablesPage = () => {
               as the value of a <NxCode>filter</NxCode> <NxCode>drop-shadow</NxCode>.
             </NxWarningAlert>
           </PropertyDocItem>
+          <PropertyDocItem propertyVar="--nx-radio-checkbox-border-width">
+            The standard checkbox border width, or radio stroke-width.
```
rpokorny, line 234: No longer using `stroke-width`

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -653,6 +659,30 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-policy-threat-slider-text">
               The default text color of <NxCode>NxPolicyThreatSlider</NxCode>.
             </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-radio-checkbox-selected">
+              The default selected color for radios and checkboxes.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-radio-disabled-border">
+              The default border color for disabled <NxCode>NxRadio</NxCode>.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-radio-disabled-box-shadow">
```
rpokorny, line 668: I feel like for these the the fact that they're box shadows is very much an implementation detail.  We might want to avoid expressing that in the variable name.

lib/src/base-styles/_nx-colors.scss
```diff
@@ -184,19 +198,19 @@
 
     --nx-color-interactive-border: var(--nx-swatch-indigo-70);
     --nx-color-interactive-border-hover: var(--nx-swatch-indigo-60);
-    --nx-color-interactive-border-disabled: var(--nx-swatch-indigo-40);
+    --nx-color-interactive-border-disabled: var(--nx-color-disabled);
```
rpokorny, line 201: Properties that are set the same in light and dark mode don't need to be repeated in the dark mode block

lib/src/base-styles/_nx-radio-checkbox.scss
```diff
@@ -30,17 +31,21 @@
     cursor: pointer;
 
     .nx-radio-checkbox__control {
-      border-color: var(--nx-color-interactive-border-hover);
-      stroke: var(--nx-color-interactive-border-hover);
+      // We are using box-shadow below instead of border
+      // so we do not have to re-calculate the focus Pseudo-element in NxRadio and NxCheckbox
+      // when the width changes in dark mode.
+      box-shadow: inset 0 0 0 var(--nx-radio-checkbox-border-width) var(--nx-color-interactive-border-hover);
```
rpokorny, line 37: You previously had a mixin for these box-shadows.  That seemed valuable due to the repetition; why'd you get rid of it?

lib/src/base-styles/_nx-variables.scss
```diff
@@ -65,13 +65,21 @@
                             0 1px 2px 0 rgba(91, 95, 141, 0.2),
                             0 1px 1px 0 rgba(124, 128, 179, 0.2);
 
+  // Radios and Checkboxes
+  --nx-radio-checkbox-border-width: 1px;
+  --nx-radio-checkbox-border-width-disabled: 1px;
```
rpokorny, line 70: Please follow the naming conventions already present in this file e.g. `--nx-border-width-radio-checkbox`

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -21,24 +21,49 @@ $-width: 16px;
...
+    top: -4px;
   }
 
   .fa-check {
-    color: var(--nx-swatch-white);
+    color: var(--nx-color-component-background);
     font-size: var(--nx-font-size-s);
-    vertical-align: 4px;
+    vertical-align: 3px;
+    margin-left: 1px;
```
rpokorny, line 32: This can be avoided with `text-align: center` on the `nx-checkbox__box`

rpokorny, line 32: These can be avoided by turning the parent into an inline flex container with `align-items` and `justify-content` both set to `center`.

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -21,24 +21,49 @@ $-width: 16px;
...
-      color: var(--nx-color-text-disabled);
+    &:disabled {
+      &, &:hover {
+        + .nx-checkbox__box {
+          background-color: var(--nx-color-checkbox-disabled-background);
+          box-shadow: inset 0 0 0 var(--nx-radio-checkbox-border-width) var(--nx-color-checkbox-disabled-box-shadow);
+    
+          .fa-check {
+            color: var(--nx-color-checkbox-disabled-icon);
+          }
```
rpokorny, line 64: The `background-color` and `box-shadow` here are redundant, and I believe the `.fa-check` block can be moved up to the disabled styles just above this without any harm.

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -6,31 +6,39 @@
...
+      }    
+    }
 
-      .nx-radio__inner-circle {
-        stroke: var(--nx-color-interactive-background-disabled);
-        fill: var(--nx-color-text-disabled);
+    &:disabled {
+      + .nx-radio__circle {
+        background-color: var(--nx-color-radio-disabled-checked-background);
+        box-shadow:
```
rpokorny, line 39: Redundant

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -6,31 +6,39 @@
...
-
-.nx-radio__input:checked {
-  + .nx-radio__circle .nx-radio__outer-circle {
-    fill: none;
+.nx-radio__input {
+  &:disabled {
+    + .nx-radio__circle {
+      background-color: var(--nx-color-radio-disabled-background);
+      box-shadow:
+        inset 0 0 0 1px var(--nx-color-radio-disabled-border),
```
rpokorny, line 24: Pretty sure this first shadow does nothing that isn't also accomplished by the second shadow

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -6,31 +6,39 @@
...
+    }
 
-      .nx-radio__inner-circle {
-        stroke: var(--nx-color-interactive-background-disabled);
-        fill: var(--nx-color-text-disabled);
+    &:disabled {
+      + .nx-radio__circle {
+        background-color: var(--nx-color-radio-disabled-checked-background);
+        box-shadow:
+          inset 0 0 0 1px var(--nx-color-radio-disabled-border),
```
rpokorny, line 40: Like above, I don't see any effect in using two shadows here.  The larger of the two seems sufficient by itself

entroform: > Like above, I don't see any effect in using two shadows here. The larger of the two seems sufficient by itself

The first shadow is for when it is in light mode, there is an additional border.

lib/src/components/NxCheckbox/NxCheckbox.scss
entroform: > The `background-color` and `box-shadow` here are redundant, and I believe the `.fa-check` block can be moved up to the disabled styles just above this without any harm.

I thought so too, but it is needed or else it doesn't show the disabled state when checked
Will investigate more.

Screenshot 2023-04-11 at 9 21 35 AM

gallery/src/styles/CssVariables/CssVariablesPage.tsx
entroform: > No longer using `stroke-width`

Fixed as part of: fa351f5ea157e03267d26df3eebf91cfb2881237

lib/src/base-styles/_nx-variables.scss
entroform: > Please follow the naming conventions already present in this file e.g. `--nx-border-width-radio-checkbox`

Renamed here: fa351f5ea157e03267d26df3eebf91cfb2881237

gallery/src/styles/CssVariables/CssVariablesPage.tsx
entroform: > I feel like for these the the fact that they're box shadows is very much an implementation detail. We might want to avoid expressing that in the variable name.

Let me know your thoughts! 29f54f813d110ac8ff28b80f676350b3d9feed62

lib/src/base-styles/_nx-colors.scss
```diff
@@ -328,6 +345,15 @@
     --nx-color-policy-threat-slider-background-disabled: var(--nx-color-interactive-border-disabled);
     --nx-color-policy-threat-slider-text-disabled: var(--nx-color-policy-threat-slider-border-disabled);
     --nx-color-policy-threat-slider-text: var(--nx-swatch-indigo-97);
+
+    // Radio Checkbox
+    --nx-color-radio-disabled-border: var(--nx-color-disabled);
```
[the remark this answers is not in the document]
entroform: > As per Ross' comment above, you can remove this and line 355 since they're values don't change

Thanks! removed here: 34c92ae8e9cd7e3092d4f05218ec3a5b1414a659

gallery/visualtests/__image_snapshots_dark__/nx-radio-js-nx-radio-attribute-disabled-nx-radio-and-attribute-disabled-checked-looks-disabled-by-default-1-dark.png
entroform: Kaleido is updated now, and new snapshots here:
8abbf72e9cc0194ead8197925600ff137e565ba8

gallery/src/styles/CssVariables/CssVariablesPage.tsx
rpokorny: > Let me know your thoughts! https://github.com/sonatype/sonatype-react-shared-components/commit/29f54f813d110ac8ff28b80f676350b3d9feed62

:thinking: It's awkward to name because the construction of the disabled radios doesn't match the non-disabled, and also doesn't match between light and dark mode.  The naming of the dot in the center of the disabled+checked state is also in play, as you have that currently named `background`.  To me, this "inner border" as you call it is actually more conceptually the background. What if we call the dot `dot`, call this middle section the `background`, and call the border `border`.  All of these are aligned more to what they look like than to how they're implemented, but I think that's the right choice here since the implementation is pretty funky and subject to change in the future.

lib/src/base-styles/_nx-radio-checkbox.scss
entroform: > > Brought it back here: [d6bab91](https://github.com/sonatype/sonatype-react-shared-components/commit/d6bab913222219f39f54653c5cf49057ff9def54)
> 
> Rather than duplicating it in each file, I'd say it should be in a helper file. The one for the more complicated disabled box-shadow should probably move to that same file

I moved it to its own file here: 66373d319bea3d6929c75b8ff76835f5cc99574a
Although, I feel that it might be simpler before (less coupling), not sure.
Let me know what you think!

gallery/visualtests/__image_snapshots_dark__/nx-radio-js-nx-radio-attribute-disabled-nx-radio-and-attribute-disabled-checked-looks-disabled-by-default-1-dark.png
rpokorny, line 1: Judging by the name of the test, isn't this one supposed to be checked?

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -21,24 +25,47 @@ $-width: 16px;
     @include nx-focus-border-helpers.radio-checkbox-focus-border($-height, $-width);
 
     border-radius: 4px;
-    left: -5px;
-    top: -5px;
+    left: -4px;
```
rpokorny, line 28:
```suggestion
    left: calc(var(--nx-focus-outline-offset) + var(--nx-focus-outline-width));
```

lib/src/base-styles/_nx-radio-checkbox.scss
entroform: > > I moved it to its own file here: [66373d3](https://github.com/sonatype/sonatype-react-shared-components/commit/66373d319bea3d6929c75b8ff76835f5cc99574a)
> > Although, I feel that it might be simpler before (less coupling), not sure.
> > Let me know what you think!
> 
> You missed the one in `_nx-radio-checkbox.scss`, which was a big part of my point as it's identical to one of the others.

Ah right!
Fixed:
02fc06491d76cc47ea2a483a6238180066df55b6
ca6524f7561dd171d85894971055cd491513c9e1

lib/src/components/NxCheckbox/NxCheckbox.scss
rpokorny: > I thought so too, but it is needed or else it doesn't show the disabled state when checked
Will investigate more.

Try reversing the order of the `checked` and `disabled` blocks in the CSS.  They have equal specificity so disabled needs to come afterwards in order to win.

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -5,32 +5,39 @@
...
+  &:checked {
+    &, &:hover {
+      + .nx-radio__circle {
+        @include nx-radio-checkbox-helpers.inset-box-shadow(var(--nx-color-radio-checkbox-selected), $-checked-border-width);
+      }    
+    }
+
+    &:disabled {
+      + .nx-radio__circle {
+        @include nx-radio-checkbox-helpers.multi-inset-box-shadow($-checked-border-width);
```
rpokorny, line 38: Can be slightly simplified similar to https://github.com/sonatype/sonatype-react-shared-components/pull/1048#discussion_r1163218108

lib/src/components/NxRadio/NxRadio.scss
```diff
@@ -5,32 +5,39 @@
...
   }
 
-  &:disabled {
-    + svg.nx-radio__circle {
-      stroke: var(--nx-color-interactive-border-disabled);
+  &:checked {
+    &, &:hover {
+      + .nx-radio__circle {
+        @include nx-radio-checkbox-helpers.inset-box-shadow(var(--nx-color-radio-checkbox-selected), $-checked-border-width);
+      }
```
rpokorny, line 33:
```suggestion
      }
```

gallery/visualtests/__image_snapshots_dark__/nx-radio-js-nx-radio-attribute-disabled-nx-radio-and-attribute-disabled-checked-looks-disabled-by-default-1-dark.png
entroform: > Judging by the name of the test, isn't this one supposed to be checked?

Yup!
it looks like the selector was incorrect, fixed: 7d8932375c91d1cfef6b33343a2590b3806fbd1c
Forgot to update it for light mode will do that momentarily

gallery/src/styles/CssVariables/CssVariablesPage.tsx
entroform: > > Let me know your thoughts! [29f54f8](https://github.com/sonatype/sonatype-react-shared-components/commit/29f54f813d110ac8ff28b80f676350b3d9feed62)
> 
> 🤔 It's awkward to name because the construction of the disabled radios doesn't match the non-disabled, and also doesn't match between light and dark mode. The naming of the dot in the center of the disabled+checked state is also in play, as you have that currently named `background`. To me, this "inner border" as you call it is actually more conceptually the background. What if we call the dot `dot`, call this middle section the `background`, and call the border `border`? All of these are aligned more to what they look like than to how they're implemented, but I think that's the right choice here since the implementation is pretty funky and subject to change in the future.

Renamed to `ring` and `dot`: 693777f1b65e65acda79a9816bccbc90a0c8e248

lib/src/components/NxCheckbox/NxCheckbox.scss
entroform: > > I thought so too, but it is needed or else it doesn't show the disabled state when checked
> > Will investigate more.
> 
> Try reversing the order of the `checked` and `disabled` blocks in the CSS. They have equal specificity so disabled needs to come afterwards in order to win.

Thanks this worked for checkbox but not radio 🤔 I'm still playing around with it
c5f9fe75186d4edcbcfd8ad29c6ba6e4a53af5a0

lib/src/components/NxRadio/NxRadio.scss
entroform: > Can be slightly simplified similar to [#1048 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/1048#discussion_r1163218108)

This is simplified here: 43a72297fbdddfc7ef57ab1e73a1a3a37d033b1c

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -653,6 +659,30 @@ const CssVariablesPage = () => {
...
+              The default selected color for radios and checkboxes.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-radio-disabled-border">
+              The default border color for disabled <NxCode>NxRadio</NxCode>.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-radio-disabled-ring">
+              The default ring color for disabled <NxCode>NxRadio</NxCode>.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-radio-disabled-dot">
+              The default dot color for disabled <NxCode>NxRadio</NxCode>.
```
rpokorny, line 672: This would be a good place to expand slightly on what we mean by ring and dot

gallery/visualtests/nxRadio.js
```diff
@@ -25,6 +25,7 @@ describe('NxRadio', function() {
 
   const simpleExampleSelector = '#nx-radio-example .gallery-example-live',
       simpleExampleLabelSelector = '#nx-radio-example .gallery-example-live label:nth-of-type(3)',
+      disabledExampleSelector = '#nx-radio-disabled-example .gallery-example-live',
       disabledExampleLabelSelector = '#nx-radio-disabled-example .gallery-example-live label:nth-of-type(1)',
       disabledExampleLabelCheckedSelector = '#nx-radio-disabled-example .gallery-example-live label:nth-of-type(2)';
```
rpokorny, line 30:
```suggestion
      disabledExampleLabelSelector = `${disabledExampleSelector} label:nth-of-type(1)`,
      disabledExampleLabelCheckedSelector = `${disabledExampleSelector} label:nth-of-type(2)`;
```

lib/src/base-styles/_nx-colors.scss
```diff
@@ -184,19 +198,18 @@
 
     --nx-color-interactive-border: var(--nx-swatch-indigo-70);
     --nx-color-interactive-border-hover: var(--nx-swatch-indigo-60);
-    --nx-color-interactive-border-disabled: var(--nx-swatch-indigo-40);
     --nx-color-interactive-border-focus: var(--nx-swatch-purple-80);
 
     --nx-color-text-stark: var(--nx-swatch-indigo-90);
     --nx-color-text: var(--nx-swatch-indigo-80);
-    --nx-color-text-disabled: var(--nx-color-interactive-border-disabled);
+    --nx-color-text-disabled: var(--nx-color-disabled);
```
rpokorny, line 205: Same as light mode
```suggestion
```

lib/src/base-styles/_nx-colors.scss
```diff
@@ -184,19 +198,18 @@
...
     --nx-color-text-placeholder: var(--nx-color-indigo-70);
 
     --nx-color-link: var(--nx-swatch-purple-80);
     --nx-color-link-focus: var(--nx-color-link);
     --nx-color-link-active: var(--nx-swatch-purple-90);
     --nx-color-link-hover: var(--nx-swatch-purple-70);
-    --nx-color-link-disabled: var(--nx-color-text-disabled);
+    --nx-color-link-disabled: var(--nx-color-disabled);
```
rpokorny, line 212: Same as light mode
```suggestion
```

lib/src/components/NxCheckbox/NxCheckbox.scss
rpokorny: The same applies to the `top` on the line right below this

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -21,24 +25,38 @@ $-width: 16px;
     @include nx-focus-border-helpers.radio-checkbox-focus-border($-height, $-width);
 
     border-radius: 4px;
```
rpokorny, line 27: I'm seeing a border-radius of `2px` in kaliedo

lib/src/components/NxCheckbox/NxCheckbox.scss
```diff
@@ -21,24 +25,38 @@ $-width: 16px;
...
 }
 
-.nx-checkbox__input:checked:disabled {
-  + .nx-checkbox__box {
-    background-color:var(--nx-color-interactive-background-disabled);
-    border-color: var(--nx-color-interactive-border-disabled);
+.nx-checkbox__input {
+  &:checked {
+    + .nx-checkbox__box {
+      @include nx-radio-checkbox-helpers.inset-box-shadow(var(--nx-color-radio-checkbox-selected));
```
rpokorny, line 41: This block could equivalently be
```suggestion
      background-color: var(--nx-color-radio-checkbox-selected);
      
      &, &:hover {
        @include nx-radio-checkbox-helpers.inset-box-shadow(var(--nx-color-radio-checkbox-selected));
      }
```

lib/src/scss-shared/_nx-radio-checkbox-helpers.scss
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+// We are using box-shadow below instead of border
+// so we do not have to re-calculate the focus Pseudo-element in NxRadio and NxCheckbox
+// when the width changes in dark mode.
+@mixin inset-box-shadow($color, $border-width: var(--nx-border-width-radio-checkbox)) {
+  box-shadow: inset 0 0 0 #{$border-width} #{$color};
+}
+
+@mixin multi-inset-box-shadow($border-width) {
```
rpokorny, line 15: This is now only used in one place so doesn't really need to be a mixin anymore

lib/src/components/NxCheckbox/NxCheckbox.scss
entroform: > The same applies to the `top` on the line right below this

Fixed: 6c8b03cdcf54133c67d8788fbc853d0685364067

lib/src/components/NxCheckbox/NxCheckbox.scss
entroform: > This block could equivalently be
> 
> ```
>       background-color: var(--nx-color-radio-checkbox-selected);
>       
>       &, &:hover {
>         @include nx-radio-checkbox-helpers.inset-box-shadow(var(--nx-color-radio-checkbox-selected));
>       }
> ```

Yup, fixed: a81ecdf26ef5489af7702df498b135c0b45e9e8a

lib/src/scss-shared/_nx-radio-checkbox-helpers.scss
entroform: > This is now only used in one place so doesn't really need to be a mixin anymore

Removed: 0f2e8e4227d56062c60222ca9abd6ffd439197c8

gallery/src/styles/CssVariables/CssVariablesPage.tsx
entroform: > This would be a good place to expand slightly on what we mean by ring and dot

Updated description here: 67e550af39a47ffe74e50c1b153ffeaa3f79742d

PR #1055, opened 2023-04-06
fix active secondary button text color in light mode- RSC-1560

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1560


rpokorny (approved):
Looks like that did the trick, assuming those are in fact the only visual test diffs

EmilyBermudez88:
> Looks like that did the trick, assuming those are in fact the only visual test diffs

Thanks for the reminder. I ran the entire set of tests and noticed bug with dropdown/filter dropdown now due to the hover selector change, so need to rework it.

rpokorny (approved):

PR #1058, opened 2023-04-07
NxForm Dark Mode - RSC-1020

cindyko226:
https://issues.sonatype.org/browse/RSC-1020


gallery/src/components/NxForm/NxFormCustomizedExample.scss
```diff
@@ -4,9 +4,16 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
+@use '~@sonatype/react-shared-components/scss-shared/nx-dark-mode-helpers';
+
 .gallery-custom-form {
   background-color: #eee;
   padding: var(--nx-spacing-4x);
+
+  // use dark mode helper here as it is for demonstration purpose only
```
rpokorny, line 13: Keep in mind that this code is part of the example, readable by the gallery user.  What is this comment intended to convey to that audience?

rpokorny:
The dark mode form mockup is missing the Required Field Asterisks.  I'm not sure what color they should be in dark mode; they should be added to a mockup so we can be sure.

gallery/src/components/NxForm/NxFormCustomizedExample.scss
cindyko226: Since we are creating specific variables for the color used in component in dark mode, leave the comments here that this code is for example purpose and no need to create a new variable.

cindyko226:
> The dark mode form mockup is missing the Required Field Asterisks. I'm not sure what color they should be in dark mode; they should be added to a mockup so we can be sure.

Mockups are updated. https://www.sketch.com/s/2393264e-1233-4909-99c8-a7dda02588cd/a/Gom1awY

gallery/visualtests/__image_snapshots__/nx-formlayout-js-nx-form-nx-form-horizontal-layout-looks-right-1.png
rpokorny, line 1: There seem to be some minute differences in how radios render now which is the cause of these light mode diffs

EmilyBermudez88 (approved):

PR #1061, opened 2023-04-12
Fix padding and disabled colors of NxDropdown - RSC-1564 

cindyko226:
https://issues.sonatype.org/browse/RSC-1564


lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -37,9 +37,8 @@
   display: flex;
   margin: 0;
   outline: none;
+  padding: calc(var(--nx-spacing-2x) - 1px) calc(var(--nx-spacing-4x) - 1px);
```
EmilyBermudez88, line 40:
```suggestion
  padding: calc(var(--nx-spacing-2x) - 1px) var(--nx-spacing-4x);
```
The left & right padding was already correct, so you should just have to override the top & bottom padding to always subtract 1px (instead of 2px in dark mode inherited from nxButton)

rpokorny: > The left & right padding was already correct

Is it though? Sketch shows 16px left and right _including the border_.  So by setting the CSS padding to 16px dropdowns are currently 2px too wide in both light and dark mode

cindyko226:
> It's not clear to me exactly what this is trying to change. Of the three mockups in the diff, on the first two it only seems like the border is different, and in the third one I can't tell what's different at all. I don't see any differences that seem related to the description of this ticket. If this ticket is about a case that isn't covered in our existing visual tests, more visual tests should probably be added in order to cover it.

I think this ticket was created to fix the bugs affected by [RSC-1007](https://issues.sonatype.org/browse/RSC-1007) and [RSC-1543](https://issues.sonatype.org/browse/RSC-1543), both are implemented after `NxDropdown`.

- [RSC-1007](https://issues.sonatype.org/browse/RSC-1007) - the border width is changed to 2px in dark mode which leads to the result of 1px less in top & bottom padding in `NxDropdown`
- [RSC-1543](https://issues.sonatype.org/browse/RSC-1543) - the color change from `nx-indigo-35` to `nx-indigo-40` which is the difference in the third snapshot

Maybe we should also check the dark mode components that were done before the changes in above two tickets which might need to update its snapshots or a fully clean up before merging Feature branch to main.

lib/src/components/NxDropdown/NxDropdown.scss
cindyko226: The left and right padding was not correct, so it needs to be overridden here to just subtract 1px instead of 2px(button border width) in dark mode.

EmilyBermudez88: >>Is it though? Sketch shows 16px left and right including the border. So by setting the CSS padding to 16px dropdowns are currently 2px too wide in both light and dark mode

> I think the left and right padding was not correct, so it would need to be overridden here to just subtract 1px instead of 2px(button border width) in dark mode.

Yep you're right! I was looking at main and actually it's currently wrong then. It's currently set to 16px of padding plus 1 px of border, so this would resolve indiscrepencies in light mode as well.

EmilyBermudez88:
> Maybe we should also check the dark mode components that were done before the changes in above two tickets which might need to update its snapshots or a fully clean up before merging Feature branch to main.

I may be wrong, but I believe the snapshots for NxDropdown & NxFilterDropdown have not been touched since RSC-1007 has gone through, so anything that RSC-1007 has affected in NxDropdown & NxFilterDropdown should not currently be reflected in the FEATURE_RSC-928's snapshots (the disabled dropdown snapshot doesn't have the incorrect styling applied)

rpokorny:
I agree the that light-mode snapshots should be updated and honestly probably quite a few more dark mode snapshots too.  You can run `yarn ci-jest-dark` to have it run all tests but avoid generating snapshots for ones we haven't done yet (you'll get a lot of failures for those which you can ignore).  Then go through the other failures and for any that are because of your `NxDropdown` changes, update those snapshots.  You may want to merge https://github.com/sonatype/sonatype-react-shared-components/pull/1065 into your branch first though as it cleans of a lot of other snapshot issues that had accumulated recently.

cindyko226:
I have updated light/dark mode snapshots affected by `NxDropdown` changes.

- Dark mode settings bc54c46d5579f8128b63ad1ec7a34f8a895ae9a6
- Dropdown 
  - 16a2e16607a14c19c9f5cd1177816a5be994ab2c
  - dfe949b65cdff88e2d088a2aaa7ad008c1d596ea
- Filter Dropdown
  - 5935b5fd93e7752f8c15d3f418732b782754e15d
  - dc863969b6c77e65e531731278aef73c0f21e1d6
- Tile
  - 61907ffe7ce9ef40360f8a43e7c1ea53e0661329
  - fc7e22c509a0e67648eb1bf9dd2f68e05c5d5160

EmilyBermudez88:
> I have updated light/dark mode snapshots affected by `NxDropdown` changes.
> 
> * Dark mode settings [bc54c46](https://github.com/sonatype/sonatype-react-shared-components/commit/bc54c46d5579f8128b63ad1ec7a34f8a895ae9a6)
> * Dropdown
>   
>   * [16a2e16](https://github.com/sonatype/sonatype-react-shared-components/commit/16a2e16607a14c19c9f5cd1177816a5be994ab2c)
>   * [dfe949b](https://github.com/sonatype/sonatype-react-shared-components/commit/dfe949b65cdff88e2d088a2aaa7ad008c1d596ea)
> * Filter Dropdown
>   
>   * [5935b5f](https://github.com/sonatype/sonatype-react-shared-components/commit/5935b5fd93e7752f8c15d3f418732b782754e15d)
>   * [dc86396](https://github.com/sonatype/sonatype-react-shared-components/commit/dc863969b6c77e65e531731278aef73c0f21e1d6)
> * Tile
>   
>   * [61907ff](https://github.com/sonatype/sonatype-react-shared-components/commit/61907ffe7ce9ef40360f8a43e7c1ea53e0661329)
>   * [fc7e22c](https://github.com/sonatype/sonatype-react-shared-components/commit/fc7e22c509a0e67648eb1bf9dd2f68e05c5d5160)

For the updated snapshots with focus borders, I think they're off slightly now as well. The outline offset is a calculation of `calc(0px - var(--nx-border-width-button) - var(--nx-focus-outline-inset) - var(--nx-focus-outline-width)`, which works for buttons but not here. The border width of `nx-dropdown__toggle` is always 1px (while `--nx-border-width-button` is 2px in dark mode), so the focus border in dark mode is slightly more inset than before. We will need to override this for NxDropdown!

cindyko226:
> For the updated snapshots with focus borders, I think they're off slightly now as well. The outline offset is a calculation of calc(0px - var(--nx-border-width-button) - var(--nx-focus-outline-inset) - var(--nx-focus-outline-width), which works for buttons but not here. The border width of nx-dropdown__toggle is always 1px (while --nx-border-width-button is 2px in dark mode), so the focus border in dark mode is slightly more inset than before. We will need to override this for NxDropdown!

Thanks for catching this, updated here
0cbc7b3b778416e78c7dafa2ac5c4a9b212ffc4d
1b853ca621cc28d699b77a92b17269445085a717

EmilyBermudez88:
You can revert the changes made to the snapshots where NxFilterDropdown is contained with NxTable. It was decided not to update snapshots that have incomplete components present. I am working on the NxTable ticket atm and have a note to update those snapshots when this ticket is complete.

EmilyBermudez88 (approved):
Minus the feedback in https://github.com/sonatype/sonatype-react-shared-components/pull/1061#issuecomment-1514960195, it LGTM!

cindyko226:
> You can revert the changes made to the snapshots where NxFilterDropdown is contained with NxTable. It was decided not to update snapshots that have incomplete components present (decided after these snapshots were initially updated). I am working on the NxTable ticket atm and have a note to update those snapshots when this ticket is complete.

Reverted 916f7b77cc0852e0ea619204a50e21912359a98a

PR #1062, opened 2023-04-12
Fix Small Icon-only Button Height in Dark Mode - RSC-1567

rpokorny:
https://issues.sonatype.org/browse/RSC-1567

This affects all "small" icon-only buttons including those in `NxList` actions, all `NxCloseButton`s, etc.


EmilyBermudez88:
I don't believe there is a specific ticket for close buttons, is it worth updating the snapshots here for dark mode now that the height problem is fixed?

rpokorny:
> I don't believe there is a specific ticket for close buttons, is it worth updating the snapshots here for dark mode now that the height problem is fixed?

Fine with me: 1e62e629b

EmilyBermudez88 (approved):
LGTM!

PR #1063, opened 2023-04-12
Dark Mode - NxPagination - RSC-1033

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1033

This PR also fixes a bug where the selected button on the NxPagination didn't have semibold text and also applies the correct color to the carets on the tertiary buttons.


lib/src/base-styles/_nx-pagination.scss
```diff
@@ -66,21 +71,22 @@
...
     outline: none;
   }
 
   &.selected {
     &, &:hover {
-      background-color: var(--nx-swatch-blue-40);
+      background-color: var(--nx-color-primary-button-background);
       border-color: transparent;
-      color: var(--nx-color-component-background);
+      color: var(--nx-color-pagination-focus-selected);
```
rpokorny, line 83: So these have the background of a primary button but a different text color?

lib/src/base-styles/_nx-pagination.scss
```diff
@@ -66,21 +71,22 @@
...
       cursor: default;
+      font-weight: 600;
     }
-    
+
     &:focus {
-      outline: 2px solid var(--nx-color-component-background);
+      outline: 2px solid var(--nx-color-pagination-focus-selected);
       outline-offset: -4px;
```
rpokorny, line 90:
```suggestion
      outline-offset: calc(0px - var(--nx-focus-outline-width) - var(--nx-focus-outline-offset));
```

lib/src/base-styles/_nx-pagination.scss
```diff
@@ -66,21 +71,22 @@
...
-      color: var(--nx-color-component-background);
+      color: var(--nx-color-pagination-focus-selected);
       cursor: default;
+      font-weight: 600;
     }
-    
+
     &:focus {
-      outline: 2px solid var(--nx-color-component-background);
+      outline: 2px solid var(--nx-color-pagination-focus-selected);
```
rpokorny, line 89: Also surprised that this doesn't just follow primary button styling

lib/src/base-styles/_nx-pagination.scss
avnvbhatta: > 

It doesn't look like the `--nx-focus-outline-offset` variable is defined anywhere. However, `NxCheckbox.scss` seems to use it ❓

lib/src/base-styles/_nx-pagination.scss
avnvbhatta: > So these have the background of a primary button but a different text color?

Talked with Jason and now they're following the primary button styles. Check c554154f716c300a3a9efce306a4a27379ed355d

lib/src/base-styles/_nx-pagination.scss
avnvbhatta: > Also surprised that this doesn't just follow primary button styling

Mockups have been updated to use primary button styling. Check c554154f716c300a3a9efce306a4a27379ed355d

lib/src/base-styles/_nx-pagination.scss
```diff
@@ -66,21 +71,22 @@
...
-      color: var(--nx-color-component-background);
+      color: var(--nx-color-primary-button-text);
       cursor: default;
+      font-weight: 600;
     }
-    
+
     &:focus {
-      outline: 2px solid var(--nx-color-component-background);
+      outline: 2px solid var(--nx-color-primary-button-focus-outline);
```
EmilyBermudez88, line 89:
```suggestion
      outline: var(--nx-focus-outline-width) solid var(--nx-color-primary-button-focus-outline);
```

EmilyBermudez88:
A ticket coming up RSC-1554 (https://issues.sonatype.org/browse/RSC-1554) is about replacing some of our hard coded 2px values to variables created in the NxButton dark mode ticket. I think it's better to update some of that here during this ticket instead of coming back afterwards as clean up. ie. `%btn-focus-outline` could use `--nx-focus-outline-width` in the outline property, and the focus states could use both `--nx-focus-outline-width` and `--nx-focus-outline-inset`)

EmilyBermudez88:
Now that the selected button matches the primary styling, should we change it's classname to `-nx-btn--primary` to see what styling can be inherited?

avnvbhatta:
> A ticket coming up RSC-1554 (https://issues.sonatype.org/browse/RSC-1554) is about replacing some of our hard coded 2px values to variables created in the NxButton dark mode ticket. I think it's better to update some of that here during this ticket instead of coming back afterwards as clean up. ie. `%btn-focus-outline` could use `--nx-focus-outline-width` in the outline property, and the focus states could use both `--nx-focus-outline-width` and `--nx-focus-outline-inset`)

Changed to use variables here: 7f0e9bb48f08faa39ce2af380ab159197389eca5

lib/src/base-styles/_nx-pagination.scss
EmilyBermudez88: > > > > 
> > > 
> > > 
> > > It doesn't look like the `--nx-focus-outline-offset` variable is defined anywhere. However, `NxCheckbox.scss` seems to use it ❓
> > 
> > 
> > Oooh that's interesting re: NxCheckbox. So I think what Ross is referring to is we made variables during the NxButton ticket to standardize some of the consistent outline pixel values. However, the variable is called `--nx-focus-outline-inset` as opposed to `---nx-focus-outline-offset`. It seemed to make sense to me at the time (since the outline is technically inside not outside), but maybe not the best variable name. That renaming however could be tackled in RSC-1554 (https://issues.sonatype.org/browse/RSC-1554)
> 
> Changed to use the `--nx-focus-outline-inset` variable here: [7f0e9bb](https://github.com/sonatype/sonatype-react-shared-components/commit/7f0e9bb48f08faa39ce2af380ab159197389eca5)

Do you want to make a ticket to fix the problem in `NxCheckbox.scss`? I just tried updating it to the correct variable, which completely threw off the focus border, so that will need to be fixed.
Screenshot 2023-04-19 at 12 59 57 PM

EmilyBermudez88 (approved):
Add in the dark mode snapshots for indeterminate pagination, & see comment https://github.com/sonatype/sonatype-react-shared-components/pull/1063#discussion_r1171621473 😄 . Otherwise LGTM!

lib/src/base-styles/_nx-pagination.scss
avnvbhatta: > -nx-focus-outline-inset

Ticket here: https://issues.sonatype.org/browse/RSC-1649

PR #1066, opened 2023-04-13
Dark Mode - NxFilterInput clear button hover - RSC-1563

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1563


lib/src/components/NxFilterInput/NxFilterInput.scss
```diff
@@ -38,7 +38,7 @@
     @include nx-button-helpers.small-icon-btn;
 
     margin: 0;
-    padding: calc(var(--nx-spacing-base) - 2px) 0;
+    padding: calc(var(--nx-spacing-base) - var(--nx-padding-offset-filter-clear-button)) 0;
```
achen5, line 41: This works, but do you mind explaining the purpose of this calculation? Perhaps it's simpler to do the following?
```
--nx-padding-filter-clear-button: 2px;

    --nx-padding-filter-clear-button: 1px;
}
```
```
padding: var(--nx-padding-filter-clear-button) 0;
```

gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -236,6 +236,10 @@ const CssVariablesPage = () => {
           <PropertyDocItem propertyVar="--nx-border-grid">
             The standard border style for <NxCode>nx-grid</NxCode> vertical borders.
           </PropertyDocItem>
+          <PropertyDocItem propertyVar="--nx-padding-offset-filter-clear-button">
+            The padding offset used to calculate the correct height of the clear button
+            in <NxCode>NxFilterInput</NxCode>.
+          </PropertyDocItem>
```
achen5, line 242: image
Looks like the gallery is having issues rendering this row. You may need to widen the first column.

lib/src/components/NxFilterInput/NxFilterInput.scss
rpokorny, line 41: I don't think we need a new variable here.  We just need to take one additional pixel off of the top and bottom paddings produced by `nx-button-helpers.small-icon-btn`.  How about giving that mixin a parameter called `extra-vertical-padding` or something, and then passing `-1px` for it?

avnvbhatta: > I don't think we need a new variable here. We just need to take one additional pixel off of the top and bottom paddings produced by `nx-button-helpers.small-icon-btn`. How about giving that mixin a parameter called `extra-vertical-padding` or something, and then passing `-1px` for it?

Thank you, that's a great idea. Changed here: ab33a4d82a32620ec803fe44ebbea22d815d66ca and e7eeb6339c172b9ffa6ae94b9fb9cc35f2058bcb

gallery/src/styles/CssVariables/CssVariablesPage.tsx
avnvbhatta: > image Looks like the gallery did not render this row correctly. You may need to widen the first column.

Removed: e7eeb6339c172b9ffa6ae94b9fb9cc35f2058bcb

PR #1065, opened 2023-04-12
update incomplete snapshots dark mode - RSC-1568

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1568

I left out NxFilterInput (snapshots involving the clear btn), since I expect they will be dealt with in RSC-1563(https://issues.sonatype.org/browse/RSC-1563).

I have added comments into Jira tickets of components that still need to be completed, but are included in snapshots of components that are already done or have no ticket (ie. NxFilterDropdown snapshots that need to be updated in NxTable, or NxFieldset snapshots that need to be updated when NxTransferList is complete)

I did discover that `--nx-color-text-placeholder` is incorrect in dark mode. It's set to `--nx-color-indigo-70` instead of `--nx-swatch-indigo-70`, so will need to be updated, which will update snapshots. I can add that to here as part of general clean up, or I can make a separate bug ticket.

PR #1067, opened 2023-04-14
Combobox Dark Mode - RSC-1518

rpokorny:
https://issues.sonatype.org/browse/RSC-1518

This PR is dependent on https://github.com/sonatype/sonatype-react-shared-components/pull/1046 which defines the dark mode value for the selected background color.


gallery/visualtests/__image_snapshots_dark__/nx-combobox-js-nx-combobox-when-disabled-looks-right-1-dark.png
rpokorny, line 1: Note: I believe this snapshot is wrong, but in a way consistent with all of our other disabled text inputs.  See https://github.com/sonatype/sonatype-react-shared-components/pull/1065#discussion_r1166796178 .  I expect that to get fixed in that PR, which this PR is thus dependent on.

lib/src/components/NxCombobox/NxCombobox.scss
```diff
@@ -60,7 +60,7 @@
   }
 
   .nx-dropdown-button.selected {
-    background-color: var(--nx-swatch-blue-90);
+    background-color: var(--nx-color-interactive-background-selected);
     color: var(--nx-color-text-dark);
```
EmilyBermudez88, line 64:
```suggestion
    color: var(--nx-color-text-stark);
```

EmilyBermudez88, line 64: In the mocks the 'selected' text color is `indigo-80`. If the mocks are wrong, then this should be `--nx-color-text-stark`

EmilyBermudez88: Actually, I suspect the mocks are wrong. All the interaction states show the button to have a text color of `indigo-80` but in NxDropdown's sketch file, hover & click both have the text color set to `indigo-90`

gallery/visualtests/__image_snapshots_dark__/nx-combobox-js-nx-combobox-when-disabled-looks-right-1-dark.png
EmilyBermudez88: > Note: I believe this snapshot is wrong, but in a way consistent with all of our other disabled text inputs. See [#1065 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/1065#discussion_r1166796178) . I expect that to get fixed in that PR, which this PR is thus dependent on.

I think this might be wrong because the value is set to 'foo' instead of a placeholder set to 'foo', so the disabled placeholder text isn't kicking in

lib/src/components/NxCombobox/NxCombobox.scss
JasonDSwearingen: I have corrected the mockups. Placeholder text should be indigo-70, and Entry text should be indigo-80. Disabled text should be the same color regardless of what the copy is.

gallery/visualtests/__image_snapshots_dark__/nx-combobox-js-nx-combobox-when-disabled-looks-right-1-dark.png
rpokorny: > I think this snapshot might be wrong because the value is set to 'foo' instead of a placeholder set to 'foo', so the disabled placeholder text isn't kicking in. I figure we add the disabled color definition inside of the .disabled .nx-text-input__input and that should fix everything.

At any rate, should be fixed in ec0c026f1

lib/src/components/NxCombobox/NxCombobox.scss
rpokorny: > I have corrected the mockups. Placeholder text should be indigo-70, and Entry text should be indigo-80. Disabled text should be the same color regardless of what the copy is.

I assume that is actually in response to the other thread on this PR.  But @JasonDSwearingen, we do need your input on the issue @EmilyBermudez88 described above: what color is the text supposed to be on a selected dropdown menu item (specifically in combobox though we assume it should be the same everywhere)?

rpokorny: > In the mocks the 'selected' text color is indigo-80. If the mocks are wrong, then this should be --nx-color-text-stark

For now, changed to use `...-stark`. 5adb2f508

EmilyBermudez88: > > In the mocks the 'selected' text color is indigo-80. If the mocks are wrong, then this should be --nx-color-text-stark
> 
> For now, changed to use `...-stark`. [5adb2f5](https://github.com/sonatype/sonatype-react-shared-components/commit/5adb2f50842eaf298b06f7a208e7ba6e205c63a7)

Looks good. I just checked the mocks and they now have the text color set to `indigo-90` when selected

PR #1071, opened 2023-04-19
Update threat colours- RSC-1576

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1576

I also changed the 'none' and 'low' threat colours as mentioned in the comments of the related ticket https://issues.sonatype.org/browse/RSC-1048 , since it will affect both light and dark mode


gallery/src/components/NxSmallThreatCounter/NxSmallThreatCounterInfiniteMaxDigitsExample.tsx
```diff
@@ -15,15 +15,15 @@ export default function NxSmallThreatCounterInfiniteMaxDigitsExample() {
...
+                              lowCount={0}
+                              noneCount={2353}
                               unspecifiedCount={1} />
       </div>
       <div style={{ marginTop: '16px' }}>
         <NxSmallThreatCounter maxDigits={Infinity}
                               criticalCount={2}
                               severeCount={50}
-                              noneCount={0}
+                              noneCount={23}
```
cindyko226, line 26: Out of curiosity, it is necessary to change the number here?

EmilyBermudez88: > Out of curiosity, it is necessary to change the number here?

Yeah the `zero` state was overriding the text color of the none count, so I changed the number to avoid any problems

cindyko226: > the zero state was overriding the text color of the none count

If I remember correctly in this use case, shouldn't the threat color remain the same even if there is a value of 0?

EmilyBermudez88: > > the zero state was overriding the text color of the none count
> 
> If I remember correctly in this use case, shouldn't the threat color remain the same even if there is a value of 0?

I totally missed that initially, thanks for catching it!  Updated [db368ae](https://github.com/sonatype/sonatype-react-shared-components/pull/1071/commits/db368ae238807d20b1332a67ada66dccc0e0135b), [062ca2d](https://github.com/sonatype/sonatype-react-shared-components/pull/1071/commits/062ca2d226c41a55fa754226e9282e32348749f6)

cindyko226:
This might be a bug in main, there is discrepancy of severe threat color between mockups and gallery.
The threat color of severe is set to `nx-indigo-55` in mockups but `nx-indigo-50` is implemented in the component. I think we should check with design.

entroform (approved):
LGTM!

PR #1076, opened 2023-04-20
Dark Mode - NxTag - RSC-1045

avnvbhatta:
https://issues.sonatype.org/browse/RSC-1045


lib/src/components/NxTag/NxTag.scss
```diff
@@ -7,17 +7,18 @@
 @use '../../scss-shared/nx-text-helpers';
 @use '../../scss-shared/nx-container-helpers';
 @use '../../scss-shared/nx-hide-checkbox-helper';
+@use '../../scss-shared/nx-dark-mode-helpers';
```
cindyko226, line 10: Do we need this?

lib/src/base-styles/_nx-colors.scss
```diff
@@ -488,6 +572,90 @@
...
+    --nx-color-selectable-color-kiwi-dark: var(--nx-swatch-green-80);
+    --nx-color-selectable-color-kiwi-light: var(--nx-swatch-green-95);
+
+    --nx-color-selectable-color-indigo-dark: var(--nx-swatch-indigo-70);
+    --nx-color-selectable-color-indigo-light: var(--nx-swatch-indigo-95);
+
+    // NxTag
+    --nx-color-tag-text: var(--nx-swatch-indigo-10);
+
+    --nx-color-tag-sky-selected-hover-border: var(--nx-color-tag-sky-selected-hover-background);
```
rpokorny, line 610: I'm not sure these variables are needed.  I think it would be ok to use the swatches directly in `_nx-tag-variables.scss`, including a `dark-mode` mixin block within each CSS block there.

lib/src/scss-shared/_nx-tag-variables.scss
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.nx-tag {
```
rpokorny, line 7: I wouldn't thing this wrapper would really be necessary

lib/src/util/selectableColors.scss
```diff
@@ -6,53 +6,53 @@
  */
 .nx-selectable-color--light-blue, // deprecated
 .nx-selectable-color--sky {
-  --nx-selectable-color-dark: var(--nx-swatch-blue-80);
-  --nx-selectable-color-light: var(--nx-swatch-blue-90);
+  --nx-selectable-color-dark: var(--nx-color-selectable-color-sky-dark);
+  --nx-selectable-color-light: var(--nx-color-selectable-color-sky-light);
```
rpokorny, line 10: Here too I think it would make sense to use swatches and `dark-mode` mixin calls directly rather than an additional layer of variables.

lib/src/scss-shared/_nx-tag-variables.scss
avnvbhatta: > I wouldn't thing this wrapper would really be necessary

f9c3f0d866d5c4b14e813c354451b5bbf031765e

lib/src/util/selectableColors.scss
avnvbhatta: > Here too I think it would make sense to use swatches and `dark-mode` mixin calls directly rather than an additional layer of variables.

2482cf1c45a68dee608e717600d57e33f811cd06

lib/src/components/NxTag/NxTag.scss
```diff
@@ -7,17 +7,18 @@
...
+@use '../../scss-shared/nx-tag-variables';
 
 .nx-tag {
   @include nx-text-helpers.semi-bold;
   @include nx-container-helpers.container-horizontal;
 
   align-items: center;
   background-color: var(--nx-selectable-color-dark);
-  border: 1px solid var(--nx-selectable-color-dark);
+  border: var(--nx-border-width) solid var(--nx-selectable-color-dark);
```
rpokorny, line 18: I believe the padding below needs to also use this variable rather than `1px`

lib/src/components/NxTag/NxTag.scss
```diff
@@ -37,22 +38,28 @@
   background-color: var(--nx-selectable-color-light);
```
rpokorny, line 38: Is the padding-right on the line above this also supposed to have the border subtracted?

lib/src/scss-shared/_nx-tag-variables.scss
```diff
@@ -0,0 +1,151 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@use '../scss-shared/nx-dark-mode-helpers';
+
+.nx-selectable-color--light-blue, // deprecated
+.nx-selectable-color--sky {
+  --nx-tag-selected-hover-border: var(--nx-color-interactive-border-hover);
```
rpokorny, line 11: Should these all be `--nx-color-tag-...`?

lib/src/components/NxTag/NxTag.scss
```diff
@@ -37,22 +38,28 @@
...
+    background-color: var(--nx-tag-unselected-hover-background);
   }
 
   &:focus-within {
     &::before {
       border: 2px solid var(--nx-color-interactive-border-focus);
       border-radius: 20px;
       content: '';
-      inset: calc(0px - (var(--nx-spacing-base) + 1px));
+      inset: calc(0px - (var(--nx-spacing-base) + var(--nx-border-width)));
```
rpokorny, line 50: I'm not sure this calculation makes sense.  The variables being used here don't really seem related to the value we're trying to compute, and I'm pretty sure it ends up with the wrong value anyway in light mode.  As far as I can tell it should be 4px, but is 6px in dark mode and 5px in light mode.  I believe the correct variables to use for this calculation would be `--nx-focus-outline-width` and `--nx-focus-outline-inset` (soon to be renamed)

lib/src/components/NxTag/NxTag.scss
```diff
@@ -37,22 +38,28 @@
...
 
   &:hover {
-    border-color: var(--nx-color-interactive-border-hover);
+    border-color: var(--nx-tag-unselected-hover-border);
+    background-color: var(--nx-tag-unselected-hover-background);
   }
 
   &:focus-within {
     &::before {
       border: 2px solid var(--nx-color-interactive-border-focus);
```
rpokorny, line 47:
```suggestion
      border: var(--nx-focus-outline-width) solid var(--nx-color-interactive-border-focus);
```

lib/src/scss-shared/_nx-tag-variables.scss
```diff
@@ -0,0 +1,151 @@
...
+  --nx-tag-selected-hover-background: var(--nx-selectable-color-dark);
+  --nx-tag-unselected-hover-border: var(--nx-color-interactive-border-hover);
+  --nx-tag-unselected-hover-background: var(--nx-selectable-color-light);
```
rpokorny, line 14: The light mode uses the same values for each one? There are probably ways to cut out that repetition, either by putting that into a `:root` block or into a placeholder selector that each of these blocks `@extends`.

rpokorny: > The light mode uses the same values for each one? There are probably ways to cut out that repetition, either by putting that into a :root block or into a placeholder selector that each of these blocks @extends.

Looks like the dark mode hover borders are also always the same.  So lets move all of the common ones into a `:root` block.  Maybe even back into `_nx-colors.scss`.

lib/src/components/NxTag/NxTag.scss
avnvbhatta: > I'm not sure this calculation makes sense. The variables being used here don't really seem related to the value we're trying to compute, and I'm pretty sure it ends up with the wrong value anyway. As far as I can tell it should be 4px, but is 6px in dark mode and 5px in light mode. I believe the correct variables to use for this calculation would be `--nx-focus-outline-width` and `--nx-focus-outline-inset` (soon to be renamed)

In dark-mode, setting the inset to 4px would bring the height of the focus-outline to be 36px. The mock shows it to be 40px (inset of 6px would be correct). 

In light mode, setting the inset to 4px would bring the height of the focus-outline to be 38px. The mock shows it to be 40px (inset of 5px would be correct).

avnvbhatta: > > In dark-mode, setting the inset to 4px would bring the height of the focus-outline to be 36px. The mock shows it to be 40px (inset of 6px would be correct).
> 
> How do you figure? `inset` is a shorthand for `top`, `bottom`, `left`, and `right`. So `inset: 4px` means `top: 4px; bottom: 4px;` (and left and right). The `.nx-tag` is 32px tall. The focus border should be 40px tall. 40 - 32 = 8. 8 / 2 = 4. So I believe `top: 4px; bottom: 4px` is correct.

I could be wrong, but I think what's happening is that currently, the 32px height of the`.nx-tag` is due to 1px border on the top and bottom. An inset of 0px does not account for the 1px border, and looks like this:
image

In order to correctly offset the 1px border, the inset needs to have an additional 1px, hence 4px+1px = 5px in light mode, and 4px+2px = 6px in dark mode (since dark mode has a 2px border instead of one).

lib/src/components/NxTag/NxTag.scss
avnvbhatta: > I believe the padding below needs to also use this variable rather than `1px`

Changed here: 53af6b3256b4a0d0845640be2d5b8b94cbbb6832

lib/src/components/NxTag/NxTag.scss
avnvbhatta: > Is the padding-right on the line above this also supposed to have the border subtracted?

9c73156f076493c0879e3b303f4d4457150ab4c4

lib/src/scss-shared/_nx-tag-variables.scss
avnvbhatta: > Should these all be `--nx-color-tag-...`?

Yup. 3899f851770c116ed1c1c254ce3e000215ebb3aa

lib/src/scss-shared/_nx-tag-variables.scss
avnvbhatta: > > The light mode uses the same values for each one? There are probably ways to cut out that repetition, either by putting that into a :root block or into a placeholder selector that each of these blocks @extends.
> 
> Looks like the dark mode hover borders are also always the same. So lets move all of the common ones into a `:root` block. Maybe even back into `_nx-colors.scss`.

Went with this approach: 9f8ead04a361a509d5708121b3c0d310de113a37

lib/src/components/NxTag/NxTag.scss
```diff
@@ -33,26 +34,32 @@
 
 .nx-tag--selectable {
   cursor: pointer;
-  padding-right: var(--nx-spacing-2x);
+  padding-right: calc(var(--nx-spacing-2x) - calc(var(--nx-border-width) - 1px));
```
rpokorny, line 37: Why the `-1px`?

lib/src/scss-shared/_nx-tag-variables.scss
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@use '../scss-shared/nx-dark-mode-helpers';
+
+%nx-tag-hover-style {
```
rpokorny, line 9: Given that this group of variables do always have the same values, I am thinking putting them in `_nx-colors.scss` really is the best approach

rpokorny:
Don't forget to document the CSS variables.  IMO even the ones that aren't defined in `:root`, such as `--nx-color-tag-selected-hover-border`, should be documented but including a note that they won't have a value outside of `.nx-selectable-color--...` elements.

lib/src/components/NxTag/NxTag.scss
avnvbhatta: > Why the `-1px`?

To account for dark-mode borders being 2px compared to 1px on light-mode.

lib/src/scss-shared/_nx-tag-variables.scss
rpokorny: > Hmm for some reason, the variables --nx-selectable-color-dark and --nx-selectable-color-light are undefined when I assign them in _nx-colors.scss thinking

Ok, it appears that CSS variables values are intepreted in the scope of the selector where they were defined, not where they are used.  That's something we should keep in mind for the future.

rpokorny:
> Don't forget to document the CSS variables. IMO even the ones that aren't defined in :root, such as --nx-color-tag-selected-hover-border, should be documented but including a note that they won't have a value outside of .nx-selectable-color--... elements.

In light of https://github.com/sonatype/sonatype-react-shared-components/pull/1076#discussion_r1181547294 , it no longer seems like it would make sense to document `--nx-color-tag-…hover-…` in the global CSS variable documentation.  How do you think we should document them? Maybe on the Selectable Colors page?

lib/src/components/NxTag/NxTag.scss
rpokorny: > To account for dark-mode borders being 2px compared to 1px on light-mode.

I'm afraid I don't follow; could you elaborate? In sketch, I see the following:
image
image

That is, 8px from the edge of the icon to the _outside_ of the tag border.  With your implementation, I'm seeing
image
in dark mode and
image
 in light mode, so 9px from the edge of the icond to the outside of the tag border.

lib/src/components/NxTag/NxTag.scss
avnvbhatta: > > An inset of 0px does not account for the 1px border, and looks like this:
> 
> Ah that's what I was forgetting that `inset` (and `top` etc) count from the parent's padding box rather than border box. In that case your numbers are correct but your use of `var(--nx-spacing-base)` still seems not ideal. I think that should be replaced by `var(--nx-focus-outline-width) + var(--nx-focus-outline-inset)`.

Changed here: b275f34f3c3528b3ab3b613520977aebc9bd4c38

lib/src/components/NxTag/NxTag.scss
avnvbhatta: > > To account for dark-mode borders being 2px compared to 1px on light-mode.
> 
> I'm afraid I don't follow; could you elaborate? In sketch, I see the following: 
> 
> That is, 8px from the edge of the icon to the _outside_ of the tag border. With your implementation, I'm seeing ! in light mode, so 9px from the edge of the icond to the outside of the tag border.

Oh yes, you're right, I've changed it here and updated both light mode and dark mode snapshots.
142c4fb6e649203f55b4538522f9e556bd38798a

avnvbhatta:
> 



> > Don't forget to document the CSS variables. IMO even the ones that aren't defined in :root, such as --nx-color-tag-selected-hover-border, should be documented but including a note that they won't have a value outside of .nx-selectable-color--... elements.
> 
> In light of [#1076 (comment)](https://github.com/sonatype/sonatype-react-shared-components/pull/1076#discussion_r1181547294) , it no longer seems like it would make sense to document `--nx-color-tag-…hover-…` in the global CSS variable documentation. How do you think we should document them? Maybe on the Selectable Colors page?

That's a good idea. Does this make sense? FYI, I didn't add any color samples as I don't think it's necessary.
7437097cc6587e5b67fda5bc62c1009da8453d17

rpokorny:
> That's a good idea. Does this make sense? FYI, I didn't add any color samples as I don't think it's necessary.
https://github.com/sonatype/sonatype-react-shared-components/commit/7437097cc6587e5b67fda5bc62c1009da8453d17

Yeah; just add a little paragraph in there to indicate that these are only defined within `.nx-selectable-color--…` CSS classes

rpokorny (approved):

PR #1077, opened 2023-04-21
Dark Mode: Remove Unset Variable : NxCheckbox - RSC-1649 

entroform:
https://issues.sonatype.org/browse/RSC-1649


EmilyBermudez88:
I think you should be able to get rid of the `position: relative` set on `-nx-checkbox__box` on line 20 as well 😄

EmilyBermudez88 (approved):
other than the comment above, LGTM!

entroform:
> I think you should be able to get rid of the `position: relative` set on `-nx-checkbox__box` on line 20 as well 😄

Thanks! 84e42d3b3ee212082357e61fb6e42c379e1a36cd

PR #1078, opened 2023-04-21
update logo height in NxGlobalSidebar- RSC-1557

EmilyBermudez88:
As per the discussion in WW today, this is a temporary stop gap to accommodate the Lifecycle logo before the visual refresh. We had the logo height set to a standard height of 34px, so I kept that as the min-height to try and keep things as consistent as possible.

I ran the full suite of visual tests, and no failures.

https://issues.sonatype.org/browse/RSC-1557


lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -10,7 +10,8 @@
 
 $-width-open: 250px;
 $-width-closed: 72px;
-$-logo-height: 34px;
+$-max-logo-height: 43px;
```
rpokorny, line 13: Where does this number come from?

rpokorny:
What exactly does this PR do? It might be easier to see if examples with logos in newer sizes are added.  That would also allow us to visual-test that it is working correctly.

rpokorny:
> I did ask Jason for the new logos to test with to make sure they fit correctly, but didn't include them here. I can send you the link to the google drive folder to take a look to make things easier

Let's not add the real product logos in this repo, but we should have dummy logos (larger Plaid Villains?) to construct visual tests for the new size with.

rpokorny:
The logo size rules in the documentation should be updated as well

EmilyBermudez88:
> The logo size rules in the documentation should be updated as well

updated here [bacb015](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/bacb015436795d7397a3f4b32d39eacb9fbb212d)

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.tsx
```diff
@@ -55,7 +56,7 @@ const NxGlobalSidebar: FunctionComponent<Props> = function NxGlobalSidebar(props
   return (
     <div className={classes} id={id}>
       <aside aria-label="global sidebar">
-        <div className="nx-global-sidebar__header">
+        <div className="nx-global-sidebar__header" ref={headerRef} style={{height: headerRef?.current?.clientHeight}}>
```
EmilyBermudez88, line 59: This was the solution I came up with to keep the header height the same whether the sidebar is open or closed to make sure the nav items don't shift up when closed. However, the height won't update if the sidebar defaults to closed and then opens with a logo image taller than 34px. Are there scenarios in product where the sidebar is closed by default?

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -102,8 +102,8 @@ export default function NxGlobalSidebarPage() {
                 <NxTable.Cell>
                   The path to an image placed in the top left corner which displays branding and product name.
                   <NxWarningAlert>
-                    Note: the image should not be larger than 145px wide and 32px tall. It should
-                    display either a logo (apx 32px x 32px) and the product name, or just the product name.
+                    Note: the image should not be larger than 155px wide and 34px to 43px tall. It should
```
rpokorny, line 105: 34? The previous max was 32 so I'd think we'd want our min to be no larger than 32.  But further, why do we even need a min? We didn't before.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -291,7 +291,7 @@ export default function NxGlobalSidebarPage() {
         <NxTile.Content>
           <NxP>
             This examples shows 10 nav items (the maximum), a middle content area with enough content to trigger
```
rpokorny, line 293:
```suggestion
            This example shows 10 nav items (the maximum), a middle content area with enough content to trigger
```

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -291,7 +291,7 @@ export default function NxGlobalSidebarPage() {
...
-            scrolling, and all footer options enabled.
+            scrolling, all footer options enabled.
```
rpokorny, line 294: Why'd you get rid of "and" here?

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -86,7 +88,10 @@ $-separator-color: var(--nx-swatch-indigo-40);
 }
 
 .nx-global-sidebar__logo {
-  height: $-logo-height;
+  // a temporary stop gap to ensure no logos break the header container - matches plaid-villain logo width
+  max-width: 155px;
```
rpokorny, line 92: I guess we should really have another logo and visual test to check this case

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.tsx
rpokorny: > This was the solution I came up with to keep the header height the same whether the sidebar is open or closed to make sure the nav items don't shift up when closed. However, the height won't update if the sidebar defaults to closed and then opens with a logo image taller than 34px. Are there scenarios in product where the sidebar is closed by default?

It should be possible to set the width of the `.nx-global-sidebar__product-info` to 0 when collapsed, rather than using `display: none`.  That way its height will still be part of the layout.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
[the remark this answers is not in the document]
EmilyBermudez88: > Why 155? I'm not seeing that in sketch. The comment above makes it sound like it's just to match the plaid-villain logo, which doesn't seem like a good basis. Given that the parent is a flex container I bet we can avoid specifying a `max-width` here and just let it shrink to fit the available space. Using that approach, we'd need design to clarify the min space between the logo and the collapse button.

Thanks for this idea. You're right, I used 155px because that was the current width of the plaid-logo to avoid breaking visual tests, which isn't a good enough reason. I'm going to reach out to Jason about the min space between logo & toggle which would resolve the problem of needing some sort of width restriction to prevent all images from increasing to 43px high, and then breaking the width of the header container.

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.tsx
EmilyBermudez88: > > This was the solution I came up with to keep the header height the same whether the sidebar is open or closed to make sure the nav items don't shift up when closed. However, the height won't update if the sidebar defaults to closed and then opens with a logo image taller than 34px. Are there scenarios in product where the sidebar is closed by default?
> 
> It should be possible to set the width of the `.nx-global-sidebar__product-info` to 0 when collapsed, rather than using `display: none`. That way its height will still be part of the layout.

I initially tried changing the width, and found an interesting problem that I couldn't figure out why it was happening. Setting the width to 0px works for the example with the 43px tall logo, but in the 34px tall logo, `.nx-global-sidebar__product-info` increased to 40px tall (while the logo img remained at 34px tall)

rpokorny: > I initially tried changing the width, and found an interesting problem that I couldn't figure out why it was happening. Setting the width to 0px works for the example with the 43px tall logo, but in the 34px tall logo, .nx-global-sidebar__product-info increased to 40px tall (while the logo img remained at 34px tall)

Hmm that's not happening for me.  See https://github.com/sonatype/sonatype-react-shared-components/pull/1090

EmilyBermudez88: > > I initially tried changing the width, and found an interesting problem that I couldn't figure out why it was happening. Setting the width to 0px works for the example with the 43px tall logo, but in the 34px tall logo, .nx-global-sidebar__product-info increased to 40px tall (while the logo img remained at 34px tall)
> 
> Hmm that's not happening for me. See #1090

Merged #1090 here [c98e395](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/c98e395d3a815eb269a2bd777b4c1bea37226d26). Looking through it I found where my initial CSS mistake was when overriding the `display:none` set on `.nx-global-sidebar__expanded-content`. Thanks for the fix!

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
EmilyBermudez88: > 34? The previous max was 32 so I'd think we'd want our min to be no larger than 32. But further, why do we even need a min? We didn't before.

Confirmed with Jason, and no minimum height is needed. Removed here: [9408439](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/940843939654746264321c4a310af4e69db68282). I will update the documentation once they clarify the padding between the logo & button

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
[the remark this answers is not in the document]
EmilyBermudez88: > Given that the parent is a flex container I bet we can avoid specifying a `max-width` here and just let it shrink to fit the available space. Using that approach, we'd need design to clarify the min space between the logo and the collapse button.

So I discussed this with Jason and they decided on a 12px padding between the logo & collapse button. Implementing that, the original plaid-logo expands to 43px high, which clearly isn't the desirable effect. 
Screenshot 2023-04-27 at 10 48 23 AM
If I set a max-width on the image to 100%, this cleans up the problem when the sidebar is open, but when closed, shrinks the image down to 0px x 0px, and therefore shrinks the height of the header. What are your thoughts on setting a max-width as a calculation of available space instead of a percent? Something like 
`calc($-width-open - calc(2 * $-left-right-spacing) - var(--nx-spacing-3x) - 32px)` ?  
(where var(--nx-spacing-3x) refers to padding between logo & button, and 32px the width of the button?)

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
EmilyBermudez88: > Why'd you get rid of "and" here?

Definitely an accident when I was thinking about whether I should add a comment in about the logo size here. [d737a1d](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/d737a1dd6adfe253aff5e9435779531dddff5ac9)

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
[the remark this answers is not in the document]
rpokorny: > Something like
calc($-width-open - calc(2 * $-left-right-spacing) - var(--nx-spacing-3x) - 32px) ?

Unfortunate, but that's probably what we'll have to do.  Initially I was going to say use `max-width:100%` but then change that `max-width` to `unset` when the sidebar is closed, but that doesn't help because then the logo returns to its natural size on collapse and pushes the other content _down_.

I also started to play around with getting the header to overflow off the left side in such a way that it hides the logo when collapsed, but ultimately, because the collapse/expand button also actually overflows the content box, and we need to keep that fully visible, it'd still involve using the width of the button as a magic number making it no better than your solution.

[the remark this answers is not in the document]
EmilyBermudez88: > Unfortunate, but that's probably what we'll have to do. Initially I was going to say use `max-width:100%` but then change that `max-width` to `unset` when the sidebar is closed, but that doesn't help because then the logo returns to its natural size on collapse and pushes the other content _down_.
> 
> I also started to play around with getting the header to overflow off the left side in such a way that it hides the logo when collapsed, but ultimately, because the collapse/expand button also actually overflows the content box, and we need to keep that fully visible, it'd still involve using the width of the button as a magic number making it no better than your solution.

I also thought `max-width: unset` would do it, and was disappointed to see that wasn't the case. Not ideal, but updated the max width [092b7eb](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/092b7eb829c85adaef9fae63f80854403b8f976b)

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
```diff
@@ -79,14 +79,18 @@ $-separator-color: var(--nx-swatch-indigo-40);
 .nx-global-sidebar__product-info {
   flex: 1 1 auto;
   display: flex;
+  overflow: hidden;
+  padding-right: var(--nx-spacing-3x);
```
rpokorny, line 83: I think it would be cleaner to use `margin` or `gap` for this spacing.  We try to avoid using padding except for things that are explicitly within the border box of the element in question.

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
EmilyBermudez88: > > 34? The previous max was 32 so I'd think we'd want our min to be no larger than 32. But further, why do we even need a min? We didn't before.
> 
> Confirmed with Jason, and no minimum height is needed. Removed here: [9408439](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/940843939654746264321c4a310af4e69db68282). I will update the documentation once they clarify the padding between the logo & button

Updated to remove min-width, and calculated the available space, which is 158px wide: [00198e7](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/00198e73dd714936edc3eab073ca5ca4ddc4f47c)

lib/src/components/NxGlobalSidebar/NxGlobalSidebar.scss
EmilyBermudez88: > I guess we should really have another logo and visual test to check this case

The regular snapshots do mostly this already, since the Plaid Villain logo breaks the container without a max-width in place. However, I did add another snapshot with the different sized logo here: [39c4f95](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/39c4f9579385ab12e53f3c20e1a4b067d13ccb5e)

rpokorny: > The regular snapshots do mostly this already, since the Plaid Villain logo breaks the container without a max-width in place

Good point

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
```diff
@@ -102,8 +102,8 @@ export default function NxGlobalSidebarPage() {
                 <NxTable.Cell>
                   The path to an image placed in the top left corner which displays branding and product name.
                   <NxWarningAlert>
-                    Note: the image should not be larger than 145px wide and 32px tall. It should
-                    display either a logo (apx 32px x 32px) and the product name, or just the product name.
+                    Note: the image should not be larger than 158px wide and 43px tall. It should
+                    display either a logo (maximum 43px x 43px) and the product name, or just the product name.
                   </NxWarningAlert>
```
rpokorny, line 107: I'm wondering if this should be adjusted to no longer be a warning/instruction.  We now have what we need to support larger images and our code simply resizes them.  Additionally, these instructions are ambiguous for sizeless images such as SVGs.  So maybe replace this warning with something saying that the image will be rendered at a size no greater than 158x43.  I think we can also drop the bit about the 43x43 logo.

gallery/visualtests/nxGlobalSidebar.js
```diff
@@ -48,5 +48,16 @@ describe('nx-global-sidebar', function() {
...
 
+  describe('with a different logo size', function() {
+    const { getPage, checkFullPageScreenshot } = setupBrowser('#/NxGlobalSidebarScrollingExample', false);
+
+    it('looks right', async function() {
+      await getPage().setViewport({ width: 1366, height: 1000 });
+      await checkFullPageScreenshot();
+    });
+
+    it('passes a11y checks', a11yTest(null, true));
```
rpokorny, line 59: Probably not necessary

rpokorny (approved):

gallery/src/components/NxGlobalSidebar/NxGlobalSidebarPage.tsx
EmilyBermudez88: >I'm wondering if this should be adjusted to no longer be a warning/instruction. We now have what we need to support larger images and our code simply resizes them. Additionally, these instructions are ambiguous for sizeless images such as SVGs. So maybe replace this warning with something saying that the image will be rendered at a size no greater than 158x43. I think we can also drop the bit about the 43x43 logo.

Yep, that makes sense! [2ec3b0a](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/2ec3b0a1cc54504504e02aeb94151a7fa64c8755)

gallery/visualtests/nxGlobalSidebar.js
EmilyBermudez88: >Probably not necessary

Removed [2ec3b0a](https://github.com/sonatype/sonatype-react-shared-components/pull/1078/commits/2ec3b0a1cc54504504e02aeb94151a7fa64c8755)

PR #1087, opened 2023-04-25
Dark Mode: Fix NxToggle focus in Safari - RSC-1653 

entroform:
https://issues.sonatype.org/browse/RSC-1653


achen5 (approved):
LGTM, but a couple of things to note:

- I'm pretty sure that the initial comment has the wrong ticket listed: RSC-1657. It should be RSC-1653. 
- I'm not sure if a snapshot update was necessary. Both the before and after snapshots look exactly the same to me. But I assume this was done since one of the CSS properties was updated and did not pass visual testing for some reason. If you have another reason why the snapshots were updated, let me know.

rpokorny (approved):

PR #1088, opened 2023-04-25
Fix Description List Disabled State - RSC-1654

cindyko226:
https://issues.sonatype.org/browse/RSC-1654

This PR includes updating the light mode snapshots of description list as it is a bug in main as well.


rpokorny:
This seems like it should be fixed on a PR targeting `main`

cindyko226:
> This seems like it should be fixed on a PR targeting main

I was thinking this would also fix the bug in `main` once the feature branch is merged but it might not be the best way.
Since the dark mode snapshots would need to be updated, should I create a separate ticket to fix the bug in `main` and keep this PR only for dark mode?

rpokorny:
> I was thinking this would also fix the bug in main once the feature branch is merged but it might not be the best way.
Since the dark mode snapshots would need to be updated, should I create a separate ticket to fix the bug in main and keep this PR only for dark mode?

It'll need to be 2 PRs at least.  Up to you whether you also want to make separate JIRA tickets, and if so whether they are subtasks or unrelated.

rpokorny (approved):

PR #1097, opened 2023-04-28
NxDrawer dark mode - RSC-1520

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1520


gallery/src/styles/CssVariables/CssVariablesPage.tsx
```diff
@@ -222,6 +222,9 @@ const CssVariablesPage = () => {
           <PropertyDocItem propertyVar="--nx-box-shadow-dropdown">
             The standard box-shadow to apply to dropdown elements.
           </PropertyDocItem>
+          <PropertyDocItem propertyVar="--nx-box-shadow-drawer">
+            The standard box-shadow to apply to dropdown elements.
```
rpokorny, line 226: The variable name says "drawer" but the description says "dropdown"

lib/src/base-styles/_nx-variables.scss
```diff
@@ -65,9 +65,14 @@
                             0 1px 2px 0 rgba(91, 95, 141, 0.2),
                             0 1px 1px 0 rgba(124, 128, 179, 0.2);
 
+  // NxDrawer
+  --nx-box-shadow-drawer: -34px 0 50px -32px hsla(0, 0%, 50%, 0.5);
```
rpokorny, line 69: Do we want to parameterize this whole thing or just the color?  The positioning is pretty specific to `NxDrawer` if I'm not mistaken

lib/src/base-styles/_nx-variables.scss
```diff
@@ -65,9 +65,14 @@
...
+
   @include nx-dark-mode-helpers.dark-mode {
     --nx-border-grid: var(--nx-border-default);
     --nx-border-width: 2px;
+    // NxDrawer
+    --nx-box-shadow-drawer: -34px 0 50px -32px hsla(231, 41%, 10%, 0.75);
```
rpokorny, line 75:
```suggestion
    --nx-box-shadow-drawer: -34px 0 50px -32px hsla(var(--nx-swatch-indigo-hs), 10%, 0.75);
```
Could do the light mode one similarly, though it's less necessary with grey

rpokorny, line 75: I'm finding the mockups of the shadow a bit confusing but at any rate I'm not seeing indigo-10 involved anywhere.  Where did you get that from?

rpokorny, line 75: We should probably add a comment explaining why the box shadow parameters are the way they are:

The mockups indicate that the shadow should be as if it is on an element that is shorter than the drawer such that its top and bottom are offset inwards by 32px each.  On this hypothetical element, the mockups indicate the shadow should have a spread of zero.  We can emulate that, without needing the extra element itself, by setting the actual spread to -32px, thus bringing in size of the shadow inwards by that amount on all sides.  But of course, we don't want it to be brought inward on the left, so we cancel that effect out on that side by adding -32px to the shadow's x-offset.  The mockups specify a further x-offset of -2px, resulting in a total of -34px.  On the right side of course, it doesn't matter how the shadow ends up as the element is against the edge of the viewport.

The y-offset can be zero as the shadow should be vertically centered relative to the drawer.  Finally, the blur comes directly from the mockup where it is specified as 50px.

EmilyBermudez88: > I'm finding the mockups of the shadow a bit confusing but at any rate I'm not seeing indigo-10 involved anywhere. Where did you get that from?

When I clicked onto the "Shadows" color 808080 it comes up with the HSLA value for `indigo-10`. I can confirm with Jason and ask them to adjust the box-shadow color to the HSLA value
Screenshot 2023-05-01 at 9 43 27 AM

rpokorny: > When I clicked onto the "Shadows" color 808080 it comes up with the HSLA value for indigo-10. I can confirm with Jason and ask them to adjust the box-shadow color to the HSLA value

They created a variable named "808080" which seems like a hex code which would be grey, and set its value to indigo? That definitely seems like a mistake one way or another.

EmilyBermudez88: > We should probably add a comment explaining why the box shadow parameters are the way they are:

Great idea. Would you like me to use the full explanation here, or pare it down where possible?

EmilyBermudez88: >Could do the light mode one similarly, though it's less necessary with grey

Updated here, including the light mode (seems nicer to have the light/dark mode variables defined the same way) [7e25ef6](https://github.com/sonatype/sonatype-react-shared-components/pull/1097/commits/7e25ef66b14f6de8c1d93ad316da5864a951b0cb)

lib/src/base-styles/_nx-variables.scss
EmilyBermudez88: > Do we want to parameterize this whole thing or just the color? The positioning is pretty specific to `NxDrawer` if I'm not mistaken

This is just for drawer, so I think you're right to just parameterize the color. Done so here: [7e25ef6](https://github.com/sonatype/sonatype-react-shared-components/pull/1097/commits/7e25ef66b14f6de8c1d93ad316da5864a951b0cb)

gallery/src/styles/CssVariables/CssVariablesPage.tsx
EmilyBermudez88: >The variable name says "drawer" but the description says "dropdown"

Based on the changes made in [7e25ef6](https://github.com/sonatype/sonatype-react-shared-components/pull/1097/commits/7e25ef66b14f6de8c1d93ad316da5864a951b0cb), variable documentation updated here: [f72177f](https://github.com/sonatype/sonatype-react-shared-components/pull/1097/commits/f72177f4d926bc0ccd6d23a6d1ffb68cc0da92e6)

lib/src/base-styles/_nx-variables.scss
rpokorny: > A spread of -32px sets the spread inward to create the effect of a smaller, vertically centred element behind the drawer, as displayed in the mocks. The x-offset counteracts the shadow’s spread (to prevent it from being brought inward on the left), plus a further -2px of x-offset defined in the mocks. The Y-offset can be zero as the shadow should be vertically centred relative to the drawer. The blur comes directly from the mocks where it is specified as 50px.

Sounds good

EmilyBermudez88: > > A spread of -32px sets the spread inward to create the effect of a smaller, vertically centred element behind the drawer, as displayed in the mocks. The x-offset counteracts the shadow’s spread (to prevent it from being brought inward on the left), plus a further -2px of x-offset defined in the mocks. The Y-offset can be zero as the shadow should be vertically centred relative to the drawer. The blur comes directly from the mocks where it is specified as 50px.
> 
> Sounds good

[1e6dd43](https://github.com/sonatype/sonatype-react-shared-components/pull/1097/commits/1e6dd43f403a4ffc331e913e7429bae247debfa8)

EmilyBermudez88: > They created a variable named "808080" which seems like a hex code which would be grey, and set its value to indigo? That definitely seems like a mistake one way or another.

Confirmed in slack that this should indeed by `indigo-10` : https://sonatype.slack.com/archives/CJUF9G7T2/p1683120481965319

rpokorny (approved):

PR #1098, opened 2023-05-01
Use --nx-focus-outline-width - RSC-1554

rpokorny:
https://issues.sonatype.org/browse/RSC-1554

I also took the opportunity to expand the use of `--nx-focus-outline-inset`


gallery/visualtests/__image_snapshots_dark__/nx-transfer-list-js-nx-transfer-list-handles-overflowing-content-with-a-tooltip-on-items-initially-scrolled-out-of-view-1-dark.png
rpokorny, line 1: The jest-image-snapshot diff on this image is this:
image

I guess either the before or after image was taken when the tooltip hadn't _quite_ yet fully appeared.  I can't tell for sure which is which but the test runs appear to be stable with the new image.

EmilyBermudez88:
> > Looking for other use cases,:
> > NxColorPIcker, I think we could set the stroke-width to --nx-focus-outline-width for .nx-color-picker__outline (line 50),
> > NxAccordion .nx-accordion__summary-wrapper when focused (line 51/52)
> > NxTable .nx-cell—header when focused that I missed in my ticket (line 278/279)
> 
> I omitted all three of those and possibly others that I knew were already in progress in other tickets. I expect those tickets' PRs to use the correct variables.

Ah, makes sense. You can update NxTable since I accidentally missed updating that in the ticket and it has since been merged.

EmilyBermudez88 (approved):

PR #1106, opened 2023-05-03
Fix Accordion Containers in Snapshot Background - RSC-1680

rpokorny:
https://issues.sonatype.org/browse/RSC-1680


gallery/visualtests/__image_snapshots_dark__/nx-combobox-js-nx-combobox-dropdown-button-has-a-blue-background-when-selected-1-dark.png
rpokorny, line 1: I'll be honest I don't see any difference in this one

rpokorny: > I'll be honest I don't see any difference in this one

Apparently, it's the pixels at the bottom corners of the dropdown menu.  This is also the case in the other seemingly identical snapshots below

EmilyBermudez88: >Apparently, it's the pixels at the bottom corners of the dropdown menu. This is also the case in the other seemingly identical snapshots below

Thanks for mentioning that, because I would have been very confused looking over some of those snapshots

EmilyBermudez88 (approved):

PR #1112, opened 2023-05-03
update snapshots - RSC-1682

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1682

One light mode snapshot updated here as well which seems as though it wasn't initially updated with the padding changes to NxThreatCounter


rpokorny:
> One light mode snapshot updated here as well which seems as though it wasn't initially updated with the padding changes to NxThreatCounter

Hmm ok.  I think what happened there is [RSC-1545](https://issues.sonatype.org/browse/RSC-1545) really should have been merged to main, but it wasn't which means CI run allowing that to go unnoticed

rpokorny (approved):

PR #1126, opened 2023-05-10
Fixed Nested NxModal Focus When Browser Alert or Focus Outside Browser - RSC-1657

entroform:
https://issues.sonatype.org/browse/RSC-1657
 
 Forgot to add tests will do that soon


lib/src/components/AbstractDialog/AbstractDialog.tsx
```diff
@@ -188,6 +195,7 @@ const AbstractDialog = forwardRef<HTMLDialogElement, Props>((props, ref) => {
           event.relatedTarget
```
rpokorny, line 195: This line no longer needed

entroform:
It turns out this solution breaks the fix for when there is a focused element inside a modal that is suddenly hidden via CSS.
This can be tested on the Nested Modals Example by pressing the "Hide Button" button is pressed.

rpokorny:
Closing in favor of https://github.com/sonatype/sonatype-react-shared-components/pull/1142

PR #1127, opened 2023-05-12
Add parameters to onChange in NxCheckbox and NxToggle - RSC-1671

rpokorny:
https://issues.sonatype.org/browse/RSC-1671

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.12.3",
+  "version": "12.13.0",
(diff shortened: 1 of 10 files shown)


Blacktiger (approved):
Should we also make a change to `NxFormSelect`? Maybe in a separate ticket since it would be a breaking change?

avnvbhatta (approved):

rpokorny:
> Should we also make a change to NxFormSelect? Maybe in a separate ticket since it would be a breaking change?

Created RSC-1719 for `NxFormSelect`.  It should definitely be a separate ticket as a breaking change.

PR #1129, opened 2023-05-12
Run dark and light test processes in parallel - RSC-1718

rpokorny:
https://issues.sonatype.org/browse/RSC-1718


rpokorny:
See the linked build results - running the tests in parallel did not make a significant difference.

PR #1133, opened 2023-05-12
update NxCopyToClipboard width- RSC-1561

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1561


lib/src/components/NxCopyToClipboard/NxCopyToClipboard.scss
```diff
@@ -14,7 +14,7 @@
       1fr      auto;
 
   margin-bottom: var(--nx-spacing-8x);
-  width: 700px;
+  max-width: 800px;
```
rpokorny, line 17: Why `max-width` and not `width`?

EmilyBermudez88: This is what is specified in Sketch. I can check with Jason though to see if they want a hard coded 800px width or just a max-width

rpokorny: > This is what is specified in Sketch. I can check with Jason though to see if they want a hard coded 800px width or just a max-width

When the designers say "max width" they don't necessarily mean the same thing as CSS's `max-width`.  It can sometimes take some discussion to figure out exactly what they mean.  They specify the behavioral requirements, we figure out the appropriate implementation.  In this case my best guess is that they meant it should be 800px by default, but overridable by downstream designers provided they don't override it to something greater than 800px.  If that's the case, the implementation can just use `width` and downstream code can customize by overriding that `width`.

EmilyBermudez88: > > This is what is specified in Sketch. I can check with Jason though to see if they want a hard coded 800px width or just a max-width
> 
> When the designers say "max width" they don't necessarily mean the same thing as CSS's `max-width`. It can sometimes take some discussion to figure out exactly what they mean. They specify the behavioral requirements, we figure out the appropriate implementation. In this case my best guess is that they meant it should be 800px by default, but overridable by downstream designers provided they don't override it to something greater than 800px. If that's the case, the implementation can just use `width` and downstream code can customize by overriding that `width`.

That makes sense. @JasonDSwearingen, would you like it to be set to a standard 800px width?

JasonDSwearingen: Yeah, that sounds good to me. I'll clarify the language in Kaleido, too.

EmilyBermudez88: > Why `max-width` and not `width`?

Updated to width: [9752deb](https://github.com/sonatype/sonatype-react-shared-components/pull/1133/commits/9752deb8dcd9a4c1e5171030d8d1d8a5e4af7054)

rpokorny (approved):

PR #1137, opened 2023-05-15
prevent style override on .nx-btn- RSC-1668

EmilyBermudez88:
https://issues.sonatype.org/browse/RSC-1668


lib/src/base-styles/_nx-btn.scss
```diff
@@ -31,7 +31,7 @@
   }
 }
 
-.nx-btn {
+.nx-btn, .nx-btn[href] {
```
rpokorny, line 34: Was there a reason to use `.nx-btn[href]` rather than `a.nx-btn`?  The latter seems cleaner and likely more performant to me.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -79,11 +79,11 @@
   }
 }
 
-.nx-btn--primary {
+.nx-btn--primary, .nx-btn--primary[href] {
```
rpokorny, line 82: If these blocks were included in the above block using `&--primary` etc, we wouldn't need to repeat the new selector

gallery/visualtests/nxButton.js
```diff
@@ -102,9 +102,41 @@ describe('NxButton', function() {
...
+      const tertiarySelector = `${selector} .nx-btn--tertiary`;
+      buttonTests(tertiarySelector);
+    });
+
+    describe('Error styled buttons', function() {
+      const errorSelector = `${selector} .nx-btn--error`;
+      buttonTests(errorSelector);
+    });
+
+    describe('Disabled styled buttons', function() {
```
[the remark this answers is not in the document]
EmilyBermudez88: > Our tests and examples have a blind spot regarding disabled non-secondary buttons. Lets fix that

Are you referring to buttons or <a>s styled as buttons, or both? Happy to write the tests for both use cases

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > Was there a reason to use `.nx-btn[href]` rather than `a.nx-btn`? The latter seems cleaner and likely more performant to me.

No reasoning for using one over the other, so changed here: [eb5d38c](https://github.com/sonatype/sonatype-react-shared-components/pull/1137/commits/eb5d38c1ad7af033d1a7a25f66d6d734df3212ee)

lib/src/base-styles/_nx-btn.scss
EmilyBermudez88: > If these blocks were included in the above block using `&--primary` etc, we wouldn't need to repeat the new selector

Great idea, much cleaner now: [eb5d38c](https://github.com/sonatype/sonatype-react-shared-components/pull/1137/commits/eb5d38c1ad7af033d1a7a25f66d6d734df3212ee)

gallery/visualtests/nxButton.js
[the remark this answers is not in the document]
EmilyBermudez88: > > Are you referring to buttons or s styled as buttons, or both? Happy to write the tests for both use cases
> 
> Oh, let's just do it for the `<a>` tags in this PR

All disabled variants added to the gallery here: [fb8cbaf](https://github.com/sonatype/sonatype-react-shared-components/pull/1137/commits/fb8cbaf0beaf640b84143c8cfe969e10f8a69e0b)
Visual tests for them here: [6a03c20](https://github.com/sonatype/sonatype-react-shared-components/pull/1137/commits/6a03c20f8670fe6b21ea165430e48ef042dd4208)
Snapshots updated here: [e2315af](https://github.com/sonatype/sonatype-react-shared-components/pull/1137/commits/e2315af5b3bedc875358d6d8ccce3706aa44b510) - for some reason it seems that by adjusting the layout of the example, all the initial snapshots of links as buttons failed with a handful of text pixels out of place, so updated again.

PR #1140, opened 2023-05-18
Update Hover State of Color Picker - RSC-1716

cindyko226:
JIRA: [RSC-1716](https://issues.sonatype.org/browse/RSC-1716)

[RSC-1716]: https://sonatype-sandbox-840.atlassian.net/browse/RSC-1716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.15.0",
+  "version": "12.15.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "12.15.0",
+  "version": "12.15.1",
   "description": "Sonatype shared UI components and utilities written in React",
   "main": "index.js",
   "repository": {


lib/src/components/NxColorPicker/NxColorPicker.scss
```diff
@@ -28,6 +28,7 @@
     &:hover {
       .nx-color-picker__color {
         stroke: var(--nx-color-text-stark);
+        stroke-width: var(--nx-border-width-alt);
```
EmilyBermudez88, line 31: If I'm not mistaken, the middle of the stroke width matches the circle's radius, so by increasing the stroke-width by 1px in dark mode, it would also slightly increase the size of the color picker that is being hovered over (you can see that a bit better when the focus line is visible- the separation between the focus outline and hover outline is a tad bit closer). In Sketch it says that the hover border should be an inside border, so you may need to investigate further.

cindyko226: Thanks for catching it!
In order to main the circle size, the radius of circle should be changed based on the different stroke width in light and dark mode when hovered, here is the updates f1bbda7c69710944b7f653d3680d315ac345c529.

rpokorny: > Thanks for catching it!
In order to main the circle size, the radius of circle should be changed based on the different stroke width in light and dark mode when hovered, here is the updates f1bbda7.

The `r` specified in the JSX presumably can be removed now

rpokorny: > The r specified in the JSX presumably can be removed now

As discussed in WW, without further changes the `r` in the JSX cannot in fact be removed because it is needed for the non-hover case.  And we can't just remove `:hover` from the selectors in this PR because the `r` values need to be slightly different in the non-hover case.

However, it still doesn't sit well with me that the `r` is being controlled in two completely different places.  I think a better option would be to just always put a stroke on the `.nx-color-picker__color`, but to have the stroke color match the fill when not hovered.  In fact it looks like this is already what's happening to some extent: there's an `&__color` scss block where this happens but the problem is that it's hard coded to use a stroke-width of 1px; that should probably be changed to `var(--nx-border-width-alt)`.  Then the `r` could also be set there (for both the selected and unselected cases) and the `r` attribute in the JSX could be removed.

cindyko226: Agreed, handling the r attributes in one place is much more cleaner.
Here are the updates, ba92021baacf7ed59a5eca404b103059ded0e1bb, thanks.

EmilyBermudez88 (approved):
LGTM - interesting to see how it all came together so cleanly 👍

cindyko226:
> +1. I guess the snapshot differences are just a browser oddity. Surprising, since we're using the same numbers as before, but not impossible.

I guess it is, when I check the diffs locally, not even all the color circles have the diffs though they have the same css properties applied.

PR #1141, opened 2023-05-22
Upgrade RSC Dependencies - RSC-1709

rpokorny:
https://issues.sonatype.org/browse/RSC-1709

Note: not all dependencies could be upgraded or upgraded all the way due to various constraints.  In particular, since we aren't upgrading React to version 18 in this ticket, we also can't fully upgrade RTL or Material.  The version of Material that we are stuck on seemed to have problems with the latest version of TypeScript, so we also couldn't upgrade that.  On the fontawesome side we couldn't upgrade further than we have without moving to fontawesome 6 which has noticeably different icon shapes.


cindyko226 (approved):
LGTM!

PR #1142, opened 2023-05-22
Improve Modal Focus Trap - RSC-1657

rpokorny:
https://issues.sonatype.org/browse/RSC-1657

In order to fix this it seemed necessary to centralize the focus trap handling, avoiding the complication of each separate focus trap having to figure out whether it's the active one.


cindyko226:
Do we need the test for focusUtil?

rpokorny:
> Do we need the test for focusUtil?

IMO no, since it isn't publicly exported.

PR #1143, opened 2023-05-30
NxSmallTag Component - RSC-1721

cindyko226:
JIRA: https://sonatype.atlassian.net/browse/RSC-1721


gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxSmallTag } from '@sonatype/react-shared-components';
+
+function NxSmallTagExample() {
+  return (
+    <>
+      <div className='nx-small-tag-container'>
```
rpokorny, line 13: I'm surprised eslint didn't catch this
```suggestion
      <div className="nx-small-tag-container">
```

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,70 @@
...
+  &--teal {
+    background-color: var(--nx-swatch-teal-30);
+  }
+
+  &--turquoise {
+    background-color: var(--nx-swatch-turquoise-20);
+  }
+}
+
+.nx-small-tag-container {
```
rpokorny, line 63: I don't think we need this class.  I think `.nx-small-tag` should simply have left and right margin, and maybe a rule to remove the left margin of a second `.nx-small-tag` in a row

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,70 @@
...
+  display: flex;
+  gap: var(--nx-spacing-2x);
+}
+
+#nx-small-tag-in-table-example .nx-table .nx-small-tag{
```
rpokorny, line 68: Example CSS belongs in the gallery, not the lib

rpokorny:
Sketch calls out the way the tags should align with text of different sizes: that it should be vertically centered.  That is not happening: 
image

gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxSmallTag } from '@sonatype/react-shared-components';
+
+function NxSmallTagExample() {
+  return (
+    <>
```
rpokorny, line 12: Redundant

gallery/src/components/NxSmallTag/NxSmallTagInTableExample.tsx
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+import { NxSmallTag, NxTable } from '@sonatype/react-shared-components';
+
+function NxSmallTagInTableExample() {
```
rpokorny, line 10: What was the motivation for having this particular example? Does it demonstrate anything meaningful aside from the alignment between an `NxSmallTag` and an adjacent `<span>`?

gallery/src/components/NxSmallTag/NxSmallTagPage.tsx
```diff
@@ -0,0 +1,83 @@
...
+const NxSmallTagInTableExampleCode = require('./NxSmallTagInTableExample?raw');
+
+const NxSmallTagPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <NxP>
+        Small tags are used only by designers and developers when making a product, rather than
+        being applied and removed by users of the product. <NxCode>NxSmallTag</NxCode> is an inline component
+        intended to call out product features ("Beta", "Custom", "Labs", etc), and is informational
+        whereas <NxCode>NxTag</NxCode> is functional. There is a set of colors for small tags that can be specified
```
rpokorny, line 25:
```suggestion
        <NxCode>NxSmallTag</NxCode> is an inline component
        intended to call out product features ("Beta", "Custom", "Labs", etc). It is informational
        whereas <NxCode>NxTag</NxCode> is functional. There is a set of colors for small tags that can be specified
```

gallery/src/components/NxSmallTag/NxSmallTagPage.tsx
```diff
@@ -0,0 +1,83 @@
...
+        via props.
+      </NxP>
```
rpokorny, line 27: Might also want to mention that small tags have their own set of colors and specifically do not use the `SelectableColors`.

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  align-items: center;
+  border-radius: 8px;
+  box-sizing: border-box;
+  color: var(--nx-swatch-white);
+  display: inline-flex;
+  font-size: var(--nx-font-size-s);
+  height: 24px;
+  margin: 0;
+  max-width: Min(100%,200px);
```
rpokorny, line 20: We need examples and visual tests for the max width behavior

rpokorny, line 20: Why the 100% here? I don't see anything corresponding to that specified in sketch

lib/src/components/NxSmallTag/NxSmallTag.tsx
```diff
@@ -0,0 +1,29 @@
...
+const NxSmallTag = forwardRef<HTMLLabelElement, Props>(function NxSmallTag(props, ref) {
+  const { children, className, color, ...attrs } = props,
+      tagClasses = classnames('nx-small-tag', className, {
+        'nx-small-tag-color--pink': !color,
+        [`nx-small-tag-color--${color}`]: color
+      });
+
+  return (
+    <label className={tagClasses} ref={ref} {...attrs}>
+      <span>{children}</span>
```
rpokorny, line 23: What's the purpose of the span?

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,70 @@
...
+ @use '../../scss-shared/nx-text-helpers';
+
+.nx-small-tag {
+  @include nx-text-helpers.semi-bold;
+
+  align-items: center;
+  border-radius: 8px;
+  box-sizing: border-box;
+  color: var(--nx-swatch-white);
+  display: inline-flex;
```
rpokorny, line 16: Does it need to be `inline-flex` or would just `inline` work?

lib/src/components/NxSmallTag/NxSmallTag.scss
cindyko226: Thanks for the suggestions, updated here ffe1364fb439eebf3e3442735d890469490acb0f.

gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
cindyko226: The unnecessary `<div className='nx-small-tag-container'>` has been removed, so keep fragment here.

gallery/src/components/NxSmallTag/NxSmallTagPage.tsx
cindyko226: Updated here 16bdea34c720239d28187b4fa8eb807a5cf20b5e.

lib/src/components/NxSmallTag/NxSmallTag.tsx
cindyko226: No specific purpose, removed here 30e0f308bfe4d3c84213449914d1df186215aa2b.

lib/src/components/NxSmallTag/NxSmallTag.scss
cindyko226: Thanks for catching it, should be 200px only, updated here 9258cf1345e1bb010a7be699cd14f11d8bb3d598.

cindyko226:
> Sketch calls out the way the tags should align with text of different sizes: that it should be vertically centered. That is not happening: image

Updated here 7978e35bcd889d4c409c6bfca03a5ccba004e73c.

lib/src/components/NxSmallTag/NxSmallTag.scss
cindyko226: Added here bb369fd0dee5453393c16a743aeadc92d9cec074.

gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
cindyko226: This `div` is no longer needed.

> I'm surprised eslint didn't catch this

According to the doc https://eslint.org/docs/latest/rules/jsx-quotes, should we also add rule `jsx-quotes: ["error", "prefer-double"]` to eslint file?

rpokorny: > According to the doc https://eslint.org/docs/latest/rules/jsx-quotes, should we also add rule jsx-quotes: ["error", "prefer-double"] to eslint file?

Yeah seems that way.  Please add a tech debt ticket for it.

gallery/src/components/NxSmallTag/NxSmallTagInTableExample.tsx
cindyko226: It was demonstrated in the mockups, thinking it wouldn't hurt to have it in the gallery and have the look and feel of a small tag in the table. But if you prefer removing this example, I am fine with it.

lib/src/components/NxSmallTag/NxSmallTag.scss
cindyko226: The small tag has the height of 24px, it would need be a block-level element, so here I have `inline-flex` along with `align-items: center` to center the text.

gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <NxSmallTag>Default</NxSmallTag>
+      <NxSmallTag color="blue">Blue</NxSmallTag>
+      <NxSmallTag color="indigo">Indigo</NxSmallTag>
+      <NxSmallTag>Wwwwwwwwwwwwwwww</NxSmallTag>
```
rpokorny, line 16: This happens to produce a tag that is right at its max size with no apparent overflow.  We want to visually test what happens when there _is_ overflow, so please make it a bit longer.  You might want to use text that makes it clear that it's broken on purpose, such as "This tag content is too long and should overflow".

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,66 @@
...
+  font-size: var(--nx-font-size-s);
+  height: 24px;
+  margin: 0 var(--nx-spacing-2x);
+  max-width: 200px;
+  padding: 0 var(--nx-spacing-2x);
+  position: relative;
+  vertical-align: middle;
```
rpokorny, line 23: Unfortunately, this does not quite produce the desired effect.  See here that it's still too low:
image

This is the description of what `vertical-align: middle` does:
> Aligns the middle of the element with the baseline plus half the x-height of the parent.

As the x-height means the height of a _lowercase_ x, this is almost never actually useful.  I'm working on a better solution.

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,66 @@
...
+
+  + .nx-small-tag {
+    margin-left: 0;
+  }
+}
+
+.nx-small-tag-color {
```
rpokorny, line 30: This shouldn't be a separate base class.  The color classes should be `nx-small-tag--blue` etc.

lib/src/components/NxSmallTag/__tests__/NxSmallTag.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+ */
+import React, { RefAttributes } from 'react';
+
+import NxSmallTag, { Props } from '../NxSmallTag';
+
+import { rtlRenderElement } from '../../../__testutils__/rtlUtils';
+
+describe('NxSmallTag', function() {
+  type PropsWithRef = Props & RefAttributes<HTMLLabelElement>;
+  const renderEl = rtlRenderElement<PropsWithRef>(NxSmallTag, { children: 'small tag'});
```
rpokorny, line 15:
```suggestion
  const renderEl = rtlRenderElement<PropsWithRef>(NxSmallTag, { children: 'small tag' });
```

lib/src/components/NxSmallTag/types.ts
```diff
@@ -0,0 +1,22 @@
...
+import { LabelHTMLAttributes, ReactNode } from 'react';
+import * as PropTypes from 'prop-types';
+
+export const NX_SMALL_TAG_COLORS = ['blue', 'green', 'indigo', 'orange', 'pink',
+  'purple', 'red', 'teal', 'turquoise'] as const;
+export type NX_SMALL_TAG_COLOR_TYPE = (typeof NX_SMALL_TAG_COLORS)[number];
+
+export interface Props extends LabelHTMLAttributes<HTMLLabelElement> {
+  children: ReactNode;
+  smallTagColor?: NX_SMALL_TAG_COLOR_TYPE | null;
```
rpokorny, line 16: This doesn't seem to match the interface in the docs and actual component code.  Isn't it just `color`?

lib/src/components/NxSmallTag/NxSmallTag.scss
rpokorny: > As the x-height means the height of a lowercase x, this is almost never actually useful. I'm working on a better solution.
See here: https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-1721_small_tag_component...RSC-1721_small_tag_component_ross

I think an example with `NxSmallTag` inline with other text of varying sizes (and in multiple lines both of other text and other `NxSmallTag`s should be added and tested.

gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
cindyko226: Created here https://sonatype.atlassian.net/browse/RSC-1723.

gallery/src/components/NxSmallTag/NxSmallTagInTableExample.tsx
cindyko226: Removed here d6ee65a1f24a7c3e96a420490e625b4bc9ddb5be

lib/src/components/NxSmallTag/types.ts
cindyko226: Yes it should be `color`, updated here 0fe003abf2cbaef0bb1d32876e5d23c78f26c39c

lib/src/components/NxSmallTag/NxSmallTag.scss
cindyko226: Updated here 9eda9c3227c6741d2f3bbfd27aa1af289d5fcd59

gallery/src/components/NxSmallTag/NxSmallTagExample.scss
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.gallery-overflow-example {
+  background-color: var(--nx-swatch-indigo-40);
```
rpokorny, line 8: What's the purpose of this blue background?

gallery/visualtests/__image_snapshots__/nx-small-tag-js-nx-small-tag-looks-right-1.png
rpokorny, line 1: Although they don't use ellipsis truncation, I would have expected the tags to hide overflowing text

lib/src/components/NxSmallTag/types.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import { LabelHTMLAttributes, ReactNode } from 'react';
+import * as PropTypes from 'prop-types';
+
+export const NX_SMALL_TAG_COLORS = ['blue', 'green', 'indigo', 'orange', 'pink',
+  'purple', 'red', 'teal', 'turquoise'] as const;
```
rpokorny, line 11: A cleaner way to break the lines
```suggestion
export const NX_SMALL_TAG_COLORS =
  ['blue', 'green', 'indigo', 'orange', 'pink', 'purple', 'red', 'teal', 'turquoise'] as const;
```

gallery/src/components/NxSmallTag/NxSmallTagExample.scss
cindyko226: I think this can be removed if the overflow text should be hidden.

lib/src/components/NxSmallTag/NxSmallTag.scss
cindyko226: Examples added here 219c4dfa4ad15a82dd54c873f0ad99cdb653e6ee, f12fbdfffae05fbf5b4605e0f6bf2981139ef059.

rpokorny: > I think an example with NxSmallTag inline with other text of varying sizes (and in multiple lines both of other text and other NxSmallTags should be added and tested.
 
 I've realized that my previous suggested implementation still wasn't quite right - I needed to set the (line) height of `.nx-small-tag` to the _line-height_ of the parent, not the font-size.  See https://github.com/sonatype/sonatype-react-shared-components/compare/RSC-1721_small_tag_component...RSC-1721_small_tag_component_ross again for further suggestions which further improve the alignment.

cindyko226: Updated css and snapshots 4e88bd0b9317f7024c506d16793d983546be0df1, 795d274c15bc4a28c7d18c6382020373fe84ea6d.

gallery/src/components/NxSmallTag/NxSmallTagExample.tsx
cindyko226: Updated here 95a0d8db60872e6bf8ee325930863d30bba6fbe3, be0238cbe083e05e4bafdfc87982cb303fc97a3d.

gallery/src/components/NxSmallTag/NxSmallTagExample.scss
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+.gallery-overflow-example {
```
rpokorny, line 7: Unused now, right?

gallery/src/components/NxSmallTag/NxSmallTagLayoutExample.tsx
```diff
@@ -0,0 +1,25 @@
...
+ */
+import React from 'react';
+import { NxH1, NxP, NxSmallTag } from '@sonatype/react-shared-components';
+
+function NxSmallTagLayoutExample() {
+  return (
+    <>
+      <NxH1>H1 Header / 32px / Bold<NxSmallTag>Pink</NxSmallTag></NxH1>
+      <NxP>Paragraph Text / 16px / Regular<NxSmallTag>Pink</NxSmallTag></NxP>
+      <NxP className='small-font-size'>Paragraph Text / 14px / Regular<NxSmallTag>Pink</NxSmallTag></NxP>
```
rpokorny, line 15:
```suggestion
      <NxP className="small-font-size">Paragraph Text / 14px / Regular<NxSmallTag>Pink</NxSmallTag></NxP>
```

gallery/src/components/NxSmallTag/NxSmallTagLayoutExample.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <NxP>Paragraph of wrapping text - Toffee liquorice sugar plum halvah carrot cake. Powder jelly-o marzipan
```
rpokorny, line 16:
```suggestion
      <NxP>
        Paragraph of wrapping text - Toffee liquorice sugar plum halvah carrot cake. Powder jelly-o marzipan
```

gallery/src/components/NxSmallTag/NxSmallTagPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+        </NxTable.Body>
+      </NxTable>
+    </GalleryDescriptionTile>
+
+    <GalleryExampleTile title="Small Tag Example"
+                        id="nx-small-tag-example"
+                        liveExample={NxSmallTagExample}
+                        codeExamples={NxSmallTagExampleCode}>
+      Example of small tags in all available colors. Note that the small tags have a maximum width
+      of 200px and should be designed as short as possible, they do not truncate.
```
rpokorny, line 72:
```suggestion
      of 200px and should be designed as short as possible.  They do not use ellipsis truncation and should never be given content exceeding their max width.  The over-long content in this example is only present as a test of the max-width behavior.
```

gallery/src/components/NxSmallTag/NxSmallTagExample.scss
cindyko226: Removed here e2e3da97137ca3defa9e5cc22a7ee56f755d7066

gallery/src/components/NxSmallTag/NxSmallTagExample.scss
```diff
@@ -0,0 +1,23 @@
...
+  background-color: var(--nx-swatch-indigo-40);
+  margin-left: var(--nx-spacing-2x);
+  margin-top: var(--nx-spacing-2x);
+  padding: var(--nx-spacing-2x) 0;
+  width: 400px;
+}
+
+.nx-h1, .nx-h2, .nx-h3 {
+  > :first-child {
+    margin-left: var(--nx-spacing-2x);
```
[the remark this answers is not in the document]
cindyko226: Yes you are right, this is not correct and it affect other headers elements on this page. 
Updated here 24fd036f7b0c060ca77f5ef5e86a77d024215244

gallery/visualtests/nxSmallTag.js
```diff
@@ -0,0 +1,20 @@
...
+describe('NxSmallTag', function() {
+  const { simpleTest, a11yTest } = setupBrowser('#/pages/Small%20Tag');
+
+  const smallTagSelector = '#nx-small-tag-example .gallery-example-live',
+      defaultTagSelector = `${smallTagSelector} .nx-small-tag--pink .nx-small-tag__text`,
+      smallTagLayoutSelector = '#nx-small-tag-layout-example .gallery-example-live';
+
+  it('looks right', simpleTest(smallTagSelector));
+  it('renders a default pink colored NxSmallTag if no color prop is passed', simpleTest(defaultTagSelector));
+  it('looks right', simpleTest(smallTagLayoutSelector));
```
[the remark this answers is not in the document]
cindyko226: Updated here 657f89f2f7a3cfcfe63c001564d4516d07370d54

lib/src/components/NxSmallTag/NxSmallTag.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  align-items: center;
+  display: inline-flex;
+  height: calc(1em * var(--nx-line-height));
+  margin: 0 var(--nx-spacing-2x);
+  vertical-align: top;
+
+  + .nx-small-tag {
+    margin-left: 0;
+  }
```
cindyko226, line 21: I did notice this affected the margin when multiple small tags are included in the paragraph(not in sequence, some text in between), there is no margin on the  second small tag, seems like the `+` css selector ignored the plain text. 
In this case, should we move these to the gallery example css file and apply the styles within the example block instead?
Screenshot 2023-06-05 at 6 43 12 PM

cindyko226: Ah right, updated here 6907f882554eaab2d08605229ad73180402d1b9f

gallery/src/components/NxSmallTag/NxSmallTagLayoutExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+ */
+import React from 'react';
+import { NxH1, NxP, NxSmallTag } from '@sonatype/react-shared-components';
+
+function NxSmallTagLayoutExample() {
+  return (
+    <>
+      <NxH1>
+        <span>H1 Header / 32px / Bold</span>
+        <NxSmallTag>Pink</NxSmallTag>
```
rpokorny, line 15: Remove whitespace here too
```suggestion
        <span>H1 Header / 32px / Bold</span><NxSmallTag>Pink</NxSmallTag>
```

lib/src/components/NxSmallTag/NxSmallTag.scss
rpokorny: > Oh but you also need to remove the text whitespace before and after each NxSmallTag in that paragraph, since they handle their spacing with margin. If you want you could even mention that in the example description.

I was wrong about this.

gallery/src/components/NxSmallTag/NxSmallTagPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+      Example of small tags in all available colors. Note that the small tags have a maximum width
+      of 200px and should be designed as short as possible. They do not use ellipsis truncation and
+      should never be given content exceeding their max width. The over-long content in this example
+      is only present as a test of the max-width behavior.
+    </GalleryExampleTile>
+
+    <GalleryExampleTile title="Small Tag Layout Example"
+                        id="nx-small-tag-layout-example"
+                        liveExample={NxSmallTagLayoutExample}
+                        codeExamples={NxSmallTagLayoutExampleCode}>
```
rpokorny, line 80: This example relies on custom CSS that isn't being depicted in the code display.  Either add `NxSmallTagExample.scss` to the `codeExamples` here or just use a `style` attribute to set the font size rather than a class.

cindyko226: Updated here 75333fa37fd47b436f40662367d468845c0df899.

JasonDSwearingen (approved):

PR #1144, opened 2023-06-06
Update Active Text Color and Dropshadow for NxDropdownMenu - RSC-1672

cindyko226:
Jira: https://sonatype.atlassian.net/browse/RSC-1672


lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
```diff
@@ -68,6 +68,7 @@
 
   &:active {
     background-color: var(--nx-color-dropdown-menu-background-active);
+    color: var(--nx-color-text-stark);
```
rpokorny, line 71: I'm not finding and example of the active state in the mockups, where is it?

lib/src/base-styles/_nx-variables.scss
```diff
@@ -62,9 +62,9 @@
   --nx-border-grid: var(--nx-border-2x);
 
   // for dropdown
-  --nx-box-shadow-dropdown: 0 2px 12px 0 rgba(74, 77, 113, 0.2),
-                            0 1px 2px 0 rgba(91, 95, 141, 0.2),
-                            0 1px 1px 0 rgba(124, 128, 179, 0.2);
+  --nx-box-shadow-dropdown: 0 2px 12px 0 Hsla(232, 43%, 80%, 0.2),
+                            0 1px 2px 0 Hsla(232, 42%, 70%, 0.2),
+                            0 1px 1px 0 Hsla(231, 43%, 60%, 0.2);
```
rpokorny, line 67: All of these are 231 in figma.  Moreover I'm assuming they're all supposed to be indigo and any variation from our standard indigo hue and saturation is a rounding error.  There are similar issues in the dark mode value for this variable.

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
cindyko226: There are some discrepancies between Sketch and Figma(e.g. Figma is using rgba on the box-shadow value instead of hsla) which @JasonDSwearingen  is aware of and will take care of it, hence, will refer to sketch instead for this ticket.
However, there were issues with Sketch that it is kinda broken and the active state example is cut off, I checked with Jason and it seems only happened in the browser, not in the desktop app. He confirmed that the text color of active state would be `nx-grey-30` in light mode and `nx-grey-05` in dark mode.

lib/src/base-styles/_nx-variables.scss
cindyko226: Confirmed with Design that it should be the Indigo. 
Updated here fa72e1ab71dd4515dd3824abac032458d8dbdfec.

lib/src/base-styles/_nx-variables.scss
```diff
@@ -62,9 +62,9 @@
   --nx-border-grid: var(--nx-border-2x);
 
   // for dropdown
-  --nx-box-shadow-dropdown: 0 2px 12px 0 rgba(74, 77, 113, 0.2),
-                            0 1px 2px 0 rgba(91, 95, 141, 0.2),
-                            0 1px 1px 0 rgba(124, 128, 179, 0.2);
+  --nx-box-shadow-dropdown: 0 2px 12px 0 Hsla(231, 43%, 80%, 0.2),
```
rpokorny, line 65:
```suggestion
  --nx-box-shadow-dropdown: 0 2px 12px 0 Hsla(var(--nx-swatch-indigo-hs), 80%, 0.2),
```
and likewise on the other lines in both light and dark mode

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
rpokorny: > He confirmed that the text color of active state would be nx-grey-30 in light mode and nx-grey-05 in dark mode.

I suspect you misspoke there as those numbers (the dark mode one in particular) don't make sense and don't match what you've implemented.  Could you clarify again what the active text colors are supposed to be?

cindyko226: You are right, my brain prob was not clear at that moment.
The correct text color in Active state should be

- Light mode: `nx-grey-05`

- Dark mode: `nx-indigo-90`

rpokorny (approved):

JasonDSwearingen (approved):

PR #1146, opened 2023-06-08
Draft: Adding searchable tags to page config

chriswininger:
This will allow the component filter to better surface pages, for example a page called Alert that discusses the NxAlert component, could match on NxAlert as well as Alert.

Note: This is not quite ready yet, but pushing it up for discussion

### DIFF
--- a/gallery/src/filterUtil.tsx
+++ b/gallery/src/filterUtil.tsx
@@ -4,7 +4,7 @@
  * the terms of the Eclipse Public License 2.0 which accompanies this
  * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
  */
-import { reduce, head, tail, toLower, isEmpty, splitAt, drop, join } from 'ramda';
+import {reduce, head, tail, toLower, isEmpty, splitAt, drop, join, map} from 'ramda';
 import React, { ReactNode } from 'react';
 
 /**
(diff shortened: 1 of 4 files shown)


gallery/src/filterUtil.tsx
```diff
@@ -67,14 +67,22 @@ export function markByFilter(filter: string | undefined, text: string): ReactNod
 /**
  * @return true if the pageName contains every character present in the filter in the same order, case insensitive
  */
-export function matchesFilter(filter: string, pageName: string) {
+export function matchesFilter(filter: string, pageName: string, tags: string[] = []): boolean {
+  const matchesPageName = isStringIncludedInSecondStringCaseInsensitive(filter, pageName);
+  const matchesAnyTags: boolean = map((tag: string) => {
```
rpokorny, line 72: Ramda has an `any` function that would be simpler than the map/reduce here

rpokorny:
Approach looks good.  You might want to consult with the platform designers (`@platform-design` on slack) about the details of how it should work.  For instance I wonder if tags should match on exact strings or substrings rather than same characters in order like the page titles.  I also wonder what it should look like when it matches by tag.  With the page titles it bolds the characters that were matched.

gallery/src/filterUtil.tsx
chriswininger: I figured it had something but struggled to figure it out, I'll poke at that

chriswininger:
> Approach looks good. You might want to consult with the platform designers (`@platform-design` on slack) about the details of how it should work. For instance I wonder if tags should match on exact strings or substrings rather than same characters in order like the page titles. I also wonder what it should look like when it matches by tag. With the page titles it bolds the characters that were matched.

yeah wondered this too, I was thinking it would be less confusing if it matched on all strings in the same way, but perhaps it should be different

PR #1148, opened 2023-06-21
Update Focus State for Links in NxDropdownMenu - RSC-1541

cindyko226:
Jira: https://sonatype.atlassian.net/browse/RSC-1541


cindyko226:
I have implemented the focus style of links, with this approach that the link would have the margins instead of paddings, but I do see some align issues of tooltips, the tooltips now aligned to the left of the link which is expected since there are no paddings, should we consider this is correct and update the mockups accordingly, any thoughts? Thanks.
Breadcrumb_SS

cindyko226:
> > I have implemented the focus style of links, with this approach that the link would have the margins instead of paddings, but I do see some align issues of tooltips, the tooltips now aligned to the left of the link which is expected since there are no paddings, should we consider this is correct and update the mockups accordingly, any thoughts? Thanks.
> 
> What were they aligned with before; the edge of the dropdown menu? This change seems ok to me but check with @JasonDSwearingen

Yes, they aligned to the edge of the dropdown menu before. 
Confirmed with @JasonDSwearingen the changes are ok. https://sonatype.slack.com/archives/CJUF9G7T2/p1687374036238569

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -14,7 +14,7 @@
   position: relative;
   width: nx-dropdown-variables.$dropdown-width;
 
-  .nx-icon:not(.nx-icon--colorful, .nx-dropdown-link .nx-icon, .nx-dropdown :is(.disabled, [disabled]) .nx-icon) {
+  .nx-icon:not(.nx-icon--colorful, .nx-dropdown-link .nx-icon, .nx-filter-dropdown__reset .nx-icon, .nx-dropdown :is(.disabled, [disabled]) .nx-icon) {
```
rpokorny, line 17: What's this change about?

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
```diff
@@ -85,7 +85,11 @@
 
 .nx-dropdown-link {
   color: var(--nx-color-link);
+  margin: var(--nx-spacing-base) var(--nx-spacing-4x);
+  max-width: calc(100% - var(--nx-spacing-8x));
+  padding: 0;
   text-decoration: underline;
+  width: fit-content;
```
rpokorny, line 92: What's this for? I don't see a meaningful difference with it

rpokorny: > What's this for? I don't see a meaningful difference with it

Oh the width of the focus box.  Let's use `min-content` instead, as it works just as well and `fit-content` is much more complex to think about.

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
```diff
@@ -102,6 +106,23 @@
       cursor: default;
     }
   }
+
+  &:focus {
+    border-radius: 6px;
```
rpokorny, line 111:
```suggestion
    border-radius: var(--nx-border-radius);
```

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
```diff
@@ -102,6 +106,23 @@
       cursor: default;
     }
   }
+
+  &:focus {
```
rpokorny, line 110: The other focus block in this file now effectively only applies to `.nx-dropdown-button` and not also `.nx-dropdown-link`.  Let's move it out to be specific to `.nx-dropdown-button`.

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
```diff
@@ -85,7 +85,11 @@
 
 .nx-dropdown-link {
   color: var(--nx-color-link);
+  margin: var(--nx-spacing-base) var(--nx-spacing-4x);
```
rpokorny, line 88: Since doing this with margin is substantially more complicated, let's add a comment explaining why we're using margin.

lib/src/components/NxFilterDropdown/NxFilterDropdown.scss
```diff
@@ -40,12 +40,15 @@
   }
 
   &__reset {
+    color: var(--nx-color-link);
```
rpokorny, line 43: Similar to the changes in `NxDropdown.scss`, I'm not understanding what this is for.  I'm not noticing any behavioral difference between your branch and main on this element.

rpokorny: > Similar to the changes in NxDropdown.scss, I'm not understanding what this is for. I'm not noticing any behavioral difference between your branch and main on this element.

Oh I think I see.  It's because you changed this element to be an `.nx-dropdown-button`.

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
cindyko226: Updated here a8e7a0c7b241020d6dce431d470a18c1bc5666cc.

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
cindyko226: Updated here 694650c86648fc123ff136c037e180f703fd59d3.

lib/src/components/NxDropdownMenu/NxDropdownMenu.scss
cindyko226: Updated here d671ceee7b4c09fe71af33dc1f2ee456af49a248.

PR #1153, opened 2023-07-18
Clean up table height calculation - RSC-1726

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1726

The bug only appeared in cases of fractional scaling: when browser zoom or OS configuration result in CSS pixels mapping to a non-whole-number of device pixels.  In that case, when it comes to borders it seems the browsers favor crispness over author intent.  That is, they'll override the number of CSS pixels for a border-width such that it matches a whole number of device pixels.  This can affect the total height of elements including our table rows, making them (on my system with 150% global scaling) 0.667px shorter than the pagination-table-height mixin expects.  This deficit compounds per table row and can build up to a very noticeable gap between the bottom table row and the pagination footer.

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "12.16.5",
+  "version": "12.16.6",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components",
-  "version": "12.16.5",
+  "version": "12.16.6",
   "description": "Sonatype shared UI components and utilities written in React",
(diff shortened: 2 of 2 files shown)


lib/src/scss-shared/_nx-table-variables.scss
```diff
@@ -9,5 +9,5 @@
 // nx-tables css
 $table-cell-vertical-padding: var(--nx-spacing-3x);
 $table-cell-header-height: 48px;
-$table-cell-min-height: 48px; //content box only
+$table-cell-min-height: calc(var(--nx-font-size) * var(--nx-line-height) + #{$table-cell-vertical-padding} * 2 + 1px);
```
EmilyBermudez88, line 12: Just to confirm here, the `table-cell-min-height` is moving from 48px to 49px, because it will now include the 1px border bottom?

rpokorny: > Just to confirm here, the table-cell-min-height is moving from 48px to 49px, because it will now include the 1px border bottom?

Yes

PR #1155, opened 2023-07-19
Allow tooltip to be accessible name or description - RSC-1130

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1130

### DIFF
--- a/gallery/src/components/NxTextInput/NxTextInputPage.tsx
+++ b/gallery/src/components/NxTextInput/NxTextInputPage.tsx
@@ -395,7 +395,9 @@ const NxTextInputPage = () =>
       Examples of <NxCode>NxTextInput</NxCode>s using
       the <NxCode>long</NxCode> modifier, which makes them wider. Additionally, the second input demonstrates that
       the <NxCode>id</NxCode>, <NxCode>aria-required</NxCode>, and <NxCode>aria-describedby</NxCode> props are
-      passed down directly to the <NxCode>&lt;input&gt;</NxCode>/<NxCode>&lt;textarea&gt;</NxCode>.
+      passed down directly to the <NxCode>&lt;input&gt;</NxCode>/<NxCode>&lt;textarea&gt;</NxCode>. That input
+      is additional wrapped within a label – this is done to appease accessibility standards checking tools which
+      require inputs with an accessible description to also have an accessible name.
     </GalleryExampleTile>
 
     <GalleryExampleTile title="Short example"
--- a/gallery/src/components/NxTextInput/NxTextInputPage.tsx
+++ b/gallery/src/components/NxTextInput/NxTextInputPage.tsx
@@ -396,7 +396,7 @@ const NxTextInputPage = () =>
       the <NxCode>long</NxCode> modifier, which makes them wider. Additionally, the second input demonstrates that
       the <NxCode>id</NxCode>, <NxCode>aria-required</NxCode>, and <NxCode>aria-describedby</NxCode> props are
       passed down directly to the <NxCode>&lt;input&gt;</NxCode>/<NxCode>&lt;textarea&gt;</NxCode>. That input
-      is additional wrapped within a label – this is done to appease accessibility standards checking tools which
+      is additionally wrapped within a label – this is done to appease accessibility standards checking tools which
       require inputs with an accessible description to also have an accessible name.
     </GalleryExampleTile>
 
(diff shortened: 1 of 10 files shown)


gallery/src/components/NxTextInput/NxTextInputPage.tsx
```diff
@@ -395,7 +395,9 @@ const NxTextInputPage = () =>
       Examples of <NxCode>NxTextInput</NxCode>s using
       the <NxCode>long</NxCode> modifier, which makes them wider. Additionally, the second input demonstrates that
       the <NxCode>id</NxCode>, <NxCode>aria-required</NxCode>, and <NxCode>aria-describedby</NxCode> props are
-      passed down directly to the <NxCode>&lt;input&gt;</NxCode>/<NxCode>&lt;textarea&gt;</NxCode>.
+      passed down directly to the <NxCode>&lt;input&gt;</NxCode>/<NxCode>&lt;textarea&gt;</NxCode>. That input
+      is additional wrapped within a label – this is done to appease accessibility standards checking tools which
```
EmilyBermudez88, line 399:
```suggestion
      is additionally wrapped within a label – this is done to appease accessibility standards checking tools which
```

EmilyBermudez88: This inclusion in the PR confused me for a second until I realized it's being merged into RSC-1550 and not main!

gallery/src/components/NxTooltip/NxTooltipPage.tsx
```diff
@@ -108,6 +123,14 @@ export default function NxTooltipPage() {
                           liveExample={NxTooltipPlacementExample}>
         This example shows all possible placements for tooltips
       </GalleryExampleTile>
+
+      <GalleryExampleTile title="Accessible Name Example"
+                          id="nx-tooltip-placement-example"
+                          codeExamples={nameTooltipsExampleCode}
+                          liveExample={NxTooltipNameExample}>
+        This example shows two tooltips attached to SVG elements. One sets the <NxCode>isName</NxCode> prop while the
```
EmilyBermudez88, line 131:
```suggestion
        This example shows two tooltips attached to <button> elements. One sets the <NxCode>isName</NxCode> prop while the
```

lib/src/components/NxButton/NxButton.tsx
```diff
@@ -30,10 +31,10 @@ const NxButton = forwardRef<HTMLButtonElement, Props>(
       if (disabled && title) {
         throw new TypeError('NxButton cannot contain both the \'disabled\' and \'title\' props.');
       }
-      if (variant === 'icon-only' && !title && !disabled && !alreadyHasTooltip) {
+      if (isIconOnly && !title && !disabled && !alreadyHasTooltip) {
         console.warn('Using icon-only buttons without the title prop is deprecated');
       }
-      return wrapInTooltip ? <NxTooltip title={title}>{getBtn()}</NxTooltip> : getBtn({ title });
+      return wrapInTooltip ? <NxTooltip isName={isIconOnly} title={title}>{getBtn()}</NxTooltip> : getBtn({ title });
```
EmilyBermudez88, line 37: this works perfectly on VoiceOver

lib/src/components/NxTooltip/__tests__/NxTooltip.test.tsx
```diff
@@ -315,4 +315,147 @@ describe('NxTooltip', function() {
...
+        expect(buttonWithNoDescription).not.toHaveAccessibleDescription();
+        expect(buttonWithNoDescription).not.toHaveAttribute('title');
+      });
+    });
+
+    describe('when title is JSX', function() {
+      const title = <span data-testid="tooltip-content">tiptip</span>,
+          quickRender = rtlRender(NxTooltip, { ...minimalProps, isName: true, title });
+
+      it('sets the title as the accessible name of the child only when hovered', async function() {
```
EmilyBermudez88, line 394: why does this differ from when the title is a string?

rpokorny: > why does this differ from when the title is a string?

Admittedly, this test is more of a recording of what MUI happens to do rather than an assertion that this is what we _want_ it to do.  But essentially when you use a JSX tooltip, MUI doesn't make an attempt to work out the text content of that JSX and so can only set the accessible name via `aria-labelledby` once the tooltip is actually rendered.  I imagine they _could_ render into the DOM all the time, and just set it to display: none when not active, in order to set the accessible name when not hovered, but it doesn't seem they've elected to do that.

EmilyBermudez88 (approved):

PR #1156, opened 2023-07-19
NxTable sort icons interaction states - RSC-1646

EmilyBermudez88:
https://sonatype.atlassian.net/browse/RSC-1646

Figma: https://www.figma.com/file/SMeWSFCKTQbyarldXKVzro/Kaleido-V2?type=design&node-id=0-24636&mode=design&t=BUCp81dCupSxf0Ce-0

After speaking to Jason, they confirmed that the full table mocks are currently incorrect and still need to be updated, so refer to the new 'Sort Function Interaction States' section for proper implementation of the table's outer border (an outer border around the entire table, with no border line between the header & body, which is the current implementation in the gallery)

### DIFF
--- a/lib/src/base-styles/_nx-tables.scss
+++ b/lib/src/base-styles/_nx-tables.scss
@@ -367,6 +367,41 @@
 
 .nx-cell__sort-btn {
   height: nx-table-variables.$table-cell-header-height;
+  position: relative;
+
+  > * {
+    position: relative;
+  }
+
+  &::before {
+    box-sizing: border-box;
+    border-radius: 4px;
+    content: '';
+    // offset by 1px on top and 2px on the bottom
+    height: calc(100% - 3px);
+    position: absolute;
+    left: 0;
+    top: 1px;
+    // the button is already 1px narrower than the <th> element, so width set to the full 100%
+    width: 100%;
+  }
+
+  &:hover {
+    &::before {
+      background-color: var(--nx-color-table-header-button-background-hover);
+    }
+  }
+  &:focus {
+    &::before {
+      border: var(--nx-focus-outline-width) solid var(--nx-color-interactive-border-focus);
+    }
+  }
+  &:active {
+    &::before {
+      background-color: var(--nx-color-table-header-button-background-active);
+      border: none;
+    }
+  }
 }
 
 // adjust button sizing for ends of rows, where padding is different and there are borders
--- a/gallery/src/styles/CssVariables/CssVariablesPage.tsx
+++ b/gallery/src/styles/CssVariables/CssVariablesPage.tsx
@@ -816,6 +816,12 @@ const CssVariablesPage = () => {
             <ColorDocRow colorVar="--nx-color-table-header-background">
               The background color of the <NxCode>NxTable</NxCode> header.
             </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-table-header-button-background-hover">
+              The background color of the <NxCode>NxTable</NxCode> header button when hovered.
+            </ColorDocRow>
+            <ColorDocRow colorVar="--nx-color-table-header-button-background-active">
+              The background color of the <NxCode>NxTable</NxCode> header button when clicked.
+            </ColorDocRow>
             <ColorDocRow colorVar="--nx-color-table-row-background-hover">
               The background color of an <NxCode>NxTable</NxCode> row on hover.
             </ColorDocRow>
--- a/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-looks-right-with-an-active-sort-column-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-looks-right-with-an-active-sort-column-1.png
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:b5f26546140911bafe3879907523f170fe76109013a000b88595470e776348f0
-size 11067
--- a/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-sortable-columns-looks-right-when-clicked-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-sortable-columns-looks-right-when-clicked-1.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:5dcb4b0be88744f28ad5a6393ee82d89f67a62f75db4c3c854cde31d01033436
+size 11268
--- a/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-sortable-columns-looks-right-when-focused-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-sortable-columns-looks-right-when-focused-1.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:e3ab211b5396de6622cf7eadf88068c27f040e4ad4ed67bc60719bb2809930a3
+size 11240
--- a/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-sortable-columns-looks-right-when-focused-and-hovered-1.png
+++ b/gallery/visualtests/__image_snapshots__/nx-table-js-nx-table-sortable-columns-looks-right-when-focused-and-hovered-1.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:37d73dfcb00a9a19b22bfe23ccb9dd75dc67a5607b03214a07370e028e639527
+size 11414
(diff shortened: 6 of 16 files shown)


lib/src/base-styles/_nx-tables.scss
```diff
@@ -367,6 +367,41 @@
 
 .nx-cell__sort-btn {
   height: nx-table-variables.$table-cell-header-height;
+  position: relative;
+
+  > * {
+    position: relative;
```
rpokorny, line 373: What's this for?

rpokorny:
Should the click target correspond with the background that changes on click/hover? It doesn't exactly right now, both in the sense that that background doesn't perfectly correspond with the bounds of the `<button>` and also that the click target is the whole cell rather than the button.  I'm not sure we'd want to change the latter; it's just the `onClick` of `NxTable.Cell` which can of course be applicable in other situations besides sort headers.  We probably want to be consistent about that prop applying to the whole `<td>` regardless of context.

lib/src/base-styles/_nx-tables.scss
EmilyBermudez88: > What's this for?

Without this, the stacking context was off. The `::before` element was covering up the button's `<span>` element containing the text.

EmilyBermudez88:
> Should the click target correspond with the background that changes on click/hover? It doesn't exactly right now, both in the sense that that background doesn't perfectly correspond with the bounds of the `<button>` and also that the click target is the whole cell rather than the button. I'm not sure we'd want to change the latter; it's just the `onClick` of `NxTable.Cell` which can of course be applicable in other situations besides sort headers. We probably want to be consistent about that prop applying to the whole `<td>` regardless of context.

So then would you thought be to apply the `::before` to the `<th>` element instead for consistency between the background changes & click target? I can bring this up with Jason

EmilyBermudez88:
> Should the click target correspond with the background that changes on click/hover? It doesn't exactly right now, both in the sense that that background doesn't perfectly correspond with the bounds of the `<button>` and also that the click target is the whole cell rather than the button. I'm not sure we'd want to change the latter; it's just the `onClick` of `NxTable.Cell` which can of course be applicable in other situations besides sort headers. We probably want to be consistent about that prop applying to the whole `<td>` regardless of context.

Personally, I think for consistency sake with other `onClick` events that we should leave the click target as the entire `NxTable.Cell`. As for the size of the button, I think it would make sense to expand it to the full size of the table cell (instead of being 1px shy as it currently is), which would then at least match the pointer cursor with the click target. To me the 1px inset of the `::before` still suggests the entire cell is clickable (and that the slight difference in background size is a style choice)

lib/src/base-styles/_nx-tables.scss
```diff
@@ -365,29 +365,89 @@
...
 // adjust button sizing for ends of rows, where padding is different and there are borders
 .nx-cell {
   $edge-spacing: calc(var(--nx-spacing-4x) - 1px);
 
   &:first-child {
-    > .nx-cell__sort-btn, > .nx-cell__row-btn {
+    > .nx-cell__row-btn {
       @include btn-cell-sizing($edge-spacing, var(--nx-spacing-3x));
     }
+    > .nx-cell__sort-btn {
```
EmilyBermudez88, line 431: I wasn't sure about separating `.nx-cell__sort-btn` out of this `.nx-cell` code block  as they no longer use `$edgespacing`. I thought for clarity sake it made sense to leave them with other button sizing adjustments for ends of rows.

EmilyBermudez88:
>Should the click target correspond with the background that changes on click/hover? It doesn't exactly right now, both in the sense that that background doesn't perfectly correspond with the bounds of the <button> and also that the click target is the whole cell rather than the button. I'm not sure we'd want to change the latter; it's just the onClick of NxTable.Cell which can of course be applicable in other situations besides sort headers. We probably want to be consistent about that prop applying to the whole <td> regardless of context.

lib/src/base-styles/_nx-tables.scss
```diff
@@ -365,29 +365,89 @@
...
+    &::before {
+      box-sizing: border-box;
+      border-radius: 4px;
+      content: '';
+      // offset by 1px on top and 2px on the bottom
+      height: calc(100% - 3px);
+      position: absolute;
+      left: 0;
+      top: 1px;
+      width: calc(100% - 1px);
```
rpokorny, line 386: I think this is too wide: 
image

lib/src/base-styles/_nx-tables.scss
```diff
@@ -365,29 +365,89 @@
...
+      &::before {
+        background-color: var(--nx-color-table-header-button-background-active);
+        border: none;
+      }
+    }
+  }
+  &:first-child {
+    .nx-cell__sort-btn {
+      &::before {
+        left: 1px;
```
rpokorny, line 409: Shouldn't this apply to all of them? I assume the hover visual should be inset 1px on the top, left, and right not matter which column it's in

EmilyBermudez88: > Shouldn't this apply to all of them? I assume the hover visual should be inset 1px on the top, left, and right (and 2px from the bottom as you have) no matter which column it's in

Yep, I confirmed this with Jason, my mistake. [f488f71](https://github.com/sonatype/sonatype-react-shared-components/pull/1156/commits/f488f71b61f1f4d8dad07cb12bc5981bbeead1eb)

lib/src/base-styles/_nx-tables.scss
EmilyBermudez88: > I think this is too wide:

Fixed here: [f488f71](https://github.com/sonatype/sonatype-react-shared-components/pull/1156/commits/f488f71b61f1f4d8dad07cb12bc5981bbeead1eb)

JasonDSwearingen (approved):

lib/src/base-styles/_nx-tables.scss
```diff
@@ -365,29 +365,74 @@
...
 
-.nx-cell__sort-btn {
-  height: nx-table-variables.$table-cell-header-height;
+.nx-cell {
+  .nx-cell__sort-btn {
+    height: nx-table-variables.$table-cell-header-height;
+    position: relative;
+  
+    > * {
+      position: relative;
```
rpokorny, line 374: Fairly sure this isn't needed.  The `.nx-cell__sort-btn` has three children:
1. the `::before`, which overrides this with `position: absolute`
2. A plain `span`, where I can't imagine the `position` mattering.
3. The `.nx-cell__sort-icons.fa-layers`, which already has `position: relative` set by `.fa-layers`.

rpokorny (approved):
+1 except for https://github.com/sonatype/sonatype-react-shared-components/pull/1156/files#r1290138885

lib/src/base-styles/_nx-tables.scss
```diff
@@ -365,29 +365,74 @@
   }
 }
 
-.nx-cell__sort-btn {
-  height: nx-table-variables.$table-cell-header-height;
+.nx-cell {
+  .nx-cell__sort-btn {
+    height: nx-table-variables.$table-cell-header-height;
+    position: relative;
+
```
rpokorny, line 372:
```suggestion

```

lib/src/base-styles/_nx-tables.scss
```diff
@@ -365,29 +365,74 @@
...
+      border-radius: 4px;
+      content: '';
+      // offset by 1px on top and 2px on the bottom
+      height: calc(100% - 3px);
+      position: absolute;
+      left: 1px;
+      top: 1px;
+      width: calc(100% - 2px);
+    }
+
```
rpokorny, line 388:
```suggestion

    > * {
      position: relative;
    }

    &::before {
      box-sizing: border-box;
      border-radius: 4px;
      content: '';
      // offset by 1px on top and 2px on the bottom
      height: calc(100% - 3px);
      position: absolute;
      left: 1px;
      top: 1px;
      width: calc(100% - 2px);
    }

```

lib/src/base-styles/_nx-tables.scss
rpokorny: > In this case I was using the * selector, but it is to help with the stacking context for the span. Because the ::before has position: absolute applied (and if the span has no positioning), is disrupts the stacking context and the ::before ends up covering the span element.

Ok I'm able to reproduce that now.  I think a cleaner solution (in that it avoids a `*` selector and styling a bare `<span>`) would be to add `z-index: 0` to the `.nx-cell__sort-btn` in order to make it a stacking context, and then to add `z-index: -1` to the `::before` so that it takes its correct place behind everything else.

EmilyBermudez88: > Ok I'm able to reproduce that now. I think a cleaner solution (in that it avoids a * selector and styling a bare <span>) would be to add z-index: 0 to the .nx-cell__sort-btn in order to make it a stacking context, and then to add z-index: -1 to the ::before so that it takes its correct place behind everything else.

Works and I agree, much cleaner: [ac13c94](https://github.com/sonatype/sonatype-react-shared-components/pull/1156/commits/ac13c94ee0d959ce2910a922fb9cb800e1a5b301)

rpokorny (approved):

PR #1157, opened 2023-08-10
Node 18 - RSC-1732

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1732


Jenkinsfile
```diff
@@ -20,6 +20,9 @@ dockerizedBuildPipeline(
   // expose gallery port and nextjs dev port on host so selenium container can hit it
   dockerArgs: '-p 4043:4043 -p 3000:3000',
 
+  // an npmrc file that's compatible with the npm from node 18
+  npmConfigFileId: 'rsc-ro-npmrc-v9',
```
EmilyBermudez88, line 24: I'm looking up what a Jenkinsfile is/ how to make one, but can't seem to find the `npmConfigFileId` property to read more about what it is. Is it's purpose is to point to a `npmrc` file or create one?

rpokorny: > I'm looking up what a Jenkinsfile is/ how to make one, but can't seem to find the npmConfigFileId property to read more about what it is. Is it's purpose is to point to a npmrc file or create one?

Notice that syntactically, what we are in right now is the parameter block to the `dockerizedBuildPipeline` function call.  So `npmConfigFileId` is a named parameter (technically a key in a map parameter) to that function call.  `dockerizedBuildPipeline` comes from Sonatype's internal `jenkins-shared` libary (imported at the top of this file).  I can never remember where the docs are for these functions, but the Build and Release Team manages them and can always point me in the right direction.  In fact this line was added at their suggestion.  But anyway, the meaning of the `npmConfigFileId` field is intrinsic to that internal` dockerizedBuildPipeline` function, and won't be found in general resources on the Internet.

EmilyBermudez88: Thanks for the explanation!

EmilyBermudez88 (approved):

PR #1162, opened 2023-08-17
Support JSX option name for NxCollapsibleItemsSelect - RSC-1283

EmilyBermudez88:
https://sonatype.atlassian.net/browse/RSC-1283

At this point, unit tests are passing and the gallery is compiling, but am getting an error with the a11y accessibility checks: "Ensure elements with an ARIA role that require child roles contain them", "Certain ARIA roles must contain particular children". The link the error refers to is : https://dequeuniversity.com/rules/axe/4.7/aria-required-children?application=axe-puppeteer. 
I suspect this has something to do with the either the `aria-describedby` set to the counter element, or the Radio element requiring its children to have phrasing content- which an `NxFontAwesomeIcon` doesn't fall within this category (but could be wrong). If that's the case, I am not sure how to proceed from here if we want to allow for a `ReactNode`s in `option.name`. 
Screenshot 2023-08-17 at 10 55 30 AM


rpokorny:
> I suspect this has something to do with the either the aria-describedby set to the counter element, or the Radio element requiring its children to have phrasing content- which an NxFontAwesomeIcon doesn't fall within this category (but could be wrong). If that's the case, I am not sure how to proceed from here if we want to allow for a ReactNodes in option.name.

The issue is that the svg has a `role` of img, which axe seems to think isn't allowed here.  As far as I can tell, this is because it's within a `<label>`.  Consider that Chrome shows the accessibility tree for the `<label>` and its children as follows (the highlighted element is the svg).  Notice that the svg is showing up as "Ignored" which would seem to conflict with it having `role="img"`.

rpokorny:
> The issue is that the svg has a role of img, which axe seems to think isn't allowed here. As far as I can tell, this is because it's within a <label>. Consider that Chrome shows the accessibility tree for the <label> and its children as follows (the highlighted element is the svg). Notice that the svg is showing up as "Ignored" which would seem to conflict with it having role="img".


Actually it's not that.  After reading [some comments on the axe implementation](https://github.com/dequelabs/axe-core/issues/3842#issuecomment-1385999084) I now understand that the `target` field in the error JSON tells us which element is the parent which is disallowing the child.  So it seems that `role="menu"` cannot have `role="img"` (or most other things other then specifically menu children) as children.  I'm investigating now how that is supposed to work with checkbox labels.

rpokorny:
> Actually it's not that. After reading https://github.com/dequelabs/axe-core/issues/3842#issuecomment-1385999084 I now understand that the target field in the error JSON tells us which element is the parent which is disallowing the child. So it seems that role="menu" cannot have role="img" (or most other things other then specifically menu children) as children. I'm investigating now how that is supposed to work with checkbox labels.

See the example here: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/ .  It seems the `menuitemcheckbox` role is really supposed to wrap the entire checkbox and its label.  So we need to move `role="menuitemcheckbox"` from the `<input>` to the label, and then we need to propagate the checked and disabled states from the `<input>` to the `label[role=menuitemcheckbox]` as `aria-checked` and `aria-disabled`.  And then potentially also add `role="none"` to the `<input>`.

gallery/src/components/NxCollapsibleMultiSelect/NxCollapsibleMultiSelectDisabledExample.tsx
```diff
@@ -36,7 +36,9 @@ const NxCollapsibleMultiSelectDisabledExample = () => {
   const [filter, setFilter] = useState('');
 
   function filterPredicate(option: NxCollapsibleMultiSelectOption) {
-    return includes(toLower(filter), toLower(option.name));
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const name = typeof option.name === 'string' ? option.name : (option.name as any).props.children[1].props.children;
```
rpokorny, line 40: Same as above, just cast it

gallery/src/components/NxCollapsibleMultiSelect/NxCollapsibleMultiSelectExample.tsx
```diff
@@ -45,7 +52,9 @@ const NxCollapsibleMultiSelectExample = () => {
   const [filter, setFilter] = useState('');
 
   function filterPredicate(option: NxCollapsibleMultiSelectOption) {
-    return includes(toLower(filter), toLower(option.name));
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const name = typeof option.name === 'string' ? option.name : (option.name as any).props.children[1].props.children;
```
rpokorny, line 56: Let's make a follow-up ticket to improve this. It should be possible to add some type magic to both remove the need to cast to string in the other examples, and make a better way to check the name in the JSX here.  But that doesn't need to be done today in RSC 13.0

gallery/src/components/NxCollapsibleRadioSelect/NxCollapsibleRadioSelectCustomTooltipExample.tsx
```diff
@@ -52,7 +52,9 @@ const NxCollapsibleRadioSelectCustomTooltipExample = () => {
   const [filter, setFilter] = useState('');
 
   function filterPredicate(option: NxCollapsibleRadioSelectOption) {
-    return includes(toLower(filter), toLower(option.name));
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const name = typeof option.name === 'string' ? option.name : (option.name as any).props.children[1].props.children;
```
rpokorny, line 56: Cast to string

gallery/src/components/NxCollapsibleMultiSelect/NxCollapsibleMultiSelectExample.tsx
EmilyBermudez88: > Let's make a follow-up ticket to improve this. It should be possible to add some type magic to both remove the need to cast to string in the other examples, and make a better way to check the name in the JSX here. But that doesn't need to be done today in RSC 13.0

Agreed, this is hardly ideal. Ticket made: https://sonatype.atlassian.net/browse/RSC-1736

lib/src/components/NxCollapsibleItemsSelect/commonTypes.ts
```diff
@@ -34,7 +35,10 @@ export interface CommonProps<T extends Option = Option> {
 
 export const optionPropType = PropTypes.shape({
   id: requiredNullableString,
-  name: PropTypes.string.isRequired
+  name: PropTypes.oneOfType([
+    PropTypes.string.isRequired,
+    PropTypes.node.isRequired
+  ]).isRequired
```
rpokorny, line 41: Pretty sure node includes string
```suggestion
  name: PropTypes.node.isRequired
```

gallery/visualtests/testUtils.js
```diff
@@ -394,7 +394,12 @@ module.exports = {
 
           // the color contrast checker seems to be buggy, it has many complaints about overlapping items when
           // there aren't any
-          const builder = new AxePuppeteer(page).disableRules('color-contrast'),
+          // Aria does not allow for an element with role=menu to have any children with role=img (amongst others),
+          // such as is the case for NxCollapsibleRadioSelect & NxCollapsibleMultiSelect. Seeing that the role of
+          // ‘menuitemcheckbox’ cannot be assigned to the <label> element (as guidelines suggest 'menuitemcheckbox' to
+          // wrap the checkbox & its label), and an input[type=checkbox/radio] cannot be assigned role=none, we need to
+          // override this error
+          const builder = new AxePuppeteer(page).disableRules(['color-contrast', 'aria-required-children']),
```
rpokorny, line 402: We only want to disable `aria-required-children` in the collapsible*Select tests.  Doing it here disables it everywhere.

rpokorny (approved):
+1 aside from https://github.com/sonatype/sonatype-react-shared-components/pull/1162#discussion_r1300384449

gallery/visualtests/testUtils.js
```diff
@@ -394,7 +394,7 @@ module.exports = {
 
           // the color contrast checker seems to be buggy, it has many complaints about overlapping items when
           // there aren't any
-          const builder = new AxePuppeteer(page).disableRules('color-contrast'),
+          const builder = new AxePuppeteer(page).disableRules(['color-contrast']),
```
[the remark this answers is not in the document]
EmilyBermudez88: > Minor: remove the brackets so this file is gone from the diff entirely

[95ce24a](https://github.com/sonatype/sonatype-react-shared-components/pull/1162/commits/95ce24a15f03ddde6e91d145718dd621addbcceb)

PR #1168, opened 2023-08-23
avoid typecasting for NxCollapsibleItemsSelect options- RSC-1736

EmilyBermudez88:
Ticket: https://sonatype.atlassian.net/browse/RSC-1736
This stems from already completed ticket: https://sonatype.atlassian.net/browse/RSC-1283.


lib/src/components/NxCollapsibleItemsSelect/commonTypes.ts
```diff
@@ -4,15 +4,14 @@
...
+import {ReactNode, ReactElement} from 'react';
 import * as PropTypes from 'prop-types';
 
 import { requiredNullableString } from '../../util/customPropTypes';
-import RequiredReactNode from '../../util/RequiredReactNode';
 
 export interface Option {
   id: string | null;
-  name: RequiredReactNode;
+  name: string | ReactElement;
```
EmilyBermudez88, line 14: Since the initial ticket https://sonatype.atlassian.net/browse/RSC-1283 was looking for IQ to have the ability to add in JSX as an option.name, what are your thoughts of changing the `name` type to `string | ReactElement` instead of `ReactNode`? (since we won't need number, boolean etc). This cleans up the casting of `option.any` required wherever we need to use the `Option` interface with the name typed as `RequiredReactNode`

EmilyBermudez88: If this is an acceptable route forward, I'd need to make a slight adjustment to the gallery docs as well

EmilyBermudez88:
So currently trying to implement the cleaner suggested solution for the stateful components. I have tried to allow the components in the lib to take a generic type, and updated the `types.ts` files for each stateful component to have the `optionTooltipGenerator` take the generic as well. What I've done so far is here: 
[e9b2397](https://github.com/sonatype/sonatype-react-shared-components/pull/1168/commits/e9b23979a3a8164d8af27aa34717b5619b5ff9af)

EmilyBermudez88:
So currently trying to implement the cleaner suggested solution for the stateful components. I have tried to allow the components in the lib to take a generic type, and updated the `types.ts` files for each stateful component to have the `optionTooltipGenerator` take the generic as well. What I've done so far is here: 
[e9b2397](https://github.com/sonatype/sonatype-react-shared-components/pull/1168/commits/e9b23979a3a8164d8af27aa34717b5619b5ff9af). 
However, I am still getting typescript errors for optionTooltipGenerator: 
`Type '(option: Option) => string' is not assignable to type '(option: import("/Users/emilybermudez/Desktop/RSC/sonatype-react-shared-components/gallery/node_modules/@sonatype/react-shared-components/components/NxCollapsibleItemsSelect/commonTypes").Option) => string'.
  Types of parameters 'option' and 'option' are incompatible.
    Property 'stringName' is missing in type 'import("/Users/emilybermudez/Desktop/RSC/sonatype-react-shared-components/gallery/node_modules/@sonatype/react-shared-components/components/NxCollapsibleItemsSelect/commonTypes").Option' but required in type 'Option'.ts(2322)
NxStatefulCollapsibleRadioSelectExample.tsx(17, 3): 'stringName' is declared here.
types.d.ts(15, 5): The expected type comes from property 'optionTooltipGenerator' which is declared here on type 'IntrinsicAttributes & Props<Option>'`. Any suggestions as to where I've gone wrong?

rpokorny:
>  Any suggestions as to where I've gone wrong?

The `optionTooltipGenerator` within `NxStatefulCollapsibleMultiSelect` fails to parameterize the type of its parameter.  That is, it should be `T` not `Option`.

rpokorny:
> The optionTooltipGenerator within NxStatefulCollapsibleMultiSelect fails to parameterize the type of its parameter. That is, it should be T not Option.

However as I test fixing that myself I see that it leads to other challenges.  I'll look into it

EmilyBermudez88:
I believe I've cleaned up the extra code here: [fa7a065](https://github.com/sonatype/sonatype-react-shared-components/pull/1168/commits/fa7a065160f97b991b160207bea77fd37131ccc9)

gallery/src/components/NxCollapsibleMultiSelect/NxCollapsibleMultiSelectCustomTooltipExample.tsx
```diff
@@ -16,6 +16,7 @@ import {
 } from '@sonatype/react-shared-components';
 
 interface CustomOption extends NxCollapsibleMultiSelectOption {
+  name: string;
```
rpokorny, line 19: Let's add a comment explaining why we're including `name` here: to narrow its type to string

rpokorny: > Let's add a comment explaining why we're including name here: to narrow its type to string

And let's add that same comment to the other examples that do the same thing.

rpokorny (approved):
Needs a version bump but otherwise +1.  I think it should be a minor version as technically this does change the API by adding the type parameters.

PR #1174, opened 2023-11-02
Add underline to NxBackButton for consistency - RSC-1741

antoniolokchan:
https://sonatype.atlassian.net/browse/RSC-1741

I will bump the version before merging the PR


lib/src/components/NxBackButton/NxBackButton.scss
```diff
@@ -13,6 +13,6 @@
 
   .nx-text-link {
     @include nx-container-helpers.container-horizontal;
-    text-decoration: none;
+    text-decoration: underline;
```
rpokorny, line 16: I notice that `.nx-text-link`'s styles don't explicitly give it an underline but rather rely on the browser's default styles.  I'm not sure that's a great idea but at any rate I don't think that `NxBackButton` should be declaring its _own_ underline, it should just follow from `NxTextLink`.  That is, I think this line should be removed and we should consider explicitly adding `text-decoration: underline` to all `.nx-text-link`.  @glenhunter wdyt?

glenhunter: > That is, I think this line of code should be removed and we should consider explicitly adding `text-decoration: underline` to all `.nx-text-link`.

That makes the most sense to me as well.

rpokorny (approved):

PR #1177, opened 2023-11-10
Update color picker hover colors - RSC-1547

Deniro5:
https://sonatype.atlassian.net/jira/software/c/projects/RSC/boards/207?selectedIssue=RSC-1547


rpokorny:
Unfortunately, now that the hover color is different from the regular color, we need a separate visual test for each hover color.

lib/src/scss-shared/_nx-color-picker-variables.scss
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+
```
rpokorny, line 7: I see that this follows the same pattern as the tag variables and that's good.  However, I'm now realizing that that pattern violations one of the rules of our folder organization.  See the following from the readme in this directory:
> The files in this directory must not emit anything on their own when compiled, as anything that they
do emit would be duplicated every time one of the files is imported

We should create a separate ticket to clean this up; but continue as-is for this PR.

PR #1180, opened 2024-07-24
Remove inline styles from fontawesome for CSP header compatibility - RSC-1763

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1763

This PR adjusts our fontawesome configuration so that it does not inject or rely on inline CSS styles, as these are incompatible with CSP header best practices.  Note that there are other libraries (Material-UI) that also make us incompatible with CSP header best practices, which will need to be fixed in future work.


lib/src/base-styles/_base.scss
```diff
@@ -9,6 +9,8 @@
  * any consumer of the react-shared-components lib
  */
 
+@use '~@fortawesome/fontawesome-svg-core/styles';
```
rpokorny, line 12: Important, open question: how comfortable are we that this syntax is supported in all supported downstream build configurations? It should work anywhere a typical webpack sass-loader setup is being used, as far as I know.

gallery/visualtests/__image_snapshots__/nx-date-input-js-nx-date-input-simple-nx-date-input-has-a-blue-inner-outline-and-darker-border-when-hovered-and-focused-1.png
rpokorny, line 1: As reviewers may recall, our chrome version in which we run visual tests in unfortunately a moving target.  It seems the newer version has changed the highlight color since the last time any work on RSC was done.

marushkevych (approved):

PR #1182, opened 2024-08-20
Fix fontawesome warning in new react - RSC-1768

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1768

Fixes 
```react-dom.development.js:86 Warning: FontAwesomeIcon: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.```
as seen in downstream projects running React 18.3.1

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -89,8 +89,8 @@
     "prop-types": "^15.8.1",
     "query-string": "^8.1.0",
     "ramda": "^0.29.0",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: gallery/package.json [hunk 1]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "13.1.8",
+  "version": "13.2.0",
```
rpokorny, line 3: Since React is so core of a library, I consider upgrading it to be more than a patch change.

PR #1188, opened 2024-09-18
Allow placeholder prop at top level for NxTextInput and NxStatefulTextInput for later typescript versions

mealingr:
This is to fix a typescript issue where `placeholder` was not being allowed as a top level property on `NxTextInput` or `NxStatefulTextInput` i.e. when using `"typescript": "^5.5.4"` we were seeing this error

```
TS2322: Type
...
Property placeholder does not exist on type
IntrinsicAttributes & PublicProps & RefAttributes<HTMLDivElement>
```


rpokorny:
I don't suppose you tried upgrading typescript in this repo to reproduce the issue here?  If that doesn't cause other issues I'd be in favor of including that upgrade in this PR.

lib/src/components/NxTextInput/types.ts
```diff
@@ -34,8 +34,8 @@ export interface StateProps {
 // Imported props from react to provide support for all the regular html attributes
 type FusedProps = InputHTMLAttributes<HTMLInputElement> & TextareaHTMLAttributes<HTMLTextAreaElement>;
 // Leave out props to be re-defined
-export type InputAttrs = Omit<FusedProps, 'onChange'|'onKeyPress'|'type'|'defaultValue'|'disabled'>;
-export type DivAttrs = Omit<HTMLAttributes<HTMLDivElement>, 'onChange'|'onKeyPress'|'defaultValue'>;
+export type InputAttrs = Omit<FusedProps, 'onChange'|'onKeyPress'|'type'|'defaultValue'|'disabled'|'placeholder'>;
+export type DivAttrs = Omit<HTMLAttributes<HTMLDivElement>, 'onChange'|'onKeyPress'|'defaultValue'|'placeholder'>;
```
rpokorny, line 38: `placeholder` isn't a `<div>` attribute in the first place, so the change on this line seems unnecessary.

mealingr: > `placeholder` isn't a `<div>` attribute in the first place, so the change on this line seems unnecessary.

You're right I was able to take that off https://github.com/sonatype/sonatype-react-shared-components/pull/1188/commits/44ab170a697c7dd60d2568d4b97d498529c2d418, I had to change it to `string | undefined;` otherwise it gave this complaint
image
```
TS2322: Type
{ onChange: (newValue: string) => void; value: string; trimmedValue: string; isPristine: boolean; validationErrors?: ValidationErrors | undefined; type?: "textarea" | ... 4 more ... | undefined;  ... 267 more ...; ref: ForwardedRef<...>; }
is not assignable to type DivAttrs
Types of property placeholder are incompatible.
Type string | null | undefined is not assignable to type string | undefined
Type null is not assignable to type string | undefined
```

rpokorny: > You're right I was able to take that off

Oh huh that error seems to imply that `placeholder` _is_ a member of `HTMLAttributes<HTMLDivElement>`.  In that case what you had before was more correct.  In RSC, all optional props can accept `undefined _or_ `null`, and we want that ability on this prop too.

mealingr: > Oh huh that error seems to imply that `placeholder` _is_ a member of `HTMLAttributes<HTMLDivElement>`. In that case what you had before was more correct. In RSC, all optional props can accept `undefined` _or_ `null`, and we want that ability on this prop too.

It looks like it's in there by mistake and has been removed in https://github.com/DefinitelyTyped/DefinitelyTyped/commit/b954269038de46b4b2f1756a9f2f020cbc66a326 but not realeased yet if I'm reading that right

rpokorny: > It looks like it's in there by mistake and has been removed in https://github.com/DefinitelyTyped/DefinitelyTyped/commit/b954269038de46b4b2f1756a9f2f020cbc66a326 but not released yet if I'm reading that right

Huh good to know

rpokorny (approved):

PR #1191, opened 2024-09-19
CSDI-2092 Remove deprecated ~                                                                                               

rpanman-sonatype:
This change is for two reasons:
1. Building of RSC was failing in Vite environment with `Can't find stylesheet to import`
1. Webpack has deprecated `~` in `@use` (see [resolving import at rules](https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules)).

This PR addresses both these issues

### DIFF
--- a/gallery/package.json
+++ b/gallery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "13.4.0",
+  "version": "13.3.1",
   "description": "Gallery application to demonstrate usage and look of Sonatype shared UI components",
   "main": "src/main.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


gallery/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sonatype/react-shared-components-gallery",
-  "version": "13.3.0",
+  "version": "13.4.0",
```
rpokorny, line 3: IMO this is a patch release not a minor. 
```suggestion
  "version": "13.3.1",
```
(and change the other file too)

rpokorny (approved):
+1 after fixing the version number, and after CI passes.  Thanks for the contribution!

PR #1196, opened 2024-10-03
New Global Sidebar - RSC-1781

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1781


lib/src/base-styles/_nx-page-layout.scss
```diff
@@ -35,11 +35,11 @@
...
-    'old-header     old-header  old-header' auto
-    'global-sidebar   notices      notices' auto
-    'global-sidebar   header        header' auto
-    'global-sidebar old-sidebar    content' 1fr /
-     auto    $old-sidebar-col-width    1fr;
+    'old-header        old-header          old-header  old-header' auto
+    'global-sidebar     notices             notices       notices' auto
+    'global-sidebar      header              header        header' auto
+    'global-sidebar global-sidebar-2       old-sidebar    content' 1fr /
+     auto                 auto       $old-sidebar-col-width   1fr;
```
mealingr, line 42: Any chance you could explain these changes? (I'm somewhat struggling to follow how this grid-template influences the layout of the new sidebar)

mealingr (approved):
+1 nice I've been comparing the live examples to https://www.figma.com/design/UpWZYPAaaOwafLeYytZ99l/Updated-Side-and-Top-Nav-Component---Q2-2024---JS?node-id=0-1&node-type=canvas&t=YH0r7DXqMgXbIDNh-0 and they look great

lib/src/base-styles/_nx-page-layout.scss
rpokorny: > Any chance you could explain these changes? (I'm somewhat struggling to follow how this grid-template influences the layout of the new sidebar)

So our overall template has 4 rows and 4 columns with these changes, though not all of them would be used at the same time.  The new sidebar inhabits the grid area named "global-sidebar-2".  In proper usage, you wouldn't also have the old global sidebar, the column to the left of this would be empty and thus sized to zero.  So it would have the header and notices above it, and the inner sidebar and main content area to its right.  Note that the old global sidebar, which uses the "global-sidebar" grid area, goes up alongside the header and notices, not underneath them like the new global sidebar.  Also note that in this template, "old-header" means the _old old_ header for the legacy page layouts (as seen in the layout of the RSC gallery itself), not the soon-to-be-old global sidebar header.

rpokorny: To be even more clear, what these changes add is the second column in the template, which has a width of `auto` meaning it will size to its content or zero if it has no distinct content (when the new sidebar is not being used).

mealingr: Thanks for the explanation that's much clearer (the sizing to 0 when not in-use was the main thing I was missing :+1:  )

hjbarbieri (approved):

JasonDSwearingen:
My feedback:
1. Icons and text should be nx-indigo-20
2. The arrow icon for expand/collapse menu is inconsistent with the mocks. Is this a FA version and/or RSC version issue?
3. The margins at the top of the list don’t overlap, so there is too much space above the top link item.

rpokorny:
> The arrow icon for expand/collapse menu is inconsistent with the mocks. Is this a FA version and/or RSC version issue?

Yeah we've discussed that, the mocks use pro FA icons which RSC can't use.  So the arrow icons used in these examples are demonstrative only.  A closed-source app that does have access to FA pro should specify the icons that match the mocks.

JasonDSwearingen:
Excellent. I will make note of that in the Figma. I don't have the PR up anymore, but if it's not mentioned in the documentation, can you note it there, too?

PR #1197, opened 2024-10-04
RSC-1757 Update Color Variables

snh9905:
Updated color variables based on [this design in figma](https://www.figma.com/design/SMeWSFCKTQbyarldXKVzro/Kaleido-V2?node-id=7212-875&node-type=frame&t=hZw8fo8mbfOFMut4-0). 

Jira: https://sonatype.atlassian.net/browse/RSC-1757


lib/src/base-styles/_nx-color-swatches.scss
```diff
@@ -81,8 +136,29 @@
   --nx-swatch-pink-95: hsl(var(--nx-swatch-pink-hs), 95%);
   --nx-swatch-pink-97: hsl(var(--nx-swatch-pink-hs), 97%);
 
+  // Red
+  --nx-swatch-red-hs:0, 100%;
```
rpokorny, line 140: Saturation here is supposed to be 80%

rpokorny:
Since you're replacing all of the snapshots, I need to ask you to do something extra diligent with them:  read and understand the comment [here](https://github.com/sonatype/sonatype-react-shared-components/blob/main/gallery/setupTests.js#L10), and then temporarily decrease that `retryTimes` value to, say 4.  Re-run the visual tests and attempt to identify tests where the newly accepted snapshot wasn't correct and needs to be updated (notably, there are probably some new snapshots in your current set that have partially visible tooltips. This will help identify those).  If you can, try to get it down to where only a few tests per run don't pass with `retryTimes = 4`.  I do expecte it to be necessary to increase the `retryTimes` again when you're done in order to get ongoing build stability (though if somehow you don't have to do that then great!).

lib/src/base-styles/_nx-color-swatches.scss
snh9905: > Saturation here is supposed to be 80%

Good catch. FIxed here: [c86d8c2](https://github.com/sonatype/sonatype-react-shared-components/pull/1197/commits/c86d8c2fc0500daf71260954824476501fda904a)

rpokorny (approved):

rpokorny (approved):

PR #1199, opened 2024-10-08
RSC-1772: Vulnerability Indicator and Vulnerability Indicator Legend

ValeriaRios02:
Jira ticket: https://sonatype.atlassian.net/browse/RSC-1769

Subtasks: 
 - https://sonatype.atlassian.net/browse/RSC-1772
 - https://sonatype.atlassian.net/browse/RSC-1773


lib/src/components/NxVulnerabilityIndicator/types.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import {
+  VulnerabilityScore,
+  VulnerabilitySeverityRating,
+  allVulnerabilitySeverityRatings
+} from '../../util/vulnerabilitySeverityRatings';
+
+export interface Props {
+  severityRating?: VulnerabilitySeverityRating;
+  score?: VulnerabilityScore;
```
rpokorny, line 18: I think I'd rather this be `number` to more closely match the interface for `NxThreatIndicator`.  I know we can't cleanly lock it down to a discrete set of numbers like we do there, but I'm ok with that (as long as its checked at runtime).

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <NxTable.Body>
+          <NxTable.Row>
+            <NxTable.Cell>severityRating</NxTable.Cell>
+            <NxTable.Cell>One of 'none', 'low', 'medium', 'high', or 'critical'</NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>A Vulnerability Severity Rating off of which to base the indicator color</NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell>score</NxTable.Cell>
+            <NxTable.Cell>0.0{'\u2014'}10.0 inclusive</NxTable.Cell>
```
rpokorny, line 54:
```suggestion
            <NxTable.Cell>0.0—10.0 inclusive</NxTable.Cell>
```

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+    nxVulnerabilityIndicatorListCode = require('./NxVulnerabilityIndicatorListExample?raw'),
+    nxVulnerabilityIndicatorTableCode = require('./NxVulnerabilityIndicatorTableExample?raw'),
+    nxVulnerabilityIndicatorWithTitleCode = require('./NxVulnerabilityIndicatorWithTitleExample?raw');
+
+const NxVulnerabilityIndicatorPage = () =>
+  <>
+    <GalleryDescriptionTile>
+      <NxP>
+        <NxCode>NxVulnerabilityIndicator</NxCode> is an inline element used
+        to indicate via color the IQ policy vulnerability rating of the information to follow that follows it.
```
rpokorny, line 29:
```suggestion
        to indicate via color the CVSS severity rating of the information that follows it.
```

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+      </NxP>
+      <NxP>
+        There are two scales to choose from: vulnerability rating by category, and vulnerability
+        rating by score. When using this component, it is expected that just one of the props will be passed. If both
```
rpokorny, line 33: Category doesn't mean anything here, so this needs to be reworded.  How about something like

> The severity rating can either be specified directly, or a CVSS score may be specified from which the component will compute the rating.

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <NxTable.Cell>0.0{'\u2014'}10.0 inclusive</NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>A Vulnerability (CVSS) Score off of which to base the indicator color</NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell>title</NxTable.Cell>
+            <NxTable.Cell>string</NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>
+              Specifies tooltip content, by default, it is set to the Vulnerability Severity Rating
```
rpokorny, line 63:
```suggestion
              Specifies tooltip content. By default, it is set to the Vulnerability Severity Rating
```
I think this also needs slightly more elaboration as the tooltip content isn't _just_ the rating name.

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+          <NxTable.Row>
+            <NxTable.Cell>presentational</NxTable.Cell>
+            <NxTable.Cell>boolean</NxTable.Cell>
+            <NxTable.Cell>No</NxTable.Cell>
+            <NxTable.Cell>
+              When this is true, the tooltip will be hidden, aria-label will not be set,
+              and the icon's role will be set to "presentation".
+              This should be set to true when <NxCode>NxVulnerabilityIndicator</NxCode> is used alongside
+              visible text that repeats its meaning.
+              For example, In the case of Critical <NxCode>NxVulnerabilityIndicator</NxCode> with "Critical"
```
rpokorny, line 81:
```suggestion
              For example, In the case of a Critical <NxCode>NxVulnerabilityIndicator</NxCode> with "Critical"
```

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendHorizontalExampleHalf.tsx
```diff
@@ -0,0 +1,18 @@
...
+ */
+import React from 'react';
+import { NxVulnerabilityIndicatorLegend } from '@sonatype/react-shared-components';
+
+const NxVulnerabilityIndicatorLegendHorizontalExample = () => (
+  <NxVulnerabilityIndicatorLegend header="Vulnerability Legend"
+                                  critical
+                                  high
+                                  medium
+    />
```
rpokorny, line 15: One line

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+        <NxCode>NxVulnerabilityIndicatorLegend</NxCode> is an element to display the legend for
+        various vulnerability ratings. The legend can be in two formats: horizontal and vertical.
+      </NxP>
+      <NxInfoAlert>
+        By default, <NxCode>NxVulnerabilityIndicatorLegend</NxCode> renders a horizontal legend.
+        To render a vertical legend, the prop <NxCode>vertical</NxCode> needs to be provided.
+      </NxInfoAlert>
+      <NxP>
+        <NxCode>NxVulnerabilityIndicatorLegend</NxCode> supports displaying the legend for
+        'unspecified', 'none', 'low', 'moderate', 'severe', and 'critical' vulnerability ratings.
```
rpokorny, line 43:
```suggestion
        'none', 'low', 'moderate', 'severe', and 'critical' vulnerability ratings.
```

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <NxTable.Cell>Displays a vertical legend.</NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
+            <NxTable.Cell>critical</NxTable.Cell>
+            <NxTable.Cell>boolean</NxTable.Cell>
+            <NxTable.Cell>no</NxTable.Cell>
+            <NxTable.Cell>false</NxTable.Cell>
+            <NxTable.Cell>Displays the legend item for critical vulnerability items.</NxTable.Cell>
+          </NxTable.Row>
+          <NxTable.Row>
```
rpokorny, line 77: A bunch of these still have the policy threat levels, and need to be updated

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendHorizontalExampleHalf.tsx
rpokorny: > One line

And likewise in the other examples

gallery/src/pageConfig.ts
```diff
@@ -183,7 +187,10 @@ const pageConfig: PageConfig = {
     'Status Indicator': { content: NxStatusIndicatorPage, type: 'html' },
     'Threat Indicator': { content: NxThreatIndicatorPage, type: 'react' },
     'Threat Indicator Legend': { content: NxThreatIndicatorLegendPage, type: 'react' },
+    'Vulnerability Indicator': { content: NxVulnerabilityIndicatorPage, type: 'react' },
+    'Vulnerability Indicator Legend': { content: NxVulnerabilityIndicatorLegendPage, type: 'react' },
     'Threat Number': { content: NxThreatNumberPage, type: 'html' },
+    'Vulnerability Score': { content: NxVulnerabilityScorePage, type: 'html' },
```
rpokorny, line 193: Move all of the vuln components down to where they'd normally sort under the letter V

gallery/src/styles/NxVulnerabilityScore/NxVulnerabilityScoreListExample.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <NxVulnerabilityIndicator severityRating="low" />
+      <NxList.Text>
+        <span className="nx-vulnerability-score">{formatVulnerabilityScore(1)}</span> is the vulnerability (CVSS) score.
+      </NxList.Text>
+      <NxList.Subtext>
+        In this row, there is other content aside from the vulnerability score value within
+        the <code className="nx-code">.nx-list__text</code>, so a separate element is needed in order
+        to ensure that the <code className="nx-code">.nx-vulnerability-score</code> styles are applied only to the
+        vulnerability score value itself. Note that as of RSC 2.0.0, nx-vulnerability-score does not actually
+        have any styles that are distinct from the rest of the main list text. That may change in the future however, so
```
rpokorny, line 33: Talking about an ancient version of RSC that predates this component doesn't make a ton of sense.
```suggestion
        vulnerability score value itself. nx-vulnerability-score does not actually
        have any styles that are distinct from the rest of the main list text. That may change in the future however, so
```

gallery/src/styles/NxVulnerabilityScore/NxVulnerabilityScoreTableExample.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+import React from 'react';
+
+import { NxVulnerabilityIndicator, formatVulnerabilityScore } from '@sonatype/react-shared-components';
+
+const NxVulnerabilityNumberTableExample = () =>
```
rpokorny, line 11:
```suggestion
const NxVulnerabilityScoreTableExample = () =>
```

gallery/visualtests/nxVulnerabilityIndicator.js
```diff
@@ -0,0 +1,27 @@
...
+      vulnerabilityIndicatorsScoreExampleSelector =
+        '#nx-vulnerability-indicator-score-example .gallery-example-live';
+
+  it('looks right', simpleTest(vulnerabilityIndicatorsSimpleExampleSelector));
+  it('looks right in a list', simpleTest(vulnerabilityIndicatorsListExampleSelector));
+  it('looks right in a table', simpleTest(vulnerabilityIndicatorsTableExampleSelector));
+  it(
+      'looks right when score prop is provided',
+      simpleTest(vulnerabilityIndicatorsScoreExampleSelector)
+  );
```
rpokorny, line 24: One line

gallery/visualtests/nxVulnerabilityIndicator.js
```diff
@@ -0,0 +1,27 @@
...
+const { setupBrowser } = require('./testUtils');
+
+describe('NxVulnerabilityIndicator', function() {
+  const { simpleTest, a11yTest } = setupBrowser('#/pages/Vulnerability Indicator');
+
+  const vulnerabilityIndicatorsListExampleSelector = '#nx-vulnerability-indicator-list-example .nx-list',
+      vulnerabilityIndicatorsTableExampleSelector = '#nx-vulnerability-indicator-table-example .nx-table',
+      vulnerabilityIndicatorsSimpleExampleSelector = '#nx-vulnerability-indicator-simple-example .gallery-example-live',
+      vulnerabilityIndicatorsScoreExampleSelector =
+        '#nx-vulnerability-indicator-score-example .gallery-example-live';
```
rpokorny, line 16:
```suggestion
      vulnerabilityIndicatorsScoreExampleSelector = '#nx-vulnerability-indicator-score-example .gallery-example-live';
```

gallery/visualtests/nxVulnerabilityIndicator.js
rpokorny: There've been a lot of weird line breaks in these PRs.  Considering changing your text editor's settings.

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2019-present Sonatype, Inc.
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ @use '../../scss-shared/nx-dark-mode-helpers';
+ 
+// eyeballed, assumes $nx-font-size surrounding text
+$-vertical-alignment: -0.1em;
```
rpokorny, line 10: I'll want to come back to this once I see the mockups.  I suspect it is not correct

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,47 @@
...
+.nx-vulnerability-indicator {
+  font-size: var(--nx-font-size-s);
+  margin: 0 var(--nx-spacing-2x);
+  vertical-align: $-vertical-alignment;
+}
+
+.nx-list__item {
+  .nx-vulnerability-indicator {
+    -ms-grid-row: 1;
+    -ms-grid-column: 1;
```
rpokorny, line 21: We don't support the browsers that would require these anymore
```suggestion
```

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
```diff
@@ -0,0 +1,63 @@
...
+          severityRating,
+          score,
+          className: classNameProp,
+          presentational
+        } = props,
+        severity: VulnerabilitySeverityRating = severityRating
+          ? severityRating
+          : score !== undefined
+            ? getVulnerabilitySeverityRating(score) || 'none'
+            : 'none',
```
rpokorny, line 38: Format this the way it was in the other one.  In that formatting it flows:
```
cond1 ? result1 :
cond2 ? result2 :
…
baseCase
```

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const NxVulnerabilityIndicator: FunctionComponent<Props> =
+  function NxVulnerabilityIndicator(props): ReactElement<Props> {
+    const {
+          title,
+          severityRating,
+          score,
+          className: classNameProp,
+          presentational
+        } = props,
```
rpokorny, line 33: One line

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
ValeriaRios02: That seems right, thank you!

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const icon = (
+      <NxFontAwesomeIcon aria-label={presentational ? undefined : label}
+                         className={className}
+                         icon={faSquare}
+                         role={presentational ? 'presentation' : 'graphics-symbol img'} />
+    );
+
+    return presentational ? icon : (
+      <NxTooltip title={
+        title || `Vulnerability: ${capitalize(severity)} (${vulnerabilitySeverityScoreRangeMap[severity]})`}
```
rpokorny, line 54: This is too complex and long to put inside the JSX like this.  Extract it to a local variable.

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  it('has a fallback role of img', async function() {
+    const el = quickRender();
+    await runTimers();
+    const vulnerabilityIndicator = el.getByRole('img', { queryFallbacks: true });
+    expect(vulnerabilityIndicator).toBeInTheDocument();
+  });
+
+  it('takes precedence of severityLevel if both props are provided', async function() {
```
rpokorny, line 28: Something's wrong if this is passing; the prop is called `severityRating`

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const user = userEvent.setup(),
+        vulnerabilityIndicator = el.getByRole('graphics-symbol')!;
+
+    await user.hover(vulnerabilityIndicator);
+    const tooltip = await screen.findByRole('tooltip');
+    expect(tooltip).toHaveTextContent('Low');
+  });
+
+  it('sets the accessible name based on the severity level', async function() {
+    const el = quickRender({ score: '9.0', severityLevel: 'low' });
```
rpokorny, line 40: level -> rating

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      await runTimers();
+      const user = userEvent.setup(),
+          vulnerabilityIndicator = el.getByRole('graphics-symbol');
+
+      await user.hover(vulnerabilityIndicator);
+      const tooltip = await screen.findByRole('tooltip');
+      return tooltip.textContent;
+    };
+
+    describe('when vulnerabilityScore prop is provided', function() {
```
rpokorny, line 59: Fewer tests of values in the middle of each range, more tests of values at the edges of each range.

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      });
+      it('for vulnerability score 9.0', async function() {
+        expect(await getTooltipTextForProps({ score: '9.0' })).toBe('Critical');
+      });
+      it('for vulnerability score 10.0', async function() {
+        expect(await getTooltipTextForProps({ score: '10.0' })).toBe('Critical');
+      });
+    });
+
+    describe('when vulnerabilitySeverityLevel prop is provided', function() {
```
rpokorny, line 95: Rating.

lib/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegend.scss
```diff
@@ -0,0 +1,39 @@
...
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+@use '../../scss-shared/nx-container-helpers';
+@use '../../scss-shared/nx-text-helpers';
+@use '../../scss-shared/nx-dark-mode-helpers';
+
+.nx-vulnerability-indicator-legend {
+  $max-grid-columns: 6;
```
rpokorny, line 12: Shouldn't this be 5 now

lib/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegend.tsx
```diff
@@ -0,0 +1,66 @@
...
+        critical,
+        high,
+        medium,
+        low,
+        none,
+        vertical,
+        header,
+        className,
+        ...attrs
+      } = props;
```
rpokorny, line 40: One line

lib/src/components/NxVulnerabilityIndicatorLegend/__tests__/NxVulnerabilityIndicatorLegend.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+      high: null,
+      medium: null,
+      low: null,
+      none: null
+    });
+    expect(el).not.toBeInTheDocument();
+  });
+
+  it('renders correct legend items', function() {
+    const el = quickRender({ critical: true,
```
rpokorny, line 73:
```suggestion
    const el = quickRender({ 
      critical: true,
```

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
ValeriaRios02: Updated here: [07faf88](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/07faf8866585f8a06cc8e83275201a80e1b2d556)

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+          </NxTable.Row>
+        </NxTable.Body>
+      </NxTable>
+
+      <NxH3>
+        formatVulnerabilityScore Utility
+      </NxH3>
+      <NxP>
+        To help ensuring consistency, we provide a utility{' '}
+        <NxTextLink href="#/pages/Vulnerability Rating Utils/formatVulnerabilityScoreExample">
```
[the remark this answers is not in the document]
ValeriaRios02: Updated the link on: [5f1cc5e](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/5f1cc5e4ec31498503c505cc4b83e2d28126d622)

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendHorizontalExampleHalf.tsx
ValeriaRios02: Fixed in: [2af62e0](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/2af62e0a71426434c93b12cf760ce6d59db8989b)

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendPage.tsx
ValeriaRios02: Updated in: [cc2bfac](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/cc2bfac71027ac309f17c387261135fb65f7eedd)

gallery/src/pageConfig.ts
ValeriaRios02: Updated here: [0ea3129](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/0ea31298e43d965f4d28e9acd0e52459dbc87b44)

gallery/src/styles/NxVulnerabilityScore/NxVulnerabilityScoreTableExample.tsx
ValeriaRios02: Sorry about that, fixed here: [f460489](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/f460489a77bb351c6e936b902dc3f9a1e212de74)

gallery/visualtests/nxVulnerabilityIndicator.js
ValeriaRios02: Thank for pointing that out, fixed in: [a32275b](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/a32275bd3f95f5dc52a205829c697b3fd709a186)

gallery/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegendHorizontalExample.tsx
```diff
@@ -0,0 +1,19 @@
...
+import React from 'react';
+import { NxVulnerabilityIndicatorLegend } from '@sonatype/react-shared-components';
+
+const NxVulnerabilityIndicatorLegendHorizontalExample = () => (
+  <NxVulnerabilityIndicatorLegend critical
+                                  high
+                                  medium
+                                  low
+                                  none
+    />
```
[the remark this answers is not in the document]
ValeriaRios02: Fixed in: [2af62e0](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/2af62e0a71426434c93b12cf760ce6d59db8989b)

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
ValeriaRios02: Fixed formatting here: [71c9993](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/71c9993f1af37895cae88b020c99f9047badf26a)

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
ValeriaRios02: Fixed formatting here: [71c9993](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/71c9993f1af37895cae88b020c99f9047badf26a)

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
ValeriaRios02: Fixed formatting here: [71c9993](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/71c9993f1af37895cae88b020c99f9047badf26a)

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const icon = (
+      <NxFontAwesomeIcon aria-label={presentational ? undefined : label}
+                         className={className}
+                         icon={faSquare}
+                         role={presentational ? 'presentation' : 'graphics-symbol img'} />
+    );
+
+    const tooltipTitle = title || `Vulnerability: ${capitalize(severity)} (${
+      vulnerabilitySeverityScoreRangeMap[severity]
+    })`;
```
rpokorny, line 46: Try to break lines at the highest syntactic level that makes sense:
```suggestion
    const tooltipTitle = 
        title || `Vulnerability: ${capitalize(severity)} ${vulnerabilitySeverityScoreRangeMap[severity]})`;
```

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
ValeriaRios02: Fixed tests for `NxVulnerabilityIndicator` here: [3be4a99](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/3be4a99165a0ea9d3c233f1b1e3dba4a97999b79)

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
ValeriaRios02: Changed in: [bcf6d2d](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/bcf6d2dba3b42601307d6007ff24ff8e84fd137d)

lib/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegend.tsx
ValeriaRios02: Fixed in: [39456a8](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/39456a8c538c2cc72dde789504f5a8bb8340e3b1)

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,45 @@
...
+ * the terms of the Eclipse Public License 2.0 which accompanies this
+ * distribution and is available at https://www.eclipse.org/legal/epl-2.0/.
+ */
+ @use '../../scss-shared/nx-dark-mode-helpers';
+ 
+// eyeballed, assumes $nx-font-size surrounding text
+$-vertical-alignment: -0.1em;
+
+.nx-vulnerability-indicator {
+  font-size: var(--nx-font-size-s);
```
rpokorny, line 13: The visible shape of this fontawesome icon doesn't take up the full height of its `<svg>` box: the `<svg>` is 14px tall but the visible square is only 12.5px tall.  As it happens, this means that if you use the regular 16px font size, the visible square comes out to the desired 14px.  And then, since the "font size" of the icon is the same as that of the adjacent text, we don't have to use a custom `vertical-align`.

lib/src/components/NxVulnerabilityIndicatorLegend/NxVulnerabilityIndicatorLegend.scss
entroform: > Shouldn't this be 5 now

Ah yes: [64ca095](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/64ca09587889eadcb265b0b8d95ab40526f14c77)

lib/src/components/NxVulnerabilityIndicator/types.ts
entroform: > I think I'd rather this be `number` to more closely match the interface for `NxThreatIndicator`. I know we can't cleanly lock it down to a discrete set of numbers like we do there, but I'm ok with that (as long as its checked at runtime).

I've updated the prop here: [bc1c460](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/bc1c460debb93d085727545318472bb9521685c2)

lib/src/components/NxVulnerabilityIndicator/__tests__/NxVulnerabilityIndicator.test.tsx
entroform: > Fewer tests of values in the middle of each range, more tests of values at the edges of each range.

Updated test: [7d9d07f](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/7d9d07f4f44bfdee3e1d62a911acc76f4df13d1b)

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
entroform: > The visible shape of this fontawesome icon doesn't take up the full height of its `<svg>` box: the `<svg>` is 14px tall but the visible square is only 12.5px tall whereas it's supposed to be 14px. As it happens, this can be fixed by using the regular 16px font size, causing the visible square to come out to the desired 14px. And then, since the "font size" of the icon is the same as that of the adjacent text, we don't have to use a custom `vertical-align`.

Hmm currently discussing this with Kiran will get back to you,
but I think the default font awesome icon svg size determined by font-size should match the threat indicator 
so we can keep the desired visual weight balance benefit
(For example: if the circle without the container is exactly the same dimension as the square, it would be visually smaller. fa intentionally made the circle overshoot so it is visually balanced with the other shapes)

Screenshot 2024-10-15 at 12 27 30 PM

entroform: Just had a discussion with Kiran, and yes this is what we want.
Kiran is updating the mockup to use fa-square and there's going to be a wider discussions on how fa are imported in Figma to keep the boundary so the intended relative sizing is consistent.

I would think the (eyeballing) vertical alignment will be similar as well since they both have now the same relative alignment but will additionally confirm that with design if they are satisfied.

entroform: > Just had a discussion with Kiran, and yes this is what we want. Kiran is updating the mockup to use fa-square and there's going to be a wider discussions on how fa are imported in Figma to keep the boundary so the intended relative sizing is consistent.
> 
> I would think the (eyeballing) vertical alignment will be similar as well since they both have now the same relative alignment but will additionally confirm that with design if they are satisfied.

We added sidebearing to the icon so it aligns with NxThreatIndicator:
[70aa706](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/70aa706afe3e9c60f8784673aecdbd2b5422c094)

Screenshot 2024-10-15 at 5 21 05 PM

gallery/src/pageConfig.ts
rpokorny: > Updated here: [0ea3129](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/0ea31298e43d965f4d28e9acd0e52459dbc87b44)

Move them after "Overflow Tooltip"

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
rpokorny: > We added sidebearing to the icon so it aligns with NxThreatIndicator:

Is that a requirement? In what circumstance would a threat indicator and a vulnerability indicator be used together like this where their relative alignment is relevant?

> Kiran is updating the mockup to use fa-square and there's going to be a wider discussions on how fa are imported in Figma to keep the boundary so the intended relative sizing is consistent.

I've yet to find a way to get the icon name out of figma, but at any rate, the rounded box that I see there is still 14px tall.

entroform: > Is that a requirement? In what circumstance would a threat indicator and a vulnerability indicator be used together like this where their relative alignment is relevant?

It would be unlikely for these two be used together,
but it it was done for consistency sake.

> I've yet to find a way to get the icon name out of figma, but at any rate, the rounded box that I see there is still 14px tall.

Yes, I believe this is because they have not been properly imported into Figma.
We're going to have a wider discussion on this with design
but some of the fa icons imported in Figma have their boundary box discarded (like the em square in type)
so its just the shape that's scaled to 14px.

So in this case the rounded box should be 12.25px within a boundary box that is 12.25px by 14px
but because the boundary box is discarded,
it's left with the 12.25px square scaled to 14px becoming 14x14px in Figma.

I've talked about this with Kiran and I think it's still being worked on the current mockup.
But the fix has been agreed upon.

entroform: > > We added sidebearing to the icon so it aligns with NxThreatIndicator:
> 
> Is that a requirement? In what circumstance would a threat indicator and a vulnerability indicator be used together like this where their relative alignment is relevant?
> 
> > Kiran is updating the mockup to use fa-square and there's going to be a wider discussions on how fa are imported in Figma to keep the boundary so the intended relative sizing is consistent.
> 
> I've yet to find a way to get the icon name out of figma, but at any rate, the rounded box that I see there is still 14px tall.

Had a meeting with Kiran and we addressed the icon issue here for this component:
Kiran said that this is in the process of moving this to Kaleido: along with the rest:
https://www.figma.com/design/jG8RQhjPyIbisYAxDlOyt7/Vulnerability-Threat-Level-Count?node-id=265-213&node-type=frame&t=74prgOHcbyzxWgup-0

gallery/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicatorPage.tsx
rpokorny: > That seems right, thank you!

It doesn't appear that this change was actually made.

JasonDSwearingen:
It looks like there is 8.875 px of spacing between the vuln indicator square and its accompanying text. Is that intentional due to the icon shape issues discussed above?

JasonDSwearingen:
Other than my previous comment, these look good to me!

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.nx-vulnerability-indicator {
+  $-icon-width: 12.25px;
+  $-font-size: var(--nx-font-size-s);
+
+  font-size: $-font-size;
+  // This adds side bearing to the icon
+  // so that it aligns with the fa-circle
+  // in the NxThreatIndicator.
+  margin: 0 calc(
```
rpokorny, line 20: This still seems sketchy to me.  There's nothing in this formula that inherently ties it to the width of the circle icon (unless there's an assumption that the circle's diameter equals the font size), it just kind of happens to work.  If we really want the icons to align (which I still think is an unnecessary requirement), I'd think the right way would be to use FA's `fixedWidth` mode.

lib/src/components/NxVulnerabilityIndicator/types.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as PropTypes from 'prop-types';
+
+import {
+  VulnerabilitySeverityRating,
+  allVulnerabilitySeverityRatings
+} from '../../util/vulnerabilitySeverityRatings';
+
+export interface Props {
+  severityRating?: VulnerabilitySeverityRating;
+  score?: number;
```
rpokorny, line 17:
```suggestion
  score?: number | null;
```

rpokorny:
I see that the mockups have been added to Kaleido at https://www.figma.com/design/SMeWSFCKTQbyarldXKVzro/Kaleido-V2?node-id=8762-15535&node-type=frame&t=1Yi9JF8CzSpqPewI-0 .  Those need improvements though.  They don't depict the legend component, the dark mode styles aren't actually depicted on a dark background, and none of the space it explicitly called out.  Your designer should work with @JasonDSwearingen to get the mockups aligned with how Kaleido mockups are expected to look.

lib/src/components/NxVulnerabilityIndicator/NxVulnerabilityIndicator.scss
entroform: > This still seems sketchy to me. There's nothing in this formula that inherently ties it to the width of the circle icon (unless there's an assumption that the circle's diameter equals the font size), it just kind of happens to work. If we really want the icons to align (which I still think is an unnecessary requirement), I'd think the right way would be to use FA's `fixedWidth` mode.

I guess I should've rephrased the comment so it's not directly tied with making it aligned with NxThreatIndicator but visually consistency with this this type/category of component (fa icon next to label as part in this style). The fontsize is a good anchor for this because it's what determines the canonical boundary size (em square size of the icon) not just the diameter of the circle of fa-circle. This was discussed with Kiran but let me reach out with Jason to get their opinion on this.

And depending on their decision I can remove it or change it to `fixedWidth` mode
and update the comment to reflect the intent better.

rpokorny: > The fontsize is a good anchor for this because it's what determines the canonical boundary size (em square size of the icon) not just the diameter of the circle of fa-circle

Well, except that as demonstrated by the square here, FA icons are not at all guaranteed to be 1em in width.

entroform: > > The fontsize is a good anchor for this because it's what determines the canonical boundary size (em square size of the icon) not just the diameter of the circle of fa-circle
> 
> Well, except that as demonstrated by the square here, FA icons are not at all guaranteed to be 1em in width.

Yes, but for this specific type of component, it was with the intent of icons within that 1em
for example the square or circle which are simplified and pure shapes.

The FA icons that extend beyond 1em width to me fall into a different category and usage.

entroform: Sorry, I miss-read and argued the wrong thing here.

Here we're using that 1em as an "anchor" because the spacing around the icon also contributes to how it is perceived. The overall intent is that we want "indicator type" components essentially this and the threat component to be visually weighted the same.

Yes it is not a requirement to make the two "align" but we still need to make them visually weighted equally.
If we remove the side-bearing it will no longer be visually the same because despite the spacing being equal, it will be perceived as such that there is "less" spacing between the square and the circle. causing more tension which unbalance it visually compared to the circle. This is because the circle has more space around its curvature, whereas the square takes up that area.

rpokorny: That's quite abstract and makes me uncertain what concrete goals you are trying to solve for.  Are you talking about making the size (area) of the icons match? Or are you talking about making the distance from the nearest edge of the icon to the adjacent text match, or are you talking about making the distance from the center of the icon to the adjacent text match? Different parts of what you've said seem to potentially be alluding to all of these, but I don't think it's possible to have all three match at the same time.  

So I am uncertain what you are trying to do, but I thought, in this thread at least, the question was how to make it so that when we have two lines of content, one starting with a Threat Indicator and one starting with a Vulnerability Indicator, both the centers of the two icons and the beginnings of the two text spans are horizontally aligned, respectively.  This would imply a consistent distance from the center of the icons to their left and right margin box edges.  But we disagree about how that should be implemented, and how effective the current implementation is.  The standard way to accomplish this with with FA's `fixedWidth` setting (plus consistent margins).  What you have currently is an adjustment to the l/r margin of the square icon that is based on an _assumption_ of the width of the circle icon, even though that width is not something that we are controlling.  That is risky, and also complicated. Whether this assumed width is 1em or some other value doesn't really matter, it's still a risky assumption that seems unnecessary given the existence of a more standard mechanism for this.

> but we still need to make them visually weighted equally which is a requirement.

Is it? Why? I thought it was established that we don't actually expect Threat Indicators and Vulnerability Indicators to be used near each other.

rpokorny: I also think this isn't the level at which this debate should be had, nor are we the people to have it.  I am still awaiting precise, well documented mockups, where the answers to these questions should be worked by the designers.  Once those exist we can debate how to make the pixels in the implementation match those in the mockup.

rpokorny: > The 1em width is a natural anchor because, both shapes are symmetrical, the offset between the top and bottom of the square is the intended weighted offset by the icon designer, and we're just extending that symmetrically on the transverse axis.

OK I think I finally get what you're saying with this part.  You aren't trying to directly make one icon the size of the other, but are more trying to make both icons be `1em` x `1em` in size.  In that case, a simpler implementation of that goal might be to simply set `width: 1em` rather than the complex margin calculation.  And we'd want to do that in Threat Indicator's styles too, rather than relying on an assumption that the circle is already 1em wide.  However, I still contend that the more standard approach would be the `fixedWidth` setting.  It'll depend on what spacing the mockups call for once they are finished.

entroform: > That's quite abstract and makes me uncertain what concrete goals you are trying to solve for. Are you talking about making the size (area) of the icons match? Or are you talking about making the distance from the nearest edge of the icon to the adjacent text match, or are you talking about making the distance from the center of the icon to the adjacent text match? Different parts of what you've said seem to potentially be alluding to all of these, but I don't think it's possible to have all three match at the same time.
 
Yes the concrete goal is to keep the intended visual weight of the icons and spacing between the icon and tex consistent as the icon designer intended which included the area and the offset between the nearest edge of the icon and adjacent text. On the topic of area, this was already set by the icon designer. They purposefully drew the icon within the 1em square and each shape are relative to that. The area of the square may not match 100% mathematically with the circle, but this was the decision of the icon designer, either artistically or stylistically. We should respect that same as we would with typography so that's why we want to keep the 1em bounding box vertically. The point here is that this also effects the spacing horizontally between other elements and the icon.
Since the icon is symmetrical, I thought the best way to do this was to keep the 1em horizontally as well because the optical offset would be the same.

> So I am uncertain what you are trying to do, but I thought, in this thread at least, the question was how to make it so that when we have two lines of content, one starting with a Threat Indicator and one starting with a Vulnerability Indicator, both the centers of the two icons and the beginnings of the two text spans are horizontally aligned, respectively. This would imply a consistent distance from the center of the icons to their left and right margin box edges. But we disagree about how that should be implemented, and how effective the current implementation is. The standard way to accomplish this with with FA's fixedWidth setting (plus consistent margins). What you have currently is an adjustment to the l/r margin of the square icon that is based on an assumption of the width of the circle icon, even though that width is not something that we are controlling. That is risky, and also complicated. Whether this assumed width is 1em or some other value doesn't really matter, it's still a risky assumption that seems unnecessary given the existence of a more standard mechanism for this.

On the topic of implementation, I do agree to switching to fixedWidth setting is a better approach, I'm just waiting for confirmation from Jason before making that change.

> Is it? Why? I thought it was established that we don't actually expect Threat Indicators and Vulnerability Indicators to be used near each other.

Well, it's part of good UX is strive for consistency, although these two components are not expected to be used near each other, they are of a similar style. Both use the same class of icons, and have a label next to them.
Furthermore, we would be doing the icon designer and design in general a disservice to not optically place this icon with other elements. I guess why would we not, knowing the intent of how the icons are weighted relative to each other we don't want to ignore that.

entroform: Got confirmation from Jason, updated implementation here: [0ab0aad](https://github.com/sonatype/sonatype-react-shared-components/pull/1199/commits/0ab0aad44d91be976c108e4bf5fb8bae5c5e0a25)

entroform: Ah sorry Ross, just saw the other two comments, for some reason I didn't see them when I responded.

We had that discussion with Kiran and Jason, but were considering having a wider discussion later down the road on addressing the icon import issues in Figma and alignment issues later after this initiative.
But we did all agree to make both 1em by 1em for this vulnerability indicator though.

Kiran and Jason are currently working together on getting all the stuff into Kaleido up to standard. I know they're quite busy so I think we may get the update towards the end of this week.

In terms of implementation I do agree with your suggestion, I made a commit above but will make further adjustments per your recommendation.

rpokorny (approved):

entroform (approved):

PR #1203, opened 2024-10-11
NxGlobalHeader2 - RSC-1780

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1780

New header to go along with `NxGlobalSidebar2`.  This does not implement the new styling on the action buttons, that is a separate ticket.


gallery/src/components/NxGlobalHeader2/NxGlobalHeader2Page.tsx
```diff
@@ -0,0 +1,177 @@
...
+      <NxP>
+        Like all parts of the <NxCode>NxGlobalSidebar2</NxCode> layout, this component is only to be
+        used in full-width, section-scrolling layouts. Although the contents of <NxCode>NxGlobalHeader2</NxCode> are
+        flexible, there are strong design guidelines for the content that should be present. From left to right,
+        the header <em>should</em> contain the following items:
+      </NxP>
+      <NxList bulleted>
+        <NxList.Item>The official product logo depicting its name</NxList.Item>
+        <NxList.Item>
+          <NxList bulleted>
```
mealingr, line 35: Should this be nested under a `<NxList.Item>`? (looks a bit odd to me but doesn't really matter)
image

rpokorny: > Should this be nested under a <NxList.Item>? (looks a bit odd to me but doesn't really matter)

Whoops yeah I didn't have that structured correctly at all. 1ff4d4ec5

benflodge (approved):
+1 Great work

JasonDSwearingen:
Notes:

1. The icons should be nx-indigo-20.
2. Are the margins around the icon going to be part of the icon styling? There's a bit too much space between it and the search dropdown, but I'm guessing that will be part of that work. If so, disregard this note.

rpokorny:
> The icons should be nx-indigo-20.

Styles of the icon-only buttons will be implemented in https://sonatype.atlassian.net/browse/RSC-1784 .  There's more wrong with them than just the icon color.

> Are the margins around the icon going to be part of the icon styling? There's a bit too much space between it and the search dropdown, but I'm guessing that will be part of that work. If so, disregard this note.

Fixed in 0b9cdc3e

JasonDSwearingen (approved):
+1

PR #1208, opened 2024-11-04
NxNavigationDropdown - RSC-1784

rpokorny:
https://sonatype.atlassian.net/browse/RSC-1784

This adds a new styling variation for dropdowns, intended only to be used within `NxGlobalHeader2`.


lib/src/base-styles/_nx-btn.scss
```diff
@@ -19,7 +19,7 @@
     color: var(--nx-color-#{$variant}-button-text);
   }
 
-  &:focus:not(:active) {
```
rpokorny, line 22: These `:not` selectors were causing unnecessary specificity that was getting in the way so I refactored it a bit.  You can see by the fact that no unrelated button visual tests changed that the change did not have unintended consequences.

lib/src/components/NxDropdown/NxDropdown.scss
```diff
@@ -14,7 +14,13 @@
   position: relative;
   width: nx-dropdown-variables.$dropdown-width;
 
-  .nx-icon:not(.nx-icon--colorful, .nx-dropdown-link .nx-icon, .nx-filter-dropdown__reset .nx-icon, .nx-dropdown :is(.disabled, [disabled]) .nx-icon) {
+  .nx-icon:not(
+    .nx-icon--colorful,
+    .nx-dropdown-link .nx-icon,
+    .nx-filter-dropdown__reset .nx-icon,
+    .nx-dropdown :is(.disabled, [disabled]) .nx-icon,
+    .nx-navigation-dropdown .nx-icon
```
rpokorny, line 22: It is increasingly unfortunate that this bit of code has to be aware of these various special cases, but that's the pattern for the time being.

lib/src/components/NxDropdownMenu/types.tsx
```diff
@@ -7,8 +7,9 @@
 import { HTMLAttributes } from 'react';
 import * as PropTypes from 'prop-types';
 
-export interface Props extends HTMLAttributes<HTMLDivElement> {
+export interface Props extends Omit<HTMLAttributes<HTMLDivElement>, 'className'> {
   onClosing: () => void;
+  className?: string | null;
```
rpokorny, line 12: Allows `null` in addition to `undefined` which always should have been permitted.

JasonDSwearingen (approved):
+1

mealingr (approved):
+1 (also ran locally again with https://www.figma.com/design/UpWZYPAaaOwafLeYytZ99l/Updated-Side-and-Top-Nav-Component---Q2-2024---JS?node-id=0-1&node-type=canvas&t=YH0r7DXqMgXbIDNh-0 as reference and it looks great)

PR #1220, opened 2025-03-14
Rsc 1795 small button

Qiming-C:


gallery/src/styles/NxBtn/NxBtnPage.tsx
```diff
@@ -145,6 +153,13 @@ const NxBtnPage = () =>
       error buttons. Note that the standard disabled styles override the error styles.
     </GalleryExampleTile>
 
+    <GalleryExampleTile title="Small"
+                        htmlExample={nxBtnSmallCode}
+                        codeExamples={nxBtnSmallCode}>
+      A demonstration of an <NxCode>nx-btn</NxCode> using "small" styles, along with a disabled
+      error buttons. Note that the standard disabled styles override the small styles.
```
mealingr, line 160: What does it mean for the standard disabled styles to override the small styles? (I know this is similar wording to the above, but if the small style is overridden that reads like it'll become normal sized :sweat_smile: )

mealingr, line 160: "a disabled error buttons" > "a disabled small button"

gallery/visualtests/nxButton.js
```diff
@@ -114,8 +115,20 @@ describe('NxButton', function() {
           focusAndHoverTest(selector));
     };
 
+    describe('Small NxButton section', function() {
+      const selector = '#nx-button-small-example .nx-btn--small:first-child';
+
+      it('has a white background by default', simpleTest(selector));
+      it('has a light blue background when hovered', hoverTest(selector));
+      it('has a light blue background when clicked', clickTest(selector));
+      it('has a blue inner outline when focused', focusTest(selector));
```
mealingr, line 124: Looks more like a grey inner outline when focussed for the light theme
image

but does look light blue for the dark theme
image

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,17 @@
...
+  &--small {
+    font-size: var(--nx-font-size-s);
+    padding: var(--nx-spacing-base) var(--nx-spacing-3x);
+    margin: 0 var(--nx-spacing-3x) 0 0;
+
+    .nx-icon {
+      height: var(--nx-spacing-3x);
+      margin-right: var(--nx-spacing-2x);
+    }
+  }
```
mealingr, line 114: The colors look fine to me but are slightly different to figma e.g.

Figma
image

Actual
image

and

Figma
image


Actual
image

gallery/src/styles/NxBtn/NxBtnPage.tsx
mealingr: Well it looks fine on the page, just the wording is a bit odd
image

mealingr (approved):
+1

gallery/src/styles/NxBtn/NxBtnLinkExample.html
```diff
@@ -11,10 +11,14 @@
...
 </div>
 <div class="nx-btn-bar">
   <a href="#/" class="nx-btn nx-btn--primary disabled">Primary Disabled</a>
   <a href="#/" class="nx-btn disabled">Secondary Disabled</a>
   <a href="#/" class="nx-btn nx-btn--tertiary disabled">Tertiary Disabled</a>
   <a href="#/" class="nx-btn nx-btn--error disabled">Error Disabled</a>
 </div>
+<div class="nx-btn-bar">
+  <a href="#/" class="nx-btn nx-btn--small disabled">Small Disabled</a>
+</div>
```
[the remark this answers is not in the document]
Qiming-C: when I placed in the second button bar, the existing buttons will be wrap where there isn't enough width for them to fit.
image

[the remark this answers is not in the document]
mealingr: Oh interesting, ok that makes sense :+1:

lib/src/base-styles/_nx-btn.scss
Qiming-C: updated the style to align with design

gallery/visualtests/nxButton.js
```diff
@@ -114,8 +115,20 @@ describe('NxButton', function() {
...
+      it('has a darker blue inner outline and light blue background when focused and hovered',
```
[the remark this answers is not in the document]
Qiming-C: updated the text

gallery/visualtests/nxButton.js
Qiming-C: it gives me a blueish feeling. but yes it is grey when i use color picker ;)

gallery/src/styles/NxBtn/NxBtnPage.tsx
Qiming-C: fixed

rpokorny:
Check with Design but I assume the intent is that the smallness is an independent axis from the button `variant`.  That is, I would think we are supposed to be adding support for small secondary buttons and also small primary buttons, small tertiary buttons, and small error buttons.  All of which need examples and visual tests.  More mockups may also be needed since the primary and error buttons aren't totally similar to the secondary and tertiary buttons since they don't have borders, making it potentially ambiguous how small primary buttons should look.

gallery/src/styles/NxBtn/NxBtnPage.tsx
```diff
@@ -62,6 +63,13 @@ const NxBtnPage = () =>
               <NxTable.Cell>Modifier on <NxCode>.nx-btn</NxCode></NxTable.Cell>
               <NxTable.Cell>Makes the button a "tertiary" button</NxTable.Cell>
             </NxTable.Row>
+            <NxTable.Row>
+              <NxTable.Cell><NxCode>nx-btn--small</NxCode></NxTable.Cell>
+              <NxTable.Cell>Modifier on <NxCode>.nx-btn</NxCode></NxTable.Cell>
+              <NxTable.Cell>
+                Buttons that typically are used in situations where patterns that contain a button are repeated
```
rpokorny, line 70: I find this description confusing.  I see you copied it from the mockup, but left off part of the description which, in my opinion, made it a lot clearer.  Consider restoring that extra text.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,34 @@
       height: 16px;
     }
   }
+
+  &--small {
+    font-size: var(--nx-font-size-s);
+    padding: var(--nx-spacing-base) var(--nx-spacing-3x);
```
rpokorny, line 107: Like the full-size buttons, you need to keep in mind whether the spacing you can see in figma includes the borders or not.  In this case, the 12px and 4px padding values in the mockup _do_ include the border, which means in an HTML sense the actual padding is supposed to be 11px and 3px.  See the individual pixel boxes below:
image

You can see the way that we deal with this for the full-size buttons on line 47.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,34 @@
...
+    margin: 0 var(--nx-spacing-3x) 0 0;
```
rpokorny, line 108: I'm not seeing a spec for this in the mockup, where did you get it from?

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,34 @@
...
+    border-color: var(--nx-color-small-button-border);
+    color: var(--nx-color-small-button-text);
```
rpokorny, line 110: Related to https://github.com/sonatype/sonatype-react-shared-components/pull/1220#issuecomment-2730338949, these colors should depend on whether it's a primary, secondary, etc button.

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,34 @@
...
+
+    .nx-icon {
+      height: var(--nx-spacing-3x);
```
rpokorny, line 113: What's this about? The icon should just follow the text height which it does automatically

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,34 @@
...
+      margin-right: var(--nx-spacing-2x);
```
rpokorny, line 114: Should be redundant with `.nx-icon`'s default margin

lib/src/base-styles/_nx-btn.scss
```diff
@@ -101,6 +101,34 @@
...
+    }
+
+    &:focus {
+      outline-color: var(--nx-color-small-button-focus-outline);
+    }
+
+    &:active {
+      background-color: var(--nx-color-small-button-background-active);
+      border-color: var(--nx-color-small-button-border-active);
+      color: var(--nx-color-small-button-text);
```
rpokorny, line 129: I'm expecting that no small-specific color styles will be necessary once smallness is separated out from button variant

lib/src/components/NxButton/__tests__/NxButton.test.tsx
```diff
@@ -65,4 +65,37 @@ describe('NxButton', function() {
     expect(() => render(<NxButton variant="icon-only" title="Save" disabled />))
         .toThrow(TypeError);
   });
+
+  it('applies small modifier class when provided in className', function() {
```
rpokorny, line 69: Redundant with "it adds specified classNames to the element in addition to the defaults" test on line 24

lib/src/components/NxButton/__tests__/NxButton.test.tsx
```diff
@@ -65,4 +65,37 @@ describe('NxButton', function() {
...
+    const smallBtn = quickRender({ className: 'nx-btn--small' }).getByRole('button');
+    expect(smallBtn).toHaveClass('nx-btn--small');
+  });
+
+  it('works with small modifier across different button variants', function() {
```
rpokorny, line 74: IMO this test isn't necessary either.  While it's true that we don't have separate versions of the "it adds specified classNames to the element in addition to the defaults" for each `variant`, I think it would be excessive to do so.  Also the classnames for the variants (such as `.nx-btn--secondary`) are implementation details and should not be asserted in unit tests.  Only the stylistic results of those classes matter externally, and those are checked in the visual tests.

gallery/visualtests/nxButton.js
```diff
@@ -114,8 +115,20 @@ describe('NxButton', function() {
           focusAndHoverTest(selector));
     };
 
+    describe('Small NxButton section', function() {
```
rpokorny, line 118: For consistency with the others
```suggestion
    describe('small styled buttons', function() {
```
Note when updating test names that you'll want to update the names of the images too, to avoid generating unnecessary new images and taking up even more space in git-lfs.

rpokorny, line 118: Wait why is this nested within the "describe nx-btn class on <a>" block?

rpokorny: > Note when updating test names that you'll want to update the names of the images too, to avoid generating unnecessary new images and taking up even more space in git-lfs.

I see there is such a block online 150 below.  Is that section redundant with this one?

lib/src/base-styles/_nx-btn.scss
Qiming-C: from the design, the icon is smaller than the text height(21px) whereas the icon is 12px

lib/src/base-styles/_nx-btn.scss
Qiming-C: removed. https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/d6eeafa238b8da4d8e0ff5f0725ffb2148206c44
the mockup show a spacing between text and icon. but the default style handle that already

lib/src/base-styles/_nx-btn.scss
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/1a8d7d3e82a75132d6e193e3bae6877b22181660

lib/src/base-styles/_nx-btn.scss
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/1a8d7d3e82a75132d6e193e3bae6877b22181660

lib/src/base-styles/_nx-btn.scss
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/1a8d7d3e82a75132d6e193e3bae6877b22181660

lib/src/base-styles/_nx-btn.scss
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/1a8d7d3e82a75132d6e193e3bae6877b22181660

gallery/src/styles/NxBtn/NxBtnPage.tsx
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/a368678059041651ca20de36e398d78bd968e5c9

gallery/visualtests/nxButton.js
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/77fe815c43acda9601e579b05302e602784189c6

lib/src/components/NxButton/__tests__/NxButton.test.tsx
Qiming-C: removed

lib/src/components/NxButton/__tests__/NxButton.test.tsx
Qiming-C: good point, removed

lib/src/base-styles/_nx-btn.scss
rpokorny: > from the design, the icon is smaller than the text height(21px) whereas the icon is 12px

An example with an icon needs to be added to the gallery so I can further investigate this (and so visual tests can check it).

gallery/visualtests/nxButton.js
```diff
@@ -99,9 +99,68 @@ describe('NxButton', function() {
     it('shows tooltips', hoverTest(iconOnlyBtnExample, selector, true));
   });
 
+  describe('Small NxButton section', function() {
```
rpokorny, line 102: As I suggested on slack at one point, you may want to separate these into a separate test file so that they can be run in parallel with the existing tests.  `nxButton.js` was already the longest-running visual test before these additions, so anything further added to this file is going to have a significant impact on the runtime of the visual tests overall.

gallery/visualtests/nxButton.js
```diff
@@ -115,7 +174,8 @@ describe('NxButton', function() {
     };
 
     describe('Secondary styled button', function() {
-      const secondarySelector = `${selector} .nx-btn:not(.nx-btn--primary):not(.nx-btn--tertiary):not(.nx-btn--error)`;
+      const secondarySelector = `${selector} .nx-btn:not(.nx-btn--primary):not(.nx-btn--tertiary):not(.nx-btn--error)`
+      + ':not(.nx-btn--small)';
```
rpokorny, line 178: I _think_ this can be simplified with the new-ish `:is` selector (which I've never used before, so I'm not certain):
```suggestion
      const secondarySelector = `${selector} .nx-btn:not(:is(.nx-btn--primary, .nx-btn--tertiary, .nx-btn--error, .nx-btn--small))';
```

Qiming-C: do you mean `.nx-btn:is` or `.nx-btn:not(:is..`

rpokorny: > do you mean .nx-btn:is or .nx-btn:not(:is..

I meant not is.  We want the button that is _not_ primary, tertiary, error, or small.

Qiming-C: it works! 🎆  https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/d1edf6629c3d640d8f16dafce7bcbdacb0194ce0

gallery/visualtests/nxButton.js
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/d1edf6629c3d640d8f16dafce7bcbdacb0194ce0

lib/src/base-styles/_nx-btn.scss
Qiming-C: updated icon case to example

rpokorny: > updated icon case to example

This needs some consultation with Design regarding their intent I think.  It's not common for us to use icons with a different size than the adjacent text, especially as a general rule.  At the same time, I notice that the icon in question in the mockup is an external link icon.  Looking at the way we use that particular icon in `NxTextLink`, I see that we pass it the `size="sm"` fontawesome prop, which changes the `font-size` (and thus the height) of that icon.  So it could be the case that our designers don't like that particular icon at full size (relative to the adjacent text), but I'm not confident at all that that means they want all icons to be smaller than the font size here.


Additionally, I see you now have `max-width: 12px` as well, which is probably not desired.  Icons vary in width and I see no reason to assume we can't or don't want to allow wider ones in this spot. @JasonDSwearingen please weigh in.

JasonDSwearingen: Yep, we need to add examples of icons in the small buttons. It definitely shouldn't have a max width. I'll experiment with height and DM @Qiming-C  when it's ready.

gallery/visualtests/nxButton-2.js
```diff
@@ -0,0 +1,74 @@
...
+      '#nx-button-small-example .nx-btn-bar:first-child .nx-btn--small[disabled]';
+
+    it('has disabled styling', simpleTest(disabledSelector));
+    it('maintains disabled styling when hovered', hoverTest(disabledSelector));
+    it('maintains disabled styling when clicked', clickTest(disabledSelector));
+    it('maintains disabled styling when focused', focusTest(disabledSelector));
+  });
+
+  describe('Small Button with Icon', function () {
+    const withIconSelector = `${selector} .nx-btn-bar:nth-child(2) .nx-btn--small:has(span, svg.nx-icon)`;
```
rpokorny, line 71: From MDN:
> The functional :has() CSS [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) represents an element if any of the [relative selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_selectors/Selector_structure#relative_selector) that are passed as an argument match at least one element when anchored against this element

So this'll match a button that contains _either_ a span or an svg.  Is that really what you want? Maybe it would be cleaner to just give this button an id.

gallery/visualtests/nxButton.js
```diff
@@ -115,7 +115,8 @@ describe('NxButton', function() {
     };
 
     describe('Secondary styled button', function() {
-      const secondarySelector = `${selector} .nx-btn:not(.nx-btn--primary):not(.nx-btn--tertiary):not(.nx-btn--error)`;
+      const secondarySelector = `${selector} .nx-btn:not`
+      + '(:is(.nx-btn--primary, .nx-btn--tertiary, .nx-btn--error, .nx-btn--small))';
```
rpokorny, line 119: This is a really odd place to break the string into two lines.  I suggest breaking it on the space before `.nx-btn`.

Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/d7363ee2fb50ba3d6f4a2047de3470128540a8f3

gallery/visualtests/nxButton-2.js
Qiming-C: https://github.com/sonatype/sonatype-react-shared-components/pull/1220/commits/d7363ee2fb50ba3d6f4a2047de3470128540a8f3

gallery/src/styles/NxBtn/NxBtnSmallExample.html
```diff
@@ -0,0 +1,25 @@
...
+    <button class="nx-btn nx-btn--primary nx-btn--small">Primary button</button>
+    <button class="nx-btn nx-btn--secondary nx-btn--small">Secondary button</button>
+    <button class="nx-btn nx-btn--tertiary nx-btn--small">Tertiary button</button>
+    <button class="nx-btn nx-btn--error nx-btn--small">Error button</button>
+    <button class="nx-btn nx-btn--small" disabled>Small button disabled</button>
+  </div>
+  <div class="nx-btn-bar">
+    <button class="nx-btn nx-btn--tertiary nx-btn--small" id="nx-button-small-icon-button">
+      <span>Button with icon</span><!--
+ --><svg class="nx-icon" style="height: 1em; width: 1em;" viewBox="-1 -1 2 2">
```
rpokorny, line 20: This looks like the correct indentation
```suggestion
   --><svg class="nx-icon" style="height: 1em; width: 1em;" viewBox="-1 -1 2 2">
```

gallery/visualtests/nxButton.js
```diff
@@ -115,7 +115,8 @@ describe('NxButton', function() {
     };
 
     describe('Secondary styled button', function() {
-      const secondarySelector = `${selector} .nx-btn:not(.nx-btn--primary):not(.nx-btn--tertiary):not(.nx-btn--error)`;
+      const secondarySelector = `${selector} ` +
+          '.nx-btn:not(:is(.nx-btn--primary, .nx-btn--tertiary, .nx-btn--error, .nx-btn--small))';
```
rpokorny, line 119:
```suggestion
      const secondarySelector = selector +
          ' .nx-btn:not(:is(.nx-btn--primary, .nx-btn--tertiary, .nx-btn--error, .nx-btn--small))';
```

gallery/visualtests/nxButton.js
```diff
@@ -134,9 +135,15 @@ describe('NxButton', function() {
...
+    describe('Small styled buttons', function() {
+      const smallSelector = `${selector} .nx-btn--small`;
+      buttonTests(smallSelector);
+    });
+
     describe('Disabled styled buttons', function() {
       describe('secondary', function() {
-        const selector = `${disabledSelector} .nx-btn:not(.nx-btn--primary):not(.nx-btn--tertiary):not(.nx-btn--error)`;
+        const selector = `${disabledSelector} ` +
+            '.nx-btn:not(:is(.nx-btn--primary, .nx-btn--tertiary, .nx-btn--error, .nx-btn--small))';
```
rpokorny, line 146:
```suggestion
        const selector = disabledSelector} +
            ' .nx-btn:not(:is(.nx-btn--primary, .nx-btn--tertiary, .nx-btn--error, .nx-btn--small))';
```

